PR #102, opened 2021-08-08
ext.with() return result

ukint-vs:
Wrap ext.with() in result.

@gear-tech/dev


shamilsan (approved):
Need to be refactored deeper but ok at the moment as a temporary solution.

PR #112, opened 2021-08-09
[skip ci] Make docker dir

ukint-vs:
- build.sh for building Docker image.
- docker-compose-local.yml runs two nodes in subnet.
- docker-compose-dev.yml runs one node in dev mode. 

@gear-tech/dev


shamilsan (approved):
:+1:

PR #117, opened 2021-08-09
Refactor queue_message/queue_reply in the runner

NikVolf:
To make all arguments explicit

### DIFF
--- a/core-runner/src/runner.rs
+++ b/core-runner/src/runner.rs
@@ -210,6 +210,64 @@ pub struct ProgramInitialization {
     pub code: Vec<u8>,
 }
 
+/// New message dispatch request.
+///
+/// Message is dispatched from some identity to some identity, both should be known in advance.
+pub struct MessageDispatch {
+    /// Identity of the message origin.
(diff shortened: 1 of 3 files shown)


core-runner/src/runner.rs
```diff
@@ -210,6 +210,64 @@ pub struct ProgramInitialization {
...
+pub struct MessageDispatch {
+    /// Identity of the message origin.
+    pub source_id: ProgramId,
+    /// Identity of the destination.
+    pub destination_id: ProgramId,
+    /// Message payload and economic data.
+    pub data: ExtMessage,
+}
+
+impl MessageDispatch {
```
shamilsan, line 225: Why to not just implement `From<MessageDispatch> for Message` and then use `from()` and `into()`?

NikVolf: because it is in the different crate?

NikVolf: `Message` is local there, not so for `gear-core-runner`

shamilsan (approved):

PR #109, opened 2021-08-08
Add different allocation checks to gtest

Anfid:
Required for correct testing of (#80)

Allows to test allocation page count and compare exact allocated page indexes.

@shamilsan

PR #122, opened 2021-08-09
Rework wait-wake API for smart-contracts

shamilsan:
- Add `gr_wake` syscall.
- Move the message to the wait list after calling `gr_wait`.
- Move the message from the wait list to the main queue after calling `gr_wake`.

Done:
- [x] Fix node tests
- [x] Fix RPC test

@gear-tech/dev


core-backend/src/funcs.rs
```diff
@@ -226,7 +226,18 @@ pub(crate) fn wait<E: Ext>(ext: LaterExt<E>) -> impl Fn() -> Result<(), &'static
     move || {
         let _ = ext.with(|ext: &mut E| ext.wait())?;
         // Intentionally return an error to break the execution
-        Err("wait")
+        Err("exit")
```
NikVolf, line 229: let's make `const EXIT_TRAP_STR: &'static str = "exit"` ?

shamilsan:
Problem seems to be here:

https://github.com/gear-tech/gear/blob/55e3be3cc1c807e193fb7dc4a652f0957ea80271/gtest/src/runner.rs#L76

Will be fixed within #109.

core-backend/src/funcs.rs
shamilsan: Done.

ukint-vs (approved):

core/src/storage.rs
```diff
@@ -173,12 +173,12 @@ impl InMemoryWaitList {
 }
 
 impl WaitList for InMemoryWaitList {
-    fn insert(&mut self, waker_id: MessageId, message: Message) {
-        self.inner.insert(waker_id, message);
+    fn insert(&mut self, id: MessageId, message: Message) {
```
NikVolf, line 176: `WaitList` should be partitioned by `program_id`

shamilsan: Ok, I'll add an issue.

PR #132, opened 2021-08-11
Remove unused code from storage

### LINKED ISSUE #126: Remove code from storage if it's not referenced by any program
### Problem to Solve

Currently programs can share the same code because the same code would produce the same key in the key-value store.
In case all programs referencing a piece of code are removed, the code itself should be cleared from storage, as well.

### Possible Solution

Use reference counting pointers.

### Notes

_No response_

ekovalev:
Fixes #126.

- Reference counter introduced for each piece of code.
- If reference counter becomes 0 after a program removal, the code is also cleared from storage.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -119,11 +119,13 @@ fn program_key(id: H256) -> Vec<u8> {
     key
 }
 
-fn code_key(code_hash: H256) -> Vec<u8> {
-    let mut key = Vec::new();
+fn code_key(code_hash: H256) -> (Vec<u8>, Vec<u8>) {
+    let (mut key, mut ref_counter) = (Vec::new(), Vec::new());
     key.extend(b"g::code::");
     code_hash.encode_to(&mut key);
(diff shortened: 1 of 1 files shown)


common/src/lib.rs
```diff
@@ -141,11 +143,44 @@ fn wait_key(id: H256) -> Vec<u8> {
...
+            u32::from_le_bytes(v)
+        })
+        .unwrap_or_default()
 }
 
-pub fn set_code(code_hash: H256, code: &[u8]) {
-    sp_io::storage::set(&code_key(code_hash), code)
+fn set_code_refs(code_hash: H256, value: u32) {
+    // let mut value_bytes = [0_u8; 4];
+    // value_bytes[0..4].copy_from_slice(&value.to_le_bytes());
```
ukint-vs, line 167: rm comments

ukint-vs (approved):
👍

NikVolf (approved):

PR #143, opened 2021-08-18
Add benchmarks to pallet-gear

ekovalev:


runtime/Cargo.toml
```diff
@@ -58,7 +58,7 @@ pallet-gear = { version = "2.0.0", default-features = false, path = "../pallets/
 gear-rpc-runtime-api = { version = "2.0.0", default-features = false, path = "../pallets/gear/rpc/runtime-api" }
 
 [build-dependencies]
-substrate-wasm-builder = { version = "4.0.0", git = "https://github.com/paritytech/substrate.git" }
+substrate-wasm-builder = { version = "4.0.0" }
```
shamilsan, line 61:
```suggestion
substrate-wasm-builder = "4.0.0"
```

pallets/gear/src/benchmarking.rs
```diff
@@ -16,27 +16,213 @@
...
 
-//! Benchmarking setup for pallet-template
+//! Gear pallet benchmarking
+
+#![cfg(feature = "runtime-benchmarks")]
 
 use super::*;
+use codec::Encode;
+use common::{IntermediateMessage, Origin};
+use parity_wasm::elements::{
```
shamilsan, line 26: Philosophic question :-)
Isn't better to use `use parity_wasm::elements::*;` when we import so many subjects?

ekovalev: Agree, it'd be more readable that way.

NikVolf:
I thought benchmarking should auto-generate weight formulas which we will attach to our runtime calls?

NikVolf (approved):

ekovalev:
> I thought benchmarking should auto-generate weight formulas which we will attach to our runtime calls?

They do: you need to run `gear-node benchmark` CLI command with `--output` flag. I thought of including default weights for extrinsics based on benchmarking in a next PR because it's essentially a different task.

PR #146, opened 2021-08-19
Remove rpc tests from CI

breathx:
At the moment RPC tests works randomly and cannot act as a good test yet.

They might be fixed later, so there is no need to delete it now, just comment it in CI check.

@gear-tech/dev

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -74,10 +74,10 @@ jobs:
       - name: Run testsuite
         run: ./target/release/gear-node runtests ./gtest/spec/*.yaml
 
-      - name: Run rpc-test
-        run: |
-          ./target/release/gear-node --dev -l 2 & sleep 5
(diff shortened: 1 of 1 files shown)


.github/workflows/master.yml
```diff
@@ -74,10 +74,10 @@ jobs:
...
         run: ./target/release/gear-node runtests ./gtest/spec/*.yaml
 
-      - name: Run rpc-test
-        run: |
-          ./target/release/gear-node --dev -l 2 & sleep 5
-          node rpc-tests/index.js ./gtest/spec/*.yaml
+      # - name: Run rpc-test
+      #   run: |
+      #     ./target/release/gear-node --dev -l 2 & sleep 5
+      #     node rpc-tests/index.js ./gtest/spec/*.yaml
```
NikVolf, line 80:
```suggestion
```

NikVolf (approved):

PR #147, opened 2021-08-20
Add interaction with meta data

breathx:


gstd/Cargo.toml
```diff
@@ -7,6 +7,9 @@ license = "GPL-3.0"
 
 [dependencies]
 dlmalloc = { git = "https://github.com/nikvolf/dlmalloc-rs", features = ["global"] }
+scale-info = { version = "0.10.0", default-features = false, features = ["derive", "serde"] }
```
NikVolf, line 10: Nothing of this should be a dependency of gstd 

this should go to gstd-meta or similar

NikVolf (changes requested):
Other than moving it all to separate crate lgtm

gstd-meta/codegen/src/lib.rs
```diff
@@ -0,0 +1,42 @@
...
+
+extern crate alloc;
+extern crate proc_macro;
+
+use alloc::string::{String, ToString};
+use proc_macro::TokenStream;
+
+#[proc_macro_attribute]
+pub fn gear_data(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let derive_string = String::from("#[derive(Debug, Decode, Encode, TypeInfo)]");
```
NikVolf, line 29: this is not needed (the whole `#[gear_data]` thing)

user is quite capable of putting `#[derive(Debug, Decode, Encode, TypeInfo)]` on his own

NikVolf:
The js-test with meta should be extended to test dynamically generated type info-s

breathx:
Moved to separate PR's with fixes

Issue #107, opened 2021-08-08
Implement codec:Output for writing message by parts 


NikVolf:
Given we have syscalls to write message by parts, we can write something like `Vec<impl codec::Encde>` by parts

cc @breathx 

https://docs.rs/parity-scale-codec/2.2.0/parity_scale_codec/trait.Output.html

Anfid:
Implemented in (#153)

PR #153, opened 2021-08-24
Move 'gstd' to 'gcore'. Allow Encode/Decode values in 'gstd' functions (#107)

Anfid:
Implements #107.

- `gstd` was renamed to `gcore`;
- New library `gstd` added, which contains higher-level binding to `gcore` with feature-gated `std` support;
- Output codec was implemented on `gstd::msg::MessageHandle` to write message by parts;
- `gstd::msg` functions accept `codec::Encode` instead of `&[u8]` as it is in `gcore`;
- Where applicable, message functions have `<function>_bytes` counterpart to support sending byte buffer without encoding;

**Note:** This solution is a breaking change, all examples had to be updated. Since `&[u8]` also implements `codec::Encode`, it is easy to misuse `msg::send` if sending byte buffer without encoding is intended. Encoded byte buffer is incompatible with unencoded, because `parity-scale-codec` prepends buffer length to encoded buffer.

@gear-tech/dev


gcore/README.md
```diff
@@ -0,0 +1,3 @@
+# gstd
```
shamilsan, line 1:
```suggestion
# gcore
```

gcore/src/general.rs
```diff
@@ -16,8 +16,8 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
-pub struct MessageHandle(pub(crate) u32);
```
breathx, line 20: Continue to use `pub(crate)` is still a rational idea, as it seems to me.

gstd/src/msg.rs
```diff
@@ -17,155 +17,87 @@
...
-            value_ptr: *const u8,
-        );
-        pub fn gr_send_init() -> u32;
-        pub fn gr_send_push(handle: u32, data_ptr: *const u8, data_len: u32);
-        pub fn gr_size() -> u32;
-        pub fn gr_source(program: *mut u8);
-        pub fn gr_value(val: *mut u8);
-        pub fn gr_wait() -> !;
-        pub fn gr_wake(waker_id_ptr: *const u8);
+pub use gcore::msg::{gas_available, id, reply_to, source, value, wait, wake};
```
shamilsan, line 23: The idiomatic way to call functions is `module::func()`, so we don't need to import functions themselves.

gstd/src/msg.rs
```diff
@@ -17,155 +17,87 @@
...
+
+pub fn reply_bytes(payload: &[u8], gas_limit: u64, value: u128) {
+    gcore::msg::reply(payload, gas_limit, value)
 }
 
-pub fn reply_push(payload: &[u8]) {
-    unsafe { sys::gr_reply_push(payload.as_ptr(), payload.len() as _) }
+pub fn reply_push<E: Encode>(payload: E) {
+    let bytes = payload.encode();
+    gcore::msg::reply_push(&bytes)
```
shamilsan, line 67: No need to return the unit type result.
```suggestion
    gcore::msg::reply_push(&bytes);
```

gstd/src/msg.rs
```diff
@@ -17,155 +17,87 @@
...
 }
 
-pub fn reply_to() -> MessageId {
-    let mut message_id = MessageId::default();
-    unsafe { sys::gr_reply_to(message_id.0.as_mut_ptr()) }
-    message_id
+pub fn reply_push_bytes(payload: &[u8]) {
+    gcore::msg::reply_push(payload)
```
shamilsan, line 71:
```suggestion
    gcore::msg::reply_push(payload);
```

tests/btree/src/lib.rs
```diff
@@ -41,15 +41,15 @@ mod wasm {
...
+        let reply = match msg::load() {
             Ok(request) => process(request),
             Err(e) => {
                 ext::debug(&format!("Error processing request: {:?}", e));
                 Reply::Error
             }
         };
 
-        msg::reply(&reply.encode()[..], 1000000, 0)
+        msg::reply(reply, 1000000, 0)
```
shamilsan, line 52:
```suggestion
        msg::reply(reply, 1_000_000, 0);
```

tests/node/src/lib.rs
```diff
@@ -76,15 +76,15 @@ mod wasm {
...
+        let reply = match msg::load() {
             Ok(request) => process(request),
             Err(e) => {
                 ext::debug(&format!("Error processing request: {:?}", e));
                 Reply::Failure
             }
         };
 
-        msg::reply(&reply.encode()[..], msg::gas_available(), 0)
+        msg::reply(reply, msg::gas_available(), 0)
```
shamilsan, line 87:
```suggestion
        msg::reply(reply, msg::gas_available(), 0);
```

breathx (commented):
In general, now I don't see a special reason for this division of the crate, but it looks good and the new features are very useful. Let's see what others think about it.

gstd/src/msg.rs
Anfid: They are imported separately to be accessible as `gstd::msg::func`, this line is a re-export.

shamilsan: Why not just `pub use gcore::msg;`?

shamilsan: > Why not just `pub use gcore::msg;`?

Ah, got it. Ok.

examples/meta/Cargo.toml
```diff
@@ -10,4 +10,4 @@ crate-type = ["cdylib"]
 
 [dependencies]
 gstd = { path = "../../gstd", features = ["debug"] }
-codec = { package = "parity-scale-codec", version = "2.0.0", default-features = false, features = ["derive"] }
```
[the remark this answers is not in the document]
Anfid: `parity-scale-codec` is imported without `std` feature in `gstd` crate. It could be explicitly enabled if needed.

gcore/src/general.rs
Anfid: I don't think giving access to internal field of `MessageHandle` is makes sense outside of `msg` module. If could be added later if that proves to be necessary.

gcore/README.md
```diff
@@ -0,0 +1,3 @@
+# gcore
+
+Standard library for Gear smart contracts.
```
NikVolf, line 3:
```suggestion
Core library for Gear smart contracts.
```

shamilsan (approved):

breathx (approved):

NikVolf (approved):
Awesome work

NikVolf:
As a follow up, I believe we can also make `codec::Input` to read input stream by parts 😏

NikVolf:
As another follow-up, we also need to split `galloc` from `gcore`, so that minimal application does not even need allocated pages, only initial.

PR #156, opened 2021-08-25
Asynchronous distributor test

NikVolf:
When program receive some amount of something (😜), it can distribute it between own sub-programs and leave everything that doesn't split well for it's own

When program is asked, how much amount of something(😜) it possesses, it collects the amount from all sub-programs and adds own value

All is done asynchronously

This can also be checked with some easy fuzz-testing like quickcheck


tests/distributor/src/lib.rs
```diff
@@ -0,0 +1,296 @@
...
+
+        async fn handle_receive(amount: u64) -> Reply {
+            let subnodes_count = Self::nodes().len() as u64;
+            if subnodes_count > 0 {
+                let distributed = amount / subnodes_count * subnodes_count;
+                let mut left_over = amount - distributed;
+
+                if distributed > 0 {
+                    for program in Program::nodes().iter() {
+                        if let Err(_) = program.do_send(distributed).await {
```
NikVolf, line 125: this is no longer safe to do since the node list can be modified by another program, thus refcell should be used and the `Request::Join` should fail if the node is processing sub-nodes at the moment

core/src/env.rs
```diff
@@ -156,6 +156,22 @@ impl<E: Ext> LaterExt<E> {
         Ok(res)
     }
 
+    /// Call fn with inner ext
+    pub fn with_fallable<R>(
```
Anfid, line 160: fallable -> fallible maybe?

NikVolf: yep done

PR #168, opened 2021-08-29
Split "galloc" crate

### LINKED ISSUE #157: Split "galloc" from gcore
`gcore` should not assume that there exist any allocator or the allocation is even possible

Some programs might want to work with fixed amount of static memory and never allocate

NikVolf:
Closes #157.
Now it is possible to write programs that don't use allocations


galloc/README.md
```diff
@@ -0,0 +1,3 @@
+# gcore
```
shamilsan, line 1: gcore -> galloc etc.

shamilsan (approved):

PR #169, opened 2021-08-30
Add a new `send_and_wait` syscall

### LINKED ISSUE #140: Send and wait syscall
`fn gr_send_wait(program_id: *const u8, payload: *const u8, payload_len: u32, value: *const u8, message_id_out: *mut u8)`

To send all available gas to the consequentive message, it is required

The signature is without a gas, it sends all gas left in the current message
it puts message in the `wait` state

shamilsan:
Fixes #140.

- Add a new `send_and_wait` function to `gcore`.

TODO: 
- [ ] Some test?
- [ ] Do we need update `gstd-asyc` to the new function?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,6 +77,7 @@ rpc-test:
 .PHONY: test
 test:
 	@cargo test --workspace
+	@cargo test -p gear-core-backend --no-default-features --features wasmi_backend
 
 .PHONY: test-release
 test-release:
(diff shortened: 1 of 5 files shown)


NikVolf:
What is the status?

shamilsan:
> What is the status?

It was suspended in favor of another task. I resume and complete it soon.

shamilsan:
I don't see the way to use it practically, so I close the PR.

PR #175, opened 2021-09-01
Configure rustfmt to be more useful

shamilsan:
- Add `rustfmt.toml` with specific settings.
- Use rustfmt v2.
- Add some useful settings.
- Update CI to conform.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -16,8 +16,17 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Install nightly tools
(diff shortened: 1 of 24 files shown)


Anfid:
What is the reason to use rustfmt v2? Also, does it work only on nightly compiler?

shamilsan:
> What is the reason to use rustfmt v2? Also, does it work only on nightly compiler?

v2 is quite stable already and we will be forced to use it anyway in the future. So let's do it now. Many famous projects already use rustfmt v2. But yes it requires a nightly compiler, unfortunately.

breathx (approved):

PR #193, opened 2021-09-04
Reduce block time and process queue at the end of a block

### LINKED ISSUE #173: Reduce block time and do queue processing at the end of a block
### Problem to Solve

Currently the system has rather high latency: a message on average spends in queue at least one block time before getting processed. 
Another downside is that the `process_queue`'s weight can be random depending on the amount of actual computation. That makes it impossible for a block producer to gauge how many normal extrinsics it should include in a block which can result in extended block times, and, eventually, even higher latency.

### Possible Solution

Reducing the block time from 6 s down to 1 s will reduce the latency.

Making `process_queue` run in the end of a block could also help.
Putting it in the `on_finalize` hook is not ideal: processing can be time consuming.
Instead we could consider two things:
- reserving a portion of the maximum block weight for normal users' extrinsics (say, 20-25%);
- placing the `process_queue` in the `on_idle` hook, adjusting the remaining gas allowance according the the actual weight taken by the normal extrinsics.

This should guarantee consistent block time and mitigate the risk for a block execution to "spill over" into the next block.

### Notes

_No response_

ekovalev:
Fixes #173.

- The expected block time is set to 1 sec;
- The call of `process_queue` is placed in the `on_idle` hook which is executed after all normal extrinsics (those with `DispatchClass::Normal`) have been processed, as opposed to having it as an inherent extrinsic which would have to be run before normal dispatchables;
- The normal extrinsics max weight is set to 25% of the total block weight which guarantees the `on_idle` hook will be executed in every block.

### DIFF
--- a/pallets/gear/src/lib.rs
+++ b/pallets/gear/src/lib.rs
@@ -35,7 +35,6 @@ mod tests;
 pub mod pallet {
     use super::*;
     use common::{self, IntermediateMessage, Message, Origin};
-    use frame_support::inherent::{InherentData, InherentIdentifier};
     use frame_support::{
         dispatch::DispatchResultWithPostInfo,
         pallet_prelude::*,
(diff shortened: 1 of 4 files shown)


NikVolf:
@ekovalev is it ok? seems fine to me

NikVolf (approved):

ekovalev:
> @ekovalev is it ok? seems fine to me

In any case we'd need to try it on the testnet to see if we do benefit from shorter block intervals. Ideally, create some tests to accurately measure the latency and the amount of messages we can cram into a 6 times smaller block

PR #213, opened 2021-09-08
Extend `wasm-proc` util functionality

breathx:
Due to the fact that soon we will constantly use the metadata of programs, it would be useful to improve the existing tools for interaction with them and their optimization.

The purpose of this PR is to make using the tool more comfortable both personally and in CI.

- `wasm-proc` now processes through multiple targets
- `wasm-proc` now contains flags:
  - `-m` or `--meta`: forces metadata optimization for targets. Optional.
  - `-o` or `--optimize`:  forces chain optimization for targets. Optional.
  - `-p` or `--path`: specifies paths for `.wasm` file targets . You may skip this flag, but it's args are required.

Note:
If none of `-o` and `-m` specified, tool does nothing, so it becomes useless. To avoid this situation, presence or absence at the same time of these two flags is equivalent:
```
./wasm-proc -mo arg1 arg2
# equivalents
./wasm-proc arg1 arg2
```

Also:
- Removed warnings `unused variable` from `pallets/gear/lib.rs` by marking them as intentional

@gear-tech/dev


utils/wasm-proc/src/main.rs
```diff
@@ -17,38 +17,36 @@
...
-
-    // Invoke optimizer for the chain
-    let mut binary_module = module.clone();
-    let binary_file_name = PathBuf::from(input).with_extension("opt.wasm");
-    utils::optimize(&mut binary_module, vec!["handle", "init"]).expect("Optimizer failed");
-    parity_wasm::serialize_to_file(binary_file_name.clone(), binary_module)
-        .expect("Serialization failed");
+/// Calls chain optimizer
+fn optimize(path: &str, mut binary_module: Module) {
+    println!("\n*** Processing chain optimization: {}", path);
```
NikVolf, line 28:
```suggestion
```

or use `log::debug!`

utils/wasm-proc/src/main.rs
```diff
@@ -17,38 +17,36 @@
...
+
+    if let Err(_) = utils::optimize(&mut binary_module, vec!["handle", "init"]) {
+        println!("Optimizer failed");
+    }
+
+    if let Err(e) = parity_wasm::serialize_to_file(binary_file_name.clone(), binary_module) {
+        println!("Serialization failed: {}", e);
+    }
 
     println!("Optimized wasm: {}", binary_file_name.to_string_lossy());
```
NikVolf, line 40:
```suggestion
```

utils/wasm-proc/src/main.rs
```diff
@@ -57,10 +55,61 @@ fn main() {
...
+    let app = App::new("wasm-proc").args(&[meta, opt, path]);
+
+    let matches = app.get_matches();
+
+    let wasm_files: Vec<&str> = matches
+        .values_of("path")
+        .expect("Path to wasm files is required")
+        .collect();
+
+    let o = matches.is_present("optimize");
```
NikVolf, line 100:
```suggestion
    let do_optimize = matches.is_present("optimize");
```

utils/wasm-proc/src/main.rs
```diff
@@ -57,10 +55,61 @@ fn main() {
...
+    let m = matches.is_present("meta");
```
NikVolf, line 101:
```suggestion
    let do_meta = matches.is_present("meta");
```

utils/wasm-proc/src/main.rs
```diff
@@ -57,10 +55,61 @@ fn main() {
...
+    for file in wasm_files {
+        if let Ok(module) = parity_wasm::deserialize_file(file) {
+            if o || !(o || m) {
+                optimize(file, module.clone());
+            }
+            if m || !(o || m) {
+                optimize_meta(file, module.clone());
+            }
+        } else {
+            println!("Failed to load wasm file: {}", file);
```
NikVolf, line 112: make `main()` return Result<(), ?>

if you print errors this way, exit code of the program would be 0

utils/wasm-proc/src/main.rs
```diff
@@ -17,38 +17,36 @@
...
     println!("Optimized wasm: {}", binary_file_name.to_string_lossy());
+}
 
-    // Invoke optimizer for the metadata
-    let mut metadata_module = module.clone();
-    let metadata_file_name = PathBuf::from(input).with_extension("meta.wasm");
-    utils::optimize(
+/// Calls metadata optimizer
+fn optimize_meta(path: &str, mut metadata_module: Module) {
+    println!("\n*** Processing metadata optimization: {}", path);
```
NikVolf, line 45:
```suggestion
```

NikVolf:
> If none of -o and -m specified, tool does nothing, so it becomes useless. To avoid this situation, presence or absence at the same time of these two flags is equivalent:

don't deliberately produce ambiguous behavior

invert flag to *skip* meta or optimisation like it is done in plasm-utils

so `wasm-proc --skip-meta file1.wasm`

and return error if it is used like that
`wasm-proc --skip-meta --skip-opt file1.wasm`

utils/wasm-proc/src/main.rs
breathx: Yes, nothing crushes, we have an error in console, and script continues work with other arguments. Do you want to break processing of the script if we ran into error?

NikVolf: Yes, like every sensible program does

utils/wasm-proc/src/main.rs
breathx: Changed according to new (skipping) flags logic

utils/wasm-proc/src/main.rs
breathx: Same

utils/wasm-proc/src/main.rs
breathx: Done

NikVolf (approved):

NikVolf:
PR description is a bit outdated now

NikVolf (approved):

PR #216, opened 2021-09-09
Guestbook program

EugenWay:
New program for check how works meta types
Tests https://github.com/gear-tech/gear-js-lib/pull/7

@reviewer-or-team


NikVolf (changes requested):
if you add an example, add a test for it so the example won't become obsolete/not working

NikVolf (approved):

PR #217, opened 2021-09-09
add rust update to dockerfile

pikodd:


pikodd:
* add rust update to dockerfile
* add dependency not to generate certs every time
* some minor updates to CD

NikVolf (approved):

PR #223, opened 2021-09-09
Add test net configuration for node

NikVolf:


ekovalev (approved):

NikVolf:
@ekovalev still waiting for valid json

ekovalev:
> @ekovalev still waiting for valid json

json seems to be valid - at least it matches the one that I have (except for the bootNodes section). Having launched a network locally with it I had no issue connecting and syncing a new node.
I guess I should just generate a brand new one. And the database on staging should be wiped clean.
In any case the public IP's of 3 out of 4 nodes have changed for some reason, therefore the json must be updated accordingly.

NikVolf:
Yep works now

PR #177, opened 2021-09-02
Provide the ability to use variable message payload in tests

breathx:
Fixes #96.

@gear-tech/dev

PR #244, opened 2021-09-15
Variable message payload in tests

### LINKED ISSUE #96: Send and expect custom payload in tests
**Problem to Solve**
Custom payload in tests.

**Possible Solution**
Parse metadata from `.meta.wasm` and encode/decode data provided from `.yaml` with SCALE.

**Notes**
Something like this:
```yaml
title: meta

programs:
  - id: 1
    path: examples/target/wasm32-unknown-unknown/release/demo_meta.wasm
    meta: examples/target/wasm32-unknown-unknown/release/demo_meta.meta.wasm
    init_message:
      kind: custom
      value:
        value: 1
        annotation: "hello"

fixtures:
  - title: meta
    messages:
      - destination: 1
        payload:
          kind: custom
          value:
            value: 1
            annotation: "hello"
    expected:
      - messages:
          - destination: 1
            payload:
              kind: custom
              value:
                old_value: 0
                new_value: 1
```

breathx:
Fixes #96.

Other solution in continuing of #177.

- New payload type `custom`
- Js module of encoding and decoding bytes
- Extend `meta` demo
- Fix `wasm-proc`. It now requires you to explicitly indicate -p flag.
- Updates on CI and scripts

NOTE:
Care with expected messages. You must mark only init output messages as `init: true` to get correct comparing output. If you miss it, expectation and real messages might be decoded in null and pass the check, cause null == null

@gear-tech/dev


gtest/src/js/decode.js
```diff
@@ -0,0 +1,31 @@
+const fs = require("fs");
+const { CreateType, getWasmMetadata } = require("@gear-js/api");
+const yargs = require("yargs");
+
+const argv = yargs
+    .option('path', {
+        alias: 'p',
+        default: "../../../examples/target/wasm32-unknown-unknown/release/demo_meta.meta.wasm",
```
NikVolf, line 8: that would be strange default, better return error if meta.wasm is not set

NikVolf, line 8: that would be strange default, better return error if meta.wasm is not set

breathx: Removed default option. Calling `node decode.js` without `-p` option occurs internal js error: 
`The "path" argument must be of type string or an instance of Buffer or URL. Received undefined`

breathx: Same with other options in `decode.js` and `encode.js`

NikVolf:
@shamilsan please review

examples/meta/src/lib.rs
```diff
@@ -1,46 +1,80 @@
...
+        unsafe {
+            for wallet in WALLETS.iter() {
+                if wallet.id.decimal == other.id.decimal {
+                    return Self {
+                        res: Some(wallet.clone()),
+                    };
+                };
+            }
+
+            Self { res: None }
```
shamilsan, line 59: I'd prefer a functional style, something like this (need to be verified!)

```suggestion
        unsafe {
            let res = WALLETS.iter().find(|w| w.id.decimal == other.id.decimal).map(Clone::clone);

            Self { res }
```

examples/meta/src/lib.rs
```diff
@@ -55,8 +89,41 @@ meta! {
     extra: Id, Person, Wallet
 }
 
+static mut WALLETS: Vec<Wallet> = Vec::new();
+
 #[no_mangle]
-pub unsafe extern "C" fn handle() {}
+pub unsafe extern "C" fn handle() {
+    let message_in: MessageIn = msg::load().unwrap();
+    let message_out = MessageOut::from(message_in);
```
shamilsan, line 97: The gold rule says "implement `From`, use `Into`". The reason is that `Into` has a blanket implementation for
any `T` that implements `From`, so regardless of whether a type explicitly implements `From` or `Into`, it implements `Into`.
```suggestion
    let message_out: MessageOut = message_in.into();
```

examples/meta/src/lib.rs
```diff
@@ -55,8 +89,41 @@ meta! {
...
+            hex: vec![2u8],
+        },
+        person: Person {
+            surname: "OtherName".into(),
+            name: "OtherSurname".into(),
+        },
+    });
+
+    let message_init_in: MessageInitIn = msg::load().unwrap();
+    let message_init_out = MessageInitOut::from(message_init_in);
```
shamilsan, line 126:
```suggestion
    let message_init_out: MessageInitOut = message_init_in.into();
```

gtest/src/check.rs
```diff
@@ -123,15 +125,52 @@ fn check_messages(
...
+        let mut messages: Vec<Message> = messages.into();
+
         expected_messages
-            .iter()
-            .zip(messages.iter())
+            .iter_mut()
+            .zip(messages.iter_mut())
             .enumerate()
             .for_each(|(position, (exp, msg))| {
+                let init = exp.init.unwrap_or(false);
```
shamilsan, line 136:
```suggestion
                let init = exp.init.unwrap_or(false);
                let meta_type = if init { MetaType::InitOutput } else { MetaType::Output };
```

gtest/src/check.rs
```diff
@@ -123,15 +125,52 @@ fn check_messages(
...
+                        PayloadVariant::Custom(_) => {
+                            for v in progs_n_paths.clone() {
+                                if ProgramId::from(v.1) == msg.source() {
+                                    let json = String::from_utf8(payload.to_bytes()).unwrap();
+
+                                    let bytes = if init {
+                                        js::get_bytes(&v.0, MetaType::InitOutput, json)
+                                    } else {
+                                        js::get_bytes(&v.0, MetaType::Output, json)
+                                    };
```
shamilsan, line 150: Avoid repeating code in loops!

```suggestion
                                    let bytes = js::get_bytes(&v.0, meta_type, json);
```

gtest/src/check.rs
```diff
@@ -123,15 +125,52 @@ fn check_messages(
...
+
+                                    let hex = hex::encode(bytes);
+
+                                    *payload = PayloadVariant::Utf8(if init {
+                                        js::get_json(&v.0, MetaType::InitOutput, hex)
+                                    } else {
+                                        js::get_json(&v.0, MetaType::Output, hex)
+                                    });
```
shamilsan, line 158:
```suggestion
                                    *payload = PayloadVariant::Utf8(js::get_json(&v.0, meta_type, hex));
```

gtest/src/check.rs
```diff
@@ -123,15 +125,52 @@ fn check_messages(
...
+
+                                    let hex = hex::encode(msg.payload.clone());
+                                    msg.payload = if init {
+                                        js::get_json(&v.0, MetaType::InitOutput, hex)
+                                    } else {
+                                        js::get_json(&v.0, MetaType::Output, hex)
+                                    }
```
shamilsan, line 165:
```suggestion
                                    msg.payload = js::get_json(&v.0, meta_type, hex)
```

shamilsan (changes requested):

gtest/src/js/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+            MetaType::InitOutput => "init_output",
+            MetaType::Input => "input",
+            MetaType::Output => "output",
+        }
+        .into()
+    }
+}
+
+pub fn gear_path() -> String {
+    let pwd = Command::new("pwd")
```
shamilsan, line 44: Why to not use the cross-platform `std::env::current_dir` instead?

gtest/src/js/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+        .output()
+        .expect("Unable to call pwd command");
+
+    let path = String::from_utf8(pwd.stdout).expect("Unable to parse pwd output bytes");
+    let path_parts: Vec<String> = path.split("/").map(|v| v.replace("\n", "")).collect();
```
shamilsan, line 49: `PathBuf` has cross-platform `components` function for similar task. This code will not work on Windows.

gtest/src/js/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+    let mut wasm_path = gear_path();
+    wasm_path.push_str("/");
+    wasm_path.push_str(path);
+
+    if !wasm_path.ends_with(".meta.wasm") {
+        wasm_path = wasm_path.replace(".wasm", ".meta.wasm");
+    }
+
+    if !Path::new(&wasm_path).exists() {
+        panic!("Could not find file {}", wasm_path);
```
shamilsan, line 68: Panic when there is no file? Bad approach as it is the user space error rather than programmer space.

gtest/src/js/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+        .output()
+        .expect("Unable to call pwd command");
+
+    let path = String::from_utf8(pwd.stdout).expect("Unable to parse pwd output bytes");
+    let path_parts: Vec<String> = path.split("/").map(|v| v.replace("\n", "")).collect();
+
+    if let Some(index) = path_parts.iter().rposition(|r| r == "gear") {
+        path_parts[..index + 1].join("/")
+    } else {
+        panic!("Gear root directory not found")
```
shamilsan, line 54: `Result` would be better than `panic`

gtest/src/js/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    if !wasm_path.ends_with(".meta.wasm") {
+        wasm_path = wasm_path.replace(".wasm", ".meta.wasm");
+    }
+
+    if !Path::new(&wasm_path).exists() {
+        panic!("Could not find file {}", wasm_path);
+    }
+
+    let mut script_path = gear_path();
```
shamilsan, line 71: Better to call expensive `gear_path()` once, then clone the result.

gtest/src/js/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+    script_path.push_str("/gtest/src/js/encode.js");
+
+    let output = Command::new("node")
+        .arg(script_path)
+        .args(&["-p", &wasm_path, "-t", &meta_type.to_string(), "-j", &json])
+        .output()
+        .expect("Unable to call node.js process");
+
+    debug!(
```
shamilsan, line 80: Use the same convection as for the functions.
```suggestion
    log::debug!(
```

gtest/src/js/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+    );
+    debug!(
+        "js get_bytes stderr:{}",
+        String::from_utf8(output.stderr).unwrap()
+    );
+
+    output.stdout
+}
+
+pub fn get_json(path: &str, meta_type: MetaType, hex: String) -> String {
```
shamilsan, line 92: This function is similar to the `get_bytes()` one. Better to generalize to avoid repeating yourself.

shamilsan (changes requested):

gtest/src/check.rs
```diff
@@ -123,15 +125,58 @@ fn check_messages(
...
+                                .find(|v| ProgramId::from(v.1) == msg.source())
+                                .map(|v| {
+                                    let path: String = v.0.replace(".wasm", ".meta.wasm");
+
+                                    let json = MetaData::Json(
+                                        String::from_utf8(payload.to_bytes()).unwrap(),
+                                    );
+
+                                    let bytes = json
+                                        .convert(&path, meta_type.clone())
```
shamilsan, line 158: We could either derive `Copy` (it is ok for primitive types) for `MetaType` or just pass a reference to avoid cloning.
```suggestion
                                        .convert(&path, &meta_type)
```

breathx: I would prefer use references

shamilsan (approved):

PR #260, opened 2021-09-17
Provide `exec` module for non-messaging api

peter-cmw:


gstd/src/exec.rs
```diff
@@ -0,0 +1,19 @@
+// This file is part of Gear.
```
breathx, line 1: Useless file

gstd/src/lib.rs
```diff
@@ -24,6 +24,7 @@
 extern crate galloc;
 
 pub mod bail;
+pub mod exec;
```
breathx, line 27:
```suggestion
pub use gcore::exec;
```

gcore/src/msg.rs
```diff
@@ -21,7 +21,6 @@ use crate::{MessageId, ProgramId};
 
 mod sys {
```
breathx, line 22: Following the idea of this pull request, wouldn't it be reasonable also to move `gr_read()` and `gr_size()` to `gcore/exec` mod?

breathx (changes requested):

NikVolf (approved):

gstd/src/exec.rs
shamilsan: Why?

shamilsan: Ah, I see. We can just reimport `gcore::exec` indeed.

gcore/src/msg.rs
shamilsan: In my opinion, both `gr_read` and `gr_size` are related to the message, so there is no reason to move them.

gstd/src/exec.rs
shamilsan: Fixed.

gstd/src/lib.rs
shamilsan: Done.

breathx (approved):

PR #265, opened 2021-09-18
Enable runtime 'debug' mode

### LINKED ISSUE #246: Runtime "debug" mode
In this particular mode activated via storage key, after each handled message, the node should put additional logs:

- full state of the message queue
- all programs and their memory

ekovalev:
Fixes #246.

- added a storage value that holds the flag indicating whether the runtime "debug" mode is on;
- added an extrinsic callable by the root origin that can modify this storage value thus switching the debug mode on and off in runtime (without a need for restarting a node or storage migration);
- the message queue and programs storage contents are wrapped as runtime events.

The following information is currently being logged:

- [x] full state of the message queue
- [x] programs storage
- [x] programs memory

### DIFF
--- a/pallets/gear/src/lib.rs
+++ b/pallets/gear/src/lib.rs
@@ -38,6 +38,7 @@ pub mod pallet {
     use frame_support::{
         dispatch::DispatchResultWithPostInfo,
         pallet_prelude::*,
+        storage::PrefixIterator,
         traits::{BalanceStatus, Currency, ExistenceRequirement, ReservableCurrency},
         weights::{IdentityFee, WeightToFeePolynomial},
     };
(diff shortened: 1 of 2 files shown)


NikVolf:
I thought about runtime events, this console output might be challenging to read in js tests

ekovalev:
> I thought about runtime events, this console output might be challenging to read in js tests

Ok, fair enough

NikVolf:
Merging master to at least fix tests ^

pallets/gear/src/lib.rs
```diff
@@ -87,6 +88,12 @@ pub mod pallet {
         /// Some number of messages processed.
         // TODO: will be replaced by more comprehensive stats
         MessagesDequeued(u32),
+        /// Debug mode has been turned on or off
+        DebugMode(bool),
+        /// A snapshot of the program storage ('debug mode' only)
```
NikVolf, line 93: All debug data should be merged in one event to make things easier on js side

NikVolf:
Nice 👍 just one thing above

@osipov-mit please also take a look

PR #267, opened 2021-09-18
Move on back to rustfmt v1

shamilsan:
- Revert `rustfmt` back to v1.
- Now we can use cargo fmt with the stable compiler as earlier.


shamilsan:
Cherry-picked in https://github.com/gear-tech/gear/pull/266.

PR #300, opened 2021-09-28
Remove useless `send()` sys call

### LINKED ISSUE #295: Remove unused code from `gcore` and `gstd`
### File Location(s)

https://github.com/gear-tech/gear/blob/master/gcore/src/lib.rs
https://github.com/gear-tech/gear/blob/master/gcore/src/msg.rs
https://github.com/gear-tech/gear/blob/master/gstd/src/msg.rs

### Proposal

- Remove the `prelude` module from `gcore` as useless.
- Remove `msg::send_with_value` as it seems to be excessive and inconsistent compared to other similar functions (e.g. `msg::reply*` etc.).
- Remove `gr_send` syscall as it can be replaced by `gr_send_init/gr_send_push/gr_send_commit`.

breathx:
Fixes #295.

TODO:
- [ ] Fix wat tests in `gear-core-runner`

@gear-tech/dev

### DIFF
--- a/core-backend/src/funcs.rs
+++ b/core-backend/src/funcs.rs
@@ -136,32 +136,6 @@ pub(crate) fn reply_to<E: Ext>(ext: LaterExt<E>) -> impl Fn(i32) -> Result<(), &
     }
 }
 
-pub(crate) fn send<E: Ext>(
-    ext: LaterExt<E>,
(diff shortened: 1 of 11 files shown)


breathx:
Deleting the syscall would increase gas rejection. Would be refactored in core api in other PR.

PR #302, opened 2021-09-28
Refactor send sys call processing under the hood

### LINKED ISSUE #295: Remove unused code from `gcore` and `gstd`

breathx:
Fixes #295.

@gear-tech/dev

### DIFF
--- a/gstd/src/msg.rs
+++ b/gstd/src/msg.rs
@@ -31,8 +31,8 @@ impl MessageHandle {
         send_init()
     }
 
-    pub fn push<E: Encode>(&self, payload: E) {
-        gcore::msg::send_push(&self.0, &payload.encode());
+    pub fn push(&self, payload: &[u8]) {
+        gcore::msg::send_push(&self.0, payload);
     }
 
     pub fn commit(self, program: ProgramId, gas_limit: u64, value: u128) -> MessageId {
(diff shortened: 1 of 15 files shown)


gstd/src/msg.rs
```diff
@@ -92,8 +86,7 @@ pub fn send_with_value<E: Encode>(
     gas_limit: u64,
     value: u128,
 ) -> MessageId {
-    let bytes = payload.encode();
-    gcore::msg::send_with_value(program, &bytes, gas_limit, value)
+    send_bytes_with_value(program, &payload.encode(), gas_limit, value)
```
shamilsan, line 89: Should we rename it to `send_bytes`? 
```suggestion
    send_bytes(program, &payload.encode(), gas_limit, value)
```

breathx: We already have `send_bytes()` function, which doesn't get value as an argument. These **_with_value** functions will disappear in further api versions.

shamilsan (approved):

PR #303, opened 2021-09-28
Adding docs to gcore and minor refactoring of syscalls

peter-cmw:
Added initial documentation draft to gcore modules
Remove send methods that do not send value
Return Message Id  from reply method

### DIFF
--- a/gcore/src/exec.rs
+++ b/gcore/src/exec.rs
@@ -18,7 +18,7 @@
 
 //! Sys calls related to program execution flow.
 //!
-//! Provides api for low-level async implementation (probably gas should be other mod)
+//! Provides api for low-level async implementation
 
 use crate::MessageId;
 
--- a/gcore/src/exec.rs
+++ b/gcore/src/exec.rs
@@ -16,6 +16,10 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+//! Sys calls related to smart contract execution.
+//!
+//! Provides api for low-level async implementation (probably gas should be other mod)
+
 use crate::MessageId;
 
 mod sys {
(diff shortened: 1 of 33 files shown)


gcore/src/exec.rs
```diff
@@ -16,6 +16,10 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+//! Sys calls related to smart contract execution.
```
NikVolf, line 19:
```suggestion
//! Sys calls related to program execution flow.
```

NikVolf:
We should avoid notion of "smart contracts"

Those are "programs"

(there are multiple issues about that in the pr)

gcore/src/exec.rs
```diff
@@ -16,6 +16,10 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+//! Sys calls related to program execution flow.
+//!
+//! Provides api for low-level async implementation (probably gas should be other mod)
```
NikVolf, line 21:
```suggestion
//! Provides api for low-level async implementation
```

NikVolf (approved):

gcore/src/exec.rs
```diff
@@ -26,14 +30,69 @@ mod sys {
...
+///
+/// Each execution of message processing consumes gas, both on instructions and memory allocations.
+/// gas_available() returns value of the gas available for spend during current execution.
+///
+/// # Examples
+///
+/// ```
+/// use gcore::{msg, exec};
+///
+/// // Perform work while gas_available is more then 1000
```
shamilsan, line 43:
```suggestion
/// // Perform work while gas_available is more than 1000
```

shamilsan (changes requested):
Please don't merge now, I'll fix typos and merge it.

Issue #293, opened 2021-09-27
Add tests for async programs


shamilsan:
- [x] Test `send_push` interruption and resuming after calling `send_and_wait` (expected to fail at the moment)
- [x] Test whether the sync code in async handler doesn't execute more than once (may fail at the moment)
- [x] Test whether two `reply`s (one before  `send_and_wait` and one after) will cause an error

PR #304, opened 2021-09-28
Test which shows that async program duplicates sync code at the moment

breathx:
Part of #293.

@gear-tech/dev


examples/sync_duplicate/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+static mut COUNTER: usize = 0;
+
+fn increase() {
+    unsafe {
+        COUNTER += 1;
+    }
+}
+
+fn get() -> usize {
+    unsafe { COUNTER.clone() }
```
shamilsan, line 15: Why clone `Copy` type?

NikVolf:
We need to employ special runner (executor) to mitigate this

examples/sync_duplicate/src/lib.rs
breathx: Removed

NikVolf:
The problem is that we manually poll top level future in the `gstd_async::block_on`

shamilsan:
Raised an issue: https://github.com/gear-tech/gear/issues/307

shamilsan (approved):

PR #308, opened 2021-09-29
Make reply work by parts

### LINKED ISSUE #293: Add tests for async programs

### LINKED ISSUE #294: Add a `msg::reply_commit` syscall to `gstd`
Currently, we can call `msg::reply_push` after `msg::reply` to add a payload chunk to the reply. But we can get an inconsistent reply if some trap occurs before the final  `msg::reply_push`.

We need to finalize the reply after calling `msg::reply_commit` when sending the reply by parts. Alternatively, we can use `msg::reply` for sending the whole reply once.

We should return a trap when trying to mix `msg::reply` with `msg::reply_push/reply_commit`.

### LINKED ISSUE #319: `gstd::msg::reply` doesn't reduce gas counter
### Problem

`msg::reply` doesn't reduce gas counter when `msg::send` does.

https://github.com/gear-tech/gear/blob/master/core-runner/src/ext.rs#L95

### Steps

See `examples/async`: `ping` program replies with greater `gas_limit` than it has received from `async`.

### Possible Solution

Call `gas_counter.reduce` in `reply`/`reply_commit`.

breathx:
Fixes #294 and fixes #319. Also closes #293.

@gear-tech/dev


NikVolf:
We should make this work between `gr-wait` / `gr-wake` -s

core-backend/src/funcs.rs
```diff
@@ -98,17 +98,39 @@ pub(crate) fn read<E: Ext>(ext: LaterExt<E>) -> impl Fn(i32, i32, i32) -> Result
...
+            let message_id =
+                ext.reply_commit(ReplyPacket::new(0, vec![].into(), gas_limit as _, value))?;
+            ext.set_mem(message_id_ptr as isize as _, message_id.as_slice());
+            Ok(())
+        })?;
+        result.map_err(|_| "Trapping: unable to send message")
+    }
+}
+
 pub(crate) fn reply_push<E: Ext>(
```
Anfid, line 134: Maybe a message handle should be added as a parameter, obtained from `reply_init` function? This would make API consistent with `send_push`

core-backend/src/wasmi/env.rs
```diff
@@ -205,8 +216,11 @@ impl<E: Ext + 'static> ModuleImportResolver for Environment<E> {
             }
             "gr_reply" => func_instance!(Reply, ValueType::I32,
                 ValueType::I32,
-                ValueType::I64,
+                ValueType::I64,ValueType::I32,
```
Anfid, line 219: Bad code formatting

Anfid: Note that rustfmt only works with a small set of widely used macros, otherwise it leaves macro contents unchanged.

core-backend/src/funcs.rs
shamilsan: No need I guess as the reply is the only one.

core/src/message.rs
```diff
@@ -653,14 +647,42 @@ impl<IG: MessageIdGenerator + 'static> MessageContext<IG> {
...
+            (payload, msg) => {
+                let mut reply = self.id_generator.borrow_mut().produce_reply(packet);
+
+                let packet_payload = reply.payload.0.clone();
+                reply.payload.0.clear();
+                reply
+                    .payload
+                    .0
+                    .extend_from_slice(&payload.as_ref().unwrap_or(&Payload::default()).0);
+                reply.payload.0.extend_from_slice(&packet_payload);
```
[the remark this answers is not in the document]
breathx: Done.

NikVolf:
> > We should make this work between `gr-wait` / `gr-wake` -s
> 
> Not only this, but also sending in parts. But I suppose, it requires separated PR.

you can stub it with panic at least for now

core/src/message.rs
```diff
@@ -662,14 +647,39 @@ impl<IG: MessageIdGenerator + 'static> MessageContext<IG> {
...
+            (payload, msg) => {
+                let mut reply = self.id_generator.borrow_mut().produce_reply(packet);
+
+                reply
+                    .payload
+                    .0
+                    .splice(0..0, payload.take().unwrap_or_default().0);
+
+                *msg = Some(reply);
+                *payload = None;
```
shamilsan, line 665: No need after `take()`.
```suggestion
```

core/src/message.rs
```diff
@@ -707,13 +717,10 @@ impl<IG: MessageIdGenerator + 'static> MessageContext<IG> {
...
-                outgoing.payload.0.clear();
                 outgoing
                     .payload
                     .0
-                    .extend_from_slice(&payload.as_ref().unwrap().0);
-                outgoing.payload.0.extend_from_slice(&packet_payload);
+                    .splice(0..0, payload.take().unwrap_or_default().0);
 
                 *msg = Some(outgoing);
                 *payload = None;
```
shamilsan, line 726:
```suggestion
```

core/src/message.rs
```diff
@@ -123,6 +123,8 @@ pub enum Error {
     OutOfBounds,
     /// An attempt to push a payload into reply that was not set
     NoReplyFound,
+    /// An attempt to interrupt execution with `wait(..)` while some messages wasn't complited
```
shamilsan, line 126:
```suggestion
    /// An attempt to interrupt execution with `wait(..)` while some messages weren't completed
```

gcore/src/msg.rs
```diff
@@ -151,6 +152,43 @@ pub fn reply(payload: &[u8], gas_limit: u64, value: u128) -> MessageId {
     }
 }
 
+/// Finalizes current reply message.
```
shamilsan, line 155: We use imperative tense despite it is not too Rust-doc-like.
```suggestion
/// Finalize current reply message.
```

shamilsan (approved):
Minor typo.

shamilsan (approved):
Excellent!
But I don't see the checks result after the last changes... :eyes:

breathx:
> Excellent!
> But I don't see the checks result after the last changes... 👀

I suppose, it's Github temporary troubles

NikVolf (approved):

PR #309, opened 2021-09-29
Make `guestbook` demo fully working and use it as test

breathx:
This demo based on enums and generic meta I/O types so it is useful to test changes on it.
Also it wasn't working correctly before (`gear-js` lib already updated, see https://github.com/gear-tech/gear-js/pull/51). Now it also uses `custom` gtest payload.

@gear-tech/dev


ukint-vs (approved):

EugenWay (approved):
Good job!

PR #316, opened 2021-10-01
Disambiguate features in `gstd`

breathx:
- Removed `std` feature
- Added support for `String` in program's metadata and custom payloads in `gtest`
- Sending payload function now requires generics, which implements `AsRef<[u8]>`. This allows us to don't convert something into bytes by ourselves. Previous variant is still available.
- Fixed bug with `bail!` macro, which appeared after splitting into `gcore` and `gstd` libs. He was throwing compile error due to wrong function execution.
- `gstd::prelude` a bit expanded 

@gear-tech/dev


NikVolf (changes requested):
This is totally against design of gstd (which should be an analogue of rust std)

gstd/src/prelude.rs
```diff
@@ -16,14 +16,32 @@
...
+    string::{String, ToString},
+    vec,
+    vec::Vec,
+};
+
+#[cfg(feature = "std")]
+pub use std::{
+    borrow::ToOwned,
+    boxed::Box,
+    collections::{BTreeMap, VecDeque},
```
shamilsan: `std` just reimports these entities from `alloc`. Should we double the code?

NikVolf (approved):

gstd/src/prelude.rs
breathx: `std` feature already removed

Anfid:
Great changes, but it would force std to all users who use gstd only with alloc. I would prefer if std could still be behind a feature, even if we enable it by default.

ukint-vs (approved):

PR #367, opened 2021-10-30
Bump spec version

ukint-vs:
@gear-tech/dev


shamilsan (approved):

ukint-vs:
I will merge it after #363

NikVolf:
I don't think it's needed because #363 does not contain migration code

Issue #12, opened 2021-02-23
Use better allocator


NikVolf:
Allocator should utilize static area to store configuration and free even first page after no more memory in heap required.

PR #369, opened 2021-11-02
Switch to new allocator

### LINKED ISSUE #317: gtest specs fail when we use allocator with enabled checks
### Problem

First I found this problem when run gtest for compiled with *debug_assertions*  wasm tests  -  some of them failed on execution.
```
$ sh scripts/test.sh async 

Fixture async-await (step: 1): Ok
DEBUG: 1 total message(s) stored: 
Fixture async-await (step: 2): Messages check [expected: 1, actual: 2]
DEBUG: 1 total message(s) stored: 
Fixture async-await (step: 3): Messages check [expected: 1, actual: 2]
DEBUG: 1 total message(s) stored: 
Fixture async-await (step: 4): Messages check [expected: 1, actual: 2]
DEBUG: 1 total message(s) stored: 
Fixture async-await (step: 5): Messages check [expected: 1, actual: 2]
DEBUG: 1 total message(s) stored: 
Fixture async-await (step: 6): Messages check [expected: 1, actual: 2]
DEBUG: 1 total message(s) stored: 
Fixture async-await (step: 7): Messages check [expected: 1, actual: 2]
DEBUG: 1 total message(s) stored: 
DEBUG: 1 total message(s) stored: 
Fixture async-await (step: 8): Messages check [expected: 1, actual: 3]
DEBUG: 1 total message(s) stored: 
DEBUG: 1 total message(s) stored: 
Fixture async-await (step: 9): Messages check [expected: 1, actual: 3]
DEBUG: 1 total message(s) stored: 
DEBUG: 1 total message(s) stored: 
Fixture async-await (step: 10): Log check [expected: 1, actual: 0]
Error: 9 tests failed

```

Investigating problem I found that *debug_assertion* is not a reason and something fails tests in checks code.
So I tried to localize problem - remove all unneeded checks and unneeded code. 
Only one check function is left and inside only one small loop is left:
```
check_malloc_state(&mut self) {
        if !DL_CHECKS { return; }
        for i in 0..11 {
            let head_chunk : *mut Chunk = self.smallbin_at(i as u32);
        }
}

#[inline(never)]
 unsafe fn smallbin_at(&mut self, idx: u32) -> *mut Chunk {
        let idx = (idx * 2) as usize;
        dlassert!(idx < self.smallbins.len());

        let smallbins_ptr = &self.smallbins as *const *mut Chunk;
        let idx_ptr = smallbins_ptr.offset(idx as isize ) as *mut Chunk;
        return idx_ptr;
}
```
Here we just take few times mutable pointer and do nothing else. This function is called sometimes in allocator.

So, if DL_CHECKS is true then gtest is failed,  else all goes right.
We can also tries to make function smallbin_at as inline  -  in that case  tests goes well.
If we change iterations number in  loop, for example 0..10, then tests also goes well.

I noticed that if iterations number >= 11 then we fails in other case goes well:  20..25 and 20..30  - goes well;   20..31 - fails

### Steps

Linux kernel:  5.4.0-86-generic   OS:  20.04.3 LTS (Focal Fossa)   Proc: x86_64

1. git fetch && git checkout gsobol_alloc_fail
2. sh scripts/test.sh async   // this must fail 
3. remove in galloc/Cargo.toml   feature "gsobol_check" which is for dlallocator
4. sh scripts/test.sh async   // this runs well 

### Possible Solution

I think that problem may be in wasmtime compiler   or   when we compile from rust to wasm code.
No other ideas. 

### Notes

_No response_

### Relevant Log Output

_No response_

grishasobol:
Fixes #317 #12 .

- Change dlmalloc source to: https://github.com/gear-tech/dlmalloc-rs . It's same allocator in general moments, but now it call system to free page as soon as possible. Also a lot of unneeded functionality is removed and code refactoring is made.  
- Change gas charging in page allocations. Now gas is charged for each new page allocation and for each wasm32 grow. If new page is free before being stored to storage, then we return gas for allocation back (not for grow).
- Add ability to add gas.
- Change gas limits in many tests to be able run them with enabled allocator checks.
- Change tests which checks gas for allocations.

@NikVolf @ukint-vs @shamilsan @Anfid  please review

### DIFF
--- a/core-runner/src/ext.rs
+++ b/core-runner/src/ext.rs
@@ -167,7 +167,8 @@ impl EnvExt for Ext {
 
         // Returns back gas for alloced page if it's new
         if !self.memory_context.is_init_page(ptr)
-            && self.gas_counter.refund(self.alloc_cost) != ChargeResult::Enough {
+            && self.gas_counter.refund(self.alloc_cost) != ChargeResult::Enough
+        {
             return self.return_with_tracing(Err("Gas limit - add too many gas"));
         }
 
--- a/core-runner/src/ext.rs
+++ b/core-runner/src/ext.rs
@@ -73,7 +73,7 @@ impl EnvExt for Ext {
         let grow_pages_num = new_mem_size - old_mem_size;
         let mut gas_to_return_back = self.alloc_cost * (pages_num.raw() - grow_pages_num) as u64;
 
-        // Returns back greedly used gas for new allocations
+        // Returns back greedly used gas for allocations
         let first_page = result.unwrap().raw();
         let last_page = first_page + pages_num.raw() - 1;
         let mut new_alloced_pages_num = 0;
--- a/core-runner/src/ext.rs
+++ b/core-runner/src/ext.rs
@@ -53,18 +53,40 @@ impl Ext {
 }
 
 impl EnvExt for Ext {
-    fn alloc(&mut self, pages: PageNumber) -> Result<PageNumber, &'static str> {
-        let res = self.memory_context.pre_alloc(pages.raw());
-        if res.is_ok() {
-            self.gas(res.unwrap() * self.alloc_cost as u32)?;
-        } else {
-            self.gas(pages.raw() * self.alloc_cost as u32)?;
-        }
+    fn alloc(&mut self, pages_num: PageNumber) -> Result<PageNumber, &'static str> {
+        // Greedly sub gas for grow and allocations
+        self.gas( 2 * pages_num.raw() * self.alloc_cost as u32)?;
+
+        let old_mem_size = self.memory_context.memory().size().raw();
 
         let result = self
             .memory_context
-            .alloc(pages)
+            .alloc(pages_num)
             .map_err(|_e| "Allocation error");
+
+        if result.is_err() {
+            return self.return_with_tracing(result);
+        }
+
+        // Returns back greedly used gas for grow
+        let new_mem_size = self.memory_context.memory().size().raw();
+        let grow_pages_num = new_mem_size - old_mem_size;
+        let mut gas_to_return_back = self.alloc_cost * (pages_num.raw() - grow_pages_num) as u64;
+
+        // Returns back greedly used gas for new allocations
+        let first_page = result.unwrap().raw();
+        let last_page = first_page + pages_num.raw() - 1;
+        let mut new_alloced_pages_num = 0;
+        for page in first_page..=last_page {
+            if !self.memory_context.is_init_page(page.into()) {
+                new_alloced_pages_num += 1;
+            }
+        }
+        gas_to_return_back += self.alloc_cost * (pages_num.raw() - new_alloced_pages_num) as u64;
+
+        if self.gas_counter.add_gas(gas_to_return_back) != ChargeResult::Enough {
+            return self.return_with_tracing(Err("Gas limit - add too many gas"));
+        }
 
         self.return_with_tracing(result)
     }
@@ -142,6 +164,13 @@ impl EnvExt for Ext {
 
     fn free(&mut self, ptr: PageNumber) -> Result<(), &'static str> {
         let result = self.memory_context.free(ptr).map_err(|_e| "Free error");
+
+        // Returns back gas for alloced page if it's new
+        if !self.memory_context.is_init_page(ptr) {
+            if self.gas_counter.add_gas(self.alloc_cost) != ChargeResult::Enough {
+                return self.return_with_tracing(Err("Gas limit - add too many gas"));
+            }
+        }
 
         self.return_with_tracing(result)
     }
--- a/core-runner/src/ext.rs
+++ b/core-runner/src/ext.rs
@@ -84,7 +84,7 @@ impl EnvExt for Ext {
         }
         gas_to_return_back += self.alloc_cost * (pages_num.raw() - new_alloced_pages_num) as u64;
 
-        if self.gas_counter.add_gas(gas_to_return_back) != ChargeResult::Enough {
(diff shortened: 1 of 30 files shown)


core/src/gas.rs
```diff
@@ -84,6 +88,10 @@ impl GasCounter for GasCounterUnlimited {
         ChargeResult::Enough
     }
 
+    fn add_gas(&mut self, _amount: u64) -> ChargeResult {
```
NikVolf, line 91: better rename to "refund"

grishasobol: Renamed

core-runner/src/ext.rs
```diff
@@ -53,14 +53,41 @@ impl Ext {
 }
 
 impl EnvExt for Ext {
-    fn alloc(&mut self, pages: PageNumber) -> Result<PageNumber, &'static str> {
-        self.gas(pages.raw() * self.alloc_cost as u32)?;
+    fn alloc(&mut self, pages_num: PageNumber) -> Result<PageNumber, &'static str> {
+        // Greedly sub gas for grow and allocations
+        self.gas(2 * pages_num.raw() * self.alloc_cost as u32)?;
```
NikVolf, line 58: there should be distinguished coefficients for grow and allocation cost

ukint-vs (approved):
Looks ok now. Great work!

core-runner/src/ext.rs
```diff
@@ -53,14 +54,43 @@ impl Ext {
...
             .map_err(|_e| "Allocation error");
 
+        if result.is_err() {
+            return self.return_with_tracing(result);
+        }
+
+        // Returns back greedly used gas for grow
+        let new_mem_size = self.memory_context.memory().size().raw();
+        let grow_pages_num = new_mem_size - old_mem_size;
+        let mut gas_to_return_back = self.mem_grow_cost * (pages_num.raw() - grow_pages_num) as u64;
```
shamilsan, line 77: I'd implement `Add` trait etc for `PageNumber` to avoid these multiple `raw()` and vice versa but we can do it as a refactoring later.

core/src/memory.rs
```diff
@@ -176,12 +179,19 @@ impl MemoryContext {
...
             memory,
+            init_allocations: allocations.clone(),
             allocations,
             max_pages,
             static_pages,
         }
     }
 
+    /// Returns whether page is initial,
+    /// means page has been already in storage.
```
shamilsan, line 190:
```suggestion
    /// Return `true` if the page is the initial page, 
    /// it means that the page was already in the storage.
```

core-runner/src/ext.rs
ukint-vs: `Add` and `Sub` traits for `PageNumber` have already been implemented.

grishasobol: Yes, I suppose it's not good idea to implement it in this review. I may brake something=)

NikVolf:
Once merged, this can be put live on test network via runtime upgrade ✌️

core-runner/src/ext.rs
```diff
@@ -53,14 +54,43 @@ impl Ext {
 }
 
 impl EnvExt for Ext {
-    fn alloc(&mut self, pages: PageNumber) -> Result<PageNumber, &'static str> {
-        self.gas(pages.raw() * self.alloc_cost as u32)?;
+    fn alloc(&mut self, pages_num: PageNumber) -> Result<PageNumber, &'static str> {
+        // Greedly sub gas for allocations
+        self.gas(pages_num.raw() * self.alloc_cost as u32)?;
+        // Greedly sub gas for grow
```
Anfid, line 60: Is greedy subtraction necessary here? It may lead to unintuitive cases where program is given just enough gas to complete, but fails if it does not have enough gas for greedy subtraction.

also minor: `Greedly` -> `Greedily`
or
`Greedy gas subtraction for ...`

core-runner/src/runner.rs
```diff
@@ -1256,7 +1283,7 @@ mod tests {
 
         assert_eq!(
             result.gas_spent[0].1,
-            runner.alloc_cost() * 1 + runner.load_page_cost() * 1 + 3000
+            runner.alloc_cost() * 2 + runner.load_page_cost() * 1 + 3000
```
Anfid, line 1286: If I understand changes correctly, shouldn't it be `runner.alloc_cost() + runner.mem_grow_cost()`? Test may fail if default values are changed.

grishasobol: A yep. Forget about this test

core-runner/src/ext.rs
grishasobol: It's not necessary, but it's the least difficult way. We cannot charge gas after allocation, because we must charge before action. So we can also charge for each grow and for each new allocated page separately, but this solution will be much bigger.

tests/gas_limit/src/lib.rs
```diff
@@ -175,47 +172,42 @@ mod tests {
...
 
-        // Allocation cost is withdrawn once for initial page and twice for first allocated page
-        // because of how allocator works. Consecutive allocations only charge once per page.
-        // Probably will be fixed by https://github.com/gear-tech/gear/issues/12
-        let gas_limit = baseline.gas_spent + 3 * config.alloc_cost;
+        let pages_num = 1;
+        let allocation_size = pages_num * PAGE_SIZE as u32;
 
+        // First run is to calculate gas spent without pages allocation cost
+        // (for code execution and e.t.c)
```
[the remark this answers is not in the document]
grishasobol: Oh, interesting

grishasobol:
@NikVolf  @Anfid  @shamilsan  should I wait all approves or can merge now?

NikVolf (approved):

PR #374, opened 2021-11-04
[wip] Offchain worker for wait list tracking

### LINKED ISSUE #373: Charge rent for keeping messages in the wait list
Programs can get into the "waiting" state while processing incoming messages. Such messages end up in the wait list until "awaken" by another message. Keeping messages in the wait list should not be free of charge for programs and subject to "rent" payment. It is the programs' owners responsibility to maintain sufficient balance on programs' accounts. Failure to pay rent can result into a program being evicted from the program storage.

ekovalev:
Fixes #373.

Work in progress.

In order to track the wait list status and "invoice" programs for having messages in the wait list once in a few blocks, using Substrate offchain workers is suggested. The rationale behind includes the following considerations:
- wait list processing can take long time; having it done on chain is infeasible from the performance perspective;
- direct inspection of the wait list from the pallet code would not help to solve the problem because the wait list is a simple set and doesn't keep track of when messages were inserted into it; therefore an additional (and, potentially, large data structure) should be maintained. Using on-chain storage would unnecessarily bloat the blockchain state;
- System events are not accessible from the runtime code, but are accessible in offchain workers.

At the end of each block an offchain worker listens for certain events corresponding to messages being inserted or removed to/from the wait list during that block. It then decides if there is a change in the wait list state and updates a shared data structure residing in the offchain storage.

Once every several blocks (currently, 10) a set of "invoices" is created to charge rent for the last period in question and sent back on-chain as an unsigned extrinsic with a signed payload. This extrinsic processes the invoices and transfers necessary amont as fees to the current block author.

Note: the whole offchain worker thing is only executed by validators' nodes. A key of KEY_TYPE="gear" is required to sign the payload.


node/src/service.rs
```diff
@@ -207,6 +207,13 @@ pub fn new_full(mut config: Configuration) -> Result<TaskManager, ServiceError>
         })?;
 
     if config.offchain_worker.enabled {
+        sp_keystore::SyncCryptoStore::sr25519_generate_new(
```
ekovalev, line 210: This is for testing purpose only so that the whole thing works when run locally with --dev.
Will be removed before merge.

PR #377, opened 2021-11-05
Fix node warnings

NikVolf:


breathx:
Wouldn't it be better then to move the content of `funcs.rs` to the wasmtime backend mod

NikVolf:
> Wouldn't it be better then to move the content of `funcs.rs` to the wasmtime backend mod

they will be shared eventually

breathx (approved):

PR #378, opened 2021-11-06
Build examples separately and provide node bindings

breathx:


scripts/src/run.sh
```diff
@@ -0,0 +1,33 @@
...
+
+    node            runs gear-node
+    purge-chain     purges gear node chain
+    purge-dev-chain purges gear dev node chain
+
+EOF
+}
+
+run_node() {
+  RUST_LOG=gwasm=debug,gear=debug cargo run -p gear-node "$@"
```
grishasobol, line 24:
```suggestion
  RUST_LOG="gwasm=debug,gear=debug" cargo run -p gear-node "$@"
```

breathx: A forget change `gear` on `gear-core-backend`, yeah, but quotes aren't necessary here

grishasobol: You can just add them and this string will be much more understandable

breathx: Ok. As you wish

NikVolf (approved):

PR #384, opened 2021-11-06
Change allocator git source

grishasobol:


breathx:
Isn't it better to specify some `gear-stable` branch, instead of every time update commit number?

breathx (approved):

PR #390, opened 2021-11-08
Skip vector destruction in vec example, so that two pages stay allocated

### LINKED ISSUE #358: Properly invoke free syscall on vec destructor
### Problem

gtest/spec/test_vec.yaml  "test-vec (20000 * size_of(i32) = 80 KiB = 2 pages)"

Hello, the test creates vec which have size 0x13880 bytes (page is 0x10000).
But as I can see in my allocator logs in the end of fn handle - destructor for vector is called, so the check for two pages is incorrect.

grishasobol:
Fixes #358.

### DIFF
--- a/examples/vec/src/lib.rs
+++ b/examples/vec/src/lib.rs
@@ -19,6 +19,10 @@ pub unsafe extern "C" fn handle() {
     msg::send(msg::source(), v.len() as i32, 10_000_000, 0);
     debug!("{:?} total message(s) stored: ", MESSAGE_LOG.len());
 
+    // Test idea is to allocate two wasm pages and check this allocation,
+    // so we must skip @v disctruction.
+    core::mem::forget(v);
+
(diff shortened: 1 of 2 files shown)


examples/vec/src/lib.rs
```diff
@@ -19,6 +19,10 @@ pub unsafe extern "C" fn handle() {
     msg::send(msg::source(), v.len() as i32, 10_000_000, 0);
     debug!("{:?} total message(s) stored: ", MESSAGE_LOG.len());
 
+    // Test idea is to allocate two wasm pages and check this allocation,
+    // so we must skip @v disctruction.
```
shamilsan, line 23:
```suggestion
    // The test idea is to allocate two wasm pages and check this allocation,
    // so we must skip `v` destruction.
```

shamilsan (approved):

PR #392, opened 2021-11-09
add memory grow gas charging in the start

### LINKED ISSUE #371: Make gas charge for wasm grow when initial pages are loaded
### Problem

In core-runner we charge gas for all pages which are in storage. But in fact, when memory inits for a program we also must make grow while some pages have numbers bigger than memory size in pages. So, this can allow users to store one page with a big number in storage, and all grows before that page will be for free. 

### Possible Solution

I think we should charge gas for each grow before and during program execution. Also, we should charge for each static page when a program is started.

grishasobol:
Fixes #371 .
- add memory grow gas charging in the start
- create test for the case in gas_limit

### DIFF
--- a/core-runner/src/runner.rs
+++ b/core-runner/src/runner.rs
@@ -715,6 +715,31 @@ fn run<E: Environment<Ext>>(
 
     let memory = env.create_memory(program.static_pages());
 
+    // Charge gas for feature memory grows.
+    let max_page = program.get_pages().iter().next_back();
+    if let Some(max_page) = max_page {
+        let max_page_num = *max_page.0;
+        let mem_size = memory.size();
+        if max_page_num >= mem_size {
+            let amount =
+                context.config.mem_grow_cost * ((max_page_num - mem_size).raw() as u64 + 1);
+            let res = gas_counter.charge(amount);
+            if res != gas::ChargeResult::Enough {
+                return RunResult {
+                    messages: Vec::new(),
+                    reply: None,
+                    waiting: None,
+                    awakening: Vec::new(),
+                    gas_left: gas_counter.left(),
+                    gas_spent: 0,
+                    outcome: ExecutionOutcome::Trap(Some("Not enough gas for grow memory size.")),
+                };
+            }
+        } else {
+            assert!(max_page_num.raw() == mem_size.raw() - 1);
+        }
+    }
+
     let ext = Ext {
         memory_context: MemoryContext::new(
             program.id(),
--- a/core-runner/src/runner.rs
+++ b/core-runner/src/runner.rs
@@ -18,6 +18,8 @@
 
 //! Module for running programs.
 
+use core::default;
+
 use alloc::boxed::Box;
 use alloc::collections::{BTreeMap, BTreeSet};
 use alloc::vec::Vec;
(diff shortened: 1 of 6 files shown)


core-runner/src/runner.rs
```diff
@@ -715,6 +715,31 @@ fn run<E: Environment<Ext>>(
...
+            if res != gas::ChargeResult::Enough {
+                return RunResult {
+                    messages: Vec::new(),
+                    reply: None,
+                    waiting: None,
+                    awakening: Vec::new(),
+                    gas_left: gas_counter.left(),
+                    gas_spent: 0,
+                    outcome: ExecutionOutcome::Trap(Some("Not enough gas for grow memory size.")),
+                };
```
shamilsan, line 736: Use `Default` for shortness:

```suggestion
                return RunResult {
                    gas_left: gas_counter.left(),
                    outcome: ExecutionOutcome::Trap(Some("Not enough gas for grow memory size.")),
                    ..Default::default()
                };
```

core/src/env.rs
```diff
@@ -108,8 +108,11 @@ pub trait Ext {
     /// Access currently handled message payload.
     fn msg(&mut self) -> &[u8];
 
-    /// Report that some gas has been used.
-    fn gas(&mut self, amount: u32) -> Result<(), &'static str>;
+    /// Charge some gas. TODO: why u32?
```
shamilsan, line 111: `u32` as this function was called by the instrumented function which consumes not too much gas per weight unit. 

We use the value of `1000` at the moment: https://github.com/gear-tech/gear/blob/master/core/src/gas.rs#L169

grishasobol: It's very strange decision. Because in config:  https://github.com/gear-tech/gear/blob/94a8ee76f1b62d513f4a0b570f854b68d2568927/core-runner/src/runner.rs#L54
we have u64 costs, and u64 -> u32 cast may be dangerous

shamilsan: If you intend to use this function not only by injected gas counter you may change the signature and convert `u32` to `u64` in correspondent calls.

core-runner/src/runner.rs
grishasobol: Done

core/src/env.rs
grishasobol: Sorry, but I didn't understand anything at all =)
I want to use this function only for gas counters.
May change signature of what?
As I see the only one impl of this trait is in core-runner. And it uses u64 gas counter. So, I think the most safe way is use the same types for counter and for amount which we want to charge or refund.

grishasobol: @NikVolf, can you comment this?

NikVolf: Any single gas charge from inside the wasm is u32

grishasobol: Ok, that a good argument. Can you please tell me how can I find where in gear this function is used? https://github.com/gear-tech/gear/blob/94a8ee76f1b62d513f4a0b570f854b68d2568927/core-backend/wasmtime/src/env.rs#L55

shamilsan: When you inject the gas counter into the WASM program (see the link above), the call to `gas` function is added between WASM instructions. Our embedder exports this function to the WASM environment for the program could call it (in your quote of the code). The signature of this function is dictated by the injector (`pwasm_utils::inject_gas_counter`) and is `fn gas(value: u32)`. Of course, you can call your embedded function with `u64` gas value by converting the source gas value from `u32` to `u64`. Alternatively, you may distinguish between small gas values (e.g. fees per weight unit or per alloc) and big ones (gas limit etc.) and use correspondent types (u32 and u64).

grishasobol: Ok thanks

NikVolf (approved):

ukint-vs (approved):

PR #406, opened 2021-11-11
Refactor finding program's meta for custom payload

breathx:
Sometimes we want to send message with custom payload from some external source (user imitation) in gtest. And it fails, because gtest tries to find it's metadata. This PR allows to find not only Output, but also Input types.

@gear-tech/dev


ukint-vs (approved):

breathx:
Refactored a bit since your review. Is it okay with code-readability? @ukint-vs

shamilsan (approved):
Thank you for accompanying refactoring.

ukint-vs (approved):

PR #411, opened 2021-11-12
Add a syscall to get the block timestamp in programs

### LINKED ISSUE #301: Add ability to check block timestamp in programs
Some of the programs might want to use timestamp or block height in their logic. 
For example while hanlding some reply check if it is passed less then 2 days before performing some of the actions.
We need to add sys call to gcore that returns block timestamp/block height

shamilsan:
Fixes #301.

- Add `gr_block_timestamp` syscall that returns `u64` timestamp in milliseconds since the Unix epoch.
- Rename `blolck-height` example to `block-info`.

@gear-tech/dev

### DIFF
--- a/pallets/gear/src/lib.rs
+++ b/pallets/gear/src/lib.rs
@@ -641,7 +641,7 @@ pub mod pallet {
         /// Emits the following events:
         /// - `InitMessageEnqueued(MessageInfo)` when init message is placed in the queue.
         #[pallet::weight(
-            <T as pallet::Config>::WeightInfo::submit_program(code.len() as u32, init_payload.len() as u32)
+            <T as Config>::WeightInfo::submit_program(code.len() as u32, init_payload.len() as u32)
         )]
         pub fn submit_program(
             origin: OriginFor<T>,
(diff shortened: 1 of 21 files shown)


pallets/gear/src/lib.rs
```diff
@@ -635,7 +641,7 @@ pub mod pallet {
         /// Emits the following events:
         /// - `InitMessageEnqueued(MessageInfo)` when init message is placed in the queue.
         #[pallet::weight(
-            T::WeightInfo::submit_program(code.len() as u32, init_payload.len() as u32)
+            <T as pallet::Config>::WeightInfo::submit_program(code.len() as u32, init_payload.len() as u32)
```
ekovalev, line 644: In my opinion it'd be better to use the unqualified `<T as Config>::WeightInfo` here (as long as the purpose is to resolve conflicts with other pallets' `WeightInfo` associated type). Not insisting though

ekovalev (approved):

pallets/gear/src/lib.rs
shamilsan: Thank you for clarification, fixed it.

Issue #141, opened 2021-08-17
Tests for negative scenarios in runner


NikVolf:
- gas limit
- unable to send message
- unable to allocate
- etc

breathx:
Closing because task is outdated and doesn't suit for our current core implementation: we have no runner, but also basic cases this issue meant covered in pallet unit tests and standalone tests.

PR #414, opened 2021-11-15
Add messaging negative tests

Anfid:
Fixes messaging tests #141.

Breaks `gcore` API compatibility to prevent invalid operations on message handle.

### DIFF
--- a/tests/messages/src/lib.rs
+++ b/tests/messages/src/lib.rs
@@ -32,7 +32,7 @@ mod wasm {
     extern crate alloc;
 
     use codec::{Decode, Encode};
-    use gstd::{ext, msg, prelude::*, MessageId, ProgramId};
+    use gstd::{debug, ext, msg, prelude::*, MessageId, ProgramId};
 
     use super::{Reply, Request};
 
@@ -97,7 +97,7 @@ mod wasm {
         msg::load::<Request>()
             .map(process_request)
             .unwrap_or_else(|e| {
-                ext::debug(&format!("Error processing request: {:?}", e));
+                debug!("Error processing request: {:?}", e);
             });
     }
 }
--- a/tests/messages/src/lib.rs
+++ b/tests/messages/src/lib.rs
@@ -1,0 +1,209 @@
+#![cfg_attr(not(feature = "std"), feature(alloc_error_handler))]
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use codec::{Decode, Encode};
+#[cfg(not(feature = "std"))]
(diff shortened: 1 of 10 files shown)


tests/messages/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+            pub fn gr_send_push(handle: u32, data_ptr: *const u8, data_len: u32);
+            pub fn gr_send_commit(
+                handle: u32,
+                message_id_ptr: *mut u8,
+                program: *const u8,
+                gas_limit: u64,
+                value_ptr: *const u8,
+            );
+        }
+    }
```
shamilsan, line 51: Isn't it better to use `gcore` here? Why redeclare syscalls?

tests/messages/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+    pub extern "C" fn init() {
+        msg::reply((), 0, 0);
+    }
+
+    #[no_mangle]
+    pub extern "C" fn handle() {
+        msg::load::<Request>()
+            .map(process_request)
+            .unwrap_or_else(|e| {
+                ext::debug(&format!("Error processing request: {:?}", e));
```
shamilsan, line 100: Prefer `debug!` macro as more readable.

tests/messages/src/lib.rs
Anfid: `gcore::msg::sys` is private. This PR also makes sure that `gcore` message handle is not reusable and is consumed after committing a message, so `gcore` can't be used for that particular test

NikVolf (approved):

tests/messages/src/lib.rs
shamilsan: `gcore::msg` has the same `send_init`. `send_push`, and `send_commit`.

Anfid: `gcore::msg::send_push` and `gcore::msg::send_push` require `gcore::MessageHandle`, that gets consumed in `gcore::msg::send_commit`. This PR also makes sure that `gcore::MessageHandle` does not implement `Clone`. Unless I'm missing something, it should be impossible to recreate `send_push_after_commit` test without using syscalls. Test `send_push_after_commit` validates behavior of the runner, rather than `gcore`/`gstd` libraries, making sure that correct trap gets raised during execution.

tests/messages/src/lib.rs
Anfid: I am no longer able to push to this branch, please apply the fix yourself

PR #416, opened 2021-11-15
Gear standard library restructure

breathx:
- `gstd-async` lib merged into gstd
- `meta` feature removed
- New wrappers around `gcore` general structures added

TODO:
- [x] Refactor `gcore`
- [x] Remove `async` feature
- [x] Make `gstd` and `std` preludes and imports interchangeable
- [x] Refactor things and clean the code
- [x] Fetch latest master changes
- [x] Write minimal docs for ease of writing white paper
- [x] Add default errors

@gear-tech/dev


examples/incomplete_async_payloads/src/lib.rs
```diff
@@ -1,11 +1,10 @@
 #![no_std]
 
 use gstd::{debug, msg, prelude::*};
-use gstd_async::msg as msg_async;
 
 const GAS_LIMIT: u64 = 50_000_000;
 
-#[gstd_async::main]
+#[gstd::main]
```
NikVolf, line 7: macros name or path should have something about async

breathx: What about a small refactor of macro under the hood, which will be no-op in case of non-async function?

gstd/src/async_runtime/futures.rs
```diff
@@ -0,0 +1,66 @@
...
+
+fn empty_waker() -> Waker {
+    unsafe { Waker::from_raw(RawWaker::new(ptr::null(), &VTABLE)) }
+}
+
+type LocalBoxFuture<'a, T> = Pin<Box<dyn Future<Output = T> + 'a>>;
+pub(crate) type FuturesMap = BTreeMap<MessageId, LocalBoxFuture<'static, ()>>;
+
+/// Asynchronous message handling main loop.
+pub fn event_loop<F>(future: F)
```
NikVolf, line 39: I would've called it `message_loop` or `main_loop`, there is no notion of events in our code

examples/incomplete_async_payloads/src/lib.rs
NikVolf: Please, no

Things should be explicit

If no macro required, then no macro should be used

NikVolf:
gstd::sync::* should not match std::sync::*, as they have completely different (async) semantics anyway, so no need to call "sync" naming asynchronous primitives :)

gstd/src/async_runtime/futures.rs
```diff
@@ -0,0 +1,66 @@
...
+    let mut cx = Context::from_waker(&waker);
+
+    let pinned = Pin::new(&mut current_future);
+
+    if pinned.poll(&mut cx).is_ready() {
+        // Done!
+    } else {
+        super::futures().insert(crate::msg::id(), current_future);
+        crate::exec::wait()
+    }
```
shamilsan, line 58: I know the is the old code, but maybe it is worth not making obvious things.

```suggestion
    if !pinned.poll(&mut cx).is_ready() {
        super::futures().insert(crate::msg::id(), current_future);
        crate::exec::wait()
    }
```

examples/async-sign/signed_message/src/lib.rs
```diff
@@ -37,7 +37,11 @@ pub unsafe extern "C" fn handle() {
             msg::reply_bytes(b"Uninitialized", 10_000, 0);
             return;
         }
-        Some(s) => s.0,
+        Some(s) => {
+            let mut arr = [0u8; 32];
+            arr.copy_from_slice(s.as_ref());
+            arr
+        }
```
shamilsan, line 44: Why we've made the code more difficult? It is ok for the `newtype` idiom to get the inner value:

https://doc.rust-lang.org/stable/rust-by-example/generics/new_types.html

gstd/src/common/primitives.rs
```diff
@@ -0,0 +1,130 @@
...
+use primitive_types::H256;
+use scale_info::TypeInfo;
+
+#[derive(
+    Clone, Copy, Debug, Default, Hash, Ord, PartialEq, PartialOrd, Eq, TypeInfo, Decode, Encode,
+)]
+pub struct ActorId([u8; 32]);
+
+impl ActorId {
+    pub const fn new(arr: [u8; 32]) -> Self {
```
shamilsan, line 32: I'd leave empty `new()` (same as `default()`) and add `From<[u8; 32]>` implementation or `from_array(...)` method (as it is e.g. for `std::string::String`). As we use `new([0u8; 32])` too often.

examples/incomplete_async_payloads/src/lib.rs
shamilsan: May to rename it to `#[gstd::async_main]`?

shamilsan (commented):
A lot of work done :rocket:

examples/incomplete_async_payloads/src/lib.rs
breathx: What about universal `#[gstd::async]`, which one we gonna use for handle and init in future?

breathx: I don't like underscores in macros

gstd/src/async_runtime/futures.rs
breathx: I was also refactoring code in this PR, but decided, that this place can be left untouched, due to code readability for people who don't know our async mechanisms well. Doesn't it?

gstd/src/common/primitives.rs
breathx: I don't actually want to give an ability to manipulate with inner array
Is it okay for us?
We may give that ability in case of `debug` feature

examples/async-sign/signed_message/src/lib.rs
breathx: Depend on our decision about `ActorId` constructor

breathx:
> gstd::sync::* should not match std::sync::*, as they have completely different (async) semantics anyway, so no need to call "sync" naming asynchronous primitives :)

What about `gstd::lockers::*` / `gstd::lock::*` ?

gstd/src/async_runtime/futures.rs
shamilsan: Agree.

gstd/src/common/primitives.rs
shamilsan: I don't see the problem if somebody can get the inner value and manipulate with it. But up to you.

examples/incomplete_async_payloads/src/lib.rs
shamilsan: I would be good if it fits habitual patterns like  tokio::main etc. But I unknown the best name for our case 😄

shamilsan (approved):

PR #418, opened 2021-11-15
Refactor message way to waiting list

shamilsan:
- Make the message's way to the waiting list shorter.

@gear-tech/dev

### DIFF
--- a/core-runner/src/runner.rs
+++ b/core-runner/src/runner.rs
@@ -175,7 +175,7 @@ pub struct Runner<SC: StorageCarrier, E: Environment<Ext>> {
     pub(crate) config: Config,
     env: E,
     block_info: BlockInfo,
-    wait_list: BTreeMap<Vec<u8>, MessageDispatch>,
+    wait_list: BTreeMap<(ProgramId, MessageId), MessageDispatch>,
 }
 
 /// Fully in-memory runner builder (for tests).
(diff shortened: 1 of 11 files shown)


NikVolf:
I'd prefer to remove generic argument `WL: WaitingList` for storage completely and just return it in run results

shamilsan:
> I'd prefer to remove generic argument `WL: WaitingList` for storage completely and just return it in run results

But the waitlist is stored in the storage between blocks, isn't it? How we should interact with the key-value storage in this case?

NikVolf:
Just collect messages that has to be put in waiting list as the result of `run`, and put them after `run`

core-runner/src/runner.rs
```diff
@@ -153,17 +171,18 @@ impl RunNextResult {
 pub struct Runner<SC: StorageCarrier, E: Environment<Ext>> {
     pub(crate) program_storage: SC::PS,
     pub(crate) message_queue: SC::MQ,
-    pub(crate) wait_list: SC::WL,
     pub(crate) log: Log,
     pub(crate) config: Config,
     env: E,
     block_info: BlockInfo,
+    wait_list: BTreeMap<Vec<u8>, MessageDispatch>,
```
NikVolf, line 178:
```suggestion
    wait_list: BTreeMap<(ProgramId, MessageId), MessageDispatch>,
```

at least

not sure why you need `ProgramId` though, `MessageId` is unique

shamilsan: I've changed the key type.

NikVolf (approved):

PR #421, opened 2021-11-16
Ability to initialize ProgramId with number/name/ss58

### LINKED ISSUE #417: `gtest`: allow to set string/ss58 as ProgramId
### Problem to Solve

It is useful for tests programs to pretend users

### Possible Solution

similiar to #355

### Notes

_No response_

gshep:
Resolves #417 .

One will be able to write the following:
```yaml

  - id:
      kind: ss58
      value: 5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty
    path: target/wasm32-unknown-unknown/release/test_bob.
# ...
    expected:
      - step: 1
        messages:
          - destination:
              kind: ss58
              value: 5FHneW46xGXgs5mUiveU4sbTyGBzmstUspZC92UhjJM694ty
            payload:
              kind: bytes
              
```

and use user's AccountId/ProgramId as destination of messages. This feature eases to implement bots.

@gear-tech/dev

### DIFF
--- a/gtest/src/address.rs
+++ b/gtest/src/address.rs
@@ -1,0 +1,75 @@
+// This file is part of Gear.
+
+// Copyright (C) 2021 Gear Technologies Inc.
+// SPDX-License-Identifier: GPL-3.0-or-later WITH Classpath-exception-2.0
+
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use gear_core::program::ProgramId;
+use serde::{Deserialize, Deserializer, Serialize};
+use std::str::FromStr;
+
+use sp_core::{crypto::Ss58Codec, hexdisplay::AsBytesRef, sr25519::Public};
+use sp_keyring::sr25519::Keyring;
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(tag = "kind", content = "value")]
+pub enum Address {
+    #[serde(rename = "account")]
+    Account(String),
+    #[serde(rename = "id")]
+    ProgramId(u64),
+    #[serde(rename = "ss58")]
+    SS58(String),
+}
+
+impl Default for Address {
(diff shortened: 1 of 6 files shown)


NikVolf (approved):

gtest/src/runner.rs
```diff
@@ -226,7 +226,7 @@ pub fn init_fixture<MQ: MessageQueue, PS: ProgramStorage, WL: WaitList>(
         }
         runner.queue_message(MessageDispatch {
             source_id: message_source,
-            destination_id: message.destination.into(),
+            destination_id: message.destination.clone().into_program_id(),
```
shamilsan, line 229: It is sub-optimal to clone the whole enum to get the program id only. Note that the `ProgramId` itself is the `Copy` type.

gshep: Yes, you're right. However `into_program_id` is a consuming method so `clone` of enum is used or we get 'use after move' compile-time error.

shamilsan: You are free to add some non-consuming method or implement some std trait for the purpose.

gshep: ok

gtest/src/address.rs
```diff
@@ -0,0 +1,75 @@
...
+}
+
+impl Default for Address {
+    fn default() -> Self {
+        Self::Account("alice".to_string())
+    }
+}
+
+impl Address {
+    pub fn into_program_id(&self) -> ProgramId {
```
shamilsan, line 44: Don't break naming conventions. Name it say `to_program_id(...)`.
https://rust-lang.github.io/api-guidelines/naming.html#c-conv

breathx: What about to just impl `From` ?

gtest/src/address.rs
```diff
@@ -0,0 +1,75 @@
...
+        match self {
+            Self::Account(s) => {
+                ProgramId::from_slice(Keyring::from_str(s).unwrap().to_h256_public().as_bytes())
```
breathx, line 47: I would also recommend to replace each `unwrap` with `expect`

breathx (approved):
Except minor things in comments above, looks good to me!

breathx:
Don't forget to always run `make pre-commit`

ukint-vs (approved):

gtest/src/address.rs
gshep: `ProgramId` lives in `gcore` crate/module, so impl `From` is not-trivial.
However `Into` could be implemented but again  - it is consuming too. Nevertheless `Into` could be implemented for `&Address` though the implementation doesn't bring any advantages: `message.destination.as_ref().into()` or `message.destination.borrow().into()` is required.

What is better, just rename the method?

breathx: Up to you. Renaming seems well

shamilsan (approved):

PR #430, opened 2021-11-22
Introduce gear-debug pallet

ekovalev:
Encapsulates auxiliary extrinsics not having to do with smart contracts deployment and invocation in a separate pallet in order not to overload the main gear pallet with non-essential stuff.
Furthermore, this pallet must be explicitly included in the build by enabling the `features = debug-mode` and is only supposed to be present on a testnet and not in production.


NikVolf (approved):
cool improvement 👍

Issue #353, opened 2021-10-21
RPC-tests: should be reworked and reenabled.


ukint-vs:
RPC-tests should use `gear-js` using debug-mode in node.
Ideally, they should cover the functionality that exists in the `gtest`.

shamilsan:
Maybe should we rename them to `js-tests` or `onchain-tests` to be more relevant?

ukint-vs:
> Maybe should we rename them to `js-tests` or `onchain-tests` to be more relevant?

I think we should leave the name of the protocol used for testing, but how it is done is not important.
RPC already implies that testing takes place from external sources.

PR #435, opened 2021-11-23
JSON-RPC Tests

### LINKED ISSUE #71: Upgrade RPC-tests
RPC tests are disabled at the moment.

- [x] should verify log and program memory: #135
- [x] should never hang: #73
- [x] should suppress node log and show it in case of error: #74 
- [ ] should show pretty diagnostic
- [ ] #353

ukint-vs:
Resolves #71  #353 .

- check log
- check messages
- fix gtest message source
- remove workshop test

todo:
- [ ] check memory
- [x] enable CI

@gear-tech/dev 

**Release notes**: This allows integration testing involving full stack of software


.github/workflows/CI.yaml
```diff
@@ -83,6 +83,15 @@ jobs:
       - name: "Build: Split examples by .opt and .meta"
         run: ./scripts/gear.sh build examples-proc
 
+      - name: "Test: Run RPC tests"
```
NikVolf, line 86: Better to put integration tests at the end

Since with unit-test (that might show the same problem) it will be easier to reason about error reason

shamilsan (approved):

PR #436, opened 2021-11-23
Add ansible script for purging existing chain on testnet

peter-cmw:
Added ansible script and github action to trigger manually to purge chain

### DIFF
--- a/.github/workflows/Purge.yml
+++ b/.github/workflows/Purge.yml
@@ -1,0 +1,38 @@
+name: CD
+
+on: [workflow_dispatch]
+
+env:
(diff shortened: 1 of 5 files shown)


.github/workflows/Purge.yml
```diff
@@ -0,0 +1,38 @@
+name: CD
```
osipov-mit, line 1:
```suggestion
name: Purge-chain
```

osipov-mit (changes requested):

shamilsan (approved):

osipov-mit (approved):

Issue #432, opened 2021-11-22
encode.js has wrong output when my meta.wasm file is input


grishasobol:
### Problem

when I try to encode file
[fungible_token.meta.zip](https://github.com/gear-tech/gear/files/7583397/fungible_token.meta.zip)
using script encode.js, I have output of many many strings:
...
1058696
1058696
1058696
1058696
1058696
1058696
1058696
1058696
1058696
1058696
1058696
1058696
1058696
1058696
1058696
1058696
...
It's wrong output that cannot be parsed by gtest.

### Steps

node gtest/src/js/encode.js -p fungible_token.meta.wasm -t init_input -j '{"name":"GearToken","symbol":"GRT"}'

fungible_token.meta.wasm - it's our wasm file

grishasobol:
Forget to add, that for that output encode.js returns 0

grishasobol:
I have one other example. May be the problem is in my wasm files, but this script output is completely uninformative
[fungible_token1.meta.zip](https://github.com/gear-tech/gear/files/7587182/fungible_token1.meta.zip)

The same reproduce: node gtest/src/js/encode.js -p fungible_token.meta.wasm -t init_input -j '{"name":"GearToken","symbol":"GRT"}'

But script now exit with error:
```
/home/gregory/src/gear/gtest/src/js/node_modules/@polkadot/util/assert.cjs:34
    throw new Error((0, _function.isFunction)(message) ? message() : message);
          ^

Error: Unable to register circular 0 === 0
    at assert (/home/gregory/src/gear/gtest/src/js/node_modules/@polkadot/util/assert.cjs:34:11)
    at TypeRegistry._registerObject (/home/gregory/src/gear/gtest/src/js/node_modules/@polkadot/types/create/registry.cjs:564:26)
    at TypeRegistry.register (/home/gregory/src/gear/gtest/src/js/node_modules/@polkadot/types/create/registry.cjs:549:12)
    at CreateType.registerTypes (/home/gregory/src/gear/gtest/src/js/node_modules/@gear-js/api/lib/CreateType.js:78:23)
    at CreateType.createRegistry (/home/gregory/src/gear/gtest/src/js/node_modules/@gear-js/api/lib/CreateType.js:55:14)
    at CreateType.encode (/home/gregory/src/gear/gtest/src/js/node_modules/@gear-js/api/lib/CreateType.js:93:92)
[...]

PR #448, opened 2021-11-24
Fix demo gas and switch to new version allocator

grishasobol:
Issues #432 
- Fix gas in demos - reduce gas_limit values.
- Change allocator version to new, where implemented:
   - Static buffer allocation for small requests
   - Fix problem with small chunks which restrict to free page sometimes
   - Add free in realloc and sys_alloc
   - Optimize allocator algorithms
   
**Release notes**: Improves our user-space custom allocator

### DIFF
--- a/examples/ping/src/lib.rs
+++ b/examples/ping/src/lib.rs
@@ -9,7 +9,7 @@ pub unsafe extern "C" fn handle() {
     let new_msg = String::from_utf8(msg::load_bytes()).expect("Invalid message");
 
     if new_msg == "PING" {
-        msg::reply_bytes("PONG", 10_000_000, 0);
+        msg::reply_bytes("PONG", 12_000_000, 0);
     }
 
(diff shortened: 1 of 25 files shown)


examples/ping/src/lib.rs
```diff
@@ -21,10 +21,5 @@ pub unsafe extern "C" fn handle() {
     }
 }
 
-#[no_mangle]
-pub unsafe extern "C" fn handle_reply() {
-    msg::reply("PONG", 20_000_000, 0);
-}
-
```
shamilsan, line 28: Isn't this reply used by some other program?

grishasobol: As I can see nobody use this function in tests. So, I think if somebody want to use it somewhere, then he should implement it.

NikVolf (approved):

shamilsan (approved):

PR #451, opened 2021-11-25
Remove gas left and use gas value-tree for node

NikVolf:


core-runner/src/runner.rs
```diff
@@ -313,7 +305,14 @@ impl<SC: StorageCarrier, E: Environment<Ext>> Runner<SC, E> {
             self.block_info,
         );
 
+        let outgoing_messages = context.message_buf.drain(..).collect::<Vec<_>>();
+
         if run_result.outcome.was_trap() && generate_reply_on_trap {
+
+            let gas_spent_for_outgoing: u64 = outgoing_messages.iter().map(|msg| msg.gas_limit).sum();
+            let burned_gas = run_result.gas_spent;
+            let trap_gas = incoming_message.gas_limit() - gas_spent_for_outgoing - burned_gas;
```
ekovalev, line 314: How accurate the gas counter is? Will it hurt if we use saturating subtraction here (just in case)?

core-runner/src/runner.rs
```diff
@@ -633,10 +627,10 @@ pub struct RunResult {
     /// Reply that was received during the run.
     pub reply: Option<ReplyMessage>,
     /// Messages to be woken.
-    pub awakening: Vec<(MessageId, u64)>,
-    /// Gas that was left.
-    pub gas_left: u64,
-    /// Gas that was spent.
+    pub awakening: Vec<MessageId>,
+    /// Gas that was spen.
```
ekovalev, line 631: typo: `spen` => `spent`

core-runner/src/runner.rs
```diff
@@ -313,7 +305,18 @@ impl<SC: StorageCarrier, E: Environment<Ext>> Runner<SC, E> {
             self.block_info,
         );
 
+        let outgoing_messages = context.message_buf.drain(..).collect::<Vec<_>>();
+
         if run_result.outcome.was_trap() && generate_reply_on_trap {
+            let gas_spent_for_outgoing: u64 =
+                outgoing_messages.iter().map(|msg| msg.gas_limit).sum();
```
shamilsan, line 312: Grumble: `map` + `sum` may be replaced by `fold`.

NikVolf: but it requires to write summation function by hand

PR #452, opened 2021-11-25
Remove `MessageQueue` abstraction from `Storage`

shamilsan:
- Remove obsolete `MessageQueue` abstraction from the `Storage` struct.

@gear-tech/dev

### DIFF
--- a/core-runner/src/runner.rs
+++ b/core-runner/src/runner.rs
@@ -84,12 +84,14 @@ impl Config {
 /// Result of one or more message handling.
 #[derive(Debug, Default, Clone)]
 pub struct RunNextResult {
-    /// List of resulting messages.
-    pub message_queue: Vec<Message>,
-    /// The ID of the program that has been run.
-    pub prog_id: ProgramId,
     /// How many messages were handled.
     pub handled: u32,
+    /// List of resulting messages.
+    pub messages: Vec<Message>,
+    /// List of resulting log messages.
+    pub log: Vec<Message>,
+    /// The ID of the program that has been run.
+    pub prog_id: ProgramId,
     /// Execution outcome per each message
     pub outcomes: BTreeMap<MessageId, ExecutionOutcome>,
     /// Gas that was left.
@@ -299,7 +301,7 @@ impl<SC: StorageCarrier, E: Environment<Ext>> Runner<SC, E> {
             return RunNextResult::empty();
         }
 
-        let instrumeted_code = match gas::instrument(program.code()) {
+        let instrumented_code = match gas::instrument(program.code()) {
             Ok(code) => code,
             Err(err) => {
                 log::debug!("Instrumentation error: {:?}", err);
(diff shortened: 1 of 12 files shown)


gtest/spec/test_gui.yaml
```diff
@@ -32,7 +32,13 @@ fixtures:
               kind: custom
               value:
                 ok: 3
-          
+
+          - destination: 0
```
NikVolf, line 36: is this correct @breathx ?

breathx: Yeap, for now

shamilsan:
@NikVolf The messages and waitlist processing will be moved to the pallet within @ekovalev's PR.

core-runner/src/runner.rs
```diff
@@ -84,12 +84,14 @@ impl Config {
 /// Result of one or more message handling.
 #[derive(Debug, Default, Clone)]
 pub struct RunNextResult {
+    /// How many messages were handled.
+    pub handled: u32,
```
breathx, line 88:
```suggestion
```
This should be removed, because we always handle only one message while running.

NikVolf: it can be 0

gtest/spec/test_chat.yaml
```diff
@@ -29,7 +29,7 @@ fixtures:
           value: "0x002c707269766174655f6d7367"
         destination: 2
     expected:
-      - step: 4
+      - step: 6
```
NikVolf, line 32: why has it changed?

core-runner/src/runner.rs
```diff
@@ -339,11 +297,11 @@ impl<SC: StorageCarrier, E: Environment<Ext>> Runner<SC, E> {
                 gas_limit,
                 max_gas_limit
             );
-            self.message_queue.queue(next_message);
+            self.message_queue.push(message);
             return RunNextResult::empty();
         }
 
-        let instrumeted_code = match gas::instrument(program.code()) {
+        let instrumented_code = match gas::instrument(program.code()) {
```
breathx, line 304: Might be write simpler with `.map_err(|err| { .. })?` syntax

gtest/spec/test_chat.yaml
shamilsan: This test hadn't taken into account two messages sent from `bot`'s `init` function`. I don't know why this worked before.

breathx (approved):

NikVolf (approved):

core-runner/src/runner.rs
shamilsan: Yes, but we need to move message processing to the pallet. This will be done after https://github.com/gear-tech/gear/pull/459

core-runner/src/runner.rs
shamilsan: Agree but I've just fixed a typo here :-) Let's do it later.

PR #459, opened 2021-11-26
Move queue and wait list processing to gear pallet

ekovalev:
Collect all the results of processing a message into the `ExecutionReport` (including outgoing messages, wait list and awakened messages) and do all necessary actions upon the execution in the pallet, as opposed to doing part of the work in the node-runner, which has previously been the case.

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -286,7 +286,7 @@ pub fn caller_nonce_fetch_inc(caller_id: H256) -> u64 {
     original_nonce
 }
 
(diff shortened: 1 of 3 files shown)


breathx:
Are we able strongly associate `RunNextResult` with `ExecutionReport` in a future?

ekovalev:
> Are we able strongly associate `RunNextResult` with `ExecutionReport` in a future?

Sounds reasonable.

shamilsan (approved):

PR #472, opened 2021-11-29
Add minimum Rust compiler version to gear-node

shamilsan:
- Add the minimum required Rust compiler version: 1.56 at the moment.


gshep (approved):
:fire:

PR #482, opened 2021-11-30
`gtest`: add verbose mode

ukint-vs:
`-v` to print logs from fixture runs.

@gear-tech/dev

### DIFF
--- a/gtest/src/main.rs
+++ b/gtest/src/main.rs
@@ -38,15 +38,20 @@ struct Opts {
     pub skip_memory: bool,
     /// JSON sample file(s) or dir
     pub input: Vec<std::path::PathBuf>,
+    /// A level of verbosity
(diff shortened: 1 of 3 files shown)


grishasobol (approved):

gtest/src/main.rs
```diff
@@ -38,15 +38,20 @@ struct Opts {
     pub skip_memory: bool,
     /// JSON sample file(s) or dir
     pub input: Vec<std::path::PathBuf>,
+    /// A level of verbosity
```
gshep, line 41: Do I understand coorectyl that this is not a level actually but just a flag if the output should be verbose?

ukint-vs: yep, maybe will add levels in the future

PR #483, opened 2021-11-30
Use formula in tests and also fix gas charge

NikVolf:


pallets/gear/src/tests.rs
```diff
@@ -1610,6 +1612,9 @@ fn distributor_initialize() {
         assert_eq!(Balances::free_balance(1), 95_021_000,);
 
         assert_eq!(Balances::free_balance(255), 4_979_001,);
```
shamilsan, line 1614: Should we remove these asserts?
```suggestion
```

shamilsan (approved):

PR #484, opened 2021-11-30
switch to allocator with disabled static buffer allocation

grishasobol:
switch to allocator with disabled static buffer allocation,
fix gas because without static allocation allocator use more gas.

FYI: there is some problems with allocation in static buffer and js WebAssebly module. currently investigate this problems, and I will enable static allocation in future when this problem will be solved.


NikVolf:
I would like to see test that fails before this update, and passes after

PR #485, opened 2021-11-30
Remove `Runner::log` field

gshep:
`RunNextResult::log` should be used instead

@gear-tech/dev

### DIFF
--- a/gtest/src/runner.rs
+++ b/gtest/src/runner.rs
@@ -19,6 +19,7 @@
 use crate::js::{MetaData, MetaType};
 use crate::sample::{PayloadVariant, Test};
 use gear_backend_common::Environment;
+use gear_core::storage::ProgramStorage;
 use gear_core::{
     message::Message,
     program::{Program, ProgramId},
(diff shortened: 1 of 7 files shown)


gtest/src/runner.rs
```diff
@@ -126,14 +126,16 @@ pub fn init_fixture<SC: StorageCarrier>(
     storage: Storage<SC::PS>,
     test: &Test,
     fixture_no: usize,
-) -> anyhow::Result<(WasmRunner<SC>, Vec<Message>)> {
+) -> anyhow::Result<(WasmRunner<SC>, Vec<Message>, Vec<Message>)> {
+    let storage2 = storage.clone();
```
gshep, line 130: this is a subject to change so simple naming

shamilsan: Why to clone?

shamilsan: Ah, I see, `storage` has been consumed by `run`. Need to be refactored but within other PRs.

shamilsan (approved):
Looks good as an intermediate cleaning.

ukint-vs (approved):

PR #486, opened 2021-11-30
`node-runner`: add messages from init

ukint-vs:
@gear-tech/dev

### DIFF
--- a/node-runner/src/lib.rs
+++ b/node-runner/src/lib.rs
@@ -166,7 +166,15 @@ pub fn init_program<E: Environment<Ext>>(
         value,
         reply: None,
     };
-    let result = RunNextResult::from_single(init_message, run_result);
+    let mut result = RunNextResult::from_single(init_message, run_result.clone());
+
+    result.messages.append(
+        &mut run_result
+            .messages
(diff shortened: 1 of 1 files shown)


shamilsan (approved):

node-runner/src/lib.rs
```diff
@@ -166,7 +166,15 @@ pub fn init_program<E: Environment<Ext>>(
...
     };
-    let result = RunNextResult::from_single(init_message, run_result);
+    let mut result = RunNextResult::from_single(init_message, run_result.clone());
+
+    result.messages.append(
+        &mut run_result
+            .messages
+            .into_iter()
+            .map(|msg| msg.into_message(program_id))
+            .collect(),
```
gshep, line 176: Looks pretty good, but I would suggest to refactor `Run(Next)Restut` & `from_single`. `RunResult` sould contain the processed message and `from_single` should have just one argument - `RunResult`. `messages` have to be populated withing `from_single`.

I believe this code is duplicated already =)

ukint-vs: All this things will be refactored by @breathx

PR #488, opened 2021-12-02
Provide `program_id()` syscall

vivekvpandya:
**Release Notes:** `program_id()` function is added to `gstd` which returns program_id of the program in execution. This can be used where the program wants to store funds like fungible-tokens to for itself.

### DIFF
--- a/gcore/src/exec.rs
+++ b/gcore/src/exec.rs
@@ -20,6 +20,7 @@
 //!
 //! Provides API for low-level async implementation.
 
+use crate::ActorId;
 use crate::MessageId;
 
 mod sys {
@@ -29,6 +30,7 @@ mod sys {
         pub fn gr_gas_available() -> u64;
         pub fn gr_wait() -> !;
         pub fn gr_wake(waker_id_ptr: *const u8);
+        pub fn gr_actor_id(val: *mut u8);
     }
 }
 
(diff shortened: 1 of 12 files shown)


breathx:
Don't see the use-case of this function.
Anyway it also requires sys_cal in `gcore` and `gstd` libs to be called from programs.

vivekvpandya:
> Don't see the use-case of this function. Anyway it also requires sys_cal in `gcore` and `gstd` libs to be called from programs.

DEX smart contract wants funds of fungible-token in its own account. That is why it needs to know its own program-id.

breathx (changes requested):
New syscall addition always requires demo, showing it's functionality.

Also you need to provide the wrapper around it not only for `gcore`, but for `gstd` also.

It would be good to provide `Release notes` for this changes as well.

Except these things, looks good to me

gcore/src/exec.rs
```diff
@@ -147,3 +149,9 @@ pub fn wake(waker_id: MessageId) {
         sys::gr_wake(waker_id.as_slice().as_ptr());
     }
 }
+
+pub fn actor_id() -> ActorId {
```
shamilsan, line 153: Two concerns:
1. The function name seems to be irrelevant to its purpose. It always returns the program ID so `actor_id` looks too general and confusing.
2. This function needs to be documented.

vivekvpandya: yes , but I have heard that program_id will be deprecated soon. Also `ActorId` struct is available to be used here, `ProgramId` is not.

shamilsan: You are right about *types* `ActorId` vs `ProgramId`. But the *function name* should reflect that it does. When I see `exec::actor_id()` in code, it is not obvious to me, who is this actor. So in my opinion `exec::program_id()` or `exec::my_id()` or even `exec::me()` is more speaking for the end user. I'd choose `exec::program_id()` as the most neutral name.

gcore/src/exec.rs
```diff
@@ -20,6 +20,7 @@
 //!
 //! Provides API for low-level async implementation.
```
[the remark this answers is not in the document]
shamilsan: I can't agree, sorry. `msg::actor_id()` or `msg::program_id()` looks a little bit weird. `exec::program_id()` looks quite logical.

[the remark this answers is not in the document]
vivekvpandya: @breathx is okay to mark this as resolved now?

shamilsan (approved):
I don't see the need for this callback itself 😄 but I hope I'll see it later.

PR #490, opened 2021-12-02
Add example program that calls `wait` in `init`

### LINKED ISSUE #489: Example program calling `gr::wait` in its `init`
### Problem to Solve

Create program that demonstrates usage of `wait` in its `init` method. This is a basis to implement async `init` function.

### Possible Solution

create example

### Notes

_No response_

gshep:
Resolves #489.

Currently without test_spec.yaml since there is no support in node/gtest at the moment.

@gear-tech/dev

### DIFF
--- a/examples/init-wait/Cargo.toml
+++ b/examples/init-wait/Cargo.toml
@@ -9,5 +9,4 @@ license = "GPL-3.0"
 crate-type = ["cdylib"]
 
 [dependencies]
-gcore = { path = "../../gcore", features = ["debug"] }
 gstd = { path = "../../gstd", features = ["debug"] }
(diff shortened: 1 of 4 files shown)


examples/init-wait/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+version = "0.1.0"
+authors = ["Gear Technologies"]
+edition = "2018"
+license = "GPL-3.0"
+
+[lib]
+crate-type = ["cdylib"]
+
+[dependencies]
+gcore = { path = "../../gcore", features = ["debug"] }
```
shamilsan, line 12: Do you use gcore anywhere?

shamilsan (approved):

examples/init-wait/Cargo.toml
gshep: Nope

NikVolf (approved):

PR #499, opened 2021-12-03
Add mem::forget for returned strings

NikVolf:
Otherwise allocator will corrupt memory which was returned

### DIFF
--- a/gstd/src/macros/export.rs
+++ b/gstd/src/macros/export.rs
@@ -24,7 +24,7 @@ macro_rules! export {
         #[no_mangle]
         pub unsafe extern "C" fn $f() -> *mut [i32; 2] {
             let result = gstd::macros::util::to_wasm_ptr($val);
-            core::mem::forget(result);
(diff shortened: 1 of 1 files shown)


shamilsan:
Need to remove `core::mem::forget(encoded);` from `meta_state()`:

https://github.com/gear-tech/gear/blob/master/examples/meta/src/lib.rs#L169

NikVolf:
> Need to remove `core::mem::forget(encoded);` from `meta_state()`:
> 
> https://github.com/gear-tech/gear/blob/master/examples/meta/src/lib.rs#L169

why?

shamilsan:
> why?

To avoid double forgetting I guess.

NikVolf:
> > why?
> 
> To avoid double forgetting I guess.

it does not use `export!` macro, so no double forgetting

shamilsan:
> it does not use `export!` macro, so no double forgetting

Ah, agree.

shamilsan (approved):

PR #503, opened 2021-12-03
Enable machinery for charging message senders for having messages in wait list

### LINKED ISSUE #373: Charge rent for keeping messages in the wait list

ekovalev:
Resolves #373.

- Wait list implementation modified so that it stores the block number at which a message was inserted into the list along with the message itself;
- Added a separate pallet `pallet-usage` to keep track of the wait list usage by running an offchain worker every several blocks and scanning the full list from top to bottom. However, to mitigate potential performance implications, there is not more than a certain number of messages processed during a single block. The implementation of the offchain worker guarantees some minimum interval between the blocks the same message in the wait list is potentially visited (provided the latter stays in the list long enough to be invoiced several times);
- the `pallet-usage` implements an external api to be called either by the aforementioned offchain worker (in its unsigned "flavor") or by any legitimate user who can seek to suggest a more optimal set of messages to charge rent from than a simple scan would produce (which is the way the OCW does it) in order to maximize the collected fee in exchange for a share of it.

**Release Notes**: *To be written.*


common/src/lib.rs
```diff
@@ -286,21 +311,16 @@ pub fn caller_nonce_fetch_inc(caller_id: H256) -> u64 {
...
 }
 
-pub fn remove_waiting_message(prog_id: H256, msg_id: H256) -> Option<Message> {
+pub fn remove_waiting_message(prog_id: H256, msg_id: H256) -> Option<(Message, u32)> {
     let id = wait_key(prog_id, msg_id);
-    let msg: Option<Message> = sp_io::storage::get(&id)
-        .map(|val| Message::decode(&mut &val[..]).expect("message encoded correctly"));
+    let msg = sp_io::storage::get(&id)
+        .map(|val| <(Message, u32)>::decode(&mut &val[..]).ok())
+        .flatten();
```
shamilsan, line 323: Grumble: `map` + `flatten` may be replaced with `and_then`.
```suggestion
        .and_then(|val| <(Message, u32)>::decode(&mut &val[..]).ok());
```

pallets/usage/src/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+    impl<T: Config> ValidateUnsigned for Pallet<T> {
+        type Call = Call<T>;
+        fn validate_unsigned(source: TransactionSource, call: &Self::Call) -> TransactionValidity {
+            match call {
+                Call::collect_waitlist_rent { payees_list: _ } => {
+                    // Only accept transactions from a trusted source: local OCW or already in block.
+                    match source {
+                        TransactionSource::Local | TransactionSource::InBlock => { /* allowed */ }
+                        _ => return InvalidTransaction::Call.into(),
+                    }
```
shamilsan, line 142: Maybe to use `matches!`?
```suggestion
                    if !matches!(source, TransactionSource::Local | TransactionSource::InBlock) {
                        return InvalidTransaction::Call.into(),
                    }
```

pallets/usage/src/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+
+                    // TODO: ensure necessary validiry checks hold
```
shamilsan, line 144: Typo
```suggestion
                    // TODO: ensure necessary validity checks hold
```

pallets/usage/src/offchain.rs
```diff
@@ -0,0 +1,194 @@
...
+            Ok((prog_id, msg_id))
+        });
+
+        let mut entries = vec![];
+        let mut counter = 0_u32;
+        let mut new_last_key: Option<Vec<u8>>;
+        // Iterate through the wait list storage starting from the entry following the `last_key`
+        loop {
+            new_last_key = iter
+                .next() /*.as_ref()*/
```
shamilsan, line 151: Dead code?
```suggestion
                .next()
```

pallets/gear/src/lib.rs
```diff
@@ -373,7 +371,7 @@ pub mod pallet {
                                     });
 
                                     // TODO: weight to fee calculator might not be identity fee
```
gshep, line 373: I guess the comment is not actual anymore since gas converting is defined in the Config now?

pallets/usage/src/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+mod mock;
+
+#[cfg(test)]
+mod tests;
+
+mod offchain;
+
+pub type Authorship<T> = pallet_authorship::Pallet<T>;
+
+pub const WAITLIST_FEE_PER_BLOCK: u64 = 100;
```
gshep, line 41: Looks like a candidate for Config's constant

pallets/usage/src/lib.rs
```diff
@@ -0,0 +1,384 @@
...
+                // Starting a new round
+                current_round_storage_ref.set(&now);
+            }
+
+            // Acquire the lock protecting shared offchain workers' storage
+            let mut lock = StorageLock::<'_, Time>::with_deadline(
+                offchain::STORAGE_OCW_LOCK,
+                Duration::from_millis(T::ExpirationDuration::get()),
+            );
+            let _guard = lock.lock();
```
gshep, line 235: Can it finish with an error?

common/src/native.rs
```diff
@@ -109,26 +109,24 @@ pub fn program_exists(id: ProgramId) -> bool {
     crate::program_exists(H256::from_slice(id.as_slice()))
 }
 
-pub fn insert_waiting_message(prog_id: ProgramId, msg_id: MessageId, message: CoreMessage) {
+pub fn insert_waiting_message(
+    prog_id: ProgramId,
+    msg_id: MessageId,
+    message: CoreMessage,
+    bn: u32,
```
gshep, line 116: I would suggest to make generic over BlockNumber similar how it is defined in `frame_system::Config`. Not insisting though

pallets/usage/src/tests.rs
```diff
@@ -0,0 +1,290 @@
...
+            get_offchain_storage_value(offchain::STORAGE_LAST_KEY),
+            Some(common::wait_key(
+                20_u64.into_origin(),
+                2920_u64.into_origin()
+            ))
+        );
+
+        run_to_block_with_ocw(17);
+        assert_eq!(pool.read().transactions.len(), 5);
+        // The wait list should have been completely exhauseted at this moment,
```
gshep, line 128: typo `exhauseted`

gshep (changes requested):
looks good to me

common/src/native.rs
ekovalev: Used to have it that way before. Then got rid of the generic part for the sake of simplicity. Although personally I agree the generic way is more appropriate

pallets/usage/src/lib.rs
ekovalev: It's a blocking call (if I understand correctly what you mean). There is another type of lock (`BlockAndTime`, to be specific) which returns an error in case of failure to acquire the lock. In case of the `Time` lock the next offchain worker will be able to take hold of it either after the previous holder drops it or after the lock expiry.
Using `Time` locks is more common, however, using the `BlockAndTime` lock is also an option, in case we decide it is worthwhile to do so.

pallets/gear/src/lib.rs
ekovalev: Fair enough

gshep (approved):

shamilsan (approved):
Should we purge the testnet after merging?

pallets/usage/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+                Call::collect_waitlist_rent { payees_list: _ } => {
+                    // Only accept transactions from a trusted source: local OCW or those already in block.
+                    if !matches!(
+                        source,
+                        TransactionSource::Local | TransactionSource::InBlock
+                    ) {
+                        return InvalidTransaction::Call.into();
+                    }
+
+                    // TODO: ensure necessary validity checks hold
```
NikVolf, line 148: Please add issue for all TODO-s and reference them

pallets/usage/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+
+                    if new_free_gas_limit == 0 {
+                        // Generate trap reply
+
+                        // Account for the fact that original gas balance of the message could have
+                        // already been lower than the required "existential" gas limit
+                        let trap_gas = actual_gas_limit.min(T::TrapReplyExistentialGasLimit::get());
+
+                        let mut program = common::get_program(program_id).expect(
+                            "There is always an associated program for a message in wait list",
```
NikVolf, line 323: This expect is not true, programs might be `kill`ed eventually

pallets/usage/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+                                user_reward,
+                            );
+                            let _ = T::PaymentProvider::withhold_reserved(
+                                gas_tree.origin(),
+                                &Authorship::<T>::author(),
+                                validator_reward,
+                            );
+                        }
+                        _ => {
+                            let _ = T::PaymentProvider::withhold_reserved(
```
NikVolf, line 307: There should be a warning if it is errors

pallets/usage/src/lib.rs
```diff
@@ -0,0 +1,405 @@
...
+        type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+
+        /// Gas and value transfer currency
+        type Currency: Currency<Self::AccountId> + ReservableCurrency<Self::AccountId>;
+
+        /// Gas to Currency converter
+        type GasConverter: GasToFeeConverter<Balance = BalanceOf<Self>>;
+
+        /// Type providing interface for making payment in currency units
+        type PaymentProvider: PaymentProvider<Self::AccountId, Balance = BalanceOf<Self>>;
```
ekovalev, line 80: The reason for having this associated type as opposed to using the `Currency` type's `repatriate_reserved` method directly is that we don't want to impose additional trait bound on the `T::AccountId` type (to implement the `common::Origin` trait) in order to avoid conflict, created by the pallet's `Config` trait inheriting from the `SendTransactionTypes<Call<Self>>` because the type parameter `Call<Self>` is then subject to the same trait bound `T::AccountId: common::Origin` which is not the case for the `SendTransactionTypes<T>`.

NikVolf (approved):

PR #505, opened 2021-12-03
add script to run only one example build

grishasobol:
add script to run only one example build:
$ ./script/src/build_one_example.sh /examples/your_test_name

### DIFF
--- a/scripts/src/build_one_example.sh
+++ b/scripts/src/build_one_example.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env sh
 
 # Go into gear source root dir and run:
-# $ script/src/build_one_example.sh /examples/your_test_name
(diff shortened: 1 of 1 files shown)


SabaunT:
What's your opinion on following the same style for all the sh scripts for this PR? I mean, being able to call the "build one" from entry point `scripts/gear.sh`?

grishasobol:
@SabaunT yes it's more correct way. But scripts it's not the most important part, so you can improve it in future. I just supply my script for everyone.

breathx:
> @SabaunT yes it's more correct way. But scripts it's not the most important part, so you can improve it in future. I just supply my script for everyone.

I'd recommend to follow main concept of gear scripts and provide something like `./gear.sh build demo_name`

PR #517, opened 2021-12-09
Add support to calling `wait` in `init` on the node side

### LINKED ISSUE #18: Program initial state and lifecycle
Initially, program is instantiated in "uninitialized" state. Until it achieves "initialized" state, all messages to it should not be handled.

Program can achieve "initialized" state, once it replies to the initialization message with non-trap reply

### LINKED ISSUE #514: Tests for deferred program initialization
### Problem to Solve

Add tests representing the feature #18 

### Possible Solution

- uninitialized (but not failed) program should not receive messages
- uninitialized program accepts replies
- correct reply initializes program. Then the program accepts message and replies to it

### Notes

_No response_

### LINKED ISSUE #515: Support possibility of caling `wait` in `init` method
### Problem to Solve

tl;dr - tests #514  should pass successfully

### Possible Solution

add additional information about program in storage so program's initialization process can be tracked. Adjust other places in node/node-runner

### Notes

_No response_

### LINKED ISSUE #579: Get rid of `IntermediateMessage`
### File Location(s)

https://github.com/gear-tech/gear/blob/master/common/src/lib.rs#L146

### Proposal

We can minimize the intermediate structures count.

gshep:
Resolves #18. Resolves #515. Resolves #579.

- Split examples/init-wait/lib.rs into two files: one with crate-level `no_std` attribute and the other with the program code. This trick allows to include `code.rs` in other modules/crates.
- Add `tests/init-wait` that includes `code.rs` and exports `WASM_BINARY_BLOATY`.
- Add three tests to `pallet-gear/tests`. Theay are ignored at the moment since node doesn't support calling `wait` ini `init` method.

@gear-tech/dev


tests/init_wait/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+mod native {
+    include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
+}
+
+#[cfg(feature = "std")]
+pub use native::{WASM_BINARY, WASM_BINARY_BLOATY};
+
+#[cfg(not(feature = "std"))]
+mod wasm {
+    include! {"../../../examples/init-wait/src/code.rs"}
```
gshep, line 14: This line means that pallet-gear/tests will depend on examples/init-wait though this dependency is for test/dev purposes only, so should be ok.

If this solution seems not so good then I may suggest to introduce fake empty pallet. This pallet would depend on pallet-gear & examples/tests/init-wait and implement all required testing functionality.

examples/init-wait/src/code.rs
```diff
@@ -0,0 +1,48 @@
...
+    WaitForReply,
+    Inited,
+}
+
+static mut STATE: State = State::NotInited;
+static mut INIT_MESSAGE: MessageId = MessageId::new([0; 32]);
+
+#[no_mangle]
+pub unsafe extern "C" fn handle() {
+    if STATE != State::Inited {
```
shamilsan, line 15: Just in case, you may use `if let` for not deriving `PartialEq`.

shamilsan (dismissed):
Nice approach 👍

breathx:
First of all, I don't see the reason to keep in `gear` repo examples which do nothing: don't test core features and even don't represent real use-cases. 

Moreover, I think that asynchronous program initialization will not be in user-space. We don't need static mutable variables to keep it in mind.

I would recommend to show such things as proposals for https://github.com/gear-tech/forge repo in a future.

NikVolf:
Leave just yaml test, why do we need to test same thing in 3 different places?

gshep:
> First of all, I don't see the reason to keep in `gear` repo examples which do nothing: don't test core features and even don't represent real use-cases.
> 
> Moreover, I think that asynchronous program initialization will not be in user-space. We don't need static mutable variables to keep it in mind.
> 
> I would recommend to show such things as proposals for https://github.com/gear-tech/forge repo in a future.

Thanks for the comment. However can't agree with you. The example `init-wait` demonstrates that `wait` can be used in the `init` and it tests this functionality - actually how it is supposed to function, similiary to another example `wait`.

pallets/gear/Cargo.toml
```diff
@@ -46,6 +46,7 @@ wabt = "0.10"
 gear-core = { path = "../../core" }
 hex-literal = "0.3.4"
 tests-distributor = { path = "../../tests/distributor" }
+tests-init-wait = { path = "../../tests/init_wait" }
```
NikVolf, line 49:
```suggestion
tests-init-wait = { path = "../../tests/init-wait" }
```

and corresponding changes

common/src/lib.rs
```diff
@@ -326,6 +356,44 @@ pub fn remove_waiting_message(prog_id: H256, msg_id: H256) -> Option<(Message, u
     msg
 }
 
+pub struct WaitingMessageIterator {
```
NikVolf, line 359: Add an issue to optimise it without iterator please

gshep: done = #525

pallets/gear/Cargo.toml
gshep: Sorry didn't understand your suggestion

gshep: hyphen!

common/src/lib.rs
ekovalev: There is also this "built-in" `frame_support::storage::PrefixIterator`, that can be easily customized for any prefix (in case you find it more convenient). We have examples in our code.

core-runner/src/runner.rs
```diff
@@ -244,7 +244,7 @@ impl<SC: StorageCarrier, E: Environment<Ext>> Runner<SC, E> {
     ///
     /// Runner will return actual number of messages that was handled.
     /// Messages with no destination won't be handled.
-    pub fn run_next(&mut self, message: Message) -> RunNextResult {
+    pub fn run_next(&mut self, message: Message, next_entry: RunNextEntry) -> RunNextResult {
```
ekovalev, line 247: Can we consider making the entry point a part of the Message itself? Since the runner needs to know what the entry point is anyway...

gshep: Good point. I'd prefer to implement it within another task =).

What do you think @shamilsan @breathx @NikVolf ?

common/src/lib.rs
```diff
@@ -146,6 +148,9 @@ pub enum IntermediateMessage {
         value: u128,
         reply: Option<H256>,
     },
+    WakeMessages {
```
ekovalev, line 151: Is there a reason for this variant name to be a plural?

common/src/lib.rs
```diff
@@ -190,6 +209,21 @@ pub fn set_code(code_hash: H256, code: &[u8]) {
     sp_io::storage::set(&code_key(code_hash).0, code)
 }
 
+#[derive(Decode, Encode)]
+pub enum ProgramState {
+    Uninitialized { message_id: H256 },
```
ekovalev, line 214: Would you elaborate by commenting each variant what it means for a program to be in some particular state?

common/src/lib.rs
gshep: No real reason. I thought about it like a command: "please wake messages addressed to this program". Anyway it't not hard to rename it. 'WakeMessage' or just 'Wake'?

common/src/lib.rs
gshep: ok

common/src/lib.rs
ekovalev: Whatever it is - as long as it's consistent, it's fine by me. Personally, I would give up the `...Message` ending part at all (going ahead with just `Init`, `Dispatch` and `Wake`). But it's up to you.

common/src/lib.rs
```diff
@@ -326,6 +360,41 @@ pub fn remove_waiting_message(prog_id: H256, msg_id: H256) -> Option<(Message, u
...
+
+fn decode_message_tuple(_: &[u8], value: &[u8]) -> Result<(Message, u32), codec::Error> {
+    <(Message, u32)>::decode(&mut &*value)
+}
+
+impl WaitingMessageIterator {
+    pub fn new(destination_program_id: H256, previous_message_id: Option<H256>) -> Self {
+        let key_prefix = match previous_message_id {
+            None => wait_prefix(destination_program_id),
+            Some(id) => wait_key(destination_program_id, id),
```
ekovalev, line 373: What is the purpose of this iterator? Is it supposed to iterate trough all waitlisted messages for a given program (potentially, starting from some specific message id)?
Based on this implementation, it seems that if you provide a `previous_message_id` the prefix will be set to the encoded combination of `WAIT_PREFIX::program_id::message_id` and there is at most one such key in the storage. Is this intentional?
Maybe the `prefix` should only contain the `program_id`, regardless of the starting point (which is determined by the `previous_message_id`)?

gshep: Yeah, this is an error. Good catch!

NikVolf:
Better rebase it on new runner, will be easy

core-runner/src/runner.rs
NikVolf: This is already done in runner refactoring

pallets/gear/src/lib.rs
```diff
@@ -96,6 +96,8 @@ pub mod pallet {
         InitFailure(MessageInfo, Reason),
         /// Program initialized.
         InitSuccess(MessageInfo),
+        /// Program initialization deferred.
+        InitWait(MessageInfo),
```
NikVolf, line 100: No need for this event

gshep: Emit `InitSUccess` instead?

NikVolf: Yes, but once init is success, not on wait :)

pallets/gear/src/lib.rs
```diff
@@ -260,14 +265,60 @@ pub mod pallet {
...
         /// but hasn't been explicitly removed from storage by its creator
-        pub fn is_uninitialized(program_id: H256) -> bool {
+        pub fn is_failed(program_id: H256) -> bool {
             ProgramsLimbo::<T>::get(program_id)
                 .map(|_| true)
                 .unwrap_or(false)
         }
 
+        /// Internal helper function. Moves messages from waiting list to the queue. Returns a flag if there are
+        /// more (probably) messages and remained weight.
```
ekovalev, line 284: remained => remaining

pallets/gear/src/lib.rs
```diff
@@ -260,14 +265,60 @@ pub mod pallet {
...
+            remaining_weight: Weight,
+        ) -> (Option<WaitingMessages>, Weight) {
+            let one_step_weight = T::DbWeight::get().reads(2);
+            let message_count = remaining_weight / one_step_weight;
+
+            let actual_count = common::WaitingMessageIterator::drain(program_id, None)
+                .take(message_count as usize)
+                .fold(0u64, |acc, (m, _)| {
+                    common::queue_message(m);
+                    acc + 1
```
ekovalev, line 296: Isn't `actual_count` going to always be equal to `message_count`, because you `take` exactly `message_count` items from the collection and `fold` over them afterwards?

pallets/gear/src/lib.rs
```diff
@@ -260,14 +265,60 @@ pub mod pallet {
...
+                });
+
+            let result_weight = remaining_weight - actual_count * one_step_weight;
+            if actual_count < message_count {
+                (None, result_weight)
+            } else {
+                (Some(WaitingMessages), result_weight)
```
ekovalev, line 303: Where does the `WaitingMessages` value come from?

pallets/gear/src/lib.rs
```diff
@@ -260,14 +265,60 @@ pub mod pallet {
...
+
+            let actual_count = common::WaitingMessageIterator::drain(program_id, None)
+                .take(message_count as usize)
+                .fold(0u64, |acc, (m, _)| {
+                    common::queue_message(m);
+                    acc + 1
+                });
+
+            let result_weight = remaining_weight - actual_count * one_step_weight;
+            if actual_count < message_count {
```
ekovalev, line 300: How about the case when `actual_count` == `message_count` and there are no items left in the `WaitingMessageIterator`? That is there were exactly as many messages in the wait list as we could process due to weight constraint

pallets/gear/src/lib.rs
```diff
@@ -260,14 +265,60 @@ pub mod pallet {
...
+            }
+        }
+
+        /// Wakes all messages sent to `destination_program_id` while it hasn't been initialized.
+        pub fn wake_waiting_messages(destination_program_id: H256) {
+            let (remained, weight) = Self::requeue_waiting_messages(
+                destination_program_id,
+                Self::gas_allowance() as Weight,
+            );
+            GasAllowance::<T>::put(weight);
```
ekovalev, line 313: This will set the remaining gas allowance to `weight`. Shouldn't it be like we have in other places:
`GasAllowance::<T>::mutate(|x| *x = x.saturating_sub(weight));` ?

Maybe it's better to simply return the used weight to the calling function and do all gas allowance math in one place?

pallets/gear/src/lib.rs
gshep: Nope, this is unnecessary.

`WaitingMessages` - is a private unit type, defined right before `impl Pallet`. `Option<WWaitingMessages>` is used instead of a non-informative `bool`. `None` means that there are no more messages to wake, `Some` - there may be some messages.

On the other side I could also introduce custom enum for this flag but it requires more lines of code than such `Option`.

pallets/gear/src/lib.rs
gshep: Edge-case =).

For the sake of simplicity I decided not to make additional storage read to find out if there are more message but just schedule requeueing in the next block.

Thanks for comment anyway - I wrongly calculate `one_step_wait`. There should also one storage read by `common::queue...`

pallets/gear/src/lib.rs
```diff
@@ -260,14 +265,66 @@ pub mod pallet {
...
+            }
+        }
+
+        /// Tries to wake all messages sent to `destination_program_id` while it hasn't been initialized.
+        pub fn wake_waiting_messages(destination_program_id: H256) {
+            let (remained, used_weight) = Self::requeue_waiting_messages(
+                destination_program_id,
+                Self::gas_allowance() as Weight,
+            );
+            Self::decrease_gas_allowance(used_weight);
```
gshep, line 319: @ekovalev 

I realized that we have two transformations: (1) gas -> fee, (2) gas -> weight. Here the identity mappging `gas -> weight` is assumed which is not correct in general

ekovalev: We also have an explicit `WeightToFee` mapping from `Weigth` to `BalanceOf<T>` so we need to make sure this diagram always commutes. As long as all the maps are Identity we are fine, but generally I guess we should always express the third as a composition of the first two.

ekovalev: Care to create a task for proper conversion traits implementation and mention in a `TODO` that `Self::gas_allowance() as Weight` will be replaced with a more appropriate gas to weight conversion function?

ekovalev (approved):

ekovalev:
Actually, one more thing needs adjustment: since `init_program` can now result in a waiting message we can't just `unreserve everything right away` anymore - some gas would still be locked inside the message's `value_tree`.
`unreserve` must only occur when some `value_tree` is consumed. And vice versa: whenever we `reserve` currency a corresponding `value_tree` must be created.
But I guess this might as well be done in a separate task.

shamilsan (approved):

NikVolf:
IntermediateMessage is for user input, it should not be used in any internal logic

NikVolf (changes requested):
Should not work like that

initialisation dispatch should go into regular message queue and most of this special logic should be removed

common/src/lib.rs
```diff
@@ -38,6 +38,8 @@ use storage_queue::StorageQueue;
 
 pub const STORAGE_PROGRAM_PREFIX: &[u8] = b"g::prog::";
 pub const STORAGE_PROGRAM_PAGES_PREFIX: &[u8] = b"g::pages::";
+pub const STORAGE_PROGRAM_STATE_PREFIX: &[u8] = b"g::prog_state::";
```
NikVolf, line 41: why don't we just store that in the program struct ?

NikVolf (approved):
Would like to see comments from @ekovalev and we a good to go!

common/src/lib.rs
gshep: no real reason for standalone prefix. Will change

gtest/src/manager.rs
```diff
@@ -4,17 +4,28 @@ use gear_core::{
...
 use std::collections::{BTreeMap, VecDeque};
 
+use crate::check::ProgramInitializer;
+
 #[derive(Clone, Default)]
 pub struct InMemoryExtManager {
     message_queue: VecDeque<Message>,
     log: Vec<Message>,
-    programs: BTreeMap<ProgramId, Program>,
+    programs: RefCell<BTreeMap<ProgramId, Program>>,
```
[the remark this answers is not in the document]
gshep: Since the program is stored before any processing now I had to repeat the logic in the gtest/node test suite, so `ProgramStorage` trait is introduced. It has only one method with signature `&self`.

gtest/src/manager.rs
```diff
@@ -43,9 +54,8 @@ impl JournalHandler for InMemoryExtManager {
             DispatchOutcome::MessageTrap { .. } => {
```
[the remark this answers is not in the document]
gshep: +

pallets/gear/src/lib.rs
```diff
@@ -209,6 +206,7 @@ pub mod pallet {
     pub type GasAllowance<T> = StorageValue<_, u64, ValueQuery, DefaultForGasLimit<T>>;
 
     #[pallet::storage]
+    #[pallet::getter(fn get_mailbox)]
```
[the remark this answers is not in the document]
gshep: +

pallets/gear/src/lib.rs
```diff
@@ -664,23 +582,39 @@ pub mod pallet {
...
+            let init_message_id = common::next_message_id(&init_payload);
+            common::set_program_state(
+                id,
+                ProgramState::Uninitialized {
+                    message_id: init_message_id,
+                },
+            );
 
-            <MessageQueue<T>>::append(IntermediateMessage::InitProgram {
+            common::value_tree::ValueView::get_or_create(
```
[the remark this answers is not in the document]
gshep: +

pallets/gear/src/lib.rs
```diff
@@ -748,23 +682,37 @@ pub mod pallet {
                 ExistenceRequirement::AllowDeath,
             )?;
 
-            // Only after reservation the message is actually put in the queue.
-            <MessageQueue<T>>::append(IntermediateMessage::DispatchMessage {
+            let origin = who.into_origin();
+            common::value_tree::ValueView::get_or_create(
```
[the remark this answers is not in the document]
gshep: +

pallets/gear/src/lib.rs
```diff
@@ -848,22 +796,38 @@ pub mod pallet {
             }
 
             let message_id = common::next_message_id(&payload);
-            <MessageQueue<T>>::append(IntermediateMessage::DispatchMessage {
+
+            let origin = who.into_origin();
+            common::value_tree::ValueView::get_or_create(
```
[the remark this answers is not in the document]
gshep: +

pallets/gear/src/tests.rs
```diff
@@ -778,21 +709,24 @@ fn send_reply_works() {
...
-                .flatten()
-                .expect("reply message was previously sent");
-            match intermediate_msg {
-                IntermediateMessage::DispatchMessage { id, reply, .. } => {
-                    (id, reply.expect("was a reply message"))
-                }
-                _ => unreachable!("only reply message was in mq"),
-            }
+
+        let event = match SystemPallet::<Test>::events()
```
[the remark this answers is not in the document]
gshep: if you don't mind i would leave it as is

ekovalev:
> Would like to see comments from @ekovalev and we a good to go!

Looks awesome. Nothing has caught my eye that'd make me wanna complain ).

ekovalev (approved):

gshep:
*Release notes*: Programs lifecycle has been changed so now their initialization can be deferred with `wait`/`wake` syscalls. This is the next step towards asynchronous initialization of programs.

PR #520, opened 2021-12-09
Add `submit_code` call

### LINKED ISSUE #347: Submit code transaction
Since we want to initialise programs from other programs (see #245), one of the scenarios is to initialise program from existing code (by providing `code_hash` and initialization payload)

This `submit_code` transaction should put a code to the state and return its `code_hash`.

SabaunT:
Close #347.

- Create `submit_code` extrinsic
- Change code storage: now stores code author and block number along with the code.

@gear-tech/dev

### DIFF
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -326,6 +326,11 @@ pub fn remove_waiting_message(prog_id: H256, msg_id: H256) -> Option<(Message, u
     msg
 }
 
+pub fn code_exists(code_hash: H256) -> bool {
+    let (code_key, _) = code_key(code_hash);
+    sp_io::storage::exists(&code_key)
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -69,31 +69,6 @@ pub struct Program {
     pub nonce: u64,
 }
 
-/// Code with metadata
-///
-/// Storing a code with `submit_code` extrinsic violates code invariant: we no longer
-/// have any guarantees that saved code has an initialized program referencing it (no matter
-/// initialization failed or not). We decided to store `author` and `block_number` as metadata
-/// to the storage for a future opportunity to invalidate some `code` in storage data using either
-/// TTL with `block_number` check or collecting taxes from `author` for storing uninitialized code.
-#[derive(Clone, Debug, Decode, Encode, PartialEq, TypeInfo)]
-pub struct CodeWithMetadata {
-    pub author: H256,
-    pub block_number: u32,
-    pub code: Vec<u8>,
-}
-
-impl From<(Vec<u8>, H256, u32)> for CodeWithMetadata {
-    fn from(data: (Vec<u8>, H256, u32)) -> Self {
-        let (code, author, block_number) = data;
-        CodeWithMetadata {
-            code,
-            author,
-            block_number,
-        }
-    }
-}
-
 pub trait Origin: Sized {
     fn into_origin(self) -> H256;
     fn from_origin(val: H256) -> Self;
@@ -360,10 +335,6 @@ pub fn code_exists(code_hash: H256) -> bool {
 #[cfg(test)]
 mod tests {
     use super::*;
-
-    fn create_code_with_meta(code: Vec<u8>) -> CodeWithMetadata {
-        (code, H256::from([0; 32]), 1).into()
-    }
 
     #[test]
     fn nonce_incremented() {
--- a/common/src/lib.rs
+++ b/common/src/lib.rs
@@ -42,6 +42,7 @@ pub const STORAGE_MESSAGE_PREFIX: &[u8] = b"g::msg::";
 pub const STORAGE_MESSAGE_NONCE_KEY: &[u8] = b"g::msg::nonce";
 pub const STORAGE_MESSAGE_USER_NONCE_KEY: &[u8] = b"g::msg::user_nonce";
 pub const STORAGE_CODE_PREFIX: &[u8] = b"g::code::";
+pub const STORAGE_CODE_METADATA_PREFIX: &[u8] = b"g::code::metadata";
 pub const STORAGE_CODE_REFS_PREFIX: &[u8] = b"g::code::refs";
 pub const STORAGE_WAITLIST_PREFIX: &[u8] = b"g::wait::";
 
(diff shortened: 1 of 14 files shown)


SabaunT:
@shamilsan @NikVolf 
So basically there are two todo-s I wanted to resolve before opening for a review. 
But I will appreciate if you look at how saving program code from `core_runner` was implemented (an alternative was to change `core::storage::ProgramStorage` trait to have two different `set` and `update` functions, but it could led to some copy-paste in `common::native{get_program; set_program}`).

common/src/lib.rs
```diff
@@ -182,12 +207,31 @@ pub fn wait_key(prog_id: H256, msg_id: H256) -> Vec<u8> {
...
+            block_number,
+        }
+    })
 }
 
-pub fn set_code(code_hash: H256, code: &[u8]) {
-    sp_io::storage::set(&code_key(code_hash).0, code)
+pub fn set_code(code_hash: H256, code: CodeWithMetadata) {
+    let mut data = Vec::new();
+    code.author.encode_to(&mut data);
```
ekovalev, line 231: Doesn't this mean that if the same piece of code is submitted multiple times the metadata will be overwritten every time and only contain the details of who (and when) has submitted it last? This might render the TTL-based logic potentially susceptible to misuse (although, at obvious expense for an attacker).

SabaunT: @ekovalev well it depends on how `set_code` is actually used. Calls to the function are performed from `set_program` and `submit_code`. In the first one the code is checked for existence, i.e. If it exists we don't save it. In the second function (extrinsic) code can be set only if it doesn't exist. So, no double submissions - no rewrites of metadata.

Should be stated, that this metadata is only needed to invalidate "old" code data entries either by using authors address (charge him) or block number (TTL).

SabaunT: Oh, btw, what do you think about putting metadata to other prefix in storage?

common/src/lib.rs
```diff
@@ -68,6 +68,31 @@ pub struct Program {
...
+/// Storing a code with `submit_code` extrinsic violates code invariant: we no longer
+/// have any guarantees that saved code has an initialized program referencing it (no matter
+/// initialization failed or not). We decided to store `author` and `block_number` as metadata
+/// to the storage for a future opportunity to invalidate some `code` in storage data using either
+/// TTL with `block_number` check or collecting taxes from `author` for storing uninitialized code.
+#[derive(Clone, Debug, Decode, Encode, PartialEq, TypeInfo)]
+pub struct CodeWithMetadata {
+    pub author: H256,
+    pub block_number: u32,
+    pub code: Vec<u8>,
```
ekovalev, line 82: What is the benefit of storing metadata together with the code? We kinda already have some "metadata" for each piece of code - the reference counter which is in one-to-one relation with the code. Maybe it would make more sense to extend the ref counter (like, replace it with some metadata structure). That way we could access the metadata part independently of the code which might be a bit more efficient considering a piece of code can weigh hundreds of kilobytes so we might want to avoid fetching the whole record from storage every time we need to check, say, TTL expiry or whatnot.

common/src/lib.rs
ekovalev: > Oh, btw, what do you think about putting metadata to other prefix in storage?

Please see my comment above. In short, I'd definitely do so.

common/src/lib.rs
SabaunT: Very nice point. Agree.

core/src/program.rs
```diff
@@ -98,7 +98,7 @@ impl ProgramId {
 #[derive(Clone, Debug, Decode, Encode)]
 pub struct Program {
     id: ProgramId,
-    code: Vec<u8>,
+    code_with_meta: CodeWithMetadata,
```
ekovalev, line 101: Why does it have to be `CodeWithMeta` and not just wasm bytes?

SabaunT: Will answer in discord and paste link. It requires some discussions.

The reason for that is actually a decision to store code metadata for codes, which are saved from `submit_program` extrinsic. More precisely, from `process_queue`.

SabaunT: https://discord.com/channels/891063355526217738/905188682745655317/919139083484291093

common/src/lib.rs
```diff
@@ -148,20 +164,38 @@ pub enum IntermediateMessage {
...
-    (key, ref_counter)
+fn code_key(code_hash: H256, kind: CodeKeyPrefixKind) -> Vec<u8> {
+    let code_key_impl = |prefix: &[u8]| {
+        // key's length is N bytes of code hash + M bytes of prefix
+        // currently code hash is 32 bytes
+        let mut key = Vec::with_capacity(prefix.len() + code_hash.as_bytes().len());
+        key.extend(prefix);
+        code_hash.encode_to(&mut key);
+        key
+    };
```
shamilsan, line 192: I don't see the reason for lambda here.

pallets/gear/src/lib.rs
```diff
@@ -620,42 +631,82 @@ pub mod pallet {
...
         T::AccountId: Origin,
     {
-        /// Create a `Program` from wasm code and runs its init function.
+        /// Saves program `code` in storage.
         ///
-        /// `ProgramId` is computed as Blake256 hash of concatenated bytes of `code` + `salt`.
-        /// Such `ProgramId` must not exist in the Program Storage at the time of this call.
+        /// The extrinsic was created to provide _deploy program from program_ functionality.
+        /// Anyone who wants to define a "factory" logic in program should first store the code and metadata for the "child"
+        /// program in storage. So the code for child will be initialised by program initialization request only if it exists in storage.
```
shamilsan, line 669:
```suggestion
        /// program in storage. So the code for the child will be initialized by program initialization request only if it exists in storage.
```

pallets/gear/src/weights.rs
```diff
@@ -56,6 +57,13 @@ pub trait WeightInfo {
 
 pub struct GearWeight<T>(PhantomData<T>);
 impl<T: frame_system::Config> WeightInfo for GearWeight<T> {
+    fn submit_code(c: u32) -> Weight {
+        (17_289_000_u64)
+            .saturating_add((1_000_u64).saturating_mul(c as Weight))
+            .saturating_add(T::DbWeight::get().reads(1_u64))
+            .saturating_add(T::DbWeight::get().writes(1_u64))
```
shamilsan, line 64: I don't like the visual noise where it is not obligatory. But we have a lot of such kinds of code, so you can ignore this comment :)

```suggestion
            .saturating_add(T::DbWeight::get().reads(1))
            .saturating_add(T::DbWeight::get().writes(1))
```

ekovalev (approved):

shamilsan (approved):

PR #530, opened 2021-12-16
Refactor pallet tests

### LINKED ISSUE #524: Refactor `pallet_gear` tests not to put out internals
Our tests are "smelly", because put out lots of internals, which requires fixing lots of stuff manually after adding/editing some new feature. Code becomes too dependant. An example could be chaning code set procedure (performed only in 2 extrinsics), which resulted in pasting `common::set_code` everywhere...

SabaunT:
Resolves #524 

@gear-tech/dev

### DIFF
--- a/pallets/gear/src/mock.rs
+++ b/pallets/gear/src/mock.rs
@@ -29,11 +29,10 @@ use sp_runtime::{
 type UncheckedExtrinsic = frame_system::mocking::MockUncheckedExtrinsic<Test>;
 type Block = frame_system::mocking::MockBlock<Test>;
 
-pub(crate) const BLOCK_AUTHOR: u64 = 255;
 pub(crate) const USER_1: u64 = 1;
-// todo [sab] change accounts order + change names to MSG_SENDER and etc.
-pub(crate) const LOW_BALANCE_USER: u64 = 2;
-pub(crate) const USER_2: u64 = 3;
+pub(crate) const USER_2: u64 = 2;
+pub(crate) const LOW_BALANCE_USER: u64 = 3;
+pub(crate) const BLOCK_AUTHOR: u64 = 255;
 
 // Configure a mock runtime to test the pallet.
 #[cfg(feature = "debug-mode")]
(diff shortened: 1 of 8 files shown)


SabaunT:
Draft is opened in order to show an idea. Of course there are lots of todo-s now.
But I'm open to refactoring ideas. Especially about redesigning `TestManager` to a builder pattern or something like that.

NikVolf:
Would like to see less complicated stuff there with just something that invokes to pallet calls

Somebody looking to add/modify test does not need to learn whole new concepts of managers etc.

pallets/gear/src/mock.rs
```diff
@@ -29,7 +29,11 @@ use sp_runtime::{
 type UncheckedExtrinsic = frame_system::mocking::MockUncheckedExtrinsic<Test>;
 type Block = frame_system::mocking::MockBlock<Test>;
 
-pub const BLOCK_AUTHOR: u64 = 255;
+pub(crate) const BLOCK_AUTHOR: u64 = 255;
+pub(crate) const USER_1: u64 = 1;
+// todo [sab] change accounts order + change names to MSG_SENDER and etc.
+pub(crate) const LOW_BALANCE_USER: u64 = 2;
+pub(crate) const USER_2: u64 = 3;
```
gshep, line 36: I would suggest more conventional names for cryptography like Alice, Bob & their friends

SabaunT: The fact is that we need to state in user name, that he has low balance, that's why `LOW_BALANCE_USER` was introduced. Some tests check balance stuff, so such user is needed. IMHO, `Alice`, `Bob`, `BLOCK_AUTHOR` and  `LOW_BALANCE_USER` is no difference with `USER_1`, `USER_2` and etc.

What do you think?

shamilsan (approved):
I have some concerns about `ProgramCodeKind::to_bytes()` (it seems to be not intuitive to me) but not too important. The good thing is that the code size has been decreased.

ekovalev (approved):

PR #536, opened 2021-12-20
bump spec version

ukint-vs:
@gear-tech/dev


ekovalev (approved):

runtime/src/lib.rs
```diff
@@ -125,7 +125,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     //   `spec_version`, and `authoring_version` are the same between Wasm and native.
     // This value is set to 100 to notify Polkadot-JS App (https://polkadot.js.org/apps) to use
```
SabaunT, line 126: `This value is set to 120...`

gshep: I guess this comment could be wiped out since it is a template one.

PR #538, opened 2021-12-20
Add test yaml list argument

grishasobol:
Add possibility to build/test specified list of yamls:
```
make gtest yamls=./gtest/spec/test_capacitor.yaml
./scripts/gear.sh build examples yamls="./gtest/spec/test_async.yaml ./gtest/spec/test_capacitor.yaml"
./scripts/gear.sh test gtest yamls="./gtest/spec/test_async.yaml ./gtest/spec/test_capacitor.yaml"```

### CODE REVIEW

File: scripts/src/build.sh [hunk 1]
```diff
@@ -14,7 +14,9 @@ build_usage() {
     help           show help message and exit
 
     gear           build gear workspace
-    examples       build gear program examples
+    examples       build gear program examples,
+                   you can specify yaml list to build coresponding examples
+                   using yamls="path/to/yaml1 path/to/yaml2 ..." argument
```
gshep, line 19: I've heard somewhere that 'explicit is better than implicit' so I'd prefer to have possibility to explicitly set which examples should be built. However if the very same feature to build all examples mentioned in yaml is required then maybe it is better to introduce new target?

Another concern is how to build all examples with this approach?

scripts/src/build.sh
```diff
@@ -45,7 +47,47 @@ examples_build() {
...
+  if [ -n "$1" ]
+  then
+    has_yamls=$(echo "$1" | grep "yamls=" || true)
+  else
+    has_yamls=""
+  fi
+
+  if  [ -n "$has_yamls" ]
+  then
+    if ! command -v perl &> /dev/null
```
gshep, line 61: Although one new dependency is added the main concern is that yamls parsed here with perl and in gtest. So there will be two places with parsing logic: this may lead to maintain costs in the future.

Can't suggest anything useful here: one crazy idea is to add parameter to gtest so it returns the list of all involved examples. Another one is to extract common code for parsing and create additional binary that parses yamls and returns the required list.

scripts/src/build.sh
grishasobol: > Another concern is how to build all examples with this approach?
You can skip yamls argument in that case all examples will be built.
This argument is created to build all examples deps in yamls, so if you want to build special example - you can make your own argument))) or make your own yaml file.
Also, I create this arg to be able run specified test  and avoid building all examples each time.

scripts/src/build.sh
grishasobol: Oh, I suppose we will spend much more time to do that, then to support two small codes for yaml parsing.
Also, as you now gtest use only compiled wasm files, but build.sh need to now sources locations.
So, my position is: this functionality is only for some developers needs and we can skip architecture design here.

NikVolf (approved):

PR #502, opened 2021-12-03
Migrate to new message processing logic with `core-processor`

breathx:
TODO:
- [x] write release notes
- [ ] write related issue
- [ ] open new issues based on research during this PR
- [ ] reset the network after merge

**Release notes**: This migrates to new logic of processing messages with functional approach and journalling. This allows us to use multiple ways to actually execute core logic on different environments (collator, validator and cumulus setups).

PR #549, opened 2021-12-27
Add `create_program` syscall

### LINKED ISSUE #512: Create `create_program` sys-call
### Problem to Solve

This is one of the steps of [iteration plan](https://github.com/gear-tech/gear/issues/245#issuecomment-982558284). Implementing this logic is a crucial interim step to solve #245. 

Suggested changes to be made (from the iteration plan):
- User provides to the function 1) the code hash which he got after submit_code, 2) some salt (salt management is a user space task), 3) payload data, 4) gas limit (should be checked in core for being less than max allowed), 5) value
- The sys-call returns a new program id computed from salt and code hash. The logic of program id creation should be changed from prog_id = h(code + salt) to prog_id = h(code_hash + 
- The sys-call creates a message that later is processed by the node.
- Perform pallet tests as well

SabaunT:
Resolves #512 

**Note**: base is the branch for the root issue (create program from program), not master. All the internal logic of message creation by the sys-call will be implemented in `st-create-program-from-program-iter1`. Such separation was done due to massive refactoring in core, so the rest will be implemented after refactoring in #502 is merged to master.

@gear-tech/dev 

**Release Notes**: This creates a new sys-call `create_program` and related API in `gstd` and `gcore` crates for program developers. This API can be used to create a new program from program


core-backend/common/src/funcs.rs
```diff
@@ -240,6 +240,35 @@ pub fn send_push<E: Ext>(ext: LaterExt<E>) -> impl Fn(i32, i32, i32) -> Result<(
...
 
+pub fn create_program<E: Ext>(
+    ext: LaterExt<E>,
+) -> impl Fn(i32, i32, i32, i32, i32, i64, i32, i32) -> Result<(), &'static str> {
+    move |code_hash_ptr: i32,
+          salt_ptr: i32,
+          salt_len: i32,
+          payload_ptr: i32,
+          payload_len: i32,
+          _gas_limit: i64,
```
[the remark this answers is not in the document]
SabaunT: These values are needed only for a new init message. But due to the separation concern stated in the first PR [comment](https://github.com/gear-tech/gear/pull/549#issue-1089119255) message will be created in the next PR. So current implementation is interim, which will be merged to other dev branch (not master)

core-backend/common/src/funcs.rs
```diff
@@ -240,6 +240,35 @@ pub fn send_push<E: Ext>(ext: LaterExt<E>) -> impl Fn(i32, i32, i32) -> Result<(
...
+          value_ptr: i32,
+          program_id_ptr: i32| {
+        let res = ext.with(|ext: &mut E| -> Result<(), &'static str> {
+            let code_hash = get_bytes32(ext, code_hash_ptr);
+            let salt = get_vec(ext, salt_ptr, salt_len);
+            let _payload = get_vec(ext, payload_ptr, payload_len);
+            let _value = get_u128(ext, value_ptr);
```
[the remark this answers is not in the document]
SabaunT: Same answer as here https://github.com/gear-tech/gear/pull/549#discussion_r775492566

gcore/src/general.rs
```diff
@@ -145,3 +145,31 @@ impl ActorId {
         &mut self.0[..]
     }
 }
+
+#[derive(Clone, Copy, Debug, Default, Hash, Ord, PartialEq, PartialOrd, Eq)]
+pub struct H256(pub [u8; 32]);
```
shamilsan, line 150: This type's name is too similar to the `primitive_types::H256` and can be confused I guess.

SabaunT: `Hash256` better?

shamilsan: > `Hash256` better?

Maybe it should be more application-specific like `CodeHash`?

SabaunT: > > `Hash256` better?
> 
> Maybe it should be more application-specific like `CodeHash`?

Agree.

pallets/gear/src/lib.rs
```diff
@@ -743,12 +743,15 @@ pub mod pallet {
         ) -> DispatchResultWithPostInfo {
             let who = ensure_signed(origin)?;
 
-            let mut data = Vec::new();
-            // TODO #512
-            code.encode_to(&mut data);
-            salt.encode_to(&mut data);
+            let id: H256 = {
+                let mut data = Vec::new();
```
shamilsan, line 747: It is better to use `Vec::with_capacity` when we know the length of the resulting vector. This potentially allows avoiding several reallocations.

```suggestion
                let mut data = Vec::with_capacity(code_hash.len() + salt.len());
```

shamilsan (approved):

PR #552, opened 2021-12-28
add `AccountId` field to `get_gas_spent` rpc call

ukint-vs:
Since message processing can depend on the source, specifying it will allow to more accurately calculate the cost.

```bash
curl http://localhost:9933 -H "Content-Type:application/json;charset=utf-8" -d   '{
     "jsonrpc":"2.0",
      "id":1,
      "method":"gear_getGasSpent",
      "params": ["5GrwvaEF5zXb26Fz9rcQpDWS57CtERHpNehXCPcNoHGKutQY",
        "0x91f004f8ebf242f97d52c76d92e9a6776b7f1d341491c7d1dfd1f88b072631f3",
        "0x50494e47"]
    }'
```
@gear-tech/dev

### DIFF
--- a/pallets/gear/src/lib.rs
+++ b/pallets/gear/src/lib.rs
@@ -302,7 +302,7 @@ pub mod pallet {
             Ok(message)
         }
 
-        pub fn get_gas_spent(dest: H256, payload: Vec<u8>) -> Option<u64> {
+        pub fn get_gas_spent(source: H256, dest: H256, payload: Vec<u8>) -> Option<u64> {
             let mut ext_manager = ExtManager::<T>::default();
 
             let block_info = BlockInfo {
@@ -312,7 +312,7 @@ pub mod pallet {
 
             let mut messages: VecDeque<Message> = VecDeque::from([Message {
                 id: common::next_message_id(&payload),
-                source: 100001.into_origin(),
+                source,
                 dest,
                 gas_limit: u64::MAX,
                 payload,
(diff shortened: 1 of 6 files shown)


runtime/src/lib.rs
```diff
@@ -600,12 +600,14 @@ impl_runtime_apis! {
...
     // Here we implement our custom runtime API.
-    impl pallet_gear_rpc_runtime_api::GearApi<Block, ProgramId> for Runtime {
+    impl pallet_gear_rpc_runtime_api::GearApi<Block, AccountId, ProgramId> for Runtime {
         fn get_gas_spent(
+            account_id: AccountId,
             program_id: ProgramId,
             payload: Vec<u8>,
         ) -> Option<u64> {
-            Gear::get_gas_spent(program_id, payload)
+            use gear_common::Origin;
```
breathx, line 609: Better to import it at the end of file

pallets/gear/src/tests.rs
```diff
@@ -408,10 +408,10 @@ fn block_gas_limit_works() {
```
breathx, line 408: I would recommend to calculate it from unused user for tests, to prevent future errors and test issues

pallets/gear/src/lib.rs
```diff
@@ -272,7 +272,7 @@ pub mod pallet {
             .flatten()
         }
 
-        pub fn get_gas_spent(dest: H256, payload: Vec<u8>) -> Option<u64> {
+        pub fn get_gas_spent(source: H256, dest: H256, payload: Vec<u8>) -> Option<u64> {
```
breathx, line 275: What about making the new argument 'Option<H256>' and random it or hardcode for 'None' case?

breathx (commented):
Seems good, but I think we should wait any feedback from our js-side to merge this PR @osipov-mit

shamilsan:
Closing in favor of #612. This PR has been cherry-picked to that.

PR #557, opened 2021-12-29
Keep all gas in parent's tree when sending message to mailbox

### LINKED ISSUE #556: Enable unused gas release from a value tree
Current value tree implementation will only allows the refund of the locked value to the external owner after all the nodes of the tree have been consumed. This is undesirable in situations when some downstream message gets stuck, say, in a user's mailbox indefinitely thus preventing the whole value tree to be consumed.

We need to rework the concept of value tree so that consumption of a node would trigger immediate value propagation upwards and refund to the owner, provided all the upstream nodes have already been consumed.

ekovalev:
Resolves #556.

This change creates a distinction between messages in queue or those being processed and messages in some user's mailbox: getting into a mailbox is the end of a message life cycle, where no further processing can take place. Therefore, messages in the mailbox should not bring along any gas: no new node should be `split_off` from the gas tree associated with the parent message - all gas must remain in the parent tree.

This allows to unblock gas tree consumption if the only unconsumed message in a chain is that ended up in someone's mailbox.

If a reply follows, the author of the reply message must create a separate gas tree and pay for it by reserving their currency.
This allows to avoid a potentially buggy situation with offsetting gas/value of a mailboxed message against that of a reply message: now a `send_reply` call would simply claim the value from the mailbox first (`gas_limit` must be zero based on what is discussed here above) and then send a usual message.

### DIFF
--- a/pallets/gear/src/manager.rs
+++ b/pallets/gear/src/manager.rs
@@ -266,7 +266,7 @@ where
         log::debug!("burned: {:?} from: {:?}", amount, message_id);
 
         Pallet::<T>::decrease_gas_allowance(amount);
-        // TODO: weight to fee calculator might not be identity fee
+
         let charge = T::GasConverter::gas_to_fee(amount);
 
         self.gas_handler.spend(message_id, amount);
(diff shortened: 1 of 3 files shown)


NikVolf:
This is just incorrect, the whole tree idea is to refund only when everything is consumed

pallets/gear/src/manager.rs
```diff
@@ -199,24 +199,24 @@ where
...
+            let _ = T::Currency::repatriate_reserved(
+                &<T::AccountId as Origin>::from_origin(origin.into_origin()),
+                &Authorship::<T>::author(),
+                charge,
+                BalanceStatus::Free,
+            );
         } else {
             log::error!(
-                "Message does not have associated gas tree: {:?}",
+                target: "runtime::gear",
```
NikVolf, line 215: better leave default target for such logs

pallets/gear/src/manager.rs
```diff
@@ -232,36 +232,44 @@ where
...
-        }
+        log::debug!("Sending message {:?} from {:?}", message, message_id);
 
         if common::program_exists(message.dest) {
-            common::queue_message(message);
+            if let Some(mut gas_tree) = ValueView::get(GAS_VALUE_PREFIX, message_id) {
+                let _ = gas_tree.split_off(message.id, message.gas_limit);
+                common::queue_message(message);
+            } else {
+                log::error!(
```
NikVolf, line 260: I don't think we should silently avoid queuing messages, even with `log::error`

either propagate error somewhere or still queue it

NikVolf: We may at least add runtime event that message was going to be queued but couldn't

NikVolf (approved):
other than small grumbles seems great

NikVolf:
Would like to see more extensive tests for gas (since it is sound protocol-wise now?) as a follow-up

PR #583, opened 2022-01-14
Upgrade-runtime in two steps via cli

ukint-vs:
@gear-tech/dev


shamilsan (approved):
The best code is not written code 😄

PR #604, opened 2022-01-20
Fix build with rustc 1.60.0-nightly (5e57faa78 2022-01-19)

gshep:
@gear-tech/dev

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7444,7 +7444,6 @@ dependencies = [
 name = "tests-btree"
 version = "0.1.0"
 dependencies = [
- "gear-core",
  "gstd",
  "parity-scale-codec",
  "substrate-wasm-builder",
@@ -7467,7 +7466,6 @@ name = "tests-distributor"
 version = "0.1.0"
 dependencies = [
  "env_logger 0.9.0",
- "gear-core",
  "gstd",
  "parity-scale-codec",
  "substrate-wasm-builder",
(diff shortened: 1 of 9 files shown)


shamilsan (approved):

breathx:
Need to remove these optional deps so

breathx (approved):

shamilsan:
> Need to remove these optional deps so

As I see we need to move them to `[dev-dependencies]`.

breathx:
> > Need to remove these optional deps so
> 
> As I see we need to move them to `[dev-dependencies]`.

Why? Don't see their usage in `tests/btree`, e.g.

I think they are unused stuff, that we forget to remove after migration on `core-processor`.

shamilsan:
> Why? Don't see their usage in tests.
> 
> I think they are unused stuff, that we forget to remove after migration on `core-processor`.

At least `common` is used: 

https://github.com/gear-tech/gear/blob/master/tests/btree/src/lib.rs#L103

breathx:
> > Why? Don't see their usage in tests.
> > I think they are unused stuff, that we forget to remove after migration on `core-processor`.
> 
> At least `common` is used:
> 
> https://github.com/gear-tech/gear/blob/master/tests/btree/src/lib.rs#L103

Ah, yeah, found it.

Need to recheck it. Somewhere it's unused, as I see. Somewhere we **should** replace `gear_core::MessageId` with `gstd::MessageId` and remove unused dep, because it's more correct in gear-programs.

shamilsan (changes requested):
> Ah, yeah, found it.
> 
> Need to recheck it. Somewhere it's unused, as I see. Somewhere we **should** replace `gear_core::MessageId` with `gstd::MessageId` and remove unused dep, because it's more correct in gear-programs
> 
> https://github.com/gear-tech/gear/blob/d5219307fae465d790c1814477536d134c86aaa5/tests/wait_wake/src/lib.rs#L63
> .

Agree. @gshep please review all dependencies in standalone tests and leave only these are really needed.

shamilsan (approved):
I don't mind if @breathx has no concerns about `gear_core::MessageId`.

breathx (approved):

PR #616, opened 2022-01-28
Implement `sys::gr_exit`

### LINKED ISSUE #453: New syscall `sys::gr_exit`
### Problem to Solve

Allow programs to preemptively terminate their own execution. Program called `exit` is wiped from the storage with all its information.

### Possible Solution

Implement new syscall.

### Notes

_No response_

gshep:
Resolves #453 

Add new syscall `exit`. It stops the program execution and "kills" it: Prgoram data is removed but its ID and memory pages hash are kept.

**Release notes**: to be written


core-backend/common/src/funcs.rs
```diff
@@ -93,6 +93,20 @@ pub fn gas_available<E: Ext>(ext: LaterExt<E>) -> impl Fn() -> i64 {
     move || ext.with(|ext: &mut E| ext.gas_available()).unwrap_or(0) as i64
 }
 
+pub fn kill<E: Ext>(
```
SabaunT, line 96: what do you think about changing name to `interrupt` or `stop`?
As far as I remember we plan to make a feature of `selfdestruct`, when program can be deleted (erase state, delete address and e.t.c.). I think `kill` and `selfdestruct` names point to the same semantics, although planned to differ.

gshep: To tell the truth I'm not insisting on this name, so probably it will be changed. For example, `terminate` is ok too.

What about the syscall behavior: we discussed the task last week. When the program would call `kill` its state should be removed, however the hash of the state and all waiting messages to the program should be kept (and program's id too as I guess). This is for the next feature that allows users to restore the killed program.

NikVolf: I would go with `exit` as in `std::process:exit`

common/src/lib.rs
```diff
@@ -72,6 +74,46 @@ pub struct Program {
     pub state: ProgramState,
 }
 
+#[derive(Clone, Debug, Decode, Encode, PartialEq, TypeInfo)]
+pub struct ExitedProgram {
+    pub program_id: H256,
+    pub program: Program,
+    pub pages_hash: Vec<u8>,
```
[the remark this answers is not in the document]
gshep: Yep, H256 is better.

common/src/exit.rs
```diff
@@ -0,0 +1,71 @@
...
+            .using_encoded(sp_io::hashing::blake2_256)
+            .into(),
+        program,
+        wait_list: PrefixIterator::<_, ()>::new(prefix, previous_key, decode_message_tuple)
+            .drain()
+            .map(|(m, _)| m)
+            .collect(),
+    };
+
+    remove_program(program_id);
```
gshep, line 55: Please take a look at this. `remove_program` also release the code so if there are no refs the code will be removed. At the moment it is not an issue however for restoration feature the code has to be in the storage.

gear-test/src/manager.rs
```diff
@@ -93,6 +93,11 @@ impl JournalHandler for InMemoryExtManager {
             }
         };
     }
+
+    fn exit_dispatch(&mut self, _id_exited: ProgramId, _value_destination: ProgramId) {
+        unimplemented!()
```
gshep, line 98: And this is the second place I'm uncertain with. I'm able to implement `exit_dispatch` in every `JournalHandler` though it would not be called at the moment. So I decided to go with `umiplmemented`.

PR #622, opened 2022-02-01
Move value tree gas handler into a pallet

ekovalev:
An effort to have more consistent and verifiable gas economics.

Introduced a single interface to operations with gas.
Gas is accounted for through a DAG-based ledger (something similar to a UTxO model but much simpler - the `DAGBasedLedger` trait). The previously existed `ValueTreeGasHandler` struct is one of the implementations of this trait. Another implementation is the newly introduced `pallet-gas` which is just a plain porting of the value tree. The last one is a dummy implementation `()` which doesn't actually track gas but is used in the testsuite (a temporary measure until the testsuite is able to "honestly" handle gas).

The pallet-gear now has an associated type `GasHandler` which implements this trait. Since the `ExtManager` is generic over the `Runtime` (as `pallet_gear::Config` trait), it will have the same interface to the gas handler implementation (with a small caveat that at the moment for the sake of compatibility with the testsuite the `ExtManager` is still kept generic over a gas handler type, as well, which allows to override the default gas handler defined by the Runtime).

The declared better consistency and accountability comes from the fact that the DAGBasedLedger returns imbalances when value is created or burnt (like Currency trait in frame_support) which, if not counter-balanced by a an inversely directed operation, contributes to the gas total supply increase or decrease. In a long run this quantity must return to 0: all created gas is eventually spent or consumed. This gives a tool to detect potential gas-related discrepancies.


common/src/lib.rs
```diff
@@ -20,19 +20,16 @@
 
 pub mod native;
 pub mod storage_queue;
+mod traits;
```
breathx, line 23: As @NikVolf always says, it isn't the best naming practice to call modules like `structs`, `traits` etc.

pallets/gas/src/tests.rs
```diff
@@ -0,0 +1,227 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use super::*;
+use crate::mock::*;
+use frame_support::{assert_noop, assert_ok};
+use primitive_types::H256;
+
+// use crate::Pallet as Gas;
```
breathx, line 24: Dead code

breathx (commented):
Looks very cool at first glance!

I'm now confused should we totally depreciate `gear-common`. Maybe providing of light-weight lib with all basic essences (as it done in core-processor) and traits isn't bad idea? The most considering thing for me is how to unify behavior for native testing, pallet testing and pallet itself in future.

Related only for on-chain view, I really like it. Great job 👏🏻

common/src/lib.rs
ekovalev: In the end of the day just reverted the creation of this separate module for traits and placed everything back into the `lib.rs` where it belongs.

pallets/gas/src/lib.rs
```diff
@@ -0,0 +1,498 @@
...
+#[allow(clippy::derivable_impls)]
+// this cannot be derived, despite clippy is saying this!!
+impl Default for ValueOrigin {
+    fn default() -> Self {
+        ValueOrigin::External(H256::default())
+    }
+}
+
+#[derive(Clone, Default, Decode, Debug, Encode, MaxEncodedLen, TypeInfo)]
+pub struct ValueNode {
```
gshep, line 49: Does this intentionally duplicate `common::ValueNode`?

pallets/gas/src/lib.rs
```diff
@@ -0,0 +1,498 @@
...
+    // Gas pallet error.
+    #[pallet::error]
+    pub enum Error<T> {
+        /// Gas (gas tree) has already been created for the provided key.
+        GasTreeAlreadyExists,
+
+        /// Account doesn't have enough funds to complete operation.
+        InsufficientBalance,
+
+        /// Value node doesnt' exist for a key
```
gshep, line 90: misprint

pallets/gas/src/lib.rs
breathx, line 49: Isn't it copy-pasted from common?

breathx (approved):
🔥

pallets/gas/src/lib.rs
```diff
@@ -0,0 +1,498 @@
...
+    use super::{Config, Imbalance};
+    use frame_support::{
+        traits::{SameOrOther, TryDrop},
+        RuntimeDebug,
+    };
+    use sp_runtime::traits::Zero;
+    use sp_std::{marker::PhantomData, mem};
+
+    /// Opaque, move-only struct with private field to denote that value has been created
+    /// without any equal and opposit accounting
```
gshep, line 321: `opposite`

gshep (approved):
Great work! :rocket:

pallets/gas/src/lib.rs
ekovalev: Yes. The entire pallet is, in large, a clone of the ValueView/ValueNode from common. We don't want to keep any references to common here because that stuff will go very soon (in fact it's already not being used anywhere in the code - I left it there just in case and for testing purpose to compare the results of two implementations).

ekovalev: Yes. All the related stuff in `common` is about to be removed (it's already not being used anywhere in the code and has been kept for testing purpose to be able to compare two implementations results).

common/src/value_tree.rs
```diff
@@ -281,6 +301,106 @@ impl ValueView {
     }
 }
 
+#[deprecated(note = "will be removed soon")]
+#[derive(Default)]
+pub struct ValueTreeGasHandler;
+
+impl DAGBasedLedger for ValueTreeGasHandler {
```
NikVolf, line 308: this stuff now stays only in test code ?

why not move it to the place it is used?

ekovalev: Based on our latest understanding the entire `common` crate will sooner or later go, so I just didn't bother moving it anywhere. This specific part is only needed until we've confirmed that the old gas handler implementation (which we believe is correct) produces the identical outcome with the new gas handler (pallet-babsed) for any contract execution given the same chain state. Once confirmed, it won't be needed at all.

NikVolf (approved):
Need to bump spec version

NikVolf (approved):

Issue #614, opened 2022-01-27
Introduce critical sections for atomic operations


shamilsan:
### Problem to Solve

Some critical sections may be interrupted by a trap. e.g. due to insufficient gas limit. For example:

```rust
static mut FROM_BALANCE: i32 = 1000;
static mut TO_BALANCE: i32 = 0;

fn transfer() {
    if FROM_BALANCE >= 1000 {
        FROM_BALANCE -= 1000;
        // --> Possible trap here
        TO_BALANCE += 1000;
    }
}
```

In the case of a trap, this results in losing the balance.

### Possible Solution

Need to have some tools to implement atomic operations.

NikVolf:
We've decided to go with https://github.com/gear-tech/gear/issues/108

which will make this unneeded

PR #634, opened 2022-02-07
Rollback state mechanism

### LINKED ISSUE #80: Memory should still be unallocated when panicking
see #79

### LINKED ISSUE #108: Roll back program state if submitted WASM produces trap
### Problem to Solve

Currently program state is saved regardless of execution result. It makes possible following unwanted conditions:
* Allocated memory pages may not be freed upon exit (#80)
* Program may end up in corrupted state

### Possible Solution

Prevent saving program state, outgoing and incoming messages if trap is produced. Another option is to provide default panic handler that would roll back state. Users will be able to opt-out from it and preserve state on trap.

### Notes

Following cases have to be considered:
* Should gas be deduced for unsuccessful executions? If so, how much? If not, could it be a potential DOS vulnerability?
* Program must not loose any incoming messages if state was not preserved. How should infinite loop of handling incoming message and producing a trap be prevented?
* Mechanism to recover from such errors should be present. E.g. updating submitted code with preserved state and/or messages.

breathx:
Resolves #80.
Resolves #108.
Provides under-the-hood logic for #614, making that issue useless.

Release notes: _to be written..._

@gear-tech/dev

### DIFF
--- a/core-backend/common/src/funcs.rs
+++ b/core-backend/common/src/funcs.rs
@@ -16,7 +16,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
-use crate::{EXIT_TRAP_STR, WAIT_TRAP_STR};
+use crate::{LEAVE_TRAP_STR, WAIT_TRAP_STR};
 use alloc::{string::String, vec, vec::Vec};
 use gear_core::{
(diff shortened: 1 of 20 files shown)


NikVolf (approved):
Great 👏

As a follow up, failed dispatch after wait should put message back to wait -- please create issue

PR #639, opened 2022-02-07
Introduce lazy pages

### LINKED ISSUE #379: Database-mapped memory pages
When loading program to handle one message, it is suboptimal to load all memory pages. Instead, they should be all empty initially and access guard (mprotect) should be put, and they should load from db only once accessed.

grishasobol:
Resolves #379 .
- Add gear runtime interface
- Add lazy pages support function, which can be called in runtime using gear runtime interface
- Now pages data can be loaded from storage in lazy manner:
   - This can be enabled or disabled in runtime.
   - Works not in all environments and OS.
- initial pages (or persistent pages) BTree has now Option as value, because some pages are supposed to be lazy and has no data currently. So, I add additional logic in several places to support this change. 
- Add lazy-pages runtime support in core-processor.
- Split setup_and_execute method into two: setup and execute. This is necessary to be able to setup lazy pages after memory instantiation.
- Change ExtInfo: field `pages` is changed to `accessed_pages`, which contains data only for accessed  during execution pages. If lazy pages are disabled, then suppose that all pages were accessed (at least we copy data from storage in each page).

NOTE: currently lazy-pages is enabled only in native runtime. To support it in wasm runtime we need to change embedded executor to host. In gear-tests it's also disabled.

@NikVolf @ukint-vs @shamilsan @breathx


lazy-pages/src/lib.rs
```diff
@@ -0,0 +1,187 @@
...
+    RELEASED_LAZY_PAGES.with(|x| x.borrow().get(&page).expect("Must have this page").to_vec())
+}
+
+/// Initialize lazy pages:
+/// 1) checks whether lazy pages is supported in current environment
+/// 2) set signals handler
+///
+/// # Safety
+/// use OS specific functions
+pub unsafe fn init_lazy_pages() -> bool {
```
gshep, line 153: Maybe it is better ro return `Result` with corresponding errors?

node/src/main.rs
```diff
@@ -24,5 +24,7 @@ mod command;
 mod rpc;
 
 fn main() -> sc_cli::Result<()> {
+    // gear_lazy_pages::set_lazy_pages_signals();
```
gshep, line 27: Unnecessary?

ri/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//! Runtime insterface for gear node
```
gshep, line 19: misprint

ri/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use sp_runtime_interface::runtime_interface;
+
+#[cfg(feature = "std")]
+use gear_core::memory::PAGE_SIZE;
```
gshep, line 26: Is'nt there circular dependency due to this?

gear-test/Cargo.toml
```diff
@@ -17,7 +17,7 @@ hex = "0.4.2"
 serde = { version = "*", features = ["derive"] }
 serde_json = "*"
 log = "0.4.14"
-parity-scale-codec = { version = "*", features = ["derive"], default-features = false }
+parity-scale-codec = { version = "2.2.0", features = ["derive"], default-features = false }
```
gshep, line 20: The last version in 2.x branch is 2.3.1

core-backend/sandbox/src/env.rs
```diff
@@ -44,16 +47,18 @@ impl<E: Ext> Default for SandboxEnvironment<E> {
 
 pub struct Runtime<E: Ext + Into<ExtInfo>> {
     pub(crate) ext: LaterExt<E>,
+    pub(crate) lazy_pages_enabled: bool,
     pub(crate) trap: Option<&'static str>,
 }
 
 impl<E: Ext + Into<ExtInfo> + 'static> Runtime<E> {
-    fn from_ext(ext: E) -> Self {
+    fn new(ext: E, lazy_pages_enabled: bool) -> Self {
```
gshep, line 55: I would like to go with `Option<pub struct EnableLazyPages>` here too.

NikVolf:
Good start

Would be great to make some effort so that `core-backend` is as much agnostic as possible to the fact if we have lazy pages turned on

core/src/memory.rs
```diff
@@ -115,13 +115,15 @@ pub trait Memory: Any {
...
-    fn set_pages(&self, pages: &BTreeMap<PageNumber, Box<PageBuf>>) -> Result<(), Error> {
+    fn set_pages(&self, pages: &BTreeMap<PageNumber, Option<Box<PageBuf>>>) -> Result<(), Error> {
         for (num, buf) in pages {
             let s = self.size() - 1.into();
             if s < *num {
                 self.grow(*num - s)?;
             }
-            self.write(num.offset(), &buf[..])?;
+            if let Some(buf) = buf {
+                self.write(num.offset(), &buf[..])?;
```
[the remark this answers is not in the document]
grishasobol: There is need to grow. Because this memory can be accessed by executor. So, we must allocate it first and then protect

core-backend/sandbox/src/env.rs
```diff
@@ -154,8 +192,70 @@ impl<E: Ext + Into<ExtInfo> + 'static> Environment<E> for SandboxEnvironment<E>
...
+            lazy_pages.iter().for_each(|p| {
+                common::save_page_lazy_info(prog_id_hash, *p);
+            });
+
+            gear_ri::gear_ri::mprotect_wasm_pages(
+                wasm_mem_begin_addr as u64,
+                &lazy_pages,
+                false,
+                false,
+                false,
```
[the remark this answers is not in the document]
grishasobol: You cannot use your own types in runtime interface. Only types which supports PassBy  in substrate. So, we can write a wrapper for this function, but I think it's bad practice.

gear-test/Cargo.toml
grishasobol: We cannot use 2.3.1 because it does not support String encode

ri/src/lib.rs
grishasobol: We don't use gear-ri in gear-core

core-backend/sandbox/src/env.rs
grishasobol: What ?? Do you hate bool type?

node/src/main.rs
grishasobol: Yep, removed this

common/src/native.rs
```diff
@@ -125,7 +129,10 @@ pub fn set_program(program: Program) {
         program
             .get_pages()
             .iter()
-            .map(|(num, buf)| (num.raw(), buf.to_vec()))
+            .map(|(num, buf)| {
+                let buf = buf.as_ref().expect("Strange");
```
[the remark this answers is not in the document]
grishasobol: Ok, changed it

core-backend/sandbox/src/env.rs
gshep: Yeah, I am. You caught me :smile:

gear-test/Cargo.toml
breathx: > We cannot use 2.3.1 because it does not support String encode

Even with "full" feature enabled?🧐

core-backend/sandbox/src/env.rs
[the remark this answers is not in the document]
gshep: Sorry I honestly didn't understand. `gear-runtime-interface` is just a usual dependency so it can export any entity.

Anyway I'm not insisting.

core-backend/sandbox/src/env.rs
breathx: If you don't like booleans so much, what about passing here more obvious enum LazyPages with Enabled, Disabled variants instead of strange Option

gshep: As I commented another place: `bool` is not informative type. It is ok to use it in several specific places, for example as function return type. However `bool` creates additional cognitive issues and requires comments almost in every place.

Compare this call - `Runtime::<_>::new(E, false)` - to this one - `Runtime::<_>::new(E, Some(EnableLazyPages))`. What doues mean `false` in the former case? Developer has to read function signature and return back to the place. Or it requires comment this call.

From my point of view the second call doesn't have such disadvantages - all useful information is encoded in type name.
Frankly speaking using `Option` is not so good too. The best solution is to introduce specific `enum` though it would lead to code-bloat.

In any case I'm not insisting to use types. It's up to you. I just tried to explain disadvantages of `bool` =)

gshep: > If you don't like booleans so much, what about passing here more obvious enum LazyPages with Enabled, Disabled variants instead of strange Option

Yep, this is the best "option"!

grishasobol:
> Good start
> 
> Would be great to make some effort so that `core-backend` is as much agnostic as possible to the fact if we have lazy pages turned on

@NikVolf , I moved all to core-executor. Note: I's needed to split setup_and_execute in core-backend, into two functions: setup and execute =)

core-backend/sandbox/src/env.rs
grishasobol: In my opinion: the simplest code == the best code. It's not important option or enum: you create new entities in code. Considering your example I can say the same words: to identify what is EnableLazyPages, I need to go and see definition. I don't now by default, that it's empty struct without implementations. The same words about enum.

grishasobol: I moved lazy-pages logic from sandbox to core-processor. So, now this field is in Ext. I make Option, because it's shorter))

gear-test/Cargo.toml
grishasobol: Yep) thank you - full feature fix this

lazy-pages/src/lib.rs
grishasobol: Yes, good idea, will do this today

core-backend/sandbox/src/env.rs
```diff
@@ -154,7 +135,24 @@ impl<E: Ext + Into<ExtInfo> + 'static> Environment<E> for SandboxEnvironment<E>
...
+        Ok(())
+    }
+
+    /// Setup external environment and run closure.
+    ///
+    /// Setup external environment by providing `ext`, run nenwly initialized instance created from
+    /// provided `module`, do anything inside a `func` delegate.
+    ///
+    /// This will also set the beginning of the memory region to the `static_area` content _after_
+    /// creatig instance.
```
gshep, line 150: looks like the comment should be edited

core-backend/sandbox/src/env.rs
```diff
@@ -154,7 +135,24 @@ impl<E: Ext + Into<ExtInfo> + 'static> Environment<E> for SandboxEnvironment<E>
...
+    fn execute(&mut self, entry_point: &str) -> Result<BackendReport, BackendError> {
+        let instance = self.instance.as_mut().expect("Must have instance");
+        let runtime = self.runtime.as_mut().expect("Must have runtime");
```
gshep, line 153: Better to return corresponding `Err`

core-processor/src/lazy_pages.rs
```diff
@@ -0,0 +1,136 @@
...
+
+    // Removes protections from lazy pages
+    let lazy_pages = gear_ri::gear_ri::get_wasm_lazy_pages_numbers();
+    gear_ri::gear_ri::mprotect_wasm_pages(
+        wasm_mem_begin_addr as u64,
+        &lazy_pages,
+        true,
+        true,
+        false,
+    );
```
gshep, line 110: it looks like you can call `remove_lazy_pages_prot` there

lazy-pages/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+    /// Key in storage for each lazy page
+    static LAZY_PAGES_INFO: RefCell<BTreeMap<u32, Vec<u8>>> = RefCell::new(BTreeMap::new());
+    /// Page data, which has been in storage before current execution.
+    /// For each lazy page, which has been accessed.
+    static RELEASED_LAZY_PAGES: RefCell<BTreeMap<u32, Vec<u8>>> = RefCell::new(BTreeMap::new());
+}
+
+/// Sigsegv (or sigbus for macos) handler.
+/// Before contract execution some pages from wasm memory buffer are protected,
+/// and connot be accessed anyhow. When wasm executer tries to access one of these pages,
```
gshep, line 48: misprint `connot`

gshep:
Good job! :tada:

lazy-pages/src/lib.rs
grishasobol: A, no. When start to this - release that actually it's unused. Cannot find, a point at which runtime should now the problem in native part. Information whether we can or not enable lazy pages must be enough

core-backend/sandbox/src/env.rs
grishasobol: Ok, change and move comment to trait

core-backend/sandbox/src/env.rs
grishasobol: It, is not backend error, it's software engineer mistake .

core-backend/wasmtime/src/env.rs
```diff
@@ -312,6 +313,14 @@ impl<E: Ext + Into<ExtInfo>> Environment<E> for WasmtimeEnvironment<E> {
             gas_amount: self.ext.unset().into().gas_amount,
         })?;
 
+        self.instance.replace(instance);
+
+        Ok(())
+    }
+
+    fn execute(&mut self, entry_point: &str) -> Result<BackendReport, BackendError> {
+        let instance = self.instance.as_mut().expect("Must have instance");
```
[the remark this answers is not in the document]
grishasobol: It, is not backend error, it's software engineer mistake .

lazy-pages/src/lib.rs
grishasobol: Fix

gshep (approved):

PR #654, opened 2022-02-11
Implement syscall `exit`

### LINKED ISSUE #453: New syscall `sys::gr_exit`

gshep:
Resolves #453

@gear-tech/dev 

**Release notes**: add new syscall `exit`. It allows to preemptively stop the execution of a program and remove the program from the storage. All available to program value is transferred to the address specified in `exit` argument.

### DIFF
--- a/core-backend/sandbox/src/env.rs
+++ b/core-backend/sandbox/src/env.rs
@@ -76,6 +76,7 @@ impl<E: Ext + Into<ExtInfo> + 'static> SandboxEnvironment<E> {
         env_builder.add_host_func("env", "free", funcs::free);
         env_builder.add_host_func("env", "gr_block_height", funcs::block_height);
         env_builder.add_host_func("env", "gr_block_timestamp", funcs::block_timestamp);
+        env_builder.add_host_func("env", "gr_exit", funcs::exit);
         env_builder.add_host_func("env", "gr_exit_code", funcs::exit_code);
         env_builder.add_host_func("env", "gr_send", funcs::send);
         env_builder.add_host_func("env", "gr_send_commit", funcs::send_commit);
@@ -159,14 +160,18 @@ impl<E: Ext + Into<ExtInfo> + 'static> Environment<E> for SandboxEnvironment<E> 
         let info: ExtInfo = runtime.ext.unset().into();
 
         let termination = if res.is_err() {
-            let mut reason = None;
-
-            if let Some(trap) = runtime.trap {
-                if common_funcs::is_wait_trap(trap) {
-                    reason = Some(TerminationReason::Manual { wait: true });
+            let reason = if let Some(trap) = runtime.trap {
+                if let Some(value_dest) = info.exit_argument {
+                    Some(TerminationReason::Exit(value_dest))
+                } else if common_funcs::is_wait_trap(trap) {
+                    Some(TerminationReason::Wait)
                 } else if common_funcs::is_leave_trap(trap) {
-                    reason = Some(TerminationReason::Manual { wait: false });
+                    Some(TerminationReason::Leave)
+                } else {
+                    None
                 }
+            } else {
+                None
             };
 
             reason.unwrap_or_else(|| TerminationReason::Trap {
--- a/core-backend/wasmtime/src/env.rs
+++ b/core-backend/wasmtime/src/env.rs
@@ -57,6 +57,7 @@ impl<E: Ext + 'static> WasmtimeEnvironment<E> {
         result.add_func_to_i32("gr_exit_code", funcs::exit_code);
         result.add_func_into_i64("gr_gas_available", funcs::gas_available);
         result.add_func_i32_i32("gr_debug", funcs::debug);
+        result.add_func_i32("gr_exit", funcs::exit);
         result.add_func_i32("gr_msg_id", funcs::msg_id);
         result.add_func_i32("gr_program_id", funcs::program_id);
         result.add_func_i32_i32_i32("gr_read", funcs::read);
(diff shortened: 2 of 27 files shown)


pallets/gear/src/tests.rs
```diff
@@ -1369,12 +1369,115 @@ fn wake_messages_after_program_inited() {
...
+        assert!(!Gear::is_initialized(program_id));
+
+        let actual_n = Gear::mailbox(USER_1)
+            .map(|t| t.into_values().fold(0usize, |i, _| i + 1))
+            .unwrap_or(0);
+
+        assert_eq!(actual_n, 0);
+
+        // Program is removed and can be submitted again
+        assert_ok!(GearPallet::<Test>::submit_program(
```
NikVolf, line 1404: I think it can be submitted regardless if previous one `exit`-ed?

gshep: The salt is the same. So if I'm not mistaken ID of the program should be the same and an error should be returned.

tests/exit-init/src/lib.rs
```diff
@@ -0,0 +1,26 @@
+#![cfg_attr(not(feature = "std"), feature(alloc_error_handler))]
+#![cfg_attr(not(feature = "std"), feature(const_btree_new))]
```
shamilsan, line 2: Where do you use `BTreeMap::new()` here?

core-processor/src/ext.rs
```diff
@@ -207,6 +210,15 @@ impl EnvExt for Ext {
         self.message_context.current().source()
     }
 
+    fn exit(&mut self, address: ProgramId) -> Result<(), &'static str> {
```
shamilsan, line 213: Isn't better to keep common names across the code?
```suggestion
    fn exit(&mut self, value_destination: ProgramId) -> Result<(), &'static str> {
```

tests/exit-init/src/lib.rs
gshep: Good catch: it's not used =)

NikVolf (approved):

shamilsan (approved):
Awesome 💟

core-backend/sandbox/src/env.rs
```diff
@@ -159,14 +160,18 @@ impl<E: Ext + Into<ExtInfo> + 'static> Environment<E> for SandboxEnvironment<E>
         let info: ExtInfo = runtime.ext.unset().into();
 
         let termination = if res.is_err() {
-            let mut reason = None;
-
-            if let Some(trap) = runtime.trap {
-                if common_funcs::is_wait_trap(trap) {
-                    reason = Some(TerminationReason::Manual { wait: true });
+            let reason = if let Some(trap) = runtime.trap {
+                if let Some(value_dest) = info.exit_argument {
```
breathx, line 164: I would place here something like `common::is_exit_trap` just to keep the flow. It also may cause some issues in future, if we decide to remove argument from `gr_exit`

core-backend/sandbox/src/env.rs
```diff
@@ -159,14 +160,18 @@ impl<E: Ext + Into<ExtInfo> + 'static> Environment<E> for SandboxEnvironment<E>
         let info: ExtInfo = runtime.ext.unset().into();
 
         let termination = if res.is_err() {
-            let mut reason = None;
-
-            if let Some(trap) = runtime.trap {
-                if common_funcs::is_wait_trap(trap) {
-                    reason = Some(TerminationReason::Manual { wait: true });
+            let reason = if let Some(trap) = runtime.trap {
```
breathx, line 163: Can't we somehow simplify this matching? @shamilsan

core-backend/wasmtime/src/env.rs
```diff
@@ -323,17 +324,21 @@ impl<E: Ext + Into<ExtInfo>> Environment<E> for WasmtimeEnvironment<E> {
         let info: ExtInfo = self.ext.unset().into();
 
         let termination = if let Err(e) = &res {
-            let mut reason = None;
-
-            if let Some(trap) = e.downcast_ref::<Trap>() {
+            let reason = if let Some(trap) = e.downcast_ref::<Trap>() {
```
breathx, line 327: I see u don't love mutables

core-processor/src/common.rs
```diff
@@ -163,6 +165,14 @@ pub enum JournalNote {
         /// Amount of gas burned.
         amount: u64,
     },
+    /// Exit the program.
+    ExitDispatch {
+        /// Id of the program called `exit`.
+        id_exited: ProgramId,
```
breathx, line 171: why not just `program_id`?

core-processor/src/handler.rs
```diff
@@ -69,4 +74,8 @@ pub fn handle_journal(
             handler.update_page(program_id, page_number, data);
         }
     }
+
+    for (id_exited, value_destination) in exit_list {
```
breathx, line 78: what's the reason for that vec?

core-backend/sandbox/src/env.rs
shamilsan: We could write:

```rust
let reason = match runtime.trap {
    ...
    Some(trap) if common_funcs::is_wait_trap(trap) => Some(TerminationReason::Wait),
    Some(trap) if common_funcs::is_leave_trap(trap) => Some(TerminationReason::Leave),
    ...
    _ => None,
```

But I don't think it's much better :)

PR #675, opened 2022-02-17
fix build on windows 

grishasobol:
fix build on windows
fix problems with std feature and substrate + gear


lazy-pages/src/lib.rs
```diff
@@ -56,6 +56,10 @@ thread_local! {
 /// instruction, which cause signal. Now memory which this instruction accesses
 /// is not protected and with correct data.
 extern "C" fn handle_sigsegv(_x: i32, info: *mut siginfo_t, _z: *mut c_void) {
+    if !cfg!(unix) {
+        panic!("unsupported");
```
breathx, line 60: Isn't it too hard to panic in such case? What about placing here `return` statement?

NikVolf: This code shouldn't ever hit @grishasobol ?

runtime-interface/src/lib.rs
```diff
@@ -27,6 +27,48 @@ use gear_core::memory::PAGE_SIZE;
...
+
+#[cfg(not(unix))]
+unsafe fn sys_mprotect_wasm_pages(
+    from_ptr: u64,
+    pages_nums: &[u32],
+    prot_read: bool,
+    prot_write: bool,
+    prot_exec: bool,
+) {
+    panic!("unsupported");
```
breathx, line 69: Still same question about panics

shamilsan: Yes, if it is unreachable, better would be to use `unreachable!(message)`.

.github/workflows/win.yml
```diff
@@ -36,3 +36,4 @@ jobs:
         with:
           command: build
           args: -p gear-node --release --target x86_64-pc-windows-msvc
+
```
shamilsan, line 39: Extra empty line?
```suggestion
```

grishasobol: Ok

runtime-interface/src/lib.rs
grishasobol: Ok

runtime-interface/src/lib.rs
```diff
@@ -27,6 +27,48 @@ use gear_core::memory::PAGE_SIZE;
...
+
+#[cfg(not(unix))]
+unsafe fn sys_mprotect_wasm_pages(
+    from_ptr: u64,
+    pages_nums: &[u32],
+    prot_read: bool,
+    prot_write: bool,
+    prot_exec: bool,
+) {
+    unreachable!("unsupported OS for pages protectections");
```
shamilsan, line 69: Typo?
```suggestion
    unreachable!("unsupported OS for pages protection");
```

PR #678, opened 2022-02-17
Provide script and CI checking for spec version

breathx:
@gear-tech/dev


breathx:
> Looks great if it really works 😄

Hope so :D

It checks that here are some changes in files from `common`, `core`, `core-backend`, `node`, `pallets`, `runtime-interface`, `lazy-pages`dirs, compares your spec version with `origin/master` spec version and exits with 1 if there is some troubles with it

PR #681, opened 2022-02-18
Fix clippy master

SabaunT:


breathx (changes requested):
I'll push here checking for labels while running spec checking job to prevent redundant spec version bumping

breathx (approved):
Brilliant now

shamilsan:
> Brilliant now

What about checking in master?

PR #684, opened 2022-02-18
Save message context over wake logic

### LINKED ISSUE #496: Outgoing uncommitted packets loose after `wait`/`wake`
### Problem

Uncommitted outgoing packets aren't kept in the waitlist when calling `wait` after `send_init` and `send_push` (or `reply_push`).

### Steps

1. Call `send_init` and `send_push`
2. Call `send_and_wait_for_reply`
3. Call `send_commit` after getting the reply.
4. Message sending fails.

Dedicated example: https://github.com/gear-tech/gear/tree/master/examples/incomplete_async_payloads

### Possible Solution

Keep outgoing packets on the waitlist as we do it for the incoming messages.

breathx:
Resolves #496

@gear-tech/dev


common/src/lib.rs
```diff
@@ -187,22 +187,38 @@ pub trait DAGBasedLedger {
...
     pub kind: DispatchKind,
     pub message: Message,
+    pub payload_store: Option<PayloadStore>,
 }
 
 impl Dispatch {
     pub fn new_init(message: Message) -> Self {
         let kind = DispatchKind::Init;
-        Dispatch { message, kind }
+        let payload_store: Option<PayloadStore> = None;
```
shamilsan, line 196: No need to point the type.
```suggestion
        let payload_store = None;
```

common/src/lib.rs
```diff
@@ -187,22 +187,38 @@ pub trait DAGBasedLedger {
...
+        Dispatch {
+            message,
+            kind,
+            payload_store,
```
shamilsan, line 200: Or even (as you do in some places):
```suggestion
            payload_store: None,
```

runtime/src/lib.rs
```diff
@@ -124,7 +124,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     // The version of the runtime specification. A full node will not attempt to use its native
     //   runtime in substitute for the on-chain Wasm runtime unless all of `spec_name`,
     //   `spec_version`, and `authoring_version` are the same between Wasm and native.
-    spec_version: 200,
+    spec_version: 210,
```
SabaunT, line 127: 210 is on master now. Maybe 220, for example?

SabaunT (changes requested):
:fire:

runtime/src/lib.rs
breathx: Yeah, already isn't actual value

SabaunT (approved):

ukint-vs (approved):

PR #685, opened 2022-02-18
Add no-value nodes to gas tree

NikVolf:


common/src/lib.rs
```diff
@@ -176,10 +176,22 @@ pub trait DAGBasedLedger {
...
-    fn split(key: Self::Key, new_key: Self::Key, amount: Self::Balance) -> DispatchResult;
+    fn split_with_value(
+        key: Self::Key,
+        new_key: Self::Key,
+        amount: Self::Balance,
+    ) -> DispatchResult;
+
+    /// Split underlying value.
+    ///
+    /// If `key` does not identify any value or the `amount` exceeds what's locked under that key,
```
shamilsan, line 191: There is no `amount` in this function.

ekovalev (approved):

PR #687, opened 2022-02-19
gear-wasm-builder: Always build WASM in release mode

shamilsan:
Minor fixes in `gear-wasm-builder`.

- Build WASM in release mode regardless of the selected profile.
- Beautify `wasm_binary.rs` a little bit. 

@gear-tech/dev

### DIFF
--- a/utils/wasm-builder/src/wasm_project.rs
+++ b/utils/wasm-builder/src/wasm_project.rs
@@ -99,11 +99,16 @@ impl WasmProject {
         lib.insert("name".into(), crate_info.snake_case_name.into());
         lib.insert("crate-type".into(), vec!["cdylib".to_string()].into());
 
+        let mut dev_profile = Table::new();
+        dev_profile.insert("panic".into(), "abort".into());
+
         let mut release_profile = Table::new();
         release_profile.insert("lto".into(), true.into());
         release_profile.insert("opt-level".into(), "s".into());
+        release_profile.insert("panic".into(), "abort".into());
 
         let mut profile = Table::new();
+        profile.insert("dev".into(), dev_profile.into());
         profile.insert("release".into(), release_profile.into());
 
         let mut crate_package = Table::new();
@@ -153,7 +158,7 @@ impl WasmProject {
             .expect("Run `WasmProject::create_project()` first");
         let from_path = self
             .out_dir
-            .join(&self.wasm_subdir)
+            .join("target/wasm32-unknown-unknown/release")
             .join(format!("{}.wasm", &file_base_name));
         let to_dir = self.original_dir.join(&self.wasm_subdir);
         fs::create_dir_all(&to_dir)?;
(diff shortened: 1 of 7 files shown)


utils/wasm-builder/src/wasm_project.rs
```diff
@@ -171,11 +176,13 @@ impl WasmProject {
         fs::write(
             &wasm_binary_rs,
             format!(
-                r#"
-                    pub const WASM_BINARY: &[u8] = include_bytes!("{}");
-                    pub const WASM_BINARY_OPT: &[u8] = include_bytes!("{}");
-                    pub const WASM_BINARY_META: &[u8] = include_bytes!("{}");
-                "#,
+                r#"#[allow(dead_code)]
```
breathx, line 179: I prefer `#[allow(unused)]` here, because this code isn't dead actually

utils/wasm-builder/src/wasm_project.rs
```diff
@@ -99,11 +99,16 @@ impl WasmProject {
         lib.insert("name".into(), crate_info.snake_case_name.into());
         lib.insert("crate-type".into(), vec!["cdylib".to_string()].into());
 
+        let mut dev_profile = Table::new();
+        dev_profile.insert("panic".into(), "abort".into());
+
         let mut release_profile = Table::new();
         release_profile.insert("lto".into(), true.into());
         release_profile.insert("opt-level".into(), "s".into());
+        release_profile.insert("panic".into(), "abort".into());
```
breathx, line 108: It's not the best choice, because in future we will be able set "unwind" for wasm, once it's been released.

utils/wasm-builder/src/wasm_project.rs
```diff
@@ -99,11 +99,16 @@ impl WasmProject {
...
 
         let mut profile = Table::new();
+        profile.insert("dev".into(), dev_profile.into());
         profile.insert("release".into(), release_profile.into());
```
breathx, line 112: I would go with something like this. Always optimized, always light-weight.
```suggestion
        profile.insert("dev".into(), release_profile.into());
        profile.insert("release".into(), release_profile.into());
```

utils/wasm-builder/src/wasm_project.rs
```diff
@@ -171,11 +176,13 @@ impl WasmProject {
...
-                    pub const WASM_BINARY_OPT: &[u8] = include_bytes!("{}");
-                    pub const WASM_BINARY_META: &[u8] = include_bytes!("{}");
-                "#,
+                r#"#[allow(dead_code)]
+pub const WASM_BINARY: &[u8] = include_bytes!("{}");
+#[allow(dead_code)]
+pub const WASM_BINARY_OPT: &[u8] = include_bytes!("{}");
+#[allow(dead_code)]
+pub const WASM_BINARY_META: &[u8] = include_bytes!("{}");
+"#,
```
breathx, line 185:
```suggestion

"#,
```

breathx (approved):
Things from comments are up to you, but I would prefer them more. Otherwise good job.

utils/wasm-builder/src/wasm_project.rs
shamilsan: No need for an empty new line as we already have a carriage return.

utils/wasm-builder/src/wasm_project.rs
shamilsan: Agree, fixed.

utils/wasm-builder/src/wasm_project.rs
shamilsan: Copied it from the substrate-wasm-builder, but removed it back 😄

utils/wasm-builder/src/wasm_project.rs
shamilsan: Great advice, fixed.

PR #689, opened 2022-02-19
`gstd`: Don't panic if signal not found

ukint-vs:
@gear-tech/dev

### DIFF
--- a/gstd/src/async_runtime/signals.rs
+++ b/gstd/src/async_runtime/signals.rs
@@ -68,7 +68,7 @@ impl WakeSignals {
             }
             crate::exec::wake(signal.message_id);
         } else {
-            crate::debug!("Somehow received reply for the message we never sent");
+            crate::debug!("Received reply for the message we don't expect reply to or already processed before");
         }
     }
 
--- a/gstd/src/async_runtime/signals.rs
+++ b/gstd/src/async_runtime/signals.rs
@@ -61,16 +61,13 @@ impl WakeSignals {
     }
 
     pub fn record_reply(&mut self) {
-        let mut signal = self
-            .signals
-            .get_mut(&crate::msg::reply_to())
-            .expect("Somehow received reply for the message we never sent");
-
-        signal.payload = Some((crate::msg::load_bytes(), crate::msg::exit_code()));
-        if let Some(waker) = &signal.waker {
-            waker.wake_by_ref();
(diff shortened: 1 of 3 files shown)


gstd/src/async_runtime/signals.rs
```diff
@@ -61,16 +61,13 @@ impl WakeSignals {
...
-        signal.payload = Some((crate::msg::load_bytes(), crate::msg::exit_code()));
-        if let Some(waker) = &signal.waker {
-            waker.wake_by_ref();
+        if let Some(signal) = self.signals.get_mut(&crate::msg::reply_to()) {
+            signal.payload = Some((crate::msg::load_bytes(), crate::msg::exit_code()));
+            if let Some(waker) = &signal.waker {
+                waker.wake_by_ref();
+            }
+            crate::exec::wake(signal.message_id);
         }
```
gshep, line 70: Maybe we need to make some record to log in else-case?

SabaunT (approved):

gstd/src/async_runtime/signals.rs
ukint-vs: > Maybe we need to make some record to log in else-case?

This will increase execution cost

breathx: So with `gstd::debug!(..)` only, because it has no-op with `gstd`'s `debug` feature

gstd/src/async_runtime/signals.rs
```diff
@@ -61,16 +61,15 @@ impl WakeSignals {
...
-        if let Some(waker) = &signal.waker {
-            waker.wake_by_ref();
+        if let Some(signal) = self.signals.get_mut(&crate::msg::reply_to()) {
+            signal.payload = Some((crate::msg::load_bytes(), crate::msg::exit_code()));
+            if let Some(waker) = &signal.waker {
+                waker.wake_by_ref();
+            }
+            crate::exec::wake(signal.message_id);
+        } else {
+            crate::debug!("Somehow received reply for the message we never sent");
```
breathx, line 71:
```suggestion
            crate::debug!("Received reply for the message we don't expect reply to or already processed before");
```

NikVolf (changes requested):
It should panic and shouldn't silently drop the message

NikVolf (approved):

NikVolf:
This bug when handling is broken after panic should be documented and fixed

PR #691, opened 2022-02-21
Pause program API

### LINKED ISSUE #652: Extrinsics `resume_program` and program destruction
### Problem to Solve

Allow users to stop the execution of a program but keep its ID and hash of the state. So program goes to the temporary storage and takes less space on-chain.

Second feature is to restore program from such kind of storage. Caller has to provide memory-pages (state) and enough value to continue program execution.

### Possible Solution

API in `common` crate + tests. On node side there will be two extrinsics using this API.

### Notes

after #453

### LINKED ISSUE #707: `set_program_terminated_status` doesn't remove memory pages
### Problem

Memory pages of the terminated program isn't removed

### Steps

1.
2.
3.

### Possible Solution

_No response_

### Notes

_No response_

### Relevant Log Output

<p>

```
<Paste your log here>
```

</p>
</details>

gshep:
Resolves #652, resolves #707

- `pallet_gear_program::pause_program`
- `pallet_gear_program::resume_program_impl`
- test cases in common crate
- extrinsic `resume_program` in pallet_gear_program
- pallet tests
- benchmarks

**Release notes**: Implemented internal API that allows to pause/resume programs execution. Fixed several tiny bugs and refactored the code a bit. Added new extrinsic `resume_program`. It provides an ability to resume a program by any user/program with enough funds.

This is the next step to implement rent charging for program resources.


pallets/gear/src/lib.rs
```diff
@@ -941,6 +955,52 @@ pub mod pallet {
...
+        #[pallet::weight(<T as Config>::WeightInfo::resume_program(memory_pages.values().map(|p| p.len() as u32).sum()))]
+        pub fn resume_program(
+            origin: OriginFor<T>,
+            program_id: H256,
+            memory_pages: BTreeMap<u32, Vec<u8>>,
+            value: BalanceOf<T>,
+        ) -> DispatchResultWithPostInfo {
+            let account = ensure_signed(origin)?;
+
+            ensure!(!value.is_zero(), Error::<T>::ResumeProgramNotEnoughValue);
```
gshep, line 980: This is a place I'm not certain with.

I suggest that it is nice to have some constant representing minimal rent for several blocks and check that `value` is greater or equal that constant. However it could be introduced in the corresponding task.

common/src/lib.rs
```diff
@@ -787,4 +803,343 @@ mod tests {
...
+    #[test]
+    fn pause_program_works() {
+        sp_io::TestExternalities::new_empty().execute_with(|| {
+            let code = b"pretended wasm code".to_vec();
+            let code_hash: H256 = sp_io::hashing::blake2_256(&code[..]).into();
+            set_code(code_hash, &code);
+
+            let static_pages = 16;
+            let memory_pages = {
+                let mut pages = BTreeMap::<u32, Vec<u8>>::new();
```
shamilsan, line 816: Isn't better to not point generic types explicitly?
```suggestion
            let static_pages: u32 = 16;
            let memory_pages = {
                let mut pages = BTreeMap::new();
```

common/src/lib.rs
```diff
@@ -787,4 +803,343 @@ mod tests {
...
+
+            assert_ok!(set_program_terminated_status(program_id));
+
+            assert_err!(pause_program(program_id), PauseError::ProgramTerminated);
+        });
+    }
+
+    fn create_uninitialized_program_messages(
+        static_pages: u32,
+    ) -> (H256, H256, H256, H256, H256, BTreeMap<u32, Vec<u8>>) {
```
shamilsan, line 942: The function signature is quite ugly, better would use a struct as a return parameter.

common/src/pause.rs
```diff
@@ -0,0 +1,125 @@
+// This file is part of Gear.
+
+// Copyright (C) 2021 Gear Technologies Inc.
```
shamilsan, line 3:
```suggestion
// Copyright (C) 2022 Gear Technologies Inc.
```

common/src/pause.rs
```diff
@@ -0,0 +1,125 @@
...
+#[derive(Clone, Debug, Decode, Encode, PartialEq, TypeInfo)]
+struct PausedProgram {
+    program_id: H256,
+    program: ActiveProgram,
+    pages_hash: H256,
+    wait_list: Vec<QueuedDispatch>,
+    waiting_init: Vec<H256>,
+}
+
+fn decode_dispatch_tuple(_: &[u8], value: &[u8]) -> Result<(QueuedDispatch, u32), codec::Error> {
```
shamilsan, line 31: What is the first argument for?

common/src/pause.rs
```diff
@@ -0,0 +1,125 @@
...
+    let previous_key = prefix.clone();
+
+    let paused_program = PausedProgram {
+        program_id,
+        pages_hash: memory_pages_hash(
+            &get_program_pages(program_id, program.persistent_pages.clone())
+                .expect("pause_program: active program exists, therefore pages do"),
+        ),
+        program,
+        wait_list: PrefixIterator::<_, ()>::new(prefix, previous_key, decode_dispatch_tuple)
```
ekovalev, line 61: Why not use `frame_support::storage::KeyPrefixIterator` if you only need the key and discard the value? We have a similar example in the `usage` pallet (https://github.com/gear-tech/gear/blob/master/pallets/usage/src/offchain.rs#L124)

gshep: If I'm not mistaken the situation is the opposite - only value is used.

common/src/pause.rs
gshep: This is the raw key; will rename

pallets/gear/src/lib.rs
```diff
@@ -774,7 +788,7 @@ pub mod pallet {
             T::Currency::reserve(&who, value.unique_saturated_into())
                 .map_err(|_| Error::<T>::NotEnoughBalanceForReserve)?;
 
-            if common::program_exists(destination) {
+            if common::program_exists(destination) | common::paused_program_exists(destination) {
```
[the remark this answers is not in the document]
gshep: Good remark, thanks! I check how to refactor it.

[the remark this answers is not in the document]
gshep: I thought about it a bit and it looks like current solution with different storage key is better. To store under the same key an enumeration has to be introduced, however we already have `Program`, `ActiveProgram`, `TerminatedProgram` and so on. However if you/team insists to refactor this way - I'll change it.

Another point is that even if `PausedProgram` would be merged into the same storage key we still could need such function to determine if the program paused/running.

At the moment I could suggest to call `paused_program_exists` by `program_exists`. This will eliminate such conditions and fits the current business logic.

What should I do?

ekovalev:
Generally looks good to me.
My only question is that we have agreed on moving all the code that needs to access and modify storage from `common` into pallets (to be able use Substrate built-in mechanisms as opposed to using low-level `sp_io::storage` API) keeping in `common` only some traits definitions that might be shared between components. This is rather a mid-term (or even long-term) task so I guess we could keep this PR as is. Or should we, @NikVolf, @breathx ?

tests/proxy/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+#![cfg_attr(not(feature = "std"), feature(alloc_error_handler))]
```
breathx, line 19: Is this necessary?
```suggestion
```

gshep:
> Generally looks good to me. My only question is that we have agreed on moving all the code that needs to access and modify storage from `common` into pallets (to be able use Substrate built-in mechanisms as opposed to using low-level `sp_io::storage` API) keeping in `common` only some traits definitions that might be shared between components. This is rather a mid-term (or even long-term) task so I guess we could keep this PR as is. Or should we, @NikVolf, @breathx ?

we discussed it with @breathx  and decided to go with the new pallet.

common/src/pause.rs
```diff
@@ -0,0 +1,125 @@
...
+#[derive(Clone, Debug, Decode, Encode, PartialEq, TypeInfo)]
+struct PausedProgram {
+    program_id: H256,
+    program: ActiveProgram,
+    pages_hash: H256,
+    wait_list: Vec<QueuedDispatch>,
+    waiting_init: Vec<H256>,
+}
+
+fn decode_dispatch_tuple(_key: &[u8], value: &[u8]) -> Result<(QueuedDispatch, u32), codec::Error> {
```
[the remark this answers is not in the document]
gshep: Hm, it is used to move messages from wait list

tests/proxy/src/lib.rs
gshep: removed

pallets/gear-program/README.md
```diff
@@ -0,0 +1,11 @@
+# Tracking and charging fee for on-chain resources usage
```
ekovalev, line 1: This should probably be replaced with the information relevant to the pallet in question.

pallets/gear-program/src/lib.rs
```diff
@@ -0,0 +1,61 @@
...
+    #[pallet::pallet]
+    #[pallet::generate_store(pub(super) trait Store)]
+    pub struct Pallet<T>(_);
+
+    #[pallet::error]
+    pub enum Error<T> {}
+
+    #[pallet::storage]
+    #[pallet::unbounded]
+    pub(crate) type PausedPrograms<T: Config> = StorageMap<_, Identity, H256, pause::PausedProgram>;
```
ekovalev, line 54: Wouldn't it make more sense to make this struct a `ValueQuery` kind? That way you wouldn't have to check if it's `None` every time you call `get()` on it.

pallets/gear-program/src/pause.rs
```diff
@@ -0,0 +1,119 @@
...
+fn decode_dispatch_tuple(_key: &[u8], value: &[u8]) -> Result<(QueuedDispatch, u32), codec::Error> {
+    <(QueuedDispatch, u32)>::decode(&mut &*value)
+}
+
+fn memory_pages_hash(pages: &BTreeMap<u32, Vec<u8>>) -> H256 {
+    pages.using_encoded(sp_io::hashing::blake2_256).into()
+}
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum PauseError {
```
ekovalev, line 44: Any particular reason you opted out of the Substrate idiomatic way of defining errors in pallets via `pallet::error` macro?
```
#[pallet::error]
pub enum Error<T> {
    ...
}
```

pallets/gear/src/lib.rs
```diff
@@ -934,12 +952,61 @@ pub mod pallet {
...
+            let account = ensure_signed(origin)?;
+
+            ensure!(!value.is_zero(), Error::<T>::ResumeProgramNotEnoughValue);
+
+            GearProgramPallet::<T>::resume_program(
+                program_id,
+                memory_pages,
+                <frame_system::Pallet<T>>::block_number().unique_saturated_into(),
+            )
+            .map_err(|e| match e {
```
ekovalev, line 991: Why not just return the respective error from the `gear-program` pallet (provided the public functions there themselves return a properly defined `pallet::error`s that can be sent to an external caller)?
For instance, we don't map errors from the balances pallet - just return as is.

ekovalev:
Looks cool! There's still room for some minor improvements though, like consider moving the newly introduced extrinsics from `gear` pallet to `gear-program` where they belong. This might go against our initial idea of having everything that has to do with smart contracts deployment and usage in the `gear` pallet but the upside (in my opinion) is that the code would become more readable.

pallets/gear-program/src/pause.rs
gshep: To tell the truth no real reason. Currently this function is not called anywhere except tests so I would like to refactor it within the next task.

ekovalev (approved):

breathx (changes requested):
Request changes in order to remove waitlist for paused programs. These messages should be removed and replied by system with error exit code.

breathx:
Despite this request, looks like good part of removing `gear-common` in program-management sphere

pallets/gear-program/README.md
```diff
@@ -0,0 +1,14 @@
...
+## Interface
+
+### Dispatchable Functions
+
+### Functions
+
+* `pause_program` - pause running program
+* `resume_program` - resumes previously paused program
+* `paused_program_exists` - checks if the paused program with specified id exists
+* `program_exists` - checks if the program with specified id exists
```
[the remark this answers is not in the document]
gshep: Frankly speaking `program_exists` is already implemented this way:
```rust
pub fn program_exists(program_id: H256) -> bool {
        common::program_exists(program_id) | Self::paused_program_exists(program_id)
    }
```

Anyway will rename the function to `program_paused`.

pallets/gear-program/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+    #[pallet::pallet]
+    #[pallet::generate_store(pub(super) trait Store)]
+    pub struct Pallet<T>(_);
+
+    #[pallet::event]
+    #[pallet::generate_deposit(pub(super) fn deposit_event)]
+    pub enum Event<T: Config> {
+        /// Program has been successfully resumed
+        ProgramResumed(H256),
+    }
```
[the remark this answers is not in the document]
gshep: Thanks for the remark! Will add

pallets/gear-program/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+        /// - `value`: balance to be transferred to the program once it's been resumed.
+        ///
+        /// - `ProgramResumed(H256)` in the case of success.
+        #[frame_support::transactional]
+        #[pallet::weight(<T as Config>::WeightInfo::resume_program(memory_pages.values().map(|p| p.len() as u32).sum()))]
+        pub fn resume_program(
+            origin: OriginFor<T>,
+            program_id: H256,
+            memory_pages: BTreeMap<u32, Vec<u8>>,
+            value: BalanceOf<T>,
```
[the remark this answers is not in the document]
gshep: Yes, you're right. I guess this value should be greater than some constant representing rent for several (minimal count) blocks. However I'd prefer to introduce this constant & logic within corresponding task.

pallets/gear-program/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+                memory_pages,
+                <frame_system::Pallet<T>>::block_number().unique_saturated_into(),
+            )?;
+
+            T::Currency::transfer(
+                &account,
+                &<T::AccountId as common::Origin>::from_origin(program_id),
+                value,
+                ExistenceRequirement::AllowDeath,
+            )?;
```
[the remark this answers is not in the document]
gshep: Sorry, don't understand why?

pallets/gear-program/src/pause.rs
```diff
@@ -0,0 +1,112 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use super::*;
+use codec::{Decode, Encode};
+use common::{self, QueuedDispatch};
+use frame_support::{dispatch::DispatchResult, storage::PrefixIterator};
+use scale_info::TypeInfo;
+
+#[derive(Clone, Debug, PartialEq, Decode, Encode, TypeInfo)]
+pub(super) struct PausedProgram {
```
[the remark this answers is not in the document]
gshep: I have several concerns so would like to discuss it in more details

pallets/gear-program/src/lib.rs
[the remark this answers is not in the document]
gshep: We decided to lock the transferred value temporarily

pallets/gear-program/src/lib.rs
```diff
@@ -122,16 +128,20 @@ pub mod pallet {
...
+            let program_account = &<T::AccountId as common::Origin>::from_origin(program_id);
             T::Currency::transfer(
                 &account,
-                &<T::AccountId as common::Origin>::from_origin(program_id),
+                program_account,
                 value,
                 ExistenceRequirement::AllowDeath,
             )?;
 
+            T::Currency::extend_lock(LOCK_ID, program_account, value, WithdrawReasons::FEE);
```
breathx, line 143: I don't think that this kind of locking is the best way. We don't use it in any other places. Wouldn't it better to use same mech we used to - reserve?

pallets/gear-program/src/lib.rs
```diff
@@ -140,6 +140,7 @@ pub mod pallet {
                 ExistenceRequirement::AllowDeath,
             )?;
 
+            // TODO: maybe it is sufficient just to reserve value?
```
breathx, line 143:
```suggestion
            // TODO: maybe it is sufficient just to reserve value? (#762)
```

pallets/gear-program/src/lib.rs
gshep: created corresponding issue = #762

PR #696, opened 2022-02-21
Value management in programs

breathx:
Resolves #644 and partially #649.

Left todo-s in code, related to 644 issue now will be covered in #695 to prevent making this pr so huge.

**Release notes**: _to be written..._

@gear-tech/dev

PR #698, opened 2022-02-21
Provide `require!(..)` macro

### LINKED ISSUE #643: Provide `require!()` macro for `gstd`
### Problem to Solve

Need to provide comfortable ability to terminate execution if program has not enough gas/value balance.

### Possible Solution

I would like to go with macro, but it may be function, returning `Result<(), ContractError>` also. It should compare argument passed in it with `exec::gas_available()` and panic (or return `Err(_)`) in less case.

It's functionality should cover available gas, value of the current message and total program value, once it appears in gear contracts (`exec::value_available()` see #644)

breathx:
Once #696 merged.

Resolves #643.

@gear-tech/dev

### DIFF
--- a/gstd/src/macros/require.rs
+++ b/gstd/src/macros/require.rs
@@ -16,26 +16,19 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
-//! Gear `debug!` macro.
-//! Enables output of the logs from Wasm if the `debug` feature is enabled.
+//! Gear `require!` macro.
+//! Check the available balances and panics if they less then passed argument
 
-#[cfg(feature = "debug")]
 #[macro_export]
-macro_rules! debug {
-    ($arg:literal) => {
-        $crate::ext::debug(&$crate::prelude::format!("{}", $arg));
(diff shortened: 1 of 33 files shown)


NikVolf:
why it has to be macro and not just `require_gas(...)` function?

breathx:
> why it has to be macro and not just `require_gas(...)` function?

Because it also covers value checking and probably more in a future.

I don't actually like to separate them for functions. This macro is pretty simple and not seems to be harmful

gstd/src/macros/require.rs
```diff
@@ -0,0 +1,34 @@
+// This file is part of Gear.
+
+// Copyright (C) 2021 Gear Technologies Inc.
```
shamilsan, line 3:
```suggestion
// Copyright (C) 2021 Gear Technologies Inc.
```
```suggestion
// Copyright (C) 2022 Gear Technologies Inc.
```

shamilsan (commented):
Sorry but I also don't see profits of using macro instead of two strong typed functions `require_gas/require_value`.

shamilsan (approved):
Excellent ❤️

PR #705, opened 2022-02-22
Introduce queued dispatches and store gas only in tree

NikVolf:


NikVolf:
@osipov-mit probably will need to adjust js side

pallets/gear/src/lib.rs
```diff
@@ -445,20 +434,15 @@ pub mod pallet {
...
-                        dispatch.message.id,
-                        dispatch.message.gas_limit,
-                        actual_gas_locked
-                    );
-                    dispatch.message.gas_limit = actual_gas_locked;
-                }
+
+                let gas_limit = T::GasHandler::get_limit(dispatch.message.id)
+                    .map(|(gas, _id)| gas)
+                    .unwrap_or(0);
```
breathx, line 442: Are we okay allowing case where value node doesn't exist?

breathx (approved):

PR #706, opened 2022-02-22
remove substrate part from core  processor

grishasobol:
Make new trait ProcessorExt - to work with ext in processor, actually to work with lazy pages.
Old Ext from processor does not work with lazy pages now.
Create new LazyPagesExt which can work with lazy pages and it is used in pallets


ukint-vs:
This change also fixes thread-lock after panic in async reply case 👍

breathx:
> This change also fixes thread-lock after panic in async reply case 👍

😮 Great

PR #713, opened 2022-02-23
pallet_gear: Do not process queue on idle if weight is zero

ukint-vs:
@gear-tech/dev


pallets/gear/src/tests.rs
```diff
@@ -1662,6 +1662,54 @@ fn exit_handle() {
     })
 }
 
+#[test]
+fn set_code() {
```
NikVolf, line 1666: set_core should be tested, not the fact that you set something to 0 and check if it is 0

PR #716, opened 2022-02-23
Smart spec version checking

breathx:
Script improved, it better checks need in spec version changes and has more informative logging.

It also doesn't brake depandabot's PRs


ukint-vs (approved):

shamilsan (approved):
Awesome but need to cover this shell script by tests 😄

Issue #649, opened 2022-02-09
In-memory tokenomic management 


breathx:
### Problem to Solve

At the moment `gear-test`, `gtest`, and real on-chain behavior is different due to absence of some core features. So the test-cases that are success in tests may fail in real conditions.

### Possible Solution

Need to bring `ValueTree` and rent charging into in-memory testing.

Not sure, that concrete this issue should cover unification of tests and real behavior, but it's in our closest _to do ..._ Minimal for now is to create detached in-memory implementations.

NikVolf:
avoid creating complex issue that has no design what so ever

PR #720, opened 2022-02-23
Getter for initial user of messaging

breathx:
By @LouiseMedova's request:

Provides an ability to get the id of original user who initiated communication with blockchain, during which, currently processing message was created.

Integration for `gear-test` won't be realized, until we hadn't done with #649 

@gear-tech/dev

### DIFF
--- a/core-processor/src/processor.rs
+++ b/core-processor/src/processor.rs
@@ -83,9 +83,8 @@ pub fn process_many<A: ProcessorExt + EnvExt + Into<ExtInfo> + 'static, E: Envir
     let mut journal = Vec::new();
 
     assert_eq!(dispatches.len(), initiators.len());
-    let mut initiators = initiators.into_iter();
 
-    for dispatch in dispatches {
+    for (dispatch, initiator) in dispatches.into_iter().zip(initiators.into_iter()) {
         let actor = actors
             .get_mut(&dispatch.message.dest())
             .expect("Program wasn't found in programs");
@@ -95,9 +94,7 @@ pub fn process_many<A: ProcessorExt + EnvExt + Into<ExtInfo> + 'static, E: Envir
             dispatch,
             block_info,
             existential_deposit,
-            initiators
-                .next()
-                .expect("Can't fail. Lengths checked above"),
+            initiator,
         );
 
         for note in &current_journal {
--- a/core-processor/src/processor.rs
+++ b/core-processor/src/processor.rs
@@ -40,14 +40,14 @@ pub fn process<A: ProcessorExt + EnvExt + Into<ExtInfo> + 'static, E: Environmen
     dispatch: Dispatch,
     block_info: BlockInfo,
     existential_deposit: u128,
-    initiator: ProgramId,
+    origin: ProgramId,
 ) -> Vec<JournalNote> {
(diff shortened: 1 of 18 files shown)


pallets/gear/src/lib.rs
```diff
@@ -453,9 +457,8 @@ pub mod pallet {
             while let Some(dispatch) = common::dequeue_dispatch() {
                 // Update message gas limit for it may have changed in the meantime
 
-                let gas_limit = T::GasHandler::get_limit(*dispatch.message_id())
-                    .map(|(gas, _id)| gas)
-                    .unwrap_or(0);
+                let (gas_limit, initiator) = T::GasHandler::get_limit(*dispatch.message_id())
```
SabaunT, line 460: :thumbsup:

SabaunT (approved):

SabaunT:
maybe use `origin`instead of `initiator`?

core-processor/src/processor.rs
```diff
@@ -37,12 +40,19 @@ pub fn process<A: ProcessorExt + EnvExt + Into<ExtInfo> + 'static, E: Environmen
     dispatch: Dispatch,
     block_info: BlockInfo,
     existential_deposit: u128,
+    initiator: ProgramId,
 ) -> Vec<JournalNote> {
     if let Some(actor) = actor {
         let execution_settings = ExecutionSettings::new(block_info, existential_deposit);
+        let execution_context = ExecutionContext { initiator };
```
shamilsan, line 47: Can we combine settings and context into one entity?

core-processor/src/processor.rs
```diff
@@ -67,16 +77,28 @@ pub fn process_many<A: ProcessorExt + EnvExt + Into<ExtInfo> + 'static, E: Envir
...
     existential_deposit: u128,
+    // Will go away some time soon
+    initiators: Vec<ProgramId>,
 ) -> Vec<JournalNote> {
     let mut journal = Vec::new();
 
+    assert_eq!(dispatches.len(), initiators.len());
+    let mut initiators = initiators.into_iter();
+
     for dispatch in dispatches {
```
shamilsan, line 88: Isn't better to use `zip` here?

core-processor/src/processor.rs
breathx: I don't think so.
Setting generate at the start of the queue processing and don't change from message to message inside the same block, while context relates to each message separately and will contain more data in future, e.g. available gas

shamilsan: Got it, ok.

core-processor/src/processor.rs
breathx: Sure! Done

breathx:
Still considering about naming:
* `gr_initiator`
* `gr_origin`
* `gr_origin_user`

What do you think, @LouiseMedova, which one is the best?

LouiseMedova:
> Still considering about naming:
> 
> * `gr_initiator`
> * `gr_origin`
> * `gr_origin_user`
> 
> What do you think, @LouiseMedova, which one is the best?



> Still considering about naming:
> 
> * `gr_initiator`
> * `gr_origin`
> * `gr_origin_user`
> 
> What do you think, @LouiseMedova, which one is the best?

I think `gr_origin` is the best one.

PR #725, opened 2022-02-24
skip stack memory upload to storage

### LINKED ISSUE #682: Do not save/update stack memory in storage
### Problem to Solve

Rust to wasm compilation creates own stack emulation in program.
For this stack rustc occupy some number of pages. This number is defined at compile time.
There is no reasons to store this memory in storage, because it's uses as temporary memory for function local variables (it is not wasm locals, it's rust locals).   

### Possible Solution

Rust emulated stack pointer is wasm global $__stack_pointer. Initial value for this global is stack memory end address.
If smart contract contains this global, we can use it to identify stack memory and skips stack pages storing. This rule can be part of gear api for smart contracts.

### Notes

_No response_

grishasobol:
Fix #682 

Insert export "__gear_stack_end" export to global 0. This export is added in wasm-proc or wasm-builder. It contains value which is addr of stack end.

Use this export in core processor to skip stack memory uploading to storage

@reviewer-or-team


NikVolf:
Would be great to add a simple benchmark to show the performance increase

grishasobol:
> Would be great to add a simple benchmark to show the performance increase
@NikVolf  what do you actually mean? I have made test which checks that all stack wasm pages are skipped.

ukint-vs (commented):
It would be great to add benchmarks like @NikVolf says

ukint-vs (approved):

Issue #587, opened 2022-01-17
Improve the gas spent calculation routines


shamilsan:
### File Location(s)

https://github.com/gear-tech/gear/blob/master/pallets/gear/src/lib.rs#L305

### Proposal

- [x] Make `get_gas_spent` calculate the gas spent without affecting the message queue, it should return the gas value burned by the current message processing plus the gas value to be sent to another actor (if any).
- [x] Add the internal (mock) function to calculate the gas amount burned during message processing for using it in tests.
- [x] Calculate gas spent for init messages.
- [x] Calculate gas spent for reply messages.
- [x] Calculate gas spent for async programs.
- [x] Find the solution against the node abusing while sending an infinite program to `get_init_gas_spent()`.
- [ ] Fix the problem when a program uses `gas_available()` while sending messages.

Issue #658, opened 2022-02-15
Benchmarking Gas consumption


gshep:
### Problem to Solve

#573 

### Possible Solution

Develop benchmarks for every single WASM instruction, for every syscalls. 

### Notes

After this task minimum system requirements could be made more exact

PR #727, opened 2022-02-25
Refactor the gas spent calculator

shamilsan:
Part of #587.

- Gas calculation now uses the same queue as the general message processing.
- Improve the resulting message.
- Add `calc_handle_gas_spent()` function for testing purposes
- Add `value: u128` as the last parameter. This allows to use `msg::value()` related functions in programs.

@osipov-mit  please take it into account, need to update the frontend part.

@gear-tech/dev


NikVolf:
This time, I want to see test that get gas spent calculator calculates exactly how much gas will be used for processing message, I think it's possible (when program doesn't use `gas_available` syscall )

shamilsan:
> This time, I want to see test that get gas spent calculator calculates exactly how much gas will be used for processing message, I think it's possible (when program doesn't use `gas_available` syscall )

Please take a look at the test: https://github.com/gear-tech/gear/pull/727/commits/b92098e1dbf23ef9d4a5cf57dee6b0727b6585ec

Is it sufficient or need to use some more complex program?

SabaunT (approved):

Makefile
```diff
@@ -183,6 +183,10 @@ run-dev-node:
 run-dev-node-release:
 	@ ./scripts/gear.sh run node --release -- --dev
 
+.PHONY: run-dev-node-release-log
+run-dev-node-release-log:
+	@ RUST_LOG="gwasm=debug,pallet_gear=debug,gear_common=debug,gear_core_processor=debug,gear_backend_sandbox=debug" ./scripts/gear.sh run node --release -- --dev -l0
+
```
breathx, line 189: I don't see the reason to add this make binding separately.
You always able to run `RUST_LOG=... make run-dev-node-release`
If main purpose of this bind is to avoid manual typing - provide this universally for all node runs. `gear_common` and `gear_backend_*` are excess.
All pallets and gwasm are only required

pallets/gear/rpc/runtime-api/src/lib.rs
```diff
@@ -24,6 +24,6 @@ use sp_std::vec::Vec;
 
 sp_api::decl_runtime_apis! {
     pub trait GearApi {
-        fn get_gas_spent(source: H256, kind: HandleKind, payload: Vec<u8>) -> Option<u64>;
+        fn get_gas_spent(source: H256, kind: HandleKind, payload: Vec<u8>) -> Result<u64, Vec<u8>>;
```
breathx, line 27: I think we also need value argument here with balance of source user checking

pallets/gear/src/lib.rs
```diff
@@ -326,91 +328,120 @@ pub mod pallet {
...
+
+                core_processor::handle_journal(journal.clone(), &mut ext_manager);
+
+                let (remaining_gas, _) =
+                    T::GasHandler::get_limit(root_message_id).ok_or_else(|| {
+                        b"Internal error: unable to get gas limit after execution".to_vec()
+                    })?;
+
+                // TODO: Check whether we charge gas fee for submitting code after #646
+                for note in journal {
```
breathx, line 419: I don't see here anything about were `gas_available` syscall used?

pallets/gear/src/tests.rs
```diff
@@ -2109,6 +2089,69 @@ fn exit_handle() {
...
+            init_gas_spent
+        );
+
+        let handle_gas_spent = Gear::get_gas_spent(
+            USER_1.into_origin(),
+            HandleKind::Handle(program_id),
+            request,
+        )
+        .expect("Gas spent for init calculation error");
+
```
breathx, line 2147: Tests for the fact that this call gives the amount of gas which is enough for success execution are cool. But there should be also testing that this call doesn't return some huge amount of gas

breathx: We can add such kind directly in this test. Need only manually calculate expenses by wat: represent it with mathematical inequality or exact value (it can be changed in future following #658 )

pallets/gear/src/lib.rs
```diff
@@ -326,91 +328,120 @@ pub mod pallet {
...
+                            trap,
+                            ..
+                        }) => {
+                            return Err(format!(
+                                "Program terminated with a trap: {}",
+                                trap.unwrap_or("No reason")
+                            )
+                            .into_bytes());
+                        }
+                        _ => (),
```
breathx, line 439:
```suggestion
                        _ => {},
```

breathx (changes requested):
Requesting changes mainly for checking max limit of returned value, otherwise good

pallets/gear/src/lib.rs
shamilsan: `gas_available` will be dealt in the future PR.

pallets/gear/src/lib.rs
shamilsan: Can't agree with it, the unit type at the empty branch is more idiomatic.

https://doc.rust-lang.org/stable/rust-by-example/flow_control/match/binding.html
https://doc.rust-lang.org/book/ch06-02-match.html#catch-all-patterns-and-the-_-placeholder

pallets/gear/src/lib.rs
breathx: How do you want to impl that?
If you would like to count all syscalls - yeah, seems reasonable to separate it in different PR. But this task should close sometime :D
I found this check is minor addition here.

Makefile
shamilsan: Yes, it is just a shortcut for a logged run. Fixed the crates list.

pallets/gear/rpc/runtime-api/src/lib.rs
shamilsan: Excellent idea, added it.

NikVolf:
> > This time, I want to see test that get gas spent calculator calculates exactly how much gas will be used for processing message, I think it's possible (when program doesn't use `gas_available` syscall )
> 
> Please take a look at the test: [b92098e](https://github.com/gear-tech/gear/commit/b92098e1dbf23ef9d4a5cf57dee6b0727b6585ec)
> 
> Is it sufficient or need to use some more complex program?

Looks fine 👍

pallets/gear/src/tests.rs
shamilsan: I've added a separated test: https://github.com/gear-tech/gear/pull/727/commits/95f78401fea53b1d8b859609f2f13a63c8e29da1

pallets/gear/src/lib.rs
shamilsan: Agreed to make is in the new PR.

Makefile
```diff
@@ -183,6 +183,10 @@ run-dev-node:
 run-dev-node-release:
 	@ ./scripts/gear.sh run node --release -- --dev
 
+.PHONY: run-dev-node-release-log
+run-dev-node-release-log:
```
breathx, line 187: Didn't we agree to always put logs in dev mode and remove separated make command?

breathx (approved):
Agree now.

Things about `gas_available` and other syscalls we already discussed. I will open the issue on it soon and edit this comment with it's id.

Good job!

Only thing that I don't understand - why did u remove `wasm-builder/test-program`'s `Cargo.lock`?

breathx:
> Only thing that I don't understand - why did u remove `wasm-builder/test-program`'s `Cargo.lock`?

I really think that it shouldn't be in `.gitignore` as the others examples/*

shamilsan:
> > Only thing that I don't understand - why did u remove `wasm-builder/test-program`'s `Cargo.lock`?
> 
> I really think that it shouldn't be in `.gitignore` as the others examples/*

`test-program` and `examples/*` are in different conditions. While `test-program` should fail after any breaking change in `gear-wasm-builder`, the examples/apps/etc shouldn't. The main goal of `test-program` is to reveal problems at an early stage. Therefore we don't need to rely on the old crate version using `Cargo.lock`.

Makefile
shamilsan: Misunderstanding 😄 fixed it.

PR #739, opened 2022-03-01
Workaround to falling `gear-test` for Apple M1 CPUs

SabaunT:
After releasing program from program creation feature we faced problems with the new sys-call. When wasmtime was used as core-backend, it could not work well when lots of arguments were provided for the [func-wrap](https://docs.rs/wasmtime/latest/wasmtime/struct.Func.html#method.wrap). A workaround here is to serialize together fix length params of `create_program_with_gas` function into one buffer and provide one pointer to the multiple params.

@gear-tech/dev

### DIFF
--- a/core-backend/sandbox/src/funcs.rs
+++ b/core-backend/sandbox/src/funcs.rs
@@ -74,7 +74,7 @@ pub fn send<E: Ext + Into<ExtInfo>>(ctx: &mut Runtime<E>, args: &[Value]) -> Sys
     let result = ctx
         .ext
         .with(|ext| {
-            let dest: ProgramId = funcs::get_bytes32(ext, program_id_ptr).into();
+            let dest: ProgramId = funcs::get_fixed_bytes(ext, program_id_ptr).into();
             let payload = funcs::get_vec(ext, payload_ptr, payload_len);
             let value = funcs::get_u128(ext, value_ptr);
             let message_id = ext.send(OutgoingPacket::new(dest, payload.into(), None, value))?;
@@ -102,7 +102,7 @@ pub fn send_wgas<E: Ext + Into<ExtInfo>>(ctx: &mut Runtime<E>, args: &[Value]) -
     let result = ctx
         .ext
         .with(|ext| {
-            let dest: ProgramId = funcs::get_bytes32(ext, program_id_ptr).into();
+            let dest: ProgramId = funcs::get_fixed_bytes(ext, program_id_ptr).into();
             let payload = funcs::get_vec(ext, payload_ptr, payload_len);
             let value = funcs::get_u128(ext, value_ptr);
             let message_id = ext.send(OutgoingPacket::new(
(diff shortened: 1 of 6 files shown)


gcore/src/prog.rs
```diff
@@ -109,18 +115,30 @@ pub fn create_program_with_gas(
...
             payload.len() as _,
-            gas_limit,
-            value.to_le_bytes().as_ptr(),
             program_id.as_mut_slice().as_mut_ptr(),
         );
         program_id
     }
 }
+
+fn populate_from_iterator(src: &mut [u8], data: impl Iterator<Item = u8>) {
```
shamilsan, line 140: Have you reinvented [copy_from_slice()](https://doc.rust-lang.org/core/primitive.slice.html#method.copy_from_slice)?

gcore/src/prog.rs
```diff
@@ -109,18 +115,30 @@ pub fn create_program_with_gas(
     gas_limit: u64,
     value: u128,
 ) -> ActorId {
+    let mut inner = [0u8; 56];
+    let fix_len_data = code_hash
+        .iter()
+        .copied()
+        .chain(gas_limit.to_le_bytes())
+        .chain(value.to_le_bytes());
+    populate_from_iterator(&mut inner, fix_len_data);
```
shamilsan, line 124: Why copy `fix_len_data` to `inner` instead of using `fix_len_data` as a slice?

core-backend/sandbox/src/funcs.rs
```diff
@@ -623,22 +623,26 @@ pub fn create_program_wgas<E: Ext + Into<ExtInfo>>(
...
         .ext
         .with(|ext: &mut E| -> Result<(), &'static str> {
-            let code_hash = funcs::get_bytes32(ext, code_hash_ptr);
+            // Handling data under `fl_ptr`, which is a pointer to multiple fixed len params needed to create a [`ProgramInitPacket`].
+            // For more info see [`gcore::prog::sys::gr_create_program_wgas`].
+            let code_hash = funcs::get_fixed_bytes(ext, fl_ptr);
+            let gas_limit = funcs::get_u64(ext, fl_ptr + code_hash.len());
+            let value = funcs::get_u128(
+                ext,
+                fl_ptr + code_hash.len() + gas_limit.to_be_bytes().len(),
```
shamilsan, line 642: Why big-endian and not [size_of_val()](https://doc.rust-lang.org/core/mem/fn.size_of_val.html)?

```suggestion
                fl_ptr + code_hash.len() + core::mem::size_of_val(&gas_limit),
```

gcore/src/prog.rs
SabaunT: oh man...
I forgot I could populate any region of `inner`... thanks!

shamilsan (commented):
I don't like this solution generally because it makes writing C++/etc bindings more difficult and error-prone. But I don't have a better alternative indeed 😄

gcore/src/prog.rs
SabaunT: found no way without allocations instead of re implementing that https://crates.io/crates/collect_slice (what actually was done by `populate_from_iterator` function)

SabaunT:
In draft until we decide finally on salt as function argument

SabaunT:
If we merge #746, then this PR is redundant.

breathx:
> If we merge #746, then this PR is redundant.

I would reject this approach anyway. I don't think that FFI API corruption is a good way to fix in-memory logic, while on-chain (with sandbox backend) everything is ok.

It also has other workaround: setting next salt for program creation in other syscall, but this also isn't good enough. I see two option, that we discussed with you some time ago:
- Remove salt from the syscall and let it be core concern
- Fix wasmtime by updating its version or by forking it

NikVolf:
No longer needed with #746 ?

Issue #703, opened 2022-02-21
Remove nonces from storage


breathx:
### Proposal

We can generate message ids following these rules:

1. Message ids of the outgoing messages will be generated based on:
    - Currently processing message id
    - Program id (optional)
    - Local nonce of the current execution (amount of outgoing messages, see #684)
2. System error replies generate based on:
    - Currently processing message id
    - Program id 
    Note that they are unique (sent only once)
3. Messages from users (sent via extrinsic) will be generated based on:
    - Block number
    - User id (user's H256)
    - In-block counter with amount of messages sent from users
    
This will reduce amount of storage-interactions, reduce chance for miss something while developing new features, while generate replies, remove generic from messages context in `gear-core` and remove need in chain-randomness while generating messages id.

Resolves #32 as useless.

PR #744, opened 2022-03-05
Skip fixtures feature introduced

SabaunT:
This hot fix is introduced until we decide finally on having salt in `create_program_with_gas` function. If salt remains, then the solution in #739 will be merged. Otherwise, we take salt management responsibility and introduce it in https://github.com/gear-tech/gear/issues/703

### DIFF
--- a/gear-test/spec/test_create_program.yaml
+++ b/gear-test/spec/test_create_program.yaml
@@ -1,5 +1,7 @@
 title: Test create_program sys-call
 
+must_skip: true
+
 codes:
   # Save child's code
   - path: examples/create-program/child_contract.wasm
--- a/gear-test/spec/test_create_program.yaml
+++ b/gear-test/spec/test_create_program.yaml
@@ -1,6 +1,6 @@
 title: Test create_program sys-call
 
-must_skip: true
+skip: true
 
 codes:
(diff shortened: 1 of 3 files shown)


gear-test/spec/test_create_program.yaml
```diff
@@ -1,5 +1,7 @@
 title: Test create_program sys-call
 
+must_skip: true
```
shamilsan, line 3: Maybe just `skip`?
```suggestion
skip: true
```

shamilsan (approved):
Good but I'd rename the field.

ukint-vs (approved):

gear-test/src/check.rs
```diff
@@ -668,7 +671,7 @@ where
                         skip_memory,
                     )
                 };
-                if output != "Ok".bright_green() {
+                if !(output.starts_with("Ok") | output.starts_with("Skipped")) {
```
SabaunT, line 674: Should be `||`. Sorry for that.

gear-test/spec/test_create_program.yaml
SabaunT: done

PR #749, opened 2022-03-08
Refactor gtest

SabaunT:
A little `gtest` refactoring:
1. Actors storage is changed to be more like it's on pallet. `Dormant` actor variant was introduced. Its semantic is the same as `Terminated` program in `common`. Also it will be used for programs, which are tried to be init-ed from other programs when code hash doesn't exist.
2. Processing logic for genuine and mock programs is separated to different functions.
3. Manager now works with dispatches instead of messages.
4. Some minor consequent stuff after 3 major things stated upper.  

@gear-tech/dev


gtest/src/manager.rs
```diff
@@ -30,17 +30,70 @@ use gear_core::{
...
+    }
 }
 
 #[derive(Debug)]
-pub(crate) enum Program {
-    Core(CoreProgram),
-    Mock(Box<dyn WasmProgram>),
+pub(crate) enum ActiveProgram {
+    Genuine(CoreProgram),
+    // Contract: is always `Some`, option is used to take ownership
```
breathx, line 79: Why ownership is better than mutating for this case?

gtest/src/manager.rs
```diff
@@ -30,17 +30,70 @@ use gear_core::{
 use std::collections::{BTreeMap, VecDeque};
 use std::time::{SystemTime, UNIX_EPOCH};
 
-#[derive(Clone, Debug)]
-pub(crate) enum ProgramState {
-    Initialized,
-    Uninitialized(Option<MessageId>),
-    FailedInitialization,
+#[derive(Debug)]
+pub(crate) enum Actor {
```
breathx, line 34: If we go with pallet-style, isn't this seems better to separate ProgramState for other enum variants?

gtest/src/manager.rs
```diff
@@ -30,17 +30,70 @@ use gear_core::{
...
+    }
+
+    fn get_state_mut(&mut self) -> Option<&mut ProgramState> {
+        match self {
+            Actor::Active(_, state, _) => Some(state),
+            _ => None,
+        }
+    }
+
+    fn try_into_core_prog_mut(&mut self) -> Result<&mut CoreProgram, ()> {
```
breathx, line 51: This and other methods below with `try_into` semantic aren't correct in rust world. `into` in any representation means self consume. I would go here with `as_mut_core_prog() -> Option<mut ref>`, but they look redundant for me anyway

gtest/src/manager.rs
```diff
@@ -30,17 +30,70 @@ use gear_core::{
...
+    Active(ActiveProgram, ProgramState, u128),
+    Dormant,
+}
+
+impl Actor {
+    pub(crate) fn new_active(active_prog: ActiveProgram) -> Self {
+        Actor::Active(active_prog, ProgramState::Uninitialized(None), 0)
+    }
+
+    fn get_state_mut(&mut self) -> Option<&mut ProgramState> {
```
breathx, line 44: Same here about type-safe breaks

gtest/src/manager.rs
```diff
@@ -334,80 +435,92 @@ impl JournalHandler for ExtManager {
...
-            self.message_queue.remove(index);
+            self.dispatch_queue.remove(index);
         }
     }
-    fn send_dispatch(&mut self, _message_id: MessageId, dispatch: Dispatch) {
-        let Dispatch { mut message, .. } = dispatch;
+
+    fn send_dispatch(&mut self, _message_id: MessageId, mut dispatch: Dispatch) {
+        let Dispatch {
+            ref mut message, ..
```
breathx, line 448: why do we need ref here?

breathx (changes requested):

gtest/src/manager.rs
```diff
@@ -30,17 +30,82 @@ use gear_core::{
...
-#[derive(Clone, Debug)]
-pub(crate) enum ProgramState {
-    Initialized,
-    Uninitialized(Option<MessageId>),
-    FailedInitialization,
+#[derive(Debug)]
+pub(crate) enum Actor {
+    Initialized(Program),
+    // Contract: program is always `Some`, option is used to take ownership
+    Uninitialized(Option<MessageId>, Option<Program>),
```
SabaunT, line 37: I bet you @breathx ask why `Option` for program here. You see, same program value is **moved** from `Uninitialized` to `Initialized while being under mutable reference. Will be happy to see neat alternatives.

gtest/src/manager.rs
SabaunT: `as_` by convention is cost free "borrow -> borrow" operation. I will try to name it closer to the domain

gtest/src/manager.rs
```diff
@@ -30,17 +30,82 @@ use gear_core::{
...
+    }
+
+    fn as_mut_core_prog(&mut self) -> Option<&mut CoreProgram> {
+        match self {
+            Actor::Initialized(Program::Genuine(prog)) => Some(prog),
+            _ => None,
+        }
+    }
+
+    fn pull_mock(&mut self) -> Option<Box<dyn WasmProgram>> {
```
shamilsan, line 76: Maybe better to name it `take_`?
```suggestion
    fn take_mock(&mut self) -> Option<Box<dyn WasmProgram>> {
```

gtest/src/manager.rs
```diff
@@ -30,17 +30,82 @@ use gear_core::{
...
+        match self {
+            Actor::Initialized(Program::Mock(mock)) => mock.take(),
+            Actor::Uninitialized(_, Some(Program::Mock(mock))) => mock.take(),
+            _ => None,
+        }
+    }
+
+    fn pull_executable_actor(&mut self, balance: u128) -> Option<ExecutableActor> {
```
shamilsan, line 84: I don't see why `&self` it `mut`. Also, I'd rename the method.
```suggestion
    fn get_executable_actor(&self, balance: u128) -> Option<ExecutableActor> {
```

gtest/src/manager.rs
```diff
@@ -30,17 +30,82 @@ use gear_core::{
...
 pub(crate) enum Program {
-    Core(CoreProgram),
-    Mock(Box<dyn WasmProgram>),
+    Genuine(CoreProgram),
+    // Contract: is always `Some`, option is used to take ownership
+    Mock(Option<Box<dyn WasmProgram>>),
+}
+
+impl Program {
+    pub(crate) fn new_genuine(prog: CoreProgram) -> Self {
```
shamilsan, line 102: I don't like the word `genuine` here, it may confuse.
```suggestion
    pub(crate) fn new(prog: CoreProgram) -> Self {
```

gtest/src/manager.rs
```diff
@@ -30,17 +30,82 @@ use gear_core::{
...
+}
+
+impl Actor {
+    pub(crate) fn new(prog: Program) -> Self {
+        Actor::Uninitialized(None, Some(prog))
+    }
+
+    // #Panics
+    // If actor is initialized or dormant
+    fn transform_to_initialized(&mut self) {
```
shamilsan, line 48: I guess `set_` is ok here too.

```suggestion
    fn set_to_initialized(&mut self) {
```

gtest/src/manager.rs
```diff
@@ -278,23 +265,31 @@ impl ExtManager {
...
             .get_mut(&program_id)
             .expect("Can't find existing program");
 
-        *state = ProgramState::Initialized;
+        if let Actor::Uninitialized(_, maybe_prog) = actor {
+            assert!(maybe_prog.is_some(), "there is always program by contract");
+            if let Some(Program::Genuine(p)) = maybe_prog {
+                p.set_initialized();
+            }
+            Actor::transform_to_initialized(actor);
```
shamilsan, line 278: Why such a kind of call?

```suggestion
            actor.transform_to_initialized();
```

gtest/src/manager.rs
```diff
@@ -278,23 +265,31 @@ impl ExtManager {
...
             .actors
             .get_mut(&program_id)
             .expect("Can't find existing program");
 
-        *state = ProgramState::Initialized;
+        if let Actor::Uninitialized(_, maybe_prog) = actor {
+            assert!(maybe_prog.is_some(), "there is always program by contract");
+            if let Some(Program::Genuine(p)) = maybe_prog {
+                p.set_initialized();
+            }
```
shamilsan, line 277: Isn't better to embed this code into the `transform_to_initialized`?

gtest/src/manager.rs
SabaunT: right

gtest/src/manager.rs
SabaunT: I thought it is more readable. The name of the call tells us, that something is mutated (transformed). More readable version is not "self-transform`, but "transform->what->self". After changing the name to `set_initialized` it will be like `actor.set_initialized()`.

gtest/src/manager.rs
SabaunT: great idea

gtest/src/manager.rs
breathx: or more simpler: `fn set_initialized(&mut self)`

gtest/src/manager.rs
breathx: Better to provide same-semantic method as above: `take_executable_actor`

gtest/src/manager.rs
```diff
@@ -106,7 +118,7 @@ pub(crate) struct ExtManager {
     pub(crate) id_nonce: u64,
 
     // State
-    pub(crate) actors: BTreeMap<ProgramId, Actor>,
+    pub(crate) actors: BTreeMap<ProgramId, (Actor, u128)>,
```
breathx, line 121: I would add somewhere: `pub(crate) type Balance = u128;` and replace it here for others who want to refactor is somehow.
But it's minor stuff.

```suggestion
    pub(crate) actors: BTreeMap<ProgramId, (Actor, Balance)>,
```

breathx (approved):
I like it more now.
The last thing is resolving opened conversations.

Gj 👍

gtest/src/manager.rs
breathx: Where can I collect your bet? 😆 
This impl seems okay for me for current arch

gtest/src/manager.rs
```diff
@@ -96,155 +161,77 @@ impl ExtManager {
...
-                .get_mut(&message.dest())
-                .expect("Somehow message queue contains message for user");
+        {
+            let maybe_actor = self.actors.get_mut(&message.dest());
+            if let Some((actor, _)) = maybe_actor {
+                let kind = if actor.is_dormant() {
+                    DispatchKind::Handle
+                } else {
+                    if let Actor::Uninitialized(maybe_message_id, _) = actor {
+                        *maybe_message_id = maybe_message_id.or(Some(message.id()));
```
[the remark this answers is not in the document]
SabaunT: done

gtest/src/manager.rs
```diff
@@ -30,17 +30,82 @@ use gear_core::{
...
+    Uninitialized(Option<MessageId>, Option<Program>),
+    Dormant,
+}
+
+impl Actor {
+    pub(crate) fn new(prog: Program) -> Self {
+        Actor::Uninitialized(None, Some(prog))
+    }
+
+    // #Panics
```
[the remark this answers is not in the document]
SabaunT: done

gtest/src/manager.rs
SabaunT: done

gtest/src/manager.rs
SabaunT: done

gtest/src/manager.rs
SabaunT: done

gtest/src/manager.rs
SabaunT: done

gtest/src/manager.rs
SabaunT: done

gtest/src/manager.rs
SabaunT: done

gtest/src/manager.rs
SabaunT: discussed

gtest/src/manager.rs
```diff
@@ -96,155 +161,77 @@ impl ExtManager {
...
-                        }
-                    } else {
-                        *message_id = Some(message.id());
-
-                        Some(DispatchKind::Init)
-                    }
+            .unwrap_or_else(|| {
+                assert!(
+                    !matches!(actor, Actor::Uninitialized(None, _)),
+                    "init msg id is defined for uninitialized program before dispatches to it are saved"
```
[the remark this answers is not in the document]
SabaunT: done

PR #754, opened 2022-03-12
gear-wasm-builder: Fix target WASM path when using in workspace

shamilsan:
- Write output WASM files to the workspace target directory instead of the package one when using in the workspace.

@gear-tech/dev

### DIFF
--- a/utils/wasm-builder/Cargo.toml
+++ b/utils/wasm-builder/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "gear-wasm-builder"
-version = "0.1.1"
+version = "0.2.0"
 edition = "2021"
 license = "GPL-3.0-or-later WITH Classpath-exception-2.0"
 description = "Utility for building Gear programs"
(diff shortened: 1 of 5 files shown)


utils/wasm-builder/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "gear-wasm-builder"
-version = "0.1.1"
+version = "0.2.0"
```
breathx, line 3:
```suggestion
version = "0.1.2"
```
I don't think this change requires major version bumping.

> Initial development releases starting with "0.y.z" can treat changes in "y" as a major release, and "z" as a minor release.

https://doc.rust-lang.org/cargo/reference/semver.html

breathx (approved):
👍

utils/wasm-builder/Cargo.toml
shamilsan: I was doubting about whether it is breaking changes or not, but you are right, it is not quite breaking.

Issue #22, opened 2021-04-09
Deploy utilities


NikVolf:
Something along this lines should be good to go:

`gear login <private_key/uri>`
`gear deploy <wasm or path to Cargo.toml>`

clearloop:
more use cases for this program? 

- If providing more methods for this program, for example, call contract, this `gear login` could be 
  - sending the pk to a local gear host service ( mb a grpc serinvce which has the ability interacting with gear node )
- otherwise, this `gear login` can simply write the pk to local storage

btw, what do you think about the safty part of storing this pk?
- just store it
- with a passwd
- ....

clearloop:
https://github.com/clearloop/gear-program

PR #759, opened 2022-03-15
Avoid env vars in `gtest::Program::current()`

breathx:
By @SabaunT request.

First motivation: `Program::current` wasn't working out of common `cargo test` command due to absence of `OUT_DIR` env var, what made `cargo nextest` incompatible with this feature. 

- Path to opt binary now writes in `crate_root/.binpath`
- Path in `.binpath` now not absolute - it's relative, what allows to avoid username appearing in it. 

@gear-tech/dev

### DIFF
--- a/utils/wasm-builder/src/wasm_project.rs
+++ b/utils/wasm-builder/src/wasm_project.rs
@@ -159,6 +159,7 @@ impl WasmProject {
             .file_base_name
             .as_ref()
             .expect("Run `WasmProject::create_project()` first");
+
         let from_path = self
             .out_dir
             .join("target/wasm32-unknown-unknown/release")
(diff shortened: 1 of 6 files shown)


gshep (approved):

utils/wasm-builder/src/wasm_project.rs
```diff
@@ -179,9 +180,16 @@ impl WasmProject {
             .join(format!("{}.meta.wasm", &file_base_name));
         Self::generate_meta(&from_path, &to_meta_path)?;
 
-        let wasm_binary_path = self.out_dir.join("wasm_binary_path.txt");
-        fs::write(&wasm_binary_path, format!("{}", to_opt_path.display()))
-            .context("unable to write `wasm_binary_path.txt`")?;
+        let wasm_binary_path = self.original_dir.join(".binpath");
```
shamilsan, line 183: Am I right that since now every smart contract project should individually git-ignore the `.binpath` dir?

breathx: It's not necessary, because .binpath generates every time and doesn't contain any personal info.
But in general, you're right. I don't see there any other workarounds.
I hope in future we gonna provide some cargo tooling which will generate base smart contracts by `cargo gear new` etc. (in #22)

shamilsan (approved):

PR #773, opened 2022-03-20
Remove panics from lazy pages

grishasobol:
Resolves #767 .

PR #777, opened 2022-03-21
resolve lazy pages panics

grishasobol:
Returns back https://github.com/gear-tech/gear/pull/773 but this time add versions for changed runtime interfaces.

### DIFF
--- a/core-backend/common/src/lib.rs
+++ b/core-backend/common/src/lib.rs
@@ -109,4 +109,7 @@ pub trait Environment<E: Ext + IntoExtInfo + 'static>: Sized {
     /// Run setuped instance starting at `entry_point` - wasm export function name.
     /// - IMPORTANT: env is in inconsistent state after execution.
     fn execute(&mut self, entry_point: &str) -> Result<BackendReport, BackendError>;
+
+    /// Unset env ext and returns gas amount.
+    fn drop_env(&mut self) -> GasAmount;
 }
--- a/core-processor/src/executor.rs
+++ b/core-processor/src/executor.rs
@@ -164,7 +164,16 @@ pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environ
         Default::default(),
     );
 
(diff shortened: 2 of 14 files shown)


core-processor/src/executor.rs
```diff
@@ -164,7 +164,16 @@ pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environ
...
+    let lazy_pages_enabled = match ext.try_to_enable_lazy_pages(program_id, initial_pages) {
+        Ok(enabled) => enabled,
+        Err(e) => {
+            return Err(ExecutionError {
+                program_id,
+                gas_amount: ext.into_gas_amount(),
+                reason: e,
+            })
+        }
+    };
```
SabaunT, line 176:
```suggestion
    let lazy_pages_enabled = ext.try_to_enable_lazy_pages(program_id, initial_pages).map_err(|reason| ExecutionError::from(program_id, gas_amount, reason)?;
```
So `from` or `new` for `ExecutionError` must be implemented. What do you think?

SabaunT (dismissed):

core-processor/src/executor.rs
grishasobol: What is gas_amount variable? How you wanna to initialise it?

core-backend/common/src/lib.rs
```diff
@@ -109,4 +109,7 @@ pub trait Environment<E: Ext + IntoExtInfo + 'static>: Sized {
     /// Run setuped instance starting at `entry_point` - wasm export function name.
     /// - IMPORTANT: env is in inconsistent state after execution.
     fn execute(&mut self, entry_point: &str) -> Result<BackendReport, BackendError>;
+
+    /// Unset env ext and returns gas amount.
+    fn drop_env(&mut self) -> GasAmount;
```
breathx, line 114: Why do we need this inside the common and general trait?

breathx (approved):
Requesting changes due to Clonable gas

core-backend/common/src/lib.rs
grishasobol: Because, now we has needs to be able drop env before execution

PR #778, opened 2022-03-22
Handle possible errors in `pallet gear`

### LINKED ISSUE #774: Check usage of `Result/Option` in pallet gear
### File Location(s)

_No response_

### Proposal

caused by #761

gshep:
Resolves #774 .

- Remove unused code
- Introduce `CheckedCode` and adjust the code

@gear-tech/dev


core/src/code_hash.rs
```diff
@@ -0,0 +1,83 @@
...
+        &self.0
+    }
+
+    /// Returns reference to the code hash.
+    pub fn hash(&self) -> &CodeHash {
+        &self.1
+    }
+
+    /// Decomposes this instance.
+    pub fn into_code_hash(self) -> (CheckedCode, CodeHash) {
```
SabaunT, line 80: maybe `into_inner`? Because name states that we will get `CodeHash`, not tuple.

SabaunT (commented):
I like this type level guarantees very much!

core/src/code_hash.rs
gshep: Thanks, good remark! Frankly speaking I meant `into_code_AND_hash` but omit `and` for the sake of brevity.

What you'd say about `into_parts`?

SabaunT: Maybe it's even better :+1:

ukint-vs (approved):
Awesome! Great base also for caching instrumented code.

common/src/lib.rs
```diff
@@ -527,12 +526,16 @@ pub fn wait_key(prog_id: H256, msg_id: H256) -> Vec<u8> {
     key
 }
 
-pub fn get_code(code_hash: H256) -> Option<Vec<u8>> {
+pub fn get_code(code_hash: H256) -> Option<CheckedCode> {
     sp_io::storage::get(&code_key(code_hash, CodeKeyPrefixKind::RawCode))
+        .map(|bytes| CheckedCode::decode(&mut &bytes[..]).expect("CheckedCode encoded correctly"))
```
shamilsan, line 531: Stupid question, are we trying to avoid `expect`s, aren't we?

core/src/code_hash.rs
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /// Get inner (32 bytes) array representation
+    pub fn inner(&self) -> [u8; HASH_LENGTH] {
+        self.0
+    }
+
+    /// Create new `CodeHash` bytes. Returns `None` if the provided slice has wrong length.
+    pub fn from_slice_unchecked(s: &[u8]) -> Option<Self> {
+        <[u8; HASH_LENGTH]>::try_from(s).map(|a| a.into()).ok()
```
shamilsan, line 46: I prefer the func style:

```suggestion
        <[u8; HASH_LENGTH]>::try_from(s).map(Into::into).ok()
```

common/src/lib.rs
gshep: Yeah, but this one can't be removed easily. Actually all other code in common uses the same approach and it is supposed that all modifications in the storage are done with the help of the API. So we almost sure this expect cannot panic.

Also we planned to refactor `common` create and move all storage-logic to pallets. Then these `expect`s should gone.

shamilsan: Got it, thanks. I guess we may add `qed` for `expect`s that never occurs as we do in some places:

https://github.com/gear-tech/gear/blob/master/node/src/chain_spec.rs#L39

core/src/code_hash.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+impl From<[u8; HASH_LENGTH]> for CodeHash {
+    fn from(data: [u8; HASH_LENGTH]) -> Self {
+        CodeHash(data)
+    }
+}
+
+/// Contains checked code for a program and the hash for it.
+pub struct CheckedCodeHash(CheckedCode, CodeHash);
```
shamilsan, line 57: Maybe the name `CheckedCodeWithHash` would be more clear? Also, I'm not excited with the tuple struct but it's ok for two fields.

gshep: Sounds reasonable, will change

shamilsan (approved):
Looks very nice to me 👍

pallets/gear/src/lib.rs
```diff
@@ -497,9 +498,26 @@ pub mod pallet {
...
 
-                let (gas_limit, _) = T::GasHandler::get_limit(msg_id)
-                    .expect("Should never fail if ValueNode works properly");
+                    common::queue_dispatch(dispatch);
+
+                    let consumed = T::DbWeight::get().reads(1) + T::DbWeight::get().writes(1);
+                    Self::decrease_gas_allowance(consumed);
+                    if Self::gas_allowance() < consumed {
+                        break;
+                    }
```
breathx, line 517: What's it? I think some comments wouldn't be useless

core/src/code_hash.rs
```diff
@@ -0,0 +1,80 @@
...
+}
+
+impl From<[u8; HASH_LENGTH]> for CodeHash {
+    fn from(data: [u8; HASH_LENGTH]) -> Self {
+        CodeHash(data)
+    }
+}
+
+/// Contains checked code for a program and the hash for it.
+pub struct CheckedCodeWithHash(CheckedCode, CodeHash);
```
breathx, line 57: Why do we need to store code with hash as one struct?

breathx (approved):
👍

core/src/code_hash.rs
gshep: To guarantee on type level that the hash is calculated exactly for the provided code. This struct eases error handling in pallet and eliminates redundant hash calculations.

pallets/gear/src/lib.rs
gshep: Ok

core/src/checked_code.rs
```diff
@@ -0,0 +1,63 @@
...
+
+use alloc::vec::Vec;
+use anyhow::Result;
+use codec::{Decode, Encode};
+
+/// Contains raw binary code of a program and initial memory size from memory import.
+///
+/// This entity ensures the code has passed several checks.
+#[derive(Clone, Debug, Decode, Encode, PartialEq, Eq)]
+pub struct CheckedCode(Vec<u8>, u32);
```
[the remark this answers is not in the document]
gshep: ok

SabaunT (approved):

core/src/code_hash.rs
gshep: Yes, it is. However this entity is not used in the storage: check `set_code_with_metadata` for details.

PR #779, opened 2022-03-22
Add `.gitattributes` configuration file

ark0f:
Git usually automatically converts line endings to CRLF on Windows. It is useful to keep them according to EditorConfig. Also, it can break bash scripts


breathx (approved):
👍

PR #780, opened 2022-03-22
return rpc tests

grishasobol:
To be able test changes in host node(not in runtime) we need to run rpc test.


NikVolf:
Please no, i don't want all spec tests run via http (not against to do a couple of smoky tests this way though)

Clever (and probably hard) way would be run tests through wasm executor -- it needs some actual work

NikVolf (changes requested):
We don't want integration tests to run spec tests

grishasobol:
Closed, because of NikVolf request

PR #784, opened 2022-03-22
Fix Git thinks images are text

ark0f:
Fixes images are treated as text by Git

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,6 @@
 * text eol=lf
 
+*.png binary
+
 **/Cargo.lock linguist-generated=true
 **/package-lock.json linguist-generated=true
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,7 @@
 * text eol=lf
 
 *.png binary
+*.jpg binary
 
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1,4 +1,6 @@
 * text eol=lf
 
+*.png binary
```
shamilsan, line 3: What about JPG?

ark0f: I don't think Git will treat JPG as text because of its binary format

shamilsan: > I don't think Git will treat JPG as text because of its binary format

Anyway

```
The file will have its original line endings in your working directory
warning: CRLF will be replaced by LF in images/internal_flow.jpg.
```

And

```
❯ git status
On branch al-broken-pngs
Your branch is up to date with 'origin/al-broken-pngs'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   images/internal_flow.jpg

no changes added to commit (use "git add" and/or "git commit -a")
```

shamilsan (approved):

PR #786, opened 2022-03-23
Small amendment of GasHandler API

ekovalev:
Following the introduction of messages with unspecified concrete gas limit (that is, derived from some upstream message), the `GasHandler` API has been changed:
- the `get_limit(key)` function now returns the pair of the available gas for a message and the id of the tree node that sets this limit (previously returned the external origin as the second item in the pair);
- to get the external origin, the `get_origin(key)` method has been introduced.

Apart from the API change, a possibility to actually process the mock transaction pool has been added to the tests in `pallet-usage`.


breathx (approved):
Great

PR #792, opened 2022-03-24
CI: Fix nightly toolchain version to 2022-03-22

shamilsan:
Try to fix CI.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -44,7 +44,7 @@ jobs:
       - name: "Install: Nightly toolchain"
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly
+          toolchain: nightly-2022-03-22
           components: clippy, rustfmt
           target: wasm32-unknown-unknown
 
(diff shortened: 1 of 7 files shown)


grishasobol (approved):

breathx:
So can't we avoid manual specifying of toolchain in scripts?

breathx (changes requested):
Need to somehow throw available toolchain. My version is different so all build scripts gonna fail locally. Need to think in envvars way

shamilsan:
> Need to somehow throw available toolchain. My version is different so all build scripts gonna fail locally. Need to think in envvars way

`rust-toolchain.toml` forces to install `nightly-2022-03-22` locally. So you'll have several nightly versions. Is it ok?

ark0f:
Maybe it will be easier to [patch](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html) ring using https://github.com/briansmith/ring/pull/1470?

shamilsan:
> Maybe it will be easier to [patch](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html) ring using [briansmith/ring#1470](https://github.com/briansmith/ring/pull/1470)?

`ring` is the indirect dependency as I see. May we patch some kind of crates?

ark0f:
The feature allows to patch indirect dependencies, too

shamilsan:
The solution in #795 looks better. 

@ark0f thanks for advice 👍

Issue #793, opened 2022-03-24
Rethink pallet events


breathx:
We need to discuss about usability, non-repeating and integrity of events we deposit. Provide docs for crates io and wiki also.

PR #794, opened 2022-03-24
Deposit event on program-to-program messaging

### LINKED ISSUE #791: The ability to listen to messages between programs
### Problem to Solve

In some cases, it's necessary to listen to messages between programs

### Possible Solution

Provide events that will contain info about messaging

### Notes

_No response_

breathx:
Resolves #791.

- Now on program-to-program sending we deposit enqueue event.

It's not as clear as we want it to be, but we gonna refactor this moment in #793. Just a fix for now which don't require any changes at all for js side).

@gear-tech/dev

### DIFF
--- a/pallets/gear/src/manager.rs
+++ b/pallets/gear/src/manager.rs
@@ -392,13 +392,18 @@ where
             } else {
                 let _ = T::GasHandler::split(message_id, *dispatch.message_id());
             }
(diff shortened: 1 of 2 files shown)


osipov-mit (approved):

NikVolf:
I'd avoid that

Internal machinery of handling user transactions can eventually produce thousands internal messages on one user message

Logging every message is like logging every instruction and should be avoided

breathx:
Decided to don't push superfluous data into user event log

PR #795, opened 2022-03-25
Use patched ring

gshep:
Relates to https://github.com/briansmith/ring/issues/1469

@gear-tech/dev

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5440,8 +5440,7 @@ checksum = "51dd4445360338dab5116712bee1388dc727991d51969558a8882ab552e6db30"
 [[package]]
 name = "ring"
 version = "0.16.20"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3053cf52e236a3ed746dfc745aa9cacf1b791d846bdaf412f60a8d7d6e17c8fc"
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -8096,7 +8095,7 @@ version = "1.6.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "4ee73e6e4924fe940354b8d4d98cad5231175d615cd855b758adc658c0aac6a0"
 dependencies = [
- "cfg-if 1.0.0",
+ "cfg-if 0.1.10",
```
gshep, line 8098: I don't know why this changed :thinking:

shamilsan (approved):

Cargo.lock
shamilsan: I've been faced with such behavior after running `cargo update -p ...` sometimes. 

Maybe `cargo update -p cfg-if` will help?

shamilsan: `cargo update -p cfg-if:0.1.10`reverted these changes but as I see we have both of them (v0.1.10 and v1.0.0) as indirect dependencies. Therefore it seems to be not a problem.

PR #800, opened 2022-03-25
resolve panics in core backend

grishasobol:


core-processor/src/ext.rs
```diff
@@ -187,12 +187,17 @@ impl ProcessorExt for Ext {
...
+    fn into_ext_info<F: FnMut(usize, &mut [u8]) -> Result<(), &'static str>>(
+        self,
+        mut get_page_data: F,
+    ) -> Result<ExtInfo, (&'static str, GasAmount)> {
         let accessed_pages_numbers = self.allocations_context.allocations().clone();
         let mut accessed_pages = BTreeMap::new();
         for page in accessed_pages_numbers {
             let mut buf = alloc::vec![0u8; PageNumber::size()];
-            get_page_data(page.offset(), &mut buf);
+            if let Err(err) = get_page_data(page.offset(), &mut buf) {
```
breathx, line 198: Can't we use map_err and ? syntax

breathx (approved):

core-processor/src/ext.rs
grishasobol: We cant because closure will move self. Do you have any ideas how to operate better with this case?

core-backend/sandbox/src/env.rs
```diff
@@ -177,9 +176,21 @@ impl<E: Ext + IntoExtInfo + 'static> Environment<E> for SandboxEnvironment<E> {
...
+                self.runtime
+                    .memory
+                    .read(ptr, buff)
+                    .map_err(|_err| "Cannot read sandbox mem")
+            })
+            .map_err(|(reason, gas_amount)| BackendError {
+                reason,
+                description: None,
+                gas_amount,
+            })?;
```
shamilsan, line 193: Isn't it better to implement `From<(&'static str, u64)> for BackendError` instead of multiple manual conversions?

shamilsan (commented):
The solution seems temporary. We agreed to move on to enums in error handling, didn't we?

core-backend/sandbox/src/env.rs
grishasobol: I think we must do this in separate review where we will make refactoring for all errors

PR #801, opened 2022-03-26
Lighten `gear-test` crate

### LINKED ISSUE #650: Lighten `gear-test` compilation
### File Location(s)

`gear/gear-test`

### Proposal

Our `gear-test` native environment execution, meant as base and light-weight, not like that and compiles a lot.

Remove unused dependencies.

- [ ] For `check_main()` change argument `ext: Option<Box<dyn Fn() -> sp_io::TestExternalities + Send + Sync + 'static>>` for something like this `Option<impl Fn(impl Fn() -> ()) -> ()>`, by pre-building `Externalities`, which run the given closure inside.
- [ ] Remove support for ss58 ids, because our ecosystem don't use this tool anymore, and such kind of ids is superfluous for internal core testing.

ark0f:
Resolves #650

### DIFF
--- a/gear-test/Cargo.toml
+++ b/gear-test/Cargo.toml
@@ -17,9 +17,9 @@ hex = "0.4.2"
 serde = { version = "*", features = ["derive"] }
 serde_json = "*"
 log = "0.4.16"
-sp-keyring = { version = "5.0.0", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable" }
 rayon = "1.5"
 primitive-types = { version = "0.10.1", default-features = false, features = ["scale-info"] }
+once_cell = "1.10.0"
 
 # Internal deps
 gear-core = { path = "../core" }
--- a/gear-test/Cargo.toml
+++ b/gear-test/Cargo.toml
@@ -18,7 +18,6 @@ serde = { version = "*", features = ["derive"] }
 serde_json = "*"
 log = "0.4.16"
 sp-keyring = { version = "5.0.0", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable" }
-sp-io = { version = "5.0.0", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable", default-features = false }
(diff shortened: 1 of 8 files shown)


ukint-vs (approved):

gear-test/Cargo.toml
```diff
@@ -17,7 +17,6 @@ hex = "0.4.2"
 serde = { version = "*", features = ["derive"] }
 serde_json = "*"
 log = "0.4.16"
-parity-scale-codec = { version = "2.3.1", features = ["derive"], default-features = false }
 sp-keyring = { version = "5.0.0", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable" }
```
breathx, line 21: We need to remove all substrate deps from gear-test. They start with `sp-` or `sc-`.

ark0f: Why do we need to remove them at all?

breathx: Because there is no reason to compile any substrate crates for light-weight in-memory testing. gear-test is tool for fast compile and fast off-chain checking smart contracts execution

breathx (approved):
👍

gear-test/src/address.rs
```diff
@@ -17,10 +17,31 @@
...
+    );
+    accounts.insert(
+        "bob",
+        public_key("8eaf04151687736326c9fea17e25fc5287613693c912909cb226aa4794f26a48"),
+    );
+    accounts.insert(
+        "eve",
+        public_key("e659a7a1628cdd93febc04a4e0646ea20e9f5f0ce097d9a05290d4a9e054df4e"),
+    );
+    accounts
```
shamilsan, line 43: Isn't it better to convert an array of tuples to `HashMap`?

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0064b67bf14fa9153a367fcf6f963e51

ark0f: You are right. It will look cleaner

PR #804, opened 2022-03-29
Add mailbox mock to gtest

### LINKED ISSUE #645: Sending replies in `gtest`
### Problem to Solve

Need to provide functionality for replying on mailbox messages in tests, e.g. for testing async messaging with user scenario of the program.

### Possible Solution

Need to provide proper access to mailbox of users. It may look like this:
```rust
#[test]
fn mailbox_reply() {
   let sys = gtest::System::new();
   // .. some actions and program creations .. 
   let some_user_mailbox = sys.get_mailbox_of(SOME_USER);
   
   let mailbox_message = some_user_mailbox
       .get(EXPECTED SOURCE, EXPECTED PAYLOAD); // May panic
   
   // or .reply_bytes(..);
   let res: gtest::RunResult = mailbox_message.reply(payload);
   
   // mailbox_message is no longer available
   // as an object and as a value in mailbox
}
```

dimasikOne:
Resolves #645.

- Add mailbox mock to gtest;
- minor fix to save message if its destination is unknown


gtest/src/log.rs
```diff
@@ -117,13 +147,32 @@ impl Log {
...
 
+impl PartialEq<Message> for Log {
+    fn eq(&self, other: &Message) -> bool {
+        if matches!(other.reply, Some(reply) if reply.1 != self.exit_code) {
+            return false;
+        }
+        if matches!(self.source, Some(source) if source != other.source) {
+            return false;
+        }
+        if matches!( self.dest, Some(dest) if dest != other.dest) {
```
shamilsan, line 158:
```suggestion
        if matches!(self.dest, Some(dest) if dest != other.dest) {
```

gtest/src/log.rs
```diff
@@ -21,8 +23,34 @@ impl CoreLog {
...
+            id: message_id,
+            reply: self.exit_code.map(|exit_code| (self.id, exit_code)),
+        }
+    }
+
+    pub fn get_payload(&self) -> Payload {
+        self.payload.clone().into()
+    }
+
+    pub fn get_id(&self) -> MessageId {
```
shamilsan, line 51: I'd name getters shorter.
```suggestion
    pub fn id(&self) -> MessageId {
```

gtest/src/log.rs
```diff
@@ -21,8 +23,34 @@ impl CoreLog {
...
+            dest: self.source,
+            payload,
+            gas_limit: None,
+            value,
+            id: message_id,
+            reply: self.exit_code.map(|exit_code| (self.id, exit_code)),
+        }
+    }
+
+    pub fn get_payload(&self) -> Payload {
```
shamilsan, line 47: Better to return the reference to the slice instead of cloning. Function consumer may clone it himself but with the current signature, he can't get the slice reference if he wants.
```suggestion
    pub fn get_payload(&self) -> &[u8] {
```

gtest/src/program.rs
```diff
@@ -249,10 +250,17 @@ mod tests {
         let run_result = prog.send(user_id, init_msg_payload);
         assert!(run_result.main_failed);
 
+        let run_result = prog.send(user_id, String::from("should_be_skipped"));
+
+        //let encoded_payload = init_msg_payload.encode();
```
shamilsan, line 255: Dead code?
```suggestion
```

core/src/message/stored.rs
```diff
@@ -76,6 +76,19 @@ impl StoredMessage {
         )
     }
 
+    /// Convert Stored Message into Message.
+    pub fn into_message(&self) -> Message {
```
shamilsan, line 80: Do we really need this conversion? And usually `into_` methods consume `self` according to Rust naming convention.

https://rust-lang.github.io/api-guidelines/naming.html

gtest/src/log.rs
```diff
@@ -21,8 +23,44 @@ impl CoreLog {
             destination: other.destination(),
             payload: other.payload().to_vec(),
             exit_code: other.exit_code(),
+            id: other.id(),
         }
     }
+
+    pub(crate) fn from_stored_message(other: StoredMessage) -> Self {
```
shamilsan, line 30: You can implement `From<StoredMessage> for CoreLog` as a more flexible conversion way.

gtest/src/mailbox.rs
```diff
@@ -0,0 +1,99 @@
...
+    ) -> Mailbox<'a> {
+        Mailbox {
+            program_id,
+            manager_reference,
+        }
+    }
+
+    pub fn contains(&self, log: Log) -> bool {
+        self.manager_reference
+            .borrow_mut()
```
shamilsan, line 28: Why mutable borrowing?
```suggestion
            .borrow()
```

gtest/src/mailbox.rs
```diff
@@ -0,0 +1,99 @@
...
+            .get(&self.program_id)
+            .expect("No mailbox with such program id")
+            .iter()
+            .any(|message| log.eq(message))
+    }
+
+    pub fn take_message(&self, log: Log) -> MessageReplier {
+        let index = self
+            .manager_reference
+            .borrow_mut()
```
shamilsan, line 39:
```suggestion
            .borrow()
```

gtest/src/mailbox.rs
```diff
@@ -0,0 +1,99 @@
...
+    manager_reference: &'a RefCell<ExtManager>,
+}
+
+impl<'a> MessageReplier<'a> {
+    pub(crate) fn new(
+        message: StoredMessage,
+        manager_reference: &'a RefCell<ExtManager>,
+    ) -> MessageReplier<'a> {
+        MessageReplier {
+            log: CoreLog::from_stored_message(message),
```
shamilsan, line 75: After implementing `From`:
```suggestion
            log: message.into(),
```

gtest/src/system.rs
```diff
@@ -97,4 +98,12 @@ impl System {
         let code = fs::read(&path).unwrap_or_else(|_| panic!("Failed to read file {:?}", path));
         self.0.borrow_mut().store_new_code(&code)
     }
+
+    pub fn get_mailbox<ID: Into<ProgramIdWrapper>>(&'_ self, id: ID) -> Mailbox<'_> {
```
shamilsan, line 102: Lifetime elision: https://doc.rust-lang.org/nomicon/lifetime-elision.html

```suggestion
    pub fn get_mailbox<ID: Into<ProgramIdWrapper>>(&self, id: ID) -> Mailbox {
```

gtest/src/mailbox.rs
```diff
@@ -0,0 +1,99 @@
...
+        self.take_message(log).reply_bytes(raw_payload, value)
+    }
+}
+
+pub struct MessageReplier<'a> {
+    log: CoreLog,
+    manager_reference: &'a RefCell<ExtManager>,
+}
+
+impl<'a> MessageReplier<'a> {
```
shamilsan, line 69: Why do we need a separate `MessageReplier` instead of reply methods within the `Mailbox`?

shamilsan (commented):
In general, I don't like the idea of modifying the `gear-core` crate. In my opinion, `gtest` should use core messages without introducing something new. Also, I'd add some use case example in the smart contract.

gtest/src/mailbox.rs
dimasikOne: It was @breathx idea. As for me that looks overweight, but Dima really liked this possibility:
```
message = mailbox.take_message(log);
result = message.reply(payload);
```

core/src/message/stored.rs
breathx: `StoredMessage` shouldn't conceptually be able to convert into `Message`. If you really need this conversion, you can manually call Message::new constructor

gtest/src/mailbox.rs
```diff
@@ -0,0 +1,99 @@
...
+use codec::Encode;
+use gear_core::message::{Dispatch, DispatchKind, StoredMessage};
+use gear_core::{
+    ids::{MessageId, ProgramId},
+    message::Payload,
+};
+use std::cell::RefCell;
+
+pub struct Mailbox<'a> {
+    manager_reference: &'a RefCell<ExtManager>,
```
breathx, line 11:
```suggestion
    manager: &'a RefCell<ExtManager>,
```

gtest/src/mailbox.rs
breathx: Agree with immutable borrowing

breathx: Contains commonly waits key's reference, not ownable object.
And also better to pass here something which can be meant as Log. So the function decl will be:
`pub fn contains<T: Into<Log> + Clone>(&self, log: &T) -> bool`

gtest/src/mailbox.rs
```diff
@@ -0,0 +1,99 @@
...
+            manager_reference,
+        }
+    }
+
+    pub fn contains(&self, log: Log) -> bool {
+        self.manager_reference
+            .borrow_mut()
+            .actor_to_mailbox
+            .get(&self.program_id)
+            .expect("No mailbox with such program id")
```
breathx, line 31: This expect should be infallible due to existence of the mailbox structure. So need to replace the expect message or add here get_or_insert_with_default
```suggestion
            .get(&self.program_id)
            .expect("No mailbox with such program id")
```

gtest/src/mailbox.rs
```diff
@@ -0,0 +1,99 @@
...
+            .any(|message| log.eq(message))
+    }
+
+    pub fn take_message(&self, log: Log) -> MessageReplier {
+        let index = self
+            .manager_reference
+            .borrow_mut()
+            .actor_to_mailbox
+            .get(&self.program_id)
+            .expect("No mailbox with such program id")
```
breathx, line 42: Same here

gtest/src/mailbox.rs
```diff
@@ -0,0 +1,99 @@
...
+        self.manager_reference
+            .borrow_mut()
+            .actor_to_mailbox
+            .get(&self.program_id)
+            .expect("No mailbox with such program id")
+            .iter()
+            .any(|message| log.eq(message))
+    }
+
+    pub fn take_message(&self, log: Log) -> MessageReplier {
```
breathx, line 36: Same as in contains, but without reference

gtest/src/manager.rs
```diff
@@ -133,7 +135,7 @@ pub(crate) struct ExtManager {
     pub(crate) actors: BTreeMap<ProgramId, (Actor, Balance)>,
     pub(crate) codes: BTreeMap<CodeId, Vec<u8>>,
     pub(crate) dispatches: VecDeque<StoredDispatch>,
-    pub(crate) mailbox: BTreeMap<ProgramId, Vec<StoredMessage>>,
+    pub(crate) actor_to_mailbox: HashMap<ProgramId, Vec<StoredMessage>>,
```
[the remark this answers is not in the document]
breathx: I don't see any reason for renaming.

And also don't see real reason to use hash map instead of btree map, but let it be so

gtest/src/manager.rs
```diff
@@ -422,6 +427,13 @@ impl ExtManager {
 
         core_processor::handle_journal(journal, self);
     }
+
+    pub(crate) fn take_message(&mut self, program_id: &ProgramId, index: usize) -> StoredMessage {
+        self.actor_to_mailbox
+            .get_mut(program_id)
+            .expect("No mailbox with such program id")
+            .remove(index)
+    }
```
breathx, line 436: This is useless logic for manager. I would recommend do it manually in mailbox methods

gtest/src/program.rs
```diff
@@ -278,4 +284,147 @@ mod tests {
         assert!(!run_result.main_failed());
         assert!(run_result.contains(&expected_log));
     }
+
+    #[test]
+    fn mailbox_mock_walkthrough_test() {
```
breathx, line 289: better to place this test in mailbox module

gtest/src/system.rs
breathx: Seems it works without placeholded lifetimes. Should be remove for `get_program` also if it so

runtime/src/lib.rs
```diff
@@ -124,7 +124,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     // The version of the runtime specification. A full node will not attempt to use its native
     //   runtime in substitute for the on-chain Wasm runtime unless all of `spec_name`,
     //   `spec_version`, and `authoring_version` are the same between Wasm and native.
-    spec_version: 480,
+    spec_version: 490,
```
breathx, line 127: Shouldn't be increased, because gear_core shouldn't be affected

gtest/src/system.rs
```diff
@@ -97,4 +98,12 @@ impl System {
...
+        let program_id = id.into().0;
+        if self.0.borrow_mut().actors.contains_key(&program_id) {
+            panic!("Such program id is already in actors list");
+        }
+        Mailbox::new(program_id, &self.0)
```
breathx, line 107: empty mailbox with such program_id should be created in manager.mailbox to correct work of `free_id_nonce` and infallibility of getting mailbox for existing Mailbox structure

gtest/src/log.rs
```diff
@@ -31,6 +69,7 @@ pub struct DecodedCoreLog<T: Codec + Debug> {
     destination: ProgramId,
     payload: T,
     exit_code: Option<i32>,
+    id: MessageId,
```
breathx, line 72: place this field first please

gtest/src/log.rs
```diff
@@ -21,8 +23,44 @@ impl CoreLog {
...
+            value,
+            self.exit_code.map(|exit_code| (self.id, exit_code)),
+        )
+    }
+
+    pub fn payload(&self) -> &[u8] {
+        self.payload.as_slice()
+    }
+
+    pub fn id(&self) -> MessageId {
```
breathx, line 61: first method

gtest/src/log.rs
breathx: Agree to replace this method and `from_message` in order to provide `From` impl

gtest/src/log.rs
```diff
@@ -21,8 +23,44 @@ impl CoreLog {
...
+        Message::new(
+            message_id,
+            self.destination,
+            self.source,
+            payload,
+            None,
+            value,
+            self.exit_code.map(|exit_code| (self.id, exit_code)),
+        )
+    }
```
breathx, line 55: Useless method. Should be generated manually in reply_bytes method

gtest/src/mailbox.rs
```diff
@@ -0,0 +1,99 @@
...
+                    .fetch_inc_message_nonce(),
+            ),
+            value,
+        );
+        self.manager_reference
+            .borrow_mut()
+            .run_dispatch(Dispatch::new(DispatchKind::Reply, message))
+    }
+
+    pub fn reply_bytes(&self, raw_payload: impl AsRef<[u8]>, value: u128) -> RunResult {
```
breathx, line 95: I prefer generics instead of `impl`

breathx, line 95: here should be constructed reply message, converted in Dispatch with Reply entry and run manager run_dispatch

gtest/src/mailbox.rs
```diff
@@ -0,0 +1,99 @@
...
+        message: StoredMessage,
+        manager_reference: &'a RefCell<ExtManager>,
+    ) -> MessageReplier<'a> {
+        MessageReplier {
+            log: CoreLog::from_stored_message(message),
+            manager_reference,
+        }
+    }
+
+    pub fn reply(&self, payload: impl Encode, value: u128) -> RunResult {
```
breathx, line 80: and reply should call reply_bytes with encoded payload

breathx (changes requested):

gtest/src/mailbox.rs
breathx: I prefer to let more abilities to write test code for users. So I accept both variants: reply on message from mailbox or from got from mailbox message (non Copy, non Clone).

```rust
struct Mailbox<'a> {
    reff: &'a Manager,
    id: u8, // user id
}

// no Copy, no Clone
struct MailboxedMessage<'a> {
    log: CoreLog,
    reff: &'a Manager,
}

impl <'a> MailboxedMessage<'a> {
    pub fn reply(self, v: impl Encode) -> RunResult {
        self.reply_bytes(v.encode())
    }

    pub fn reply_bytes(self, v: impl AsRef<[u8]>) -> RunResult ;
}

impl<'a> Mailbox<'a> {
    pub fn contains(&self, log: Log) -> bool { false }

    pub fn take_msg(&self, log: Log) -> MailboxedMessage {

    }

    pub fn reply(&self, log: Log, v:  impl Encode) -> RunResult  {
        self.take_msg(log).reply(v)
    }

    pub fn reply_bytes(&self, log: Log, v: impl AsRef<[u8]>) -> RunResult {
        self.take_msg(log).reply_bytes(v)
    }
}

#[test]
fn name() {
    let sys = System::new();
    let prog = Program::current(&sys);

    let log = Log::builder().from(1).to(10).payload_bytes("pong");

    // user id = 10
    let res = prog.send_bytes(10, "ping");
    assert!(res.contains(&log));

    let mb: Mailbox = sys.get_mailbox(10);
    assert!(mb.contains(&log));

    let log_on_reply = Log::builder().from(1).to(10).payload_bytes("reply pong");

   // #1
    let res2 = mb.reply_bytes(&log, "reply ping");
    assert!(!res2.contains(& log_on_reply));

    // #2

    let mboxed_msg = mb.take_msg(&log);
    let res3 = mboxed_msg.reply_bytes(&log, "reply ping");
    assert!(res3.contains(& log_on_reply));
}
```

This code is just scheme*

shamilsan: Agree, it's more flexible.

gtest/src/mailbox.rs
shamilsan: In my opinion impl is ok when we have the only one generic parameter.

breathx: > In my opinion impl is ok when we have the only one generic parameter.

Discussed. Agree

gtest/src/log.rs
```diff
@@ -8,25 +9,60 @@ use std::fmt::Debug;
...
+        self.payload.as_slice()
+    }
+
     pub(crate) fn from_message(other: Message) -> Self {
         Self {
             source: other.source(),
             destination: other.destination(),
             payload: other.payload().to_vec(),
             exit_code: other.exit_code(),
+            id: other.id(),
```
shamilsan, line 46: Better would be to have the same field order as in the struct declaration.

gtest/src/log.rs
```diff
@@ -8,25 +9,60 @@ use std::fmt::Debug;
...
+}
+
+impl From<StoredMessage> for CoreLog {
+    fn from(other: StoredMessage) -> Self {
+        Self {
+            source: other.source(),
+            destination: other.destination(),
+            payload: other.payload().to_vec(),
+            exit_code: other.exit_code(),
+            id: other.id(),
```
shamilsan, line 58:
```suggestion
            id: other.id(),
            source: other.source(),
            destination: other.destination(),
            payload: other.payload().to_vec(),
            exit_code: other.exit_code(),       
```

gtest/src/log.rs
```diff
@@ -42,6 +78,7 @@ impl<T: Codec + Debug> DecodedCoreLog<T> {
             destination: log.destination,
             payload,
             exit_code: log.exit_code,
+            id: log.id,
```
shamilsan, line 81: The same as above.

gtest/src/mailbox.rs
shamilsan: Resolved?

shamilsan (commented):
Much better!

gtest/src/mailbox.rs
dimasikOne: sorry I was sure I did the renaming

gtest/src/mailbox.rs
```diff
@@ -0,0 +1,262 @@
...
+
+        self.manager
+            .borrow_mut()
+            .run_dispatch(Dispatch::new(DispatchKind::Reply, message))
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::program::ProgramIdWrapper;
```
breathx, line 113: Also it would be useful to show panic cases. Rust test-mechanism has `#[should_panic(expected="panic message"]` proc macro for it

gtest/src/manager.rs
```diff
@@ -210,6 +214,15 @@ impl ExtManager {
                 .entry(message.destination())
                 .or_default()
                 .push(message.message().clone());
+            self.log.push(Message::new(
```
breathx, line 217: log should be `Vec<StoredMessage>` due to latest changes

breathx: So there won't be that huge constructor

gtest/src/system.rs
```diff
@@ -1,3 +1,4 @@
+use crate::mailbox::Mailbox;
 use crate::{
```
breathx, line 2: Merge it in one `use`

gtest/src/mailbox.rs
```diff
@@ -0,0 +1,262 @@
...
+                self.manager
+                    .borrow_mut()
+                    .mailbox
+                    .insert(self.program_id, Vec::default());
+                panic!("No message that satisfies log");
+            }
+            Some(mailbox) => mailbox
+                .iter()
+                .position(|message| log.eq(message))
+                .expect("No message that satisfies log"),
```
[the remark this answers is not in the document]
dimasikOne: prev agree but why this one? we have iterate through whole mailbox and found no messages equal to this log. sounds logical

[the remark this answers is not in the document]
breathx: yeah, sorry

gtest/src/mailbox.rs
```diff
@@ -0,0 +1,262 @@
...
+        };
+
+        let taken_message = self
+            .manager
+            .borrow_mut()
+            .mailbox
+            .get_mut(&self.program_id)
+            .unwrap()
```
breathx, line 57: Never use just `unwrap`. Place `expect` with infallibility explanation.

gtest/src/mailbox.rs
```diff
@@ -0,0 +1,262 @@
...
+            }
+            Some(mailbox) => mailbox.iter().any(|message| log.eq(message)),
+        }
+    }
+
+    pub fn take_message<T: Into<Log> + Clone>(&self, log: T) -> MessageReplier {
+        let log: Log = log.into();
+        let index = match self.manager.borrow().mailbox.get(&self.program_id) {
+            None => {
+                self.manager
```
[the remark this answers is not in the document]
breathx: Oh, no
I think matching is better just to increase readability of the error. panic seems more complex than expect in logs

gtest/src/manager.rs
dimasikOne: in master there is still log is still ```Vec<Message>``` 
waiting for the update)

breathx: > in master log is still ```Vec<Message>``` 
> 
> waiting for the update)

Yeah, so I think this could be done by yourself in this PR ;)

That's really simple thing

dimasikOne: then the ```message``` field in ```Dispatch``` should be also changed to ```StoredMessage```? If so it will involve a lot of changes. I'm afraid that will be hard to merge

shamilsan (approved):
Looks good to me 👍

gtest/src/log.rs
```diff
@@ -1,22 +1,47 @@
...
+
+    pub fn destination(&self) -> ProgramId {
+        self.destination
+    }
+
+    pub fn payload(&self) -> &[u8] {
+        self.payload.as_slice()
+    }
+
+    pub fn exit_code(&self) -> Option<i32> {
```
breathx, line 36:
```suggestion
    pub fn exit_code(&self) -> Option<ExitCode> {
```

gtest/src/mailbox.rs
```diff
@@ -0,0 +1,320 @@
...
+        }
+        self.manager
+            .borrow_mut()
+            .mailbox
+            .insert(self.user_id, Vec::default());
+        false
+    }
+
+    pub fn take_message<T: Into<Log>>(&self, log: T) -> MessageReplier {
+        let log: Log = log.into();
```
breathx, line 32:
```suggestion
        let log = log.into();
```

gtest/src/mailbox.rs
```diff
@@ -0,0 +1,320 @@
...
+        let mut manager = self.manager.borrow_mut();
+        let index = if let Some(mailbox) = manager.mailbox.get(&self.user_id) {
+            mailbox
+                .iter()
+                .position(|message| log.eq(message))
+                .expect("No message that satisfies log")
+        } else {
+            panic!("No mailbox associated with this program id");
```
breathx, line 40:
```suggestion
            panic!("Infallible. No mailbox associated with this user id");
```

gtest/src/mailbox.rs
```diff
@@ -0,0 +1,320 @@
...
+            .expect(
+                "Infallible exception- we've just worked with element that we are trying to get",
+            )
+            .remove(index);
+
+        MessageReplier::new(taken_message, self.manager)
+    }
+
+    pub fn reply(&self, log: Log, payload: impl Encode, value: u128) -> RunResult {
+        self.take_message(log).reply(payload, value)
```
breathx, line 55:
```suggestion
        self.reply_bytes(log, payload.encode(), value)
```

gtest/src/manager.rs
```diff
@@ -463,7 +468,8 @@ impl JournalHandler for ExtManager {
                 .entry(dispatch.destination())
                 .or_default()
                 .push(dispatch.message().clone().into_stored());
-            self.log.push(dispatch.message().clone());
+            self.log
+                .push(Message::into_stored(dispatch.message().clone()));
```
breathx, line 472:
```suggestion
                .push(dispatch.message().clone().into_stored());
```

gtest/src/mailbox.rs
dimasikOne: Are you sure about infallible? I have a test that is catching this panic :)
If we create a mailbox and then in ```ExtManeger``` in ```mailbox``` delete user associated with it "by hands"  we will catch this panic

breathx: > Are you sure about infallible? I have a test that is catching this panic :)
> If we create a mailbox and then in `ExtManeger` in `mailbox` delete user associated with it "by hands" we will catch this panic

End-user of the lib isn't able to interact with ext manager

PR #809, opened 2022-03-31
Bugfix: gas nodes are not properly consumed in some circumstances

ekovalev:
When a message is removed from the wait list due to inability to pay rent (run out of free gas), a trap reply is generated with a new gas node split from the one corresponding to the expelled message. This must be followed by explicit consumption of that node.

### DIFF
--- a/pallets/usage/src/tests.rs
+++ b/pallets/usage/src/tests.rs
@@ -24,6 +24,7 @@ use core::convert::TryInto;
 use frame_support::{assert_ok, traits::ReservableCurrency};
 use gear_core::{
     code::CheckedCode,
+    ids::MessageId,
     message::{DispatchKind, StoredDispatch, StoredMessage},
 };
 use hex_literal::hex;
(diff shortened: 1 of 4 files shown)


pallets/usage/src/tests.rs
```diff
@@ -630,3 +661,77 @@ fn dust_discarded_with_noop() {
...
+                hex!("0061736d01000000020f0103656e76066d656d6f7279020001").to_vec(),
+            )
+            .unwrap(),
+        );
+        crate::mock::set_program(program_1);
+
+        let program_2 = gear_core::program::Program::new(
+            2.into(),
+            CheckedCode::try_new(
+                hex!["0061736d01000000020f0103656e76066d656d6f7279020001"].to_vec(),
```
gshep, line 698: The only remark I would like to make is why here square brackets are used but the usual brackets are used above?

gshep (approved):
Bux fixes :wrench:  are always appreciated!

pallets/usage/src/tests.rs
ekovalev: Good question. It's done this way in Substrate's own `chain_spec.rs` so I just blindly followed this pattern. But in most other places the round brackets are used. If it bothers you that much you can align it to some "canonical" style everywhere, I guess )

shamilsan: I vote for round brackets as this macro is function-like. Square brackets are acceptable in `vec!` macro only in my opinion.

breathx (approved):

pallets/usage/src/tests.rs
breathx: Agree with circle brackets

NikVolf (approved):

pallets/usage/src/tests.rs
ekovalev: Fixed

PR #810, opened 2022-03-31
Configurable page size

### LINKED ISSUE #799: smaller size for wasm pages data in storage
### Problem to Solve

As you now wasm has 64kBytes mem page size. Mainly in os this size is smaller : 4kbt for linux and 16kbt in macos as example.

When lazy pages are enabled each access to contracts data leads to all wasm page download from storage. We can reduce this by downloading only part of page which is not smaller then host native page.
We can choose size - for example 4 or 16 kBytes as page part size and store wasm data in storage using this parts.

If somebody has native pages bigger than this part, then lazy pages must be disabled - because we cannot set protection for part of native page (TODO: some os allows this, so must be considered).

### Possible Solution

-

### Notes

_No response_

grishasobol:
Resolves #799 .

1) Reduce size of code pages in storage: 4kB now. Fixes to work correct in LazyPages, Common, CoreBackend, CoreProcessor.
2) new type: `WasmPageNumber` - which represents `wasm` pages
3) Fixes in tests to pass them.

### Changes in storage:
common/src/lib.rs: struct ActiveProgram
core/src/code.rs: struct Code
size of data for each page in storage is reduced and now equal to 0x1000 bytes (old = 0x10000)

### IMPORTANT:
after this PR is committed we must rebuild nodes in test-net and reset storage

### DIFF
--- a/common/src/benchmarking.rs
+++ b/common/src/benchmarking.rs
@@ -18,6 +18,7 @@
 
 use super::*;
 
+use gear_core::memory::{wasm_pages_to_pages_set, WasmPageNumber};
 use parity_wasm::elements::*;
 use sp_io::hashing::blake2_256;
 use sp_std::borrow::ToOwned;
@@ -34,7 +35,7 @@ pub fn account<AccountId: Origin>(name: &'static str, index: u32, seed: u32) -> 
 //     (import "env" "memory" (memory $num_pages))
 //     (func (type 0))
 //     (export "init" (func 0)))
-pub fn create_module(num_pages: u32) -> parity_wasm::elements::Module {
+pub fn create_module(num_pages: WasmPageNumber) -> parity_wasm::elements::Module {
     parity_wasm::elements::Module::new(vec![
         Section::Type(TypeSection::with_types(vec![Type::Function(
             FunctionType::new(vec![], vec![]),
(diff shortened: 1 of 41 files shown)


osipov-mit (approved):

common/src/benchmarking.rs
```diff
@@ -131,16 +132,24 @@ pub fn generate_wasm3(payload: Vec<u8>) -> Result<Vec<u8>, &'static str> {
     Ok(code)
 }
 
-pub fn set_program(program_id: H256, code: Vec<u8>, static_pages: u32) {
+pub fn set_program(program_id: H256, code: Vec<u8>, static_pages: WasmPageNumber) {
     let code_id = CodeId::generate(&code).into_origin();
+    let wasm_pages = (0..static_pages.0)
+        .map(WasmPageNumber)
+        .collect::<Vec<WasmPageNumber>>();
```
shamilsan, line 139: No need to be so expliit.
```suggestion
        .collect::<Vec<_>>();
```

core-processor/src/executor.rs
```diff
@@ -246,8 +272,13 @@ pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environ
         }
     };
 
+    log::trace!("trem reason = {:?}", termination);
```
ukint-vs, line 275:
```suggestion
    log::trace!("term reason = {:?}", termination);
```

ukint-vs (approved):

PR #813, opened 2022-04-02
CI: drop old cache when merged to master

ukint-vs:
With this change, the cache will always be up to date and we won't have to clean it manually after a while.

@gear-tech/dev


shamilsan (approved):

breathx:
Isn't this mean that cache won't be saved while PR handles unmerged?

ukint-vs:
> Isn't this mean that cache won't be saved while PR handles unmerged?

The unmerged PR already unpacks the cache from the master, but each merge is put into the existing cache. This change will create a new cache in an empty environment.

PR #814, opened 2022-04-04
Log ids as hex strings

ekovalev:


core/src/ids.rs
```diff
@@ -108,7 +107,13 @@ macro_rules! declare_id {
                     end = precision;
                 };
 
-                write!(f, "{}", hex::encode(&self.0[..end]))
+                write!(f, "0x{}", hex::encode(&self.0[..end]))
+            }
+        }
+
+        impl core::fmt::Debug for $name {
```
breathx, line 114: Disagree with that formatting for debug. I prefer to show bytes on debug and hex on display.

ekovalev: Dumping bytes bloats a log by factor of 2 at least. Also, makes it impossible to search for an id in a log file. Why would bytes be more informative than a hex string, anyway?

NikVolf: hex for debug is fine, it's everywhere (H256, etc. )

NikVolf (approved):

gshep (approved):

PR #815, opened 2022-04-04
Update substrate

ukint-vs:
- `parity-scale-codec = 3.1.2`
- `scale-info = 2.0.1`
- `primitive-types = 0.11.1`

@gear-tech/dev

### DIFF
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -6,9 +6,9 @@ edition = "2018"
 license = "GPL-3.0"
 
 [dependencies]
-codec = { package = "parity-scale-codec", version = "2", default-features = false, features = ["derive", "max-encoded-len"] }
-scale-info = { version = "1.0.0", default-features = false, features = ["derive"] }
-primitive-types = { version = "0.10.1", default-features = false, features = ["scale-info"] }
+codec = { package = "parity-scale-codec", version = "3", default-features = false, features = ["derive", "max-encoded-len"] }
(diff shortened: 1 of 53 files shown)


common/Cargo.toml
```diff
@@ -6,21 +6,21 @@ edition = "2018"
 license = "GPL-3.0"
 
 [dependencies]
-codec = { package = "parity-scale-codec", version = "2", default-features = false, features = ["derive", "max-encoded-len"] }
-scale-info = { version = "1.0.0", default-features = false, features = ["derive"] }
-primitive-types = { version = "0.10.1", default-features = false, features = ["scale-info"] }
+codec = { package = "parity-scale-codec", version = "3", default-features = false, features = ["derive", "max-encoded-len"] }
```
gshep, line 9: Maybe it is better to set full version "3.1.2"?

shamilsan (approved):
Yeah, finally! But I vote for the full version pointing too.

common/Cargo.toml
grishasobol: I think the better way use full version like in substrate: 3.0.0.   But it's not necessary

grishasobol (approved):

common/Cargo.toml
ukint-vs: > I think the better way use full version like in substrate: 3.0.0. But it's not necessary

These dependencies are not part of substrate, so they follow SemVer rules.

PR #832, opened 2022-04-08
Test for processing message during runtime-upgrade

osipov-mit:


shamilsan:
To be merged into #839 and then closed.

PR #847, opened 2022-04-15
Refactor gas pallet

ekovalev:
This change rules out any possibility for a panic in pallet-gas: the concealed `expect` that was previously called in `node_with_value(key)` function for an arbitrary `key` has been removed, hence an `Option` as the output of that function. Instead, the validity of the key passed to the `node_wtih_value()` is guaranteed by the caller context.

Besides, the `ValueView` storage map has been renamed to `GasTree`: the latter name is self-explanatory, unlike the old legacy name.


pallets/gas/src/lib.rs
```diff
@@ -208,30 +210,34 @@ pub mod pallet {
...
 
-        pub fn node_root_origin(node: &ValueNode) -> H256 {
+        /// Returns the AccountId (as Origin) of the value tree creator
+        pub(crate) fn node_root_origin(node: &ValueNode) -> H256 {
             match node.inner {
                 ValueType::External { id, .. } => id,
                 ValueType::SpecifiedLocal { parent, .. }
                 | ValueType::UnspecifiedLocal { parent } => {
-                    Self::node_root_origin(&Self::value_view(parent).expect("Parent should exist"))
+                    Self::node_root_origin(&Self::get_node(parent).expect("Parent should exist"))
```
breathx, line 225: what about returning here Option<H256> if parent doesn't exist despite the fact that this should be in storage

pallets/gas/src/lib.rs
```diff
@@ -178,15 +180,15 @@ pub mod pallet {
                             // this is specified, so it need to get to the first specified parent also
                             // going up until external or specified parent is found
 
-                            let (parent_key, mut parent_node) = Self::node_with_value(parent);
-
+                            let (parent_key, mut parent_node) = Self::node_with_value(parent)
+                                .expect("Upstream node with value always exists for this key");
```
breathx, line 184: same here. aren't we able to avoid that expect?

pallets/gas/src/lib.rs
ekovalev: Our implementation guarantees a node can't be removed from the tree as long as it has at least one child. If the parent doesn't exist for a key while it has to, it means the storage is somehow corrupted. We don't have control over it as we only connect to the underlying storage through `sp_io::storage::Storage` runtime interface. Such database corruption can probably be the case on one validator (or a block producer) but not on all the validators in the network at once. Corrupted storage in any case means a node must be shut down and restarted. But the network should be able to recover.

pallets/gas/src/lib.rs
ekovalev: This is a basic invariant: for every node in the tree there is a node with value (which can be the node itself if it has type other than `ValueType::UnpecifiedLocal`). If this invariant doesn't hold,  the validator node must be restarted due to storage corruption. I don't see a way for us to gently handle such situation rather than panicking.

NikVolf: Some guarantees is ok, maybe just add a comment why this expect will never fire @ekovalev

pallets/gas/src/lib.rs
NikVolf: Here as well

breathx:
Discussed with @ekovalev these expects.
So we decided to change a bit api of DAGBasedLedger for fallible one.

Setting this PR as in progress one .

ekovalev:
Updated the PR:
- the API of the `ValueTree` (old name `DAGBasedLedger`) changed to make fallible functions return `Result`. Despite a valid gas tree should never throw, we should account for the tree actually being corrupted (able to contain "orphan" nodes) in which case some operations that require tree traversal may fail;
- simplified some return types in `ValueTree`;
- for the meantime the branches in calling code that receive an error after calling `ValueTree` methods are marked us `unreachable!()` - need to decide how to handle such errors (that indicate gas tree corruption).

breathx (approved):
Great!

breathx (approved):
Still like it!

PR #854, opened 2022-04-18
Economic limit to prevent message queue overflow

breathx:
To make transactions fee higher for loaded chain we able to implement this solution.
It will be increasing fee cost with multiplication by 2 for each const amount (5 for this PR, but this const may be changed in future) of messages in queue.

On the empty chain sendMessage consumes `X` of weight and in cost `Y` value for fee.
We were using IdentityFee before, so `Y = X`.

Now it will be counting like:
`X` of weight will be equal `Y = X * (2 ^ N)` value for fee, where `N` can be found with:
`5 * N <= MESSAGE QUEUE LENGTH < 5 * (N + 1) where N: u32`.

_JS test for fees to be added..._

@gear-tech/dev


breathx:
Will be applied in #863

PR #857, opened 2022-04-18
Provide non-trivial fee multiplier

ekovalev:
This fixes the problem with incorrect transaction fees calculation due to the trivial implementation of the `pallet_transaction_payment::MultiplierUpdate` trait which resulted into the `NextFeeMultiplier` storage value in transaction payment pallet zeroed out. This essentially cancelled the part of the fee that depended on transactions weight leaving only the part for a tx size and some base fee.

A custom implementation of the `MultiplierUpdate` - the `ConstantFeeMultiplier` struct, is used in place of the Substrate's built-in `pallet_transaction_payment::TargetedFeeAdjustment` just for the sake of efficiency: less computations is run at the end of each block to update the multiplier (which is, in our case, a constant equal to 1.0). However, if we want an elastic multiplier update based on the overall network congestion we should opt for this `TargetedFeeAdjestment` struct, as well.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -27,14 +27,17 @@ include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
 use pallet_grandpa::{
     fg_primitives, AuthorityId as GrandpaId, AuthorityList as GrandpaAuthorityList,
 };
+pub use pallet_transaction_payment::{Multiplier, MultiplierUpdate};
 use sp_api::impl_runtime_apis;
 use sp_consensus_aura::sr25519::AuthorityId as AuraId;
 use sp_core::{crypto::KeyTypeId, OpaqueMetadata, H256};
 use sp_runtime::{
     create_runtime_str, generic, impl_opaque_keys,
-    traits::{AccountIdLookup, BlakeTwo256, Block as BlockT, IdentifyAccount, NumberFor, Verify},
+    traits::{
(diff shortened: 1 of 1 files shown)


runtime/src/lib.rs
```diff
@@ -301,14 +304,44 @@ impl pallet_balances::Config for Runtime {
...
+    }
+}
+impl<M> Convert<Multiplier, Multiplier> for ConstantFeeMultiplier<M>
+where
+    M: frame_support::traits::Get<Multiplier>,
+{
+    fn convert(previous: Multiplier) -> Multiplier {
+        let min_multiplier = M::get();
+        previous.max(min_multiplier)
+    }
```
breathx, line 336: Any matching or comparing seems redundant while we providing really **Constant** FeeMultiplier.

```suggestion
    fn convert(_previous: Multiplier) -> Multiplier {
        M::get()
    }
```

breathx (approved):
Brilliant!

runtime/src/lib.rs
ekovalev: > Any matching or comparing seems redundant while we providing really **Constant** FeeMultiplier.

In fact, it should be the other way around:
```
fn convert(previous: Multiplier) -> Multiplier {
    previous
}
```
The reason we take max from 2 values is that currently the multiplier is 0, so unless we wipe the storage it will remain 0. Another option would be to set it to 1 in `on_runtime_upgrade()` hook. In either case in a subsequent PR we can modify it accordingly (once the pallet_transaction_payment::NextFeeMultiplier is successfully set to 1)

PR #863, opened 2022-04-21
Introduce gear payment pallet

ekovalev:
Implements one of the measures in order to resolve [this issue](https://github.com/gear-tech/core-audit/issues/2).

Briefly, a solution is to increase transaction fees if the message queue inflates and reduce them back once it's shrunk down. This would raise a barrier for transactions to enter the system and relieve the pressure on the processing engine.
This is a modification of the standard approach taken in Substrate to adjust fees based on the overall network congestion level.

Generally, it is reasonable to charge transactions that affect the Message Queue based on the gas_limit they declare:

`Fee = base_fee + P⋅tx.len +  𝛂⋅F(tx.weight) + 𝛃⋅G(tx.gas_limit)`, 

where `P` - per byte cost, `F(.)` - weight to fee conversion function, `G(.)` - gas-to-fee conversion function (`gas_price`),`𝛂 ≥ 1` - fee multiplier reflecting the tx pool state, `𝛃 ≥ 1` - fee multiplier reflecting the message queue congestion level.

The problem is that the actual amount of gas spent is not known by the time the extrinsic has been dispatched so we can't yet calculate the gas leftover that needs to be refunded.

Hence the following simplification:

`Fee = base_fee + P⋅tx.len +  𝞬⋅F(tx.weight)`, where `𝞬 ≥ 1` is some empirical coefficient depending on the number of messages in the queue by the end of the previous block. If a transaction does not affect the message queue (like `set_code`, for instance), `𝞬` is set to 1.

This is essentially the formula Substrate uses for transaction fee calculation in its own `transaction-payment` pallet, only that the multiplier is calculated differently, and can vary depending on the concrete call.

### DIFF
--- a/pallets/payment/Cargo.toml
+++ b/pallets/payment/Cargo.toml
@@ -37,7 +37,7 @@ serde = "1.0.132"
 env_logger = "0.9"
 wabt = "0.10"
 gear-core = { path = "../../core" }
-sp-io = { version = "6.0.0", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable", default-features = false }
+sp-io = { version = "6.0.0", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable" }
 pallet-timestamp = { version = "4.0.0-dev", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable" }
 pallet-gear = { path = "../gear" }
 pallet-gas = { path = "../gas" }
--- a/pallets/payment/Cargo.toml
+++ b/pallets/payment/Cargo.toml
@@ -36,12 +36,12 @@ pallet-transaction-payment = { version = "4.0.0-dev", default-features = false, 
 serde = "1.0.132"
 env_logger = "0.9"
 wabt = "0.10"
-common = { package = "gear-common", path = "../../common" }
 gear-core = { path = "../../core" }
 sp-io = { version = "6.0.0", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable", default-features = false }
 pallet-timestamp = { version = "4.0.0-dev", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable" }
 pallet-gear = { path = "../gear" }
 pallet-gas = { path = "../gas" }
+pallet-gear-messenger = { path = "../gear-messenger" }
 pallet-gear-program = { path = "../gear-program" }
 
 [features]
--- a/pallets/payment/Cargo.toml
+++ b/pallets/payment/Cargo.toml
@@ -1,0 +1,67 @@
+[package]
+name = "pallet-gear-payment"
+version = "0.1.0"
+authors = ['Gear Technologies']
+edition = '2018'
+license = "GPL-3.0"
+homepage = "https://gear-tech.io"
+repository = "https://github.com/gear-tech/gear"
+description = "Gear main pallet"
+readme = "README.md"
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+
+[dependencies]
+codec = { package = "parity-scale-codec", version = "3.1.2", default-features = false, features = ["derive", "max-encoded-len"] }
+scale-info = { version = "2.0.1", default-features = false, features = ["derive"] }
+log = { version = "0.4.16", default-features = false }
+parity-wasm = { version = "0.42.2", default-features = false, optional = true }
+primitive-types = { version = "0.11.1", default-features = false, features = ["scale-info"] }
+
+# Internal dependencies
+common = { package = "gear-common", path = "../../common", default-features = false }
+
+# Substrate deps
+frame-support = { version = "4.0.0-dev", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable", default-features = false }
+frame-system = { version = "4.0.0-dev", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable", default-features = false }
+frame-benchmarking = { version = "4.0.0-dev", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable", default-features = false, optional = true }
+sp-std = { version = "4.0.0-dev", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable", default-features = false }
+sp-runtime = { version = "6.0.0", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable", default-features = false }
+pallet-authorship = { version = "4.0.0-dev", default-features = false, git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable" }
+pallet-balances = { version = "4.0.0-dev", default-features = false, git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable" }
+pallet-transaction-payment = { version = "4.0.0-dev", default-features = false, git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable" }
+
+[dev-dependencies]
+serde = "1.0.132"
+env_logger = "0.9"
+wabt = "0.10"
+common = { package = "gear-common", path = "../../common" }
+gear-core = { path = "../../core" }
+sp-io = { version = "6.0.0", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable", default-features = false }
+pallet-timestamp = { version = "4.0.0-dev", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable" }
+pallet-gear = { path = "../gear" }
+pallet-gas = { path = "../gas" }
+pallet-gear-program = { path = "../gear-program" }
+
+[features]
+default = ['std']
+std = [
+	"codec/std",
+	"log/std",
+	"frame-support/std",
+	"frame-system/std",
+	"scale-info/std",
+	"sp-std/std",
+	"sp-runtime/std",
+	"pallet-balances/std",
+	"pallet-transaction-payment/std",
+	"pallet-authorship/std",
+	"primitive-types/std",
+]
+runtime-benchmarks = [
+	"frame-benchmarking",
+	"frame-system/runtime-benchmarks",
+	"frame-support/runtime-benchmarks",
+	"parity-wasm",
+]
--- a/pallets/payment/src/lib.rs
+++ b/pallets/payment/src/lib.rs
@@ -1,0 +1,349 @@
+// This file is part of Gear.
+
+// Copyright (C) 2021-2022 Gear Technologies Inc.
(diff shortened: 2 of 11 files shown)


breathx (commented):
Discussed the way we charge extra fee: need to return multiplier instead of explicit value.
This also applies for corrected weight.

And sure waiting for unit tests.

Otherwise, great job! 👍

breathx (changes requested):
Disagree with increasing extrinsic's weight, because then we gonna waste weight before on_idle.

ekovalev:
> Disagree with increasing extrinsic's weight, because then we gonna waste weight before on_idle.

Agree with the disagreement. Reverted the change back.

pallets/payment/src/lib.rs
```diff
@@ -0,0 +1,372 @@
...
+    }
+
+    #[pallet::pallet]
+    #[pallet::generate_store(pub(super) trait Store)]
+    #[pallet::without_storage_info]
+    pub struct Pallet<T>(_);
+
+    #[pallet::type_value]
+    pub fn FeeMultiplierOnEmpty() -> CustomFeeMultiplier {
+        CustomFeeMultiplier::saturating_from_integer(1)
```
breathx, line 347: I would provide some constant for that multiplier, cause we re-use it many times

pallets/payment/src/lib.rs
```diff
@@ -0,0 +1,372 @@
...
+    #[pallet::getter(fn fee_multiplier)]
+    pub type FeeMultiplier<T: Config> =
+        StorageValue<_, CustomFeeMultiplier, ValueQuery, FeeMultiplierOnEmpty>;
+
+    #[pallet::hooks]
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        /// Initialization
+        fn on_initialize(_bn: BlockNumberFor<T>) -> Weight {
+            0_u64
+        }
```
breathx, line 360: Can be freely removed.
```suggestion
```

runtime/src/lib.rs
```diff
@@ -411,6 +380,43 @@ impl pallet_usage::Config for Runtime {
...
+                Balances::resolve_creating(&author, fees);
+            }
+            if let Some(tips) = fees_then_tips.next() {
+                if let Some(author) = Authorship::author() {
+                    Balances::resolve_creating(&author, tips);
+                }
+            }
+        }
+    }
+}
```
breathx, line 412: Can't we reuse native substrate solution instead of `DealWithFees`?

pallets/payment/src/lib.rs
```diff
@@ -0,0 +1,372 @@
...
+        };
+        let combined_fee = extra_fee_multiplier.saturating_mul_int(fee);
+
+        let withdraw_reason = if tip.is_zero() {
+            WithdrawReasons::TRANSACTION_PAYMENT
+        } else {
+            WithdrawReasons::TRANSACTION_PAYMENT | WithdrawReasons::TIP
+        };
+
+        match T::Currency::withdraw(
```
breathx, line 271: Forget to include tip to the fee

breathx (approved):
👍

pallets/payment/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+        } else {
+            Cow::Borrowed(info)
+        }
+    }
+}
+
+/// Custom fee multiplier which remains constant regardless the network congestion
+/// TODO: consider using Substrate's built-in `pallet_transaction_payment::TargetedFeeAdjustment`
+/// to allow elastic fees based on network conditions (if that's more appropriate)
+pub struct ConstantFeeMultiplier<M>(sp_std::marker::PhantomData<M>);
```
breathx, line 181: Unused now, may be deleted

pallets/payment/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+    }
+    fn target() -> Perquintill {
+        Default::default()
+    }
+    fn variability() -> Multiplier {
+        Default::default()
+    }
+}
+
+pub type CustomFeeMultiplier = FixedU128;
```
breathx, line 236: same, unused

pallets/payment/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+    fn min() -> Multiplier {
+        Default::default()
+    }
+    fn target() -> Perquintill {
+        Default::default()
+    }
+    fn variability() -> Multiplier {
+        Default::default()
+    }
+}
```
breathx, line 234: If I understand right, this impl need only because of trait bound?

runtime/src/lib.rs
```diff
@@ -411,6 +380,41 @@ impl pallet_usage::Config for Runtime {
...
+impl OnUnbalanced<NegativeImbalance> for DealWithFees {
+    fn on_unbalanceds<B>(mut fees_then_tips: impl Iterator<Item = NegativeImbalance>) {
+        if let Some(fees) = fees_then_tips.next() {
+            if let Some(author) = Authorship::author() {
+                Balances::resolve_creating(&author, fees);
+            }
+            if let Some(tips) = fees_then_tips.next() {
+                if let Some(author) = Authorship::author() {
+                    Balances::resolve_creating(&author, tips);
+                }
```
breathx, line 408: can't fee sum them in one operation?

breathx (approved):

pallets/payment/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+
+#[frame_support::pallet]
+pub mod pallet {
+    use super::*;
+
+    #[pallet::config]
+    pub trait Config:
+        frame_system::Config + pallet_transaction_payment::Config + pallet_authorship::Config
+    {
+        /// Filter for calls subbject for extra fees
```
breathx, line 374: I would also place queue len constant in this pallet's config

breathx: To be able update it's value with runtime upgrade

runtime/src/lib.rs
ekovalev: In Substrate `fee` and `tip` are treated slightly differently - could be the case for us, as well, so I kept it as is.
Anyway, I don't even remember if we have a mechanism for using tips at all - at least we don't create the signed extension explicitly with some tip value. So likely the second `if` is never triggered.

Most importantly, after we introduce staking this code will go. And this is going to happen real soon (hopefully).

pallets/payment/src/lib.rs
ekovalev: Right, overlooked it.

pallets/payment/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+        //
+        // TODO: consider reimplementing.
+        // This procedure does introduce a rounding error 𝞮 =  w - ⎣w / m⎦⋅m
+        // However, we argue that such error is negligible:
+        // - the rounding error can never exceed `m` (multiplier). Order of `w` (weight) is
+        // usually not less than 10^8 while the fee multiplier should not be greater than 10^3.
+        // Therefore the rounding error shouln't exceed 0.001% in the worst case.
+        // Note: this only applies to calls that do not affect message queue, that is are
+        // relatively rare. Still, a better solution can be found.
+        if !T::ExtraFeeCallFilter::contains(call) {
```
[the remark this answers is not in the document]
ekovalev: We pre-compensate for subsequent multiplication by the factor inside the `transaction-payment` pallet's fee calculation algorithm. Effectively, for the calls that do not incur any extra fee (that is, do not touch the message queue), the resulting weight_fee part will be as if no multiplier had been applied (up to a small rounding error which I estimate 0.001% in the worst case; likely much less than that)

pallets/payment/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+            let multiplier = TransactionPayment::<T>::next_fee_multiplier();
+            if multiplier > Multiplier::saturating_from_integer(1) {
+                let mut info: DispatchInfo = *info;
+                info.weight = multiplier
+                    .reciprocal() // take inverse
+                    .unwrap_or_else(Multiplier::max_value)
+                    .saturating_mul_int(info.weight);
```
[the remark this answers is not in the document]
ekovalev: We must "divide" by the factor we'll be taking a product with at later step. Hence multiplication by the inverse of the multiplier (with rounding).

pallets/payment/Cargo.toml
```diff
@@ -0,0 +1,67 @@
...
+sp-io = { version = "6.0.0", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable", default-features = false }
+pallet-timestamp = { version = "4.0.0-dev", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable" }
+pallet-gear = { path = "../gear" }
+pallet-gas = { path = "../gas" }
+pallet-gear-messenger = { path = "../gear-messenger" }
+pallet-gear-program = { path = "../gear-program" }
+
+[features]
+default = ['std']
+std = [
```
breathx, line 49: gear pallets, sp-io and timestamp in dev deps should also be without default features and with "std", isn't it?

pallets/payment/src/mock.rs
```diff
@@ -0,0 +1,256 @@
...
+            Call::Gear(pallet_gear::Call::submit_program { .. })
+                | Call::Gear(pallet_gear::Call::send_message { .. })
+                | Call::Gear(pallet_gear::Call::send_reply { .. })
+        )
+    }
+}
+
+impl pallet_gear_payment::Config for Test {
+    type ExtraFeeCallFilter = ExtraFeeFilter;
+    type MessageQueue = pallet_gear_messenger::DequeImpl<Test>;
```
breathx, line 222: seems better to pass LengthImpl<Test> so?

runtime/Cargo.toml
```diff
@@ -57,6 +57,7 @@ pallet-gear = { version = "2.0.0", default-features = false, path = "../pallets/
 pallet-gear-debug = { version = "2.0.0", default-features = false, path = "../pallets/gear-debug", optional = true }
 pallet-usage = { version = "2.0.0", default-features = false, path = "../pallets/usage" }
 pallet-gas = { version = "2.0.0", default-features = false, path = "../pallets/gas" }
+pallet-gear-payment = { version = "0.1.0", default-features = false, path = "../pallets/payment" }
```
breathx, line 60: std feature?

breathx (approved):

pallets/payment/Cargo.toml
ekovalev: Right. Looks like in many of our pallets we have dev dependencies with default features disabled. Need to carefully double check it and correct where appropriate.

PR #872, opened 2022-04-26
Idiomatic errors

### LINKED ISSUE #823: Idiomatic errors
### Problem to Solve

We use static strings as errors in `core` crates which is not idiomatic and doesn't allow us to forward through syscalls errors using codes

### Possible Solution

Introduce `core-errors` crate which contains only errors to be thin and might be used in `core-processor`, `gcore` crates

### Notes

_No response_

ark0f:
Resolves #823.

Also get rid of `*_TRAP_STR` constants


core-backend/common/src/funcs.rs
```diff
@@ -16,44 +16,28 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
-use crate::{GAS_ALLOWANCE_STR, LEAVE_TRAP_STR, WAIT_TRAP_STR};
 use alloc::{vec, vec::Vec};
 use gear_core::memory::Memory;
+use gear_core_errors::MemoryError;
```
SabaunT, line 21:
```suggestion
use gear_core::memory::{Memory, Error as MemoryError};
```

SabaunT (changes requested):
Interim review

common/src/lazy_pages.rs
```diff
@@ -20,20 +20,52 @@
...
+    fn from(err: MprotectError) -> Self {
+        Self::Mprotect(err)
+    }
+}
+
+impl From<GetReleasedPageError> for Error {
+    fn from(err: GetReleasedPageError) -> Self {
+        Self::GetReleasedPage(err)
+    }
+}
```
SabaunT, line 58: what about using `thiserror` crate. It really helps with removing this.

runtime-interface/src/lib.rs
```diff
@@ -21,22 +21,38 @@
...
 
-#[derive(Debug, Encode, Decode)]
+impl fmt::Display for MprotectError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            MprotectError::PageError => write!(f, "Page error"),
+            MprotectError::OsError => write!(f, "OS error"),
+        }
+    }
+}
```
SabaunT, line 45:
```
#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode, derive_more::Display)]
pub enum MprotectError {
    #[display(fmt = "Page error")]
    PageError,
    #[display(fmt = "OS error")]
    OsError,
}
```

runtime-interface/src/lib.rs
```diff
@@ -21,22 +21,38 @@
...
+
+#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode)]
 pub struct GetReleasedPageError;
 
+impl fmt::Display for GetReleasedPageError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Failed to get released page")
+    }
+}
```
SabaunT, line 54: same as with the previous error

SabaunT (changes requested):

core-backend/common/src/funcs.rs
ark0f: Is this so important?

common/src/lazy_pages.rs
ark0f: `common` crate uses `no_std` when `thiserror` relies on `std::error::Error`

SabaunT: we can still remove that `Display` implementation with `#[derive(fmt = "")]`, isn't it?

core-backend/common/src/funcs.rs
SabaunT: Of course, in comparison to significant changes that you did, this is minor stuff. During review we see minor/major things to be changed. This is minor stuff and you can leave it how it is. Or you can emphasize using re-export. :)

core/src/env.rs
```diff
@@ -173,19 +181,17 @@ impl<E: Ext> ExtCarrier<E> {
     }
 
     /// Calls infallible fn with inner ext.
-    pub fn with<R>(&self, f: impl FnOnce(&mut E) -> R) -> Result<R, &'static str> {
+    pub fn with<T>(&self, f: impl FnOnce(&mut E) -> T) -> Result<T, LaterExtWithError> {
```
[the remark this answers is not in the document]
ark0f:
```
error[E0283]: type annotations needed
   --> core-backend\wasmtime\src\funcs.rs:627:17
    |
627 |             ext.with_fallible(|ext: &mut E| Ok(ext.msg().len() as _))
    |             ----^^^^^^^^^^^^^----------------------------------------
    |             |   |
    |             |   cannot infer type for type parameter `U` declared on the associated function `with_fallible`
    |             this method call resolves to `core::result::Result<T, U>`
    |
    = note: cannot satisfy `_: core::convert::From<gear_core::env::ExtCarrierWithError>`
```
So I have to write `ext.with_fallible::<_, ExtCarrierWithError>(...)` which is quite ugly

core-backend/wasmtime/src/env.rs
```diff
@@ -299,22 +332,16 @@ impl<E: Ext + IntoExtInfo> Environment<E> for WasmtimeEnvironment<E> {
 
         let res = entry_func.call(&mut self.store, &[], &mut []);
 
+        let termination_reason = self.store.data().termination_reason.clone();
```
ark0f, line 335: Found bug - `TerminationReason::GasAllowanceExceeded` from `Ext::charge_gas*` is not considered

core/src/env.rs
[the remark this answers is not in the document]
SabaunT: ok!

SabaunT (approved):

core-backend/common/src/funcs.rs
ark0f: > During review I see minor/major things to be improved

Don't be afraid to say about them. I just wondering sometimes

core-backend/wasmtime/src/env.rs
ark0f: Considered actually

core/src/memory.rs
```diff
@@ -196,7 +165,7 @@ impl core::ops::Sub for WasmPageNumber {
 /// ```
 pub fn pages_to_wasm_pages_set<'a>(
     pages_iter: impl Iterator<Item = &'a PageNumber>,
-) -> Result<BTreeSet<WasmPageNumber>, &'static str> {
+) -> Result<BTreeSet<WasmPageNumber>, MemoryError> {
```
gshep, line 168: why not use alias `Error` and remove the second use?

core/src/memory.rs
```diff
@@ -205,7 +174,7 @@ pub fn pages_to_wasm_pages_set<'a>(
                 wasm_pages.insert(WasmPageNumber(gp.0 / PageNumber::num_in_one_wasm_page()));
                 Ok(())
             } else {
-                Err("There is wasm page, which has not all gear pages in the begin")
+                Err(MemoryError::NotAllPagesInEnd)
```
gshep, line 177: Maybe something like `NotAllPagesInBegin` should be used? The error message states `... in the begin`

gshep (approved):
Good job! :hammer_and_wrench:

common/src/lazy_pages.rs
```diff
@@ -120,7 +141,9 @@ pub fn protect_lazy_pages_and_update_wasm_mem_addr(
         );
         gear_ri::set_wasm_mem_begin_addr(new_mem_addr);
     }
-    mprotect_lazy_pages(new_mem_addr, true)
+    mprotect_lazy_pages(new_mem_addr, true)?;
```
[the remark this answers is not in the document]
ark0f: Why?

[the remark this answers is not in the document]
gshep: Return type is the same

shamilsan (approved):
👍

ark0f:
@shamilsan, why no `C2-refactoring` label?

shamilsan:
> @shamilsan, why no `C2-refactoring` label?

We use these labels in issues only, not in PRs.

PR #882, opened 2022-04-29
store contracts pages data in storage in lazy manner

### LINKED ISSUE #858: Super Lazy Pages - lazy store in storage
### Problem to Solve

Currently each wasm page for smart contract and each corresponding gear pages must be uploaded to storage if it's program submit or some new pages were allocated during handle. This  can lead to huge performance downgrade when number of pages is huge when we submit program for example. Especially if gear page size is smaller than wasm page - and number of pages is much bigger then.

We can skip upload of untouched pages  even if it's program submit. Initial static memory data  generates from wasm code. And heap can be filled by zeroes or garbage by default. So, all pages which were changed after allocation can be uploaded to storage, for all other pages we can use default data.

### Possible Solution

See above

### Notes

_No response_

grishasobol:
Resolves #858
1) Make lazy store of pages data in storage: program has data for page in storage only iff page has been changed at least once during program live time.
2) Remove update of program allocations set in UpdatePage journal note. Move it to UpdateAllocations. Also in update allocations we removes all pages which has data in storage but does not now in allocation set of program.
3) Change interface to operate with lazy pages thru extension. Now lazy pages must be enabled before contract processing and if lazy pages cannot be enabled then data for pages must be passed in actor pages_data field, which was also added in this patch.
4) persistent_pages map is not now field of core program struct. There is only wasm pages allocations btree set.
5) Returns back old gas price for pages loading: 3000 per page, because here I return back that program allocations set is wasm pages, not gear pages.
6) Fix tests and benchmarks to works right with new functions and interfaces.

Important for gear-js diffs:
1) Remove static_pages field from active program, because there is one in program code. And we have never used it in node. Only for gear-debug once.
2) allocations field in active program is for wasm pages now.

### IMPORTANT
after this patch we must rebuild test-net nodes and reset storage.


core-processor/src/configs.rs
```diff
@@ -26,7 +26,7 @@ const MAX_WASM_PAGES: u32 = 512;
 const INIT_COST: u64 = 5000;
```
NikVolf, line 26: Those numbers should be updated and ultimately benchmarked

Now it is too low

grishasobol: Append new issue for this: https://github.com/gear-tech/gear/issues/895

ukint-vs (approved):

osipov-mit (approved):

core-processor/src/common.rs
```diff
@@ -285,18 +294,30 @@ pub enum ExecutionErrorReason {
...
-    #[display(fmt = "Not enough gas for initial memory")]
+    /// Not enough gas in block to grow memory size
+    #[display(fmt = "Not enough gas in block to grow memory size")]
+    GrowMemoryBlockGasExceeded,
+    /// Not enough gas for initial memory handling
+    #[display(fmt = "Not enough gas for initial memory handling")]
     InitialMemoryGasExceeded,
+    /// Not enough gas in block for initial memory handling
+    #[display(fmt = "Not enough gas in block for initial memory handling")]
+    InitialMemoryBlockGasExceeded,
```
gshep, line 311: How errors `...GasExceeded` and `...BlockGasExceeded` differ? If I'm not confused there is no other gas except one in block.

core-processor/src/ext.rs
```diff
@@ -57,34 +59,25 @@ pub trait ProcessorExt {
         host_fn_weights: HostFnWeights,
     ) -> Self;
 
-    /// Try to enable and initialize lazy pages env
-    fn try_to_enable_lazy_pages(
-        &mut self,
-        program_id: ProgramId,
-        memory_pages: &mut BTreeMap<PageNumber, Option<Box<PageBuf>>>,
-    ) -> Result<bool, Self::Error>;
+    /// Returns whether this extention works with lazy pages
```
gshep, line 62:
```suggestion
    /// Returns whether this extension works with lazy pages
```
?

pallets/gear-debug/src/lib.rs
```diff
@@ -61,6 +61,9 @@ pub mod pallet {
 
         /// Weight information for extrinsics in this pallet.
         type WeightInfo: WeightInfo;
+
+        /// LOL
```
gshep, line 65: :rofl:

pallets/gear/src/manager.rs
```diff
@@ -437,29 +438,40 @@ where
...
-            } else {
-                persistent_pages.remove(&page_number);
-                common::remove_program_page(program_id, page_number);
+        if let Program::Active(mut program) = program {
+            for (page, data) in pages_data {
+                common::set_program_page_data(program_id, page, data);
+                program.pages_with_data.insert(page);
             }
+            common::set_program(program_id, program);
+        };
```
gshep, line 455: redundant semicolon

pallets/gear/src/manager.rs
```diff
@@ -437,29 +438,40 @@ where
...
+        if let Program::Active(mut program) = program {
+            let removed_pages = program.allocations.difference(&allocations);
+            for page in removed_pages.flat_map(|p| p.to_gear_pages_iter()) {
+                if program.pages_with_data.remove(&page) {
+                    common::remove_program_page_data(program_id, page);
+                }
+            }
+            program.allocations = allocations;
+            common::set_program(program_id, program);
         };
```
gshep, line 475:
```suggestion
        }
```

common/src/lib.rs
```diff
@@ -400,29 +402,41 @@ pub fn save_page_lazy_info(id: H256, page_num: PageNumber) {
...
+            (*p, sp_io::storage::get(&key))
+        })
+        .filter_map(|(page, data)| data.map(|data| (page, data)))
+        .collect()
+}
 
-        persistent_pages.insert(page_num, sp_io::storage::get(&key)?);
-    }
-    Some(persistent_pages)
+pub fn set_program(id: H256, program: ActiveProgram) {
```
gshep, line 423: I would mark it as non public since it looks like it doesn't get used anywhere except this crate.

core-processor/src/common.rs
grishasobol: There is one gas counter for contract and another gas counter for block.

core-processor/src/ext.rs
grishasobol: Done

pallets/gear-debug/src/lib.rs
grishasobol: Done

common/src/lib.rs
grishasobol: There are uses

pallets/gear/src/manager.rs
grishasobol: Done

gshep (approved):

PR #886, opened 2022-05-02
CI: Fix nightly toolchain version to 2022-04-28

shamilsan:
Try to fix CI.

### DIFF
--- a/utils/wasm-builder/src/cargo_command.rs
+++ b/utils/wasm-builder/src/cargo_command.rs
@@ -25,6 +25,7 @@ use crate::builder_error::BuilderError;
 pub struct CargoCommand {
     path: String,
     manifest_path: PathBuf,
+    toolchain: &'static str,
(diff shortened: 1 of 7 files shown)


utils/wasm-builder/src/cargo_command.rs
```diff
@@ -36,7 +37,8 @@ impl CargoCommand {
         CargoCommand {
             path: "cargo".to_string(),
             manifest_path: "Cargo.toml".into(),
-            args: vec!["+nightly", "rustc", "--target=wasm32-unknown-unknown"],
+            toolchain: "nightly",
```
gshep, line 40: Looks like this could be done in separate PR

shamilsan: Agree, will do it later.

shamilsan:
Not needed as the last nighly toolchain has been fixed again 😄

PR #891, opened 2022-05-02
Optimize wasm

ukint-vs:
- Use `linker-plugin-LTO` flag https://doc.rust-lang.org/rustc/linker-plugin-lto.html
- Use [binaryen](https://github.com/WebAssembly/binaryen) `wasm-opt` 

w/ `gstd` debug feature:
```
demo_async_multisig.meta.wasm 103Ki -> 98Ki
demo_async_multisig.opt.wasm 99Ki -> 94Ki
demo_async_multisig.wasm 253Ki -> 243Ki
```
w/o debug:
```
demo_async_multisig.meta.wasm 89Ki -> 70Ki
demo_async_multisig.opt.wasm 85Ki -> 66Ki
demo_async_multisig.wasm 210Ki -> 128Ki

```
@gear-tech/dev


NikVolf:
Can we use it as a part of our build tool?

ukint-vs:
> Can we use it as a part of our build tool?

Yes, I added `wasm-opt` post-processing to `wasm-builder` and `wasm-proc`

NikVolf (approved):

utils/wasm-proc/src/main.rs
```diff
@@ -36,6 +37,7 @@ impl std::fmt::Display for Error {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             Self::OptimizerFailed => write!(f, "Optimizer failed"),
+            Error::WasmOptFailed => write!(f, "wasm-opt failed"),
```
shamilsan, line 40: Need to either rename this entry or all remaining.
```suggestion
            Self::WasmOptFailed => write!(f, "wasm-opt failed"),
```

utils/wasm-builder/Cargo.toml
```diff
@@ -16,3 +16,5 @@ toml = "0.5.9"
 thiserror = "1.0.30"
 log = "0.4.16"
 pathdiff = { version = "0.2.1", default-features = false }
+which = "4.2.5"
+colored = "2.0.0"
```
breathx, line 20: newline at the end

utils/wasm-proc/src/main.rs
```diff
@@ -158,6 +160,16 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
             continue;
         }
 
+        let res = gear_wasm_builder::optimize::optimize_wasm(PathBuf::from(file), "s", true)
```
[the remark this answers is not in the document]
ukint-vs: I think it should be forced otherwise we won't have a culture of program size optimization.

> Or how about to provide such optimization functionality just inside the gear wasm builder?

I'll look into it but I think it's the simplest solution and we don't need to reinvent the wheel.

utils/wasm-builder/src/lib.rs
```diff
@@ -25,6 +25,7 @@ use crate::wasm_project::WasmProject;
 mod builder_error;
 mod cargo_command;
 mod crate_info;
+pub mod optimize;
```
breathx, line 28: Why this module should be public?

utils/wasm-proc/src/main.rs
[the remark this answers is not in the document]
breathx: Ok, but I want you to open issue for consider about inclusion of opt functionality directly in builder

utils/wasm-builder/src/wasm_project.rs
```diff
@@ -173,6 +173,9 @@ impl WasmProject {
         let to_opt_path = self
             .target_dir
             .join(format!("{}.opt.wasm", &file_base_name));
+
+        let _ = crate::optimize::optimize_wasm(to_path.clone(), "s", false);
```
breathx, line 177: Why don't we skip debug symbols here?

breathx (approved):

utils/wasm-builder/src/wasm_project.rs
ukint-vs: symbols are stripped when creating a meta and opt file, it is convenient to analyze a file in a regular wasm with tools like `twiggy` with debug symbols.

utils/wasm-builder/src/lib.rs
ukint-vs: used in `wasm-proc`

PR #894, opened 2022-05-03
Install prebuilt cargo-{hack, nextest} in CI

ark0f:
To avoid nightly compiler regressions

### DIFF
--- a/scripts/src/init.sh
+++ b/scripts/src/init.sh
@@ -45,6 +45,15 @@ js_update() {
 }
 
 cargo_init() {
-  cargo install cargo-hack
-  cargo install cargo-nextest
+  if [ -z $CI ]; then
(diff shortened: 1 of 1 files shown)


breathx:
Better to apply such changes for CI only, because our scripts used not only for CI, but for development also. Linux build will not suit for windows or macOS development

NikVolf (approved):

scripts/src/init.sh
```diff
@@ -45,6 +45,15 @@ js_update() {
 }
 
 cargo_init() {
-  cargo install cargo-hack
-  cargo install cargo-nextest
+  if [ -z $CI ]; then
```
shamilsan, line 48: The condition seems to be inverted.

shamilsan (approved):

PR #908, opened 2022-05-05
Add test case that there is no double charging

### LINKED ISSUE #506: Apply necessary unsigned extrinsic validity checks in pallet-usage
The `collect_waitlist_rent` extrinsic can be unsigned if it comes from a local transaction pool (that is, sent by the block producer). As a precaution against a malicious validator we should think whether additional validity checks (apart from the payload size) are necessary to prevent a potential attack.

gshep:
Resolves #506 

Emulate malicious validator that could send more unsigned extrinsics with the same arguments. Check that consequent calls to unsigned extrisincs in the same block don't charge additional rents.

@gear-tech/dev


breathx (approved):
Till the moment we rewrite ocw logic, this test should be. Well done!

pallets/usage/src/tests.rs
```diff
@@ -205,6 +226,72 @@ fn ocw_interval_maintained() {
     })
 }
 
+#[test]
+fn ocw_double_charge() {
+    init_logger();
+    let (mut ext, pool) = with_offchain_ext();
+    ext.execute_with(|| {
+        // Reserve some currency on users' accounts
+        for i in 1u64..=10 {
```
shamilsan, line 235: No need to set the type explicitly as it can be inferred.
```suggestion
        for i in 1..=10 {
```

pallets/usage/src/tests.rs
```diff
@@ -112,6 +112,27 @@ fn wait_list_contents() -> Vec<(StoredDispatch, u32)> {
     .collect()
 }
 
+fn decode_validate_usage_call(encoded: &[u8]) -> crate::Call<Test> {
+    let mut encoded = <&[u8]>::clone(&encoded);
```
shamilsan, line 116: Why not `encoded.clone()`?

pallets/usage/src/tests.rs
```diff
@@ -112,6 +112,27 @@ fn wait_list_contents() -> Vec<(StoredDispatch, u32)> {
...
+    let extrinsic: Extrinsic = Decode::decode(&mut encoded).unwrap();
+
+    let call = extrinsic.call;
+    let inner = match call {
```
shamilsan, line 120: Why a separate variable?
```suggestion
    let inner = match extrinsic.call {
```

pallets/usage/src/tests.rs
gshep: Clippy says so: clone of the reference is required here and clippy suggests to do it this way

pallets/usage/src/tests.rs
```diff
@@ -112,6 +112,27 @@ fn wait_list_contents() -> Vec<(StoredDispatch, u32)> {
...
+        mock::Call::Usage(inner) => inner,
+        _ => unreachable!(),
+    };
+
+    assert!(
+        <Usage as sp_runtime::traits::ValidateUnsigned>::validate_unsigned(
```
shamilsan, line 126: Isn't it better to import the trait?
```suggestion
        <Usage as ValidateUnsigned>::validate_unsigned(
```

pallets/usage/src/tests.rs
gshep: ok

pallets/usage/src/tests.rs
gshep: ok

pallets/usage/src/tests.rs
gshep: yep, ok

shamilsan (approved):

PR #919, opened 2022-05-09
append lazy pages feature in gear node crate

grishasobol:
to be able to build node without lazy pages enabled add feature "lazy-pages" in node
```
cargo build -p gear-node --no-default-features
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2192,7 +2192,7 @@ dependencies = [
  "gear-core",
  "libc",
  "log",
- "nix 0.23.1",
+ "nix",
  "page_size",
  "sp-io",
  "sp-std",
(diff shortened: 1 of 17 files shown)


breathx (commented):
I don't actually like the feature management style when need to specify some feature to exclude functionality. It seems not canonical for me. I would recommend you to provide just "lazy-pages" feature and mark it as default, so any build will contain the feature. To disable it, I will need to run build with --default-features=false.

On the other hand, do we really need the feature-dependent compilation?

grishasobol:
> I don't actually like the feature management style when need to specify some feature to exclude functionality. It seems not canonical for me. I would recommend you to provide just "lazy-pages" feature and mark it as default, so any build will contain the feature. To disable it, I will need to run build with --default-features=false.
> 
> On the other hand, do we really need the feature-dependent compilation?

When you use -default-features=false you will disable all other features in default, which may cause problems (there is default std feature for example). As a result we will need modifies files to disable lazy pages or some other.

Yes we need because lazy-pages is very unsafe part of gear. So we need have some simple and fast mechanism to disable them.

NikVolf:
Agree with @breathx, features should be additive

gshep (approved):

PR #920, opened 2022-05-10
change to wasmer sandbox with caches

### LINKED ISSUE #802: switch to wasmer in sandbox host executor
### Problem to Solve

Currently we use wasmi as executor for smart contracts. Both in wasm and in native runtime. 
Here we will talk only about wasm runtime. Sandbox in substrate allows to choose which executor to use: wasmer or wasmi. We must check whether switching to wasmer is good idea and good in performance.

### Possible Solution

Actually, all work must be done in substrate repo. When I tried to change executor I've faced some problems: 
1) default wasmer singlepass compiler does not work on arm processors.
2) wasmer with cranelift failed some tests

Substrate uses too old version of wasmer and may be we should try to change it. But because of big api changes in wasmer since old version, too much code must be changed in substrate. So, to be able resolves conflicts, when syncing with parity substrate we have to make pull request in their repo.

### Notes

_No response_

grishasobol:
Resolves #802

Use wasmer as sandbox executor for smart contracts.

I'd investigate how time of smart contracts execution is changed for our examples:
image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -344,7 +344,7 @@ dependencies = [
  "cfg-if 1.0.0",
  "libc",
  "miniz_oxide",
- "object",
+ "object 0.27.1",
  "rustc-demangle",
 ]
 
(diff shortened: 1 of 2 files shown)


ukint-vs (approved):
Great

PR #923, opened 2022-05-11
Create a salt factory in gstd

### LINKED ISSUE #693: Default salt management for `create_program`
### Problem to Solve

Users can make mistakes in managing salt. For example, for salt user can use in his factory contract some incrementor. But if someone else deploys the same factory, all the create_program calls will fail, because the code hash and salt (provided with an incrementor) values will be used again.

### Possible Solution

We can provide users with some `CodeDeployManager` from `gstd`, which will manage the salt and other stuff using available on-chain unique, unrepeated data.

### Notes

_No response_

montaglue:
Resolves #693.

- Added sp_core_hashing dependency to the gstd
- Created SaltFactory for salt generation

@reviewer-or-team

### DIFF
--- a/gstd/Cargo.toml
+++ b/gstd/Cargo.toml
@@ -16,6 +16,7 @@ hex = { version = "0.4.3", default-features = false, features = ["alloc"] }
 primitive-types = { version = "0.11.1", default-features = false, features = ["scale-info"]}
 scale-info = { version = "2.0.1", default-features = false, features = ["derive"] }
 futures = { version = "0.3", default-features = false, features = ["alloc"] }
+sp-core-hashing = { version = "4.0.0", default-features = false }
 
 [features]
 debug = ["galloc/debug", "gcore/debug"]
--- a/gstd/src/prog/salt_factory.rs
+++ b/gstd/src/prog/salt_factory.rs
@@ -1,0 +1,32 @@
+pub struct SaltFactory([u8; 32]);
+
+fn invert(array: &mut [u8]) {
+    for byte in array {
+        *byte = 255 - *byte;
+    }
+}
+
+impl SaltFactory {
+    /// This is unsafe because all [`SaltFactory`] in one message from this constructor will produse same salts.
+    pub unsafe fn new() -> Self {
+        SaltFactory(*crate::msg::id().inner())
+    }
+
+    pub fn from_salt(mut salt: [u8; 32]) -> Self {
+        invert(&mut salt);
+        SaltFactory(sp_core_hashing::blake2_256(&salt))
+    }
+
+    pub fn generate(&mut self) -> [u8; 32] {
+        let new_hash = sp_core_hashing::blake2_256(&self.0);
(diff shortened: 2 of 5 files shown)


gstd/src/prog/salt_factory.rs
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    pub fn generate(&mut self) -> [u8; 32] {
+        let new_hash = sp_core_hashing::blake2_256(&self.0);
+        let old_hash = self.0;
+        self.0 = new_hash;
+        old_hash
+    }
+
+    pub fn clone(&mut self) -> Self {
```
shamilsan, line 30: What is the use case for `clone()`? Should it be implemented using the `Clone` trait?

gstd/src/prog/salt_factory.rs
```diff
@@ -0,0 +1,35 @@
+pub struct SaltFactory([u8; 32]);
+
+fn invert(array: &mut [u8]) {
+    for byte in array {
+        *byte = 255 - *byte;
+    }
+}
+
+impl SaltFactory {
+    /// This is unsafe because all [`SaltFactory`] in one message from this
```
shamilsan, line 10: I am not sure it is a good solution to make this function unsafe.

gstd/src/prog/salt_factory.rs
```diff
@@ -0,0 +1,35 @@
+pub struct SaltFactory([u8; 32]);
+
+fn invert(array: &mut [u8]) {
```
shamilsan, line 3: What is the profit of inversion?

gstd/Cargo.toml
```diff
@@ -16,6 +16,7 @@ hex = { version = "0.4.3", default-features = false, features = ["alloc"] }
 primitive-types = { version = "0.11.1", default-features = false, features = ["scale-info"]}
 scale-info = { version = "2.0.1", default-features = false, features = ["derive"] }
 futures = { version = "0.3", default-features = false, features = ["alloc"] }
+sp-core-hashing = { version = "4.0.0", default-features = false }
```
breathx, line 19: we don't need extra hashing in gstd.

gstd/src/prog/salt_factory.rs
montaglue: I also thought about rotation, reversion, and appending itself. First, two can keep salt unchanged, and the last one makes allocations. Invert simple, and fast and the result always differs. Maybe you knew better operations?

gstd/src/prog/salt_factory.rs
montaglue: When you have to store factory in more than one place. You create first with new constructor and clone as much as you need. I can't implement Clone because cloning also mutates the original. I can change the method's name or use Mutex.

gstd/Cargo.toml
NikVolf, line 19: please no

PR #927, opened 2022-05-11
Implement `Ext` in crates

### LINKED ISSUE #921: Make a separate implementation of `Ext` for production and test environments.
### File Location(s)

_No response_

### Proposal

We have `LazyPagesExt` which is a wrapper around the default `Ext` implementation that overrides some functions, this is a dirty trick that shows us that we must implement the `Ext` directly in the modules in which we have the execution of programs. We can define a separate `Ext` for the test environments (gear-test, gtest) and for the pallet so that programs can directly access runtime environment (storage, configs, costs) without passing things like `BlockInfo`, gas counters etc. For example, right now we can't just implement the `is_program` syscall which needs access to the storage which can change state while the program is running, so `Ext` must have access to all environment from the top.

ukint-vs:
Resolves #921.

- Implement default `Ext` members.
- Remove `LazyPagesExt`.
- Add `Ext` into `gear-test` and `gtest`.

Todo:
- [ ] Reduce code duplicate.
- [ ] Try to avoid interactions directly with `Ext` struct fields.

@gear-tech/dev


gear-test/src/ext.rs
```diff
@@ -0,0 +1,356 @@
...
+    pub program_candidates_data: BTreeMap<CodeId, Vec<(ProgramId, MessageId)>>,
+    /// Weights of host functions.
+    pub host_fn_weights: HostFnWeights,
+}
+
+/// Empty implementation for non-substrate (and non-lazy-pages) using
+impl ProcessorExt for Ext {
+    type Error = ExtError;
+
+    fn new(
```
NikVolf, line 58: No need in such giant new function, just initialize structure by-field

pallets/gear/src/ext.rs
```diff
@@ -146,54 +130,116 @@ impl ProcessorExt for LazyPagesExt {
...
         prog_id: ProgramId,
     ) -> Result<(), Self::Error> {
-        lazy_pages::protect_pages_and_init_info(mem, memory_pages, prog_id)
-            .map_err(Error::LazyPages)
+        if cfg!(feature = "lazy-pages") {
+            lazy_pages::protect_pages_and_init_info(mem, memory_pages, prog_id)
+                .map_err(Error::LazyPages)
+        } else {
+            unreachable!()
+        }
```
grishasobol, line 170: А если мы решим собрать ноду без lazy-pages?

grishasobol (changes requested):

core-processor/src/executor.rs
```diff
@@ -57,8 +57,8 @@ pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environ
     let mut gas_counter = GasCounter::new(dispatch.gas_limit());
     let mut gas_allowance_counter = GasAllowanceCounter::new(context.gas_allowance);
 
-    // Checks that lazy pages are enabled in case extension A uses them.
-    if !A::check_lazy_pages_consistent_state() {
```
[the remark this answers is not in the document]
grishasobol: Can see now that it's not support two regime. But it must because we can build node with no lazy-pages feature.

grishasobol (approved):

PR #928, opened 2022-05-12
Bugfix in gas pallet

ekovalev:
A node wouldn't be marked as "consumed" unless it had no unspecified refs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1326,6 +1326,16 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


breathx (commented):
Can we have a test showing the issue?

ekovalev:
> Can we have a test showing the issue?

Added a test that highlights the issue; if cherry-picked on current master it'd fail. Works like a charm in this branch.

breathx (approved):
👍

PR #930, opened 2022-05-12
Add syscalls black list

### LINKED ISSUE #587: Improve the gas spent calculation routines

shamilsan:
Finally resolves #587

- Add functions (aka syscalls) blacklist.
- Generate a trap when trying to call a forbidden function in a program.

TODO: 

- [x] Fix test

@gear-tech/dev


breathx:
At the first blush, I like your solution of initial task, but don't rush to merge with master, because I think we need to think over processor representation. I don't see it as a struct and I don't actually like builder pattern.

shamilsan:
> At the first blush, I like your solution of initial task, but don't rush to merge with master, because I think we need to think over processor representation. I don't see it as a struct and I don't actually like builder pattern.

Builder pattern allows us to not add a default argument in every call. I don't like thousand of parameters in a function (and McConnell doesn't too). Also. `Processor` struct will allow to not allocate host func weight etc at every call but just borrow them.

shamilsan:
Superseded by #959, closing.

PR #932, opened 2022-05-12
move codecov to build job

### LINKED ISSUE #901: Remove coverage separate job
### File Location(s)

`.github/workflows/CI.yml`

### Proposal

Need to include coverage job's logic in other CI jobs to avoid double-time-spent for building and testing.

Storing caches with it's instrumentation also would work with rebuilds including that feature.

ukint-vs:
Resolves #901.

-
-
-

@gear-tech/dev


.github/workflows/CI.yaml
```diff
@@ -137,7 +152,9 @@ jobs:
           tar -xf /tmp/build_target.tar
 
       - name: "Build: Gear"
-        run: ./scripts/gear.sh build gear --release --locked
+        run: ./scripts/gear.sh build gear --locked --exclude "gear-test" --release
```
ark0f, line 155: Why `gear-test` has been excluded?

.github/workflows/CI.yaml
```diff
@@ -147,15 +164,43 @@ jobs:
...
         run: ./scripts/gear.sh test gear --release
+        env:
+          RUSTFLAGS: -Cinstrument-coverage
 
       - name: "Test: JS metadata"
         run: ./scripts/gear.sh test js
+        env:
+          RUSTFLAGS: -Cinstrument-coverage
 
       - name: "Test: Process node runtime testsuite"
```
ark0f, line 175: Missing `RUSTFLAGS`

.github/workflows/CI.yaml
ark0f: Also get rid of `--release` flag across the job because we can get worse coverage report. And I expect build will be faster

scripts/src/test.sh
```diff
@@ -70,13 +70,8 @@ gtest() {
 }
 
 # $1 - ROOT DIR
-rtest_debug() {
-  cargo run --package gear-node $CARGO_FLAGS -- runtime-spec-tests "$1"/gear-test/spec/*.yaml -l0
```
ark0f, line 74: Why `cargo` call has been removed?

.github/workflows/CI.yaml
```diff
@@ -147,15 +164,43 @@ jobs:
...
         run: ./scripts/gear.sh test rtest
 
       - name: "Test: Runtime upgrade and queue processing"
```
ark0f, line 178: This is the only step that must be compiled in release mode because block production will take too long

.github/workflows/CI.yaml
```diff
@@ -286,114 +331,3 @@ jobs:
...
-        run: ./scripts/gear.sh test gear
-        env:
-          RUSTFLAGS: -Cinstrument-coverage
-        
-      - name: "Test: Process node runtime testsuite"
-        run: ./scripts/gear.sh coverage rtest
-        env:
-          RUSTFLAGS: -Cinstrument-coverage
-          
-      - name: "Test: Process gear-test"
```
ark0f, line 373: You forgot to add this step into `build` job

ark0f: Also remove `gtest` job. Removal was discussed in chat

.github/workflows/CI.yaml
ukint-vs: > Also get rid of `--release` flag across the job because we can get worse coverage report. And I expect build will be faster

As I can see the report didn't change when building in release mode and `gear-test` covers the same lines as `gear-node runtime-spec-test` https://codecov.io/github/gear-tech/gear/commit/44c3e40cf5136b200b930776e5d57f62a5082a49

scripts/src/test.sh
ukint-vs: because `cargo run` rebuilds wasm runtime each time which adds ~2min.

.github/workflows/CI.yaml
ukint-vs: We need to build all artefacts in same mode to re-use artefacts and cache.

.github/workflows/CI.yaml
ark0f: > As I can see the report didn't change when building in release mode

Percent is ok for the whole project but if you look closely you will see we lost ~10% of coverage for `core-backend` not speaking about other crates

> gear-test covers the same lines as gear-node runtime-spec-test

Should we remove `gear-test` at all then?

scripts/src/test.sh
ark0f: It's ok for CI but not for user who just runs `./scripts/gear.sh test rtest`

.github/workflows/CI.yaml
ark0f: I suggest to move this step into seperated job

scripts/src/test.sh
grishasobol: I think user must do builds first then. demo from examples are not builds in cargo run also as you can see

ark0f: > cargo run rebuilds wasm runtime each time

Why this even happens?

.github/workflows/CI.yaml
ukint-vs: `gear-test` is a lightweight tool to doublecheck execution logic in `wasmtime` backend. We can just ignore `wasmtime` coverage.

scripts/src/test.sh
ark0f: It means `substrate` overuse `cargo:rerun-if-changed` so it might be better to patch `substrate`

.github/workflows/CI.yaml
ark0f: > We can just ignore wasmtime coverage.

This won't help. We still have bad coverage. I just reprint what bot says:
> core-processor/src/common.rs	0.00% <0.00%> (-68.00%)
core-processor/src/ext.rs	74.54% <0.00%> (-14.10%)
core/src/costs.rs	73.68% <0.00%> (-7.90%)

This is absolutely unaccaptable

ukint-vs: moved `gear-test` to build job

.github/workflows/CI.yaml
```diff
@@ -147,15 +166,48 @@ jobs:
...
 
       - name: "Test: JS metadata"
         run: ./scripts/gear.sh test js
+        env:
+          RUSTFLAGS: -Cinstrument-coverage
+
+      - name: "Test: Process gear-test"
+        run: ./scripts/gear.sh test gtest
+        env:
+          RUSTFLAGS: -Cinstrument-coverage
```
breathx, line 180: what's the reason to keep flag here?
https://github.com/gear-tech/gear/pull/932/files#r872209943

.github/workflows/CI.yaml
breathx: Disagree that this is the only one step. Runtime testsuit requires the same. And build in release everything for CI isn't the worst practice

.github/workflows/CI.yaml
```diff
@@ -147,15 +166,48 @@ jobs:
...
 
       - name: "Test: Process node runtime testsuite"
         run: ./scripts/gear.sh test rtest
 
       - name: "Test: Runtime upgrade and queue processing"
         run: ./scripts/gear.sh test runtime-upgrade
+            
+      - name: "Coverage: Aggregate"
```
breathx, line 188: There is an issue appeared for other PR, where coverage app fails job on small diff coverage, need to fix it

.github/workflows/CI.yaml
```diff
@@ -147,15 +166,48 @@ jobs:
...
+        run: >-
+          ./grcov . --binary-path ./target/release/ -s . -t lcov --branch --ignore-not-existing
+          --ignore "/*"
```
breathx, line 191: scripts/read.me-s/weight.rs/core-errors and dirs, which start with ".", should be also ignored

breathx (approved):

.github/workflows/CI.yaml
ukint-vs: It will be fixed when codecov.yml will be in master.

breathx: > It will be fixed when codecov.yml will be in master.

Great. Can't wait to merge it 😇

.github/workflows/CI.yaml
ark0f: > Runtime testsuit requires the same

Why? 

> And build in release everything for CI isn't the worst practice

Isn't the worst but also isn't the best

.github/workflows/CI.yaml
ark0f: > core-errors

Why? If unused variant appeared compiler won't say it to us because every enum is public but we will see it in coverage

ukint-vs: > scripts/read.me-s/weight.rs/core-errors and dirs, which start with ".", should be also ignored

you can see that only `*.rs` source files are included in coverage results, so only `weights.rs` and `migration.rs` should be ignored.

PR #934, opened 2022-05-12
tests for mprotect and mem pages

grishasobol:


runtime-interface/src/lib.rs
```diff
@@ -253,3 +258,86 @@ pub trait GearRI {
...
+    // Set [OLD_VALUE] as value for each first byte of gear pages
+    unsafe {
+        for p in pages_to_protect.iter() {
+            let addr = page_begin as usize + *p as usize * PageNumber::size() + 1;
+            *(addr as *mut u8) = OLD_VALUE;
+        }
+        for p in pages_unprotected.iter() {
+            let addr = page_begin as usize + *p as usize * PageNumber::size() + 1;
+            *(addr as *mut u8) = OLD_VALUE;
+        }
```
gshep, line 312: this can be shortened:
```suggestion
        for p in pages_to_protect.iter().chain(&pages_unprotected) {
            let addr = page_begin as usize + *p as usize * PageNumber::size() + 1;
            *(addr as *mut u8) = OLD_VALUE;
        }
```

runtime-interface/src/lib.rs
```diff
@@ -148,6 +148,33 @@ unsafe fn sys_mprotect_interval(
...
+    }
+
+    let mprotect = |start, count, protect: bool| unsafe {
+        let addr = mem_addr + (start * PageNumber::size()) as u64;
+        let size = count * PageNumber::size();
+        sys_mprotect_interval(addr, size, !protect, !protect, false)
+    };
+
+    // Collects continuous intervals of memory from lazy pages to protect them.
+    let mut start = pages[0] as usize; // Can't panic as we've checked `pages` is not empty
```
gshep, line 164: Comments sideways are not looking good. Moreover `first` method could be used with the explanation:
```suggestion
    let mut start = pages.first().expect("checked `pages` is not empty") as usize;
```

gshep (approved):

core/src/memory.rs
```diff
@@ -328,4 +329,23 @@ mod tests {
     fn page_number_subtraction_with_overflow() {
         let _ = PageNumber(1) - PageNumber(u32::MAX);
     }
+
+    #[test]
+    /// Test that PageNumbers subtract correctly
```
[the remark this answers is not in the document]
grishasobol: Change comment

core/src/memory.rs
```diff
@@ -328,4 +329,23 @@ mod tests {
...
+    fn wasm_pages_to_gear_pages() {
+        let wasm_pages: Vec<WasmPageNumber> =
+            [0u32, 10u32].iter().copied().map(WasmPageNumber).collect();
```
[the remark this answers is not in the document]
grishasobol: It's just a test)) I use it to be able map u32 not &u32

runtime-interface/src/lib.rs
grishasobol: Done

runtime-interface/src/lib.rs
grishasobol: Done

PR #937, opened 2022-05-13
Pass actual errors through syscalls

### LINKED ISSUE #912: Syscall error codes
### Problem to Solve

We merged #808 and #872 so we can pass errors through syscalls using error codes now

### Possible Solution

Add representation of errors from `core-errors` as codes and pass them though syscalls

### Notes

_No response_

ark0f:
Resolves #912.

At the moment when you call, for example, `msg::send` you get `Result<_, SendError>` where `SendError(())` is just tuple struct so you have no information about actual error. This PR fixes this problem

After discussion with @breathx it was decided to use SCALE codec to pass errors through FFI

Also update `@gear-js/api` to `0.18.2`


core-backend/common/src/lib.rs
```diff
@@ -136,34 +140,230 @@ pub trait Environment<E: Ext + IntoExtInfo + 'static>: Sized {
...
 }
 
-impl<E> IntoErrorCode for Result<(), E> {
-    fn into_error_code(self) -> i32 {
-        match self {
-            Ok(()) => 0,
-            // TODO: actual error codes
-            Err(_) => 1,
-        }
+impl<E> ErrorSavingExt<E>
```
SabaunT, line 194: If the only purpose of the type is to store errors, then we copy existing logic of storing errors inside `processor::ext::Ext`, which saves errors to one of it's fields. Maybe better to introduce a separate function in `Ext` trait to **get** errors? Or separate trait for **saving** errors and **getting** them from `Ext` implementors?

core-backend/sandbox/src/env.rs
```diff
@@ -100,15 +101,15 @@ impl<E: Ext + IntoExtInfo + 'static> Environment<E> for SandboxEnvironment<E> {
...
+        let ext_carrier = ExtCarrier::new(ErrorSavingExt::new(ext));
 
         let mem: DefaultExecutorMemory = match SandboxMemory::new(mem_size.0, None) {
             Ok(mem) => mem,
             Err(e) => {
                 return Err(BackendError {
                     reason: SandboxEnvironmentError::CreateEnvMemory,
                     description: Some(format!("{:?}", e).into()),
-                    gas_amount: ext_carrier.into_inner().into_gas_amount(),
+                    gas_amount: ext_carrier.into_inner().inner.into_gas_amount(),
```
breathx, line 112:
```suggestion
                    gas_amount: ext_carrier.into_inner().into_gas_amount(),
```

`into_inner().inner` seems extraordinary strange to me :D
 Can we deal with it somehow?

core-backend/sandbox/src/funcs.rs
```diff
@@ -120,11 +122,13 @@ impl<E: Ext + 'static> FuncsHandler<E> {
...
-                .map_err(FuncError::Core)
-                .on_success_code(|message_id| wto(memory, message_id_ptr, message_id.as_ref()))
+            Ok(ExtErrorProcessor::new(ext)
+                .with(|ext| ext.send(HandlePacket::new(dest, payload, value)))
+                .map_err(FuncError::Core)?
+                .on_success(|message_id| wto(memory, message_id_ptr, message_id.as_ref()))?
+                .error_len())
         })
-        .map(|code| Value::I32(code).into())
+        .map(|code| Value::I32(code as i32).into())
```
breathx, line 131: why do we need type cast here? what's return type of inner closure?

core-backend/sandbox/src/funcs.rs
```diff
@@ -120,11 +122,13 @@ impl<E: Ext + 'static> FuncsHandler<E> {
...
         .map_err(|err| {
```
breathx, line 132: Could we directly store error to ctx field from extErrorSaving implementation?

SabaunT (changes requested):
Interim review

core-backend/wasmtime/src/env.rs
```diff
@@ -199,14 +201,15 @@ where
...
+        funcs.insert("gr_error", funcs::error(&mut store, memory));
 
         let module = match Module::new(&engine, binary) {
             Ok(module) => module,
             Err(e) => {
                 return Err(BackendError {
                     reason: WasmtimeEnvironmentError::ModuleCreation,
                     description: Some(e.to_string().into()),
-                    gas_amount: ext_carrier.into_inner().into_gas_amount(),
+                    gas_amount: ext_carrier.into_inner().inner.into_gas_amount(),
```
breathx, line 212: same all comments as in sandbox backend

gcore/Cargo.toml
```diff
@@ -8,5 +8,9 @@ license = "GPL-3.0"
 [features]
 debug=[]
 
+[dependencies]
+gear-core-errors = { path = "../core-errors" }
+codec = { package = "parity-scale-codec", version = "3.1.2", default-features = false }
```
breathx, line 13: there should no be codec dependency. gcore is a minimal implementation. provide just first byte as the num of the error for it.

core-errors/src/lib.rs
```diff
@@ -22,10 +22,14 @@ use codec::{Decode, Encode};
 use core::fmt;
```
breathx, line 22: Left the comment in gcore about codec. better to provide here feature "codec" to impl `Encode + Decode + TypeInfo` if only feature appeared

core/src/env.rs
```diff
@@ -260,6 +260,14 @@ mod tests {
...
             unreachable!()
         }
+
+        fn into_ext_error(self) -> Result<ExtError, Self> {
+            unreachable!()
+        }
+
+        fn as_ext_error(&self) -> Option<&ExtError> {
+            unreachable!()
+        }
```
breathx, line 270: why do they unreachable?

gstd/src/common/errors.rs
```diff
@@ -32,8 +33,7 @@ pub enum ContractError {
     Decode(codec::Error),
     ExitCode(i32),
     Internal(&'static str),
-    Sending(SendError),
-    Reply(ReplyError),
+    Ext(ExtError),
```
breathx, line 36: Internal meant as Ext error before, what about merging it?

breathx (commented):
Please, resolve commented issues, and this will be insane improvement of our ecosystem!
Well done! 👍 🚀

gcore/Cargo.toml
ark0f: I cannot, because error like `A::B(B::C(C::Example))` will be encoded like `[0, 0, 1]`

core/src/env.rs
ark0f: Not used in tests

core-backend/common/src/lib.rs
```diff
@@ -136,34 +140,230 @@ pub trait Environment<E: Ext + IntoExtInfo + 'static>: Sized {
...
+    fn alloc(
+        &mut self,
+        pages: WasmPageNumber,
+        mem: &mut dyn Memory,
+    ) -> Result<WasmPageNumber, Self::Error> {
+        let res = self.inner.alloc(pages, mem);
+        self.return_and_store_err(res)
+    }
+
+    fn block_height(&mut self) -> Result<u32, Self::Error> {
```
shamilsan, line 226: Is there any way to make these wrappers general? Bad perspective to add a new wrapper per every new syscall.

core-backend/sandbox/src/funcs.rs
ark0f: `error_len()` returns u32 and syscalls themselves wait for it but sandbox allows to pass only signed integers

core-backend/sandbox/src/env.rs
ark0f: Impl `IntoExtInfo` for `ErrorSavingExt`

core-backend/sandbox/src/funcs.rs
ark0f: What error?

gcore/Cargo.toml
breathx: ok, throw just bytes, let it be on the user's shoulders

core/src/env.rs
breathx: Oh, sorry. Didn't see that it's tests module in diff

core-backend/sandbox/src/funcs.rs
breathx: For this case, for example, `.map_err(FuncError::Core)?`

breathx: this ext implementation already saved this error in memory, isn't it?

ark0f: Yes, it did

ark0f: `ErrorSavingExt` saves only `ExtError`. The rest errors are thrown and saved in `ctx`

core-errors/src/lib.rs
ark0f: > better to provide here feature "codec" to impl Encode + Decode + TypeInfo if only feature appeared

What's the matter if `gr_error` always need `Decode`?

core-backend/common/src/lib.rs
```diff
@@ -136,34 +140,230 @@ pub trait Environment<E: Ext + IntoExtInfo + 'static>: Sized {
...
+        self.success.take().map(f).transpose()?;
+        Ok(self)
+    }
+
+    pub fn error_len(self) -> u32 {
+        self.ext
+            .err
+            .as_ref()
+            .and_then(|err| err.as_ext_error())
+            .map(|err| err.encoded_size() as u32)
```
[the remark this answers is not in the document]
ark0f: > remove visible dependency on codec from core-backend

Why is this needed?

core-backend/common/src/lib.rs
breathx: I would provide some trait to workaround different ext wrappers.
Something like

```rust
trait WithExt {
    with_ext<T, E>( .. ) -> Result<T, E>;
    with_ext_infallible<T>( .. ) -> T;
}
```

and impl for different essences:
```rust
impl WithExt for ErrorSavingExt {
    with_ext<T, E>( .. ) -> Result<T, E> {
        let res = self.ext.with( ... );
        self.return_and_store_error(res)
    }
}
```
etc

Does it seem like real solution and code base reducer?

breathx: @ukint-vs @ark0f please, pay attention on that idea for your ext-summoning work

core-backend/common/src/lib.rs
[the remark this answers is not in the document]
breathx: Because our common part of backends is out of blockchain/wasmtime/sandbox/scale-codec-as-default-format context

core-errors/src/lib.rs
breathx: Because gcore as minimal lib as it could be.
codec brings a lot of bloat for final binary size

core-backend/common/src/lib.rs
ark0f: I applied your suggestion in a some way

ark0f:
So, I applied @SabaunT's suggestion. It leads to less code but gives us less guarantees because Ext's `error_expanation` is always stores something after the first syscall when error was replaced to `None` after error encoding earlier

gcore/src/error.rs
```diff
@@ -0,0 +1,56 @@
...
+pub struct SyscallError {
+    len: u32,
+}
+
+impl SyscallError {
+    pub fn into_result(self) -> Result<(), ExtError> {
+        if self.len == 0 {
+            Ok(())
+        } else {
+            #[cfg(feature = "codec")]
```
SabaunT, line 40: I guess `cfg!` will increase readability. What do you think? Also if we bear in mind costs of not removing code (`cfg!` doesn't remove unused branches)

SabaunT (commented):
:+1:

gcore/src/error.rs
ark0f: > Also if we bear in mind costs of not removing code

I think compiler will remove it (dead code elimination)

ark0f: I decided to leave it as is because compiler always check the code in branch with `cfg!()` and we have conditional imports

SabaunT: if there are no cons from time cost perspective, then using `cfg!()` will give you more readability, right?

SabaunT: alright

core-errors/src/lib.rs
```diff
@@ -116,7 +110,10 @@ pub enum TerminationReason {
 }
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq, Hash, derive_more::Display)]
+#[cfg_attr(feature = "codec", derive(Encode, Decode, TypeInfo))]
 pub enum ExtError {
+    #[display(fmt = "Unknown error")]
+    Unknown,
```
SabaunT, line 116: Maybe `SomeError`? I mean we need such a name, that emphasize that there is actually an error, but you don't know exactly the variant

gcore/src/error.rs
```diff
@@ -0,0 +1,56 @@
...
+        pub fn gr_error(data: *mut u8);
+    }
+}
+
+#[must_use]
+#[repr(transparent)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct SyscallError {
+    len: u32,
+}
```
SabaunT, line 33: :clap:

gcore/src/error.rs
```diff
@@ -0,0 +1,56 @@
...
+        } else {
+            #[cfg(feature = "codec")]
+            unsafe {
+                use alloc::vec;
+                use codec::Decode;
+
+                let mut data = vec![0; self.len as usize];
+                sys::gr_error(data.as_mut_ptr());
+                Err(ExtError::decode(&mut data.as_slice()).expect("error decoded successfully"))
+            }
```
SabaunT, line 48: maybe wrap this with some safe function?

SabaunT: it would be really nice to have one with some doc-comments pointing to the fact, that we get last error and bla bla

SabaunT (approved):
:100: 
Please look at minor comments I left. Moving unsafe code into safe func is really preferable.

core-backend/common/src/lib.rs
```diff
@@ -135,34 +138,47 @@ pub trait Environment<E: Ext + IntoExtInfo + 'static>: Sized {
...
-        match self {
-            Ok(t) => f(t).map(|_| 0),
-            Err(_) => Ok(1),
-        }
+        match f(self.ext) {
+            Ok(t) => {
+                self.success = Some(t);
+            }
+            Err(err) => {
+                err.into_ext_error()?;
```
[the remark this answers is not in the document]
ark0f: You're right

gcore/src/error.rs
ark0f: I don't think we need to introduce an another function for so small function. I'll just leave comment on `unsafe` block

ark0f:
Waiting on #981 so test can be passed

gcore/src/error.rs
SabaunT: well it's commonly recommended pattern, better follow commonly accepted rules.

ark0f: What pattern?

SabaunT: https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html

```md
To isolate unsafe code as much as possible, it’s best to enclose unsafe code within a safe abstraction and provide a safe API, which we’ll discuss later in the chapter when we examine unsafe functions and methods. Parts of the standard library are implemented as safe abstractions over unsafe code that has been audited. Wrapping unsafe code in a safe abstraction prevents uses of unsafe from leaking out into all the places that you or your users might want to use the functionality implemented with unsafe code, because using a safe abstraction is safe.
```

core-backend/common/src/lib.rs
```diff
@@ -130,34 +133,47 @@ pub trait Environment<E: Ext + IntoExtInfo + 'static>: Sized {
...
-    fn on_success_code<F>(self, mut f: F) -> Result<i32, E>
+impl<'a, T, E> ExtErrorProcessor<'a, T, E>
+where
+    E: Ext + IntoExtInfo,
+{
+    pub fn new(ext: &'a mut E) -> Self {
+        Self { ext, success: None }
+    }
+
+    pub fn with<U, F>(mut self, f: F) -> Result<Self, U>
```
shamilsan, line 149: I'm not a fan of such kind functions with side effects like 'with' and 'on_success' because they have a quite strict using contract and may yield errors when called improperly. May we pass 'Result' instead of 'Self' to make them more pure?

core-backend/sandbox/src/funcs.rs
```diff
@@ -754,4 +786,24 @@ impl<E: Ext + 'static> FuncsHandler<E> {
...
+
+        let Runtime { ext, memory, .. } = ctx;
+
+        ext.with_fallible(|ext| {
+            let err = ext.last_error().ok_or(FuncError::SyscallErrorExpected)?;
+            let err = err.encode();
+            wto(memory, data_ptr, &err)?;
+            Ok(())
+        })
+        .map(|()| ReturnValue::Unit)
```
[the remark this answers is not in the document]
ark0f: It's map's argument so I just wrote `()` instead of `_`

core-backend/common/src/lib.rs
ark0f: We cannot pass `T` instead because `Self` is used further. If you afreaid of robustness I can use compile-time state so contract won't be violated

shamilsan: As I see it is used for `error_len()` only. Ok, if there is no better solution, some compile-time checks would be great.

core-backend/sandbox/src/funcs.rs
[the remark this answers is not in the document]
shamilsan: Yes, you are right, `map` calls closures with argument only, so `map(|_| ...)` syntax is more usual for me though.

shamilsan (approved):
In general looks good to me 👍

core-backend/sandbox/src/funcs.rs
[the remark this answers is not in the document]
ark0f: Unit type is here because I don't want to just silently ignore `map`'s argument

NikVolf:
@ark0f Can you outline how error reporting imoroves in the PR description?

ark0f:
> @ark0f Can you outline how error reporting imoroves in the PR description?

Done

PR #941, opened 2022-05-15
update PageBuf and remove allowance_exceed

### LINKED ISSUE #881: Change pages data from Vec<u8> to PageBuf everywhere in runtime.
### File Location(s)

_No response_

### Proposal

Currently there is many places where Vec<u8> is used to represent page data, but it's not actually right way - so I think we must change all them to PageBuf which is fixed size buffer

grishasobol:
Resolves #881  .

1) Uses now PageBuf everywhere for page data representation. I have found a way (thanks to @ark0f) to transform vec to PageBuf without any relocations, so now it's almost for free. Using PageBuf is more profitable for compiler, which can apply more optimisations in case it knows buffer size in compile time.
2) Remove allowance_exceed and use error reason instead.

### DIFF
--- a/common/src/lazy_pages.rs
+++ b/common/src/lazy_pages.rs
@@ -19,8 +19,8 @@
 //! Lazy pages support runtime functions
 
 use crate::Origin;
-use gear_core::{ids::ProgramId, memory::PageBuf};
 use gear_core::memory::PageNumber;
+use gear_core::{ids::ProgramId, memory::PageBuf};
 use gear_runtime_interface::{gear_ri, GetReleasedPageError, MprotectError};
 use sp_std::{
     collections::{btree_map::BTreeMap, btree_set::BTreeSet},
--- a/common/src/lazy_pages.rs
+++ b/common/src/lazy_pages.rs
@@ -19,7 +19,7 @@
 //! Lazy pages support runtime functions
 
 use crate::Origin;
-use gear_core::ids::ProgramId;
+use gear_core::{ids::ProgramId, memory::PageBuf};
 use gear_core::memory::PageNumber;
 use gear_runtime_interface::{gear_ri, GetReleasedPageError, MprotectError};
 use sp_std::{
@@ -37,11 +37,6 @@ pub enum Error {
     VecToPageData,
     #[display(fmt = "Cannot find page data in storage")]
     NoPageDataInStorage,
-    #[display(
-        fmt = "RUNTIME ERROR: released page {:?} already has data in provided page map to core processor",
-        _0
-    )]
-    ReleasedPageHasData(PageNumber),
 }
 
 impl From<MprotectError> for Error {
@@ -99,16 +94,14 @@ pub fn protect_pages_and_init_info(
 
 /// Lazy pages contract post execution actions
 pub fn post_execution_actions(
-    memory_pages: &mut BTreeMap<PageNumber, Vec<u8>>,
+    memory_pages: &mut BTreeMap<PageNumber, PageBuf>,
     wasm_mem_begin_addr: u64,
 ) -> Result<(), Error> {
     // Loads data for released lazy pages. Data which was before execution.
     let released_pages = gear_ri::get_released_pages();
     for page in released_pages {
         let data = gear_ri::get_released_page_old_data(page)?;
-        if memory_pages.insert(page.into(), data).is_some() {
-            return Err(Error::ReleasedPageHasData(page.into()));
-        };
+        let _ = memory_pages.insert(page.into(), data).is_some();
     }
 
     // Removes protections from lazy pages
--- a/common/src/lazy_pages.rs
+++ b/common/src/lazy_pages.rs
@@ -19,12 +19,10 @@
 //! Lazy pages support runtime functions
 
 use crate::Origin;
-use core::convert::TryFrom;
 use gear_core::ids::ProgramId;
-use gear_core::memory::{PageBuf, PageNumber};
+use gear_core::memory::PageNumber;
 use gear_runtime_interface::{gear_ri, GetReleasedPageError, MprotectError};
 use sp_std::{
-    boxed::Box,
     collections::{btree_map::BTreeMap, btree_set::BTreeSet},
     vec::Vec,
 };
@@ -101,18 +99,14 @@ pub fn protect_pages_and_init_info(
 
 /// Lazy pages contract post execution actions
 pub fn post_execution_actions(
-    memory_pages: &mut BTreeMap<PageNumber, Box<PageBuf>>,
+    memory_pages: &mut BTreeMap<PageNumber, Vec<u8>>,
     wasm_mem_begin_addr: u64,
 ) -> Result<(), Error> {
     // Loads data for released lazy pages. Data which was before execution.
     let released_pages = gear_ri::get_released_pages();
     for page in released_pages {
         let data = gear_ri::get_released_page_old_data(page)?;
-        let page_data = PageBuf::try_from(data).map_err(|_| Error::VecToPageData)?;
-        if memory_pages
-            .insert(page.into(), Box::new(page_data))
-            .is_some()
-        {
+        if memory_pages.insert(page.into(), data).is_some() {
             return Err(Error::ReleasedPageHasData(page.into()));
         };
     }
--- a/core-processor/src/common.rs
+++ b/core-processor/src/common.rs
@@ -268,9 +268,7 @@ pub struct ExecutionError {
     /// Gas amount of the execution.
     pub gas_amount: GasAmount,
     /// Error text.
-    pub reason: Option<ExecutionErrorReason>,
-    /// Triggered by gas allowance exceed.
-    pub allowance_exceed: bool,
+    pub reason: ExecutionErrorReason,
 }
 
 /// Reason of execution error
@@ -318,6 +316,9 @@ pub enum ExecutionErrorReason {
     /// Ext works with lazy pages, but lazy pages env is not enabled
     #[display(fmt = "Ext works with lazy pages, but lazy pages env is not enabled")]
     LazyPagesInconsistentState,
+    /// Page with data is not allocated for program
+    #[display(fmt = "Page with data is not allocated for program: {:?}", _0)]
+    PageIsNotAllocated(PageNumber),
 }
 
 /// Executable actor.
--- a/core-processor/src/common.rs
+++ b/core-processor/src/common.rs
@@ -25,6 +25,7 @@ use alloc::{
     vec::Vec,
 };
 use codec::{Decode, Encode};
+use gear_core::memory::PageBuf;
 use gear_core::{
     gas::GasAmount,
     ids::{CodeId, MessageId, ProgramId},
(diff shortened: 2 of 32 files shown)


core-processor/src/common.rs
```diff
@@ -318,6 +316,9 @@ pub enum ExecutionErrorReason {
     /// Ext works with lazy pages, but lazy pages env is not enabled
     #[display(fmt = "Ext works with lazy pages, but lazy pages env is not enabled")]
     LazyPagesInconsistentState,
+    /// Page with data is not allocated for program
+    #[display(fmt = "Page with data is not allocated for program: {:?}", _0)]
```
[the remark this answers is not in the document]
grishasobol: Oh I think is redundant)) it's not what we really need now

ark0f:
> Remove PageBuf, because Vec into Box using try_from uses too many data copying: from Vec to stack allocated memory in PageBuf array, then back to heap from stack memory in Boxed object (I've checked this on real examples).

I see `PageBuf` is always wrapped in `Box` so you can just use `Vec::into_boxed_slice` and [`Box::from`](https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-From%3CBox%3C%5BT%5D%2C%20A%3E%3E) so we won't get reallocations

core-processor/src/common.rs
```diff
@@ -317,7 +317,7 @@ pub enum ExecutionErrorReason {
     #[display(fmt = "Ext works with lazy pages, but lazy pages env is not enabled")]
     LazyPagesInconsistentState,
     /// Page with data is not allocated for program
-    #[display(fmt = "Page with data is not allocated for program: {:?}", _0)]
+    #[display(fmt = "{:?} is not allocated for program", _0)]
```
ark0f, line 320: Now it will look like `PageNumber(123) is not allocated for program` which is kinda confusing

ark0f: Nevermind

grishasobol:
> > Remove PageBuf, because Vec into Box using try_from uses too many data copying: from Vec to stack allocated memory in PageBuf array, then back to heap from stack memory in Boxed object (I've checked this on real examples).
> 
> I see `PageBuf` is always wrapped in `Box` so you can just use `Vec::into_boxed_slice` and [`Box::from`](https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-From%3CBox%3C%5BT%5D%2C%20A%3E%3E) so we won't get reallocations

Hmm, may be you are right I should check how it works

core/src/program.rs
```diff
@@ -181,8 +166,5 @@ mod tests {
 
         // 2 static pages
         assert_eq!(program.static_pages(), 2.into());
-
-        // no allocations because we do not set them in new
```
breathx, line 185: why have been this assert removed?

grishasobol: Ok return it back

common/src/lazy_pages.rs
```diff
@@ -101,20 +94,14 @@ pub fn protect_pages_and_init_info(
...
-    memory_pages: &mut BTreeMap<PageNumber, Box<PageBuf>>,
+    memory_pages: &mut BTreeMap<PageNumber, PageBuf>,
     wasm_mem_begin_addr: u64,
 ) -> Result<(), Error> {
     // Loads data for released lazy pages. Data which was before execution.
     let released_pages = gear_ri::get_released_pages();
     for page in released_pages {
         let data = gear_ri::get_released_page_old_data(page)?;
-        let page_data = PageBuf::try_from(data).map_err(|_| Error::VecToPageData)?;
-        if memory_pages
```
ark0f, line 112: Why this check have been removed?

core/src/memory.rs
```diff
@@ -36,10 +38,23 @@ const GEAR_PAGE_SIZE: usize = 0x1000;
 /// Number of gear pages in one wasm page
 const GEAR_PAGES_IN_ONE_WASM: u32 = (WASM_PAGE_SIZE / GEAR_PAGE_SIZE) as u32;
 
-pub use gear_core_errors::MemoryError as Error;
+/// Buffer for gear page data.
+pub type PageBuf = Box<[u8; GEAR_PAGE_SIZE]>;
```
ark0f, line 42: What about to make `PageBuf` as newtype and move functions below to its impl?

pallets/gear-debug/src/lib.rs
```diff
@@ -86,7 +86,7 @@ pub mod pallet {
     #[derive(Encode, Decode, Clone, Default, PartialEq, TypeInfo)]
     pub struct ProgramInfo {
         pub static_pages: WasmPageNumber,
-        pub persistent_pages: BTreeMap<PageNumber, Vec<u8>>,
```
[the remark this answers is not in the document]
grishasobol: Not actual now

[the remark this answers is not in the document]
grishasobol: Return back to vec here because polkadot-js api does not support boxed arrays

common/src/lazy_pages.rs
grishasobol: Because now we receive PageBuf from gear_ri

core/src/memory.rs
grishasobol: Yes, it's good idea. I will make it in next PR if this will be committed

grishasobol: Ok done now

core/src/memory.rs
```diff
@@ -36,10 +42,46 @@ const GEAR_PAGE_SIZE: usize = 0x1000;
 /// Number of gear pages in one wasm page
 const GEAR_PAGES_IN_ONE_WASM: u32 = (WASM_PAGE_SIZE / GEAR_PAGE_SIZE) as u32;
 
-pub use gear_core_errors::MemoryError as Error;
+/// Buffer for gear page data.
+/// TODO: impl self debug
```
ark0f, line 46: TODO line must be comment, not doc string

pallets/gear-debug/src/lib.rs
```diff
@@ -83,6 +83,9 @@ pub mod pallet {
     #[pallet::error]
     pub enum Error<T> {}
 
+    /// Program debug info.
+    /// TODO: unfortunatelly we cannot store pages data in [PageBuf],
```
ark0f, line 87: TODO line must be comment, not doc string

core-processor/src/executor.rs
```diff
@@ -313,8 +304,8 @@ pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environ
             }
         }
 
-        if let Some(initial_data) = pages_initial_data.get(&page) {
-            if !new_data.eq(initial_data.as_ref()) {
+        if let Some(initial_data) = pages_data.get(&page) {
+            if !new_data.0.eq(&initial_data.0) {
```
ark0f, line 308:
```suggestion
            if new_data.0 != initial_data.0 {
```

core/src/memory.rs
```diff
@@ -36,10 +42,46 @@ const GEAR_PAGE_SIZE: usize = 0x1000;
...
+#[derive(Clone, Debug, Encode, Decode, PartialEq, Eq)]
+pub struct PageBuf(pub Box<[u8; GEAR_PAGE_SIZE]>);
```
ark0f, line 48: Maybe it will be better to implement `Deref` to avoid these `.0` and make field private?

pallets/gear-program/src/lib.rs
```diff
@@ -139,6 +140,9 @@ pub mod pallet {
         /// - `value`: balance to be transferred to the program once it's been resumed.
         ///
         /// - `ProgramResumed(H256)` in the case of success.
+        ///
+        /// TODO: unfortunatelly we cannot pass pages data in [PageBuf],
```
ark0f, line 144: TODO line must be comment, not doc string

core/src/memory.rs
```diff
@@ -36,10 +42,46 @@ const GEAR_PAGE_SIZE: usize = 0x1000;
...
+
+impl PageBuf {
+    /// Tries to transform vec<u8> into page buffer.
+    /// Makes it without any rellocations or memcpy: vector's buffer becames PageBuf without any changes,
+    /// except vector's buffer capacity, which is removed.
+    pub fn new_from_vec(v: Vec<u8>) -> Result<Self, Error> {
```
ark0f, line 54: Consider using `TryFrom` instead

pallets/gear-debug/src/lib.rs
grishasobol: Done, make self debug

core/src/memory.rs
grishasobol: Ok done

core-processor/src/executor.rs
grishasobol: Ok done

core/src/memory.rs
grishasobol: No, I think this is redundant

pallets/gear-program/src/lib.rs
grishasobol: Ok done

core/src/memory.rs
grishasobol: Ok done

core/src/memory.rs
```diff
@@ -36,10 +46,69 @@ const GEAR_PAGE_SIZE: usize = 0x1000;
...
 const GEAR_PAGES_IN_ONE_WASM: u32 = (WASM_PAGE_SIZE / GEAR_PAGE_SIZE) as u32;
 
-pub use gear_core_errors::MemoryError as Error;
+/// Buffer for gear page data.
+#[derive(Clone, Encode, Decode, PartialEq, Eq)]
+pub struct PageBuf(Box<[u8; GEAR_PAGE_SIZE]>);
+
+impl fmt::Debug for PageBuf {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(
```
ark0f, line 55:
```rust
f.debug_list()
	.entries(&self.0[0..10])
	.entry(format_args!(".."))
	.entries(&self.0[GEAR_PAGE_SIZE - 10..GEAR_PAGE_SIZE])
	.finish()
```

grishasobol: No, I think the better is my variant

pallets/gear-debug/src/lib.rs
grishasobol: Done

core/src/memory.rs
ark0f: What's the difference?

core/src/memory.rs
```diff
@@ -348,4 +417,19 @@ mod tests {
 
         assert!(gear_pages.eq(&expectation));
     }
+
+    #[test]
+    fn page_buf() {
```
ark0f, line 422: What is this test for?

core/src/memory.rs
grishasobol: It's more understandable

core/src/memory.rs
grishasobol: To cover page can be crated from vec and debug print works

PR #944, opened 2022-05-16
Add meta state testing tools

### LINKED ISSUE #862: Add tools to `gtest` for testing meta state functions
### Problem to Solve

We can't test `meta_state()` function in smart contracts at the moment. But it would be great to cover such kind of functions too.

### Possible Solution

Add some tools to `gtest` that execute `meta_state()` function from the `[.meta].wasm` and return the state.

### Notes

_No response_

dustyNdumb:
Resolves #862.

- Add ability to run custom meta funcs from wasm

**Release Notes**: 

- Provide ability to execute non-void meta functions from wasm binaries. These executions don't affect the binary itself and don't change memory pages. 
- For such executions was also provided public api in ```gtest::Program```. This api allows to execute functions with and without arguments. Arguments can be provided as bytes array or as struct that implements ```Encode``` and ```Decode``` traits.
- This is exclusively testing tool. Main purpose is to provide more information about your program at any testing step bun not to change it by that


gtest/src/wasm_executor.rs
```diff
@@ -0,0 +1,349 @@
...
+        let mut store = Store::<StoreData<ExtImplementedStruct>>::new(&engine, store_data);
+
+        let module = Module::new(&engine, program.code().code()).expect("Failed to create module");
+
+        let mut memory =
+            WasmtimeMemory::new(&mut store, MemoryType::new(program.static_pages().0, None))
+                .expect("Failed to create memory");
+
+        let mut funcs = BTreeMap::<&'static str, Func>::new();
+        funcs.insert("alloc", FuncsHandler::alloc(&mut store, memory));
```
dustyNdumb, line 50: I want to move code that is building funcs tree from  ```core-backend::wasmtime::env::WasmtimeEnvironment``` to new struct and reuse it here
Would be great to hear somebody's opinion about this idea

gtest/src/wasm_executor.rs
```diff
@@ -0,0 +1,349 @@
...
+            .collect()
+    }
+
+    fn get_function(&mut self, function_name: &str) -> Func {
+        self.instance
+            .get_func(&mut self.store, function_name)
+            .expect("No function with such name")
+    }
+
+    fn read_result(&mut self, ptr_to_result_data: i32) -> Vec<u8> {
```
dustyNdumb, line 204: This low-level part is pretty hard to understand. Going to add some comments here but will that be enough?

gtest/src/manager.rs
```diff
@@ -445,6 +449,61 @@ impl ExtManager {
...
+        core_processor::handle_journal(journal, self);
+        result
+    }
+
+    #[allow(dead_code)]
+    pub(crate) fn execute_custom_void_function(
+        &mut self,
+        source: ProgramId,
+        program_id: &ProgramId,
+        message_option: Option<IncomingMessage>,
```
shamilsan, line 472: No need to highlight that this is an option.
```suggestion
        message: Option<IncomingMessage>,
```

gtest/src/manager.rs
```diff
@@ -445,6 +449,61 @@ impl ExtManager {
 
         core_processor::handle_journal(journal, self);
     }
+
+    #[allow(dead_code)]
+    pub(crate) fn execute_custom_function(
```
shamilsan, line 454: I'd rename the functions. 
```suggestion
    pub(crate) fn execute_with_result(
```

gtest/src/manager.rs
```diff
@@ -445,6 +449,61 @@ impl ExtManager {
...
+        function_name: &str,
+    ) -> Vec<u8> {
+        let mut executor = self.get_executor(source, program_id, message_option);
+        let (result, journal) = executor.call_function(function_name);
+        core_processor::handle_journal(journal, self);
+        result
+    }
+
+    #[allow(dead_code)]
+    pub(crate) fn execute_custom_void_function(
```
shamilsan, line 468:
```suggestion
    pub(crate) fn execute(
```

gtest/src/manager.rs
```diff
@@ -445,6 +449,61 @@ impl ExtManager {
 
         core_processor::handle_journal(journal, self);
     }
+
+    #[allow(dead_code)]
+    pub(crate) fn execute_custom_function(
+        &mut self,
+        source: ProgramId,
+        program_id: &ProgramId,
+        message_option: Option<IncomingMessage>,
```
shamilsan, line 458:
```suggestion
        message: Option<IncomingMessage>,
```

gtest/src/manager.rs
```diff
@@ -445,6 +449,61 @@ impl ExtManager {
...
+    ) -> WasmExecutor {
+        let (actor, balance) = self.actors.get_mut(program_id).unwrap();
+        if actor.is_uninitialized() {
+            actor.set_initialized();
+        }
+        if let Some(message) = message_option.clone() {
```
shamilsan, line 490: Can we avoid clone here by borrowing `message_option`?

gtest/src/wasm_executor.rs
```diff
@@ -0,0 +1,349 @@
...
+    message::IncomingMessage,
+    program::Program,
+};
+use std::{collections::BTreeMap, convert::TryInto};
+use wasmtime::{
+    Config, Engine, Extern, Func, Instance, Memory as WasmtimeMemory, MemoryType, Module, Store,
+    Val,
+};
+
+#[allow(dead_code)]
```
shamilsan, line 17: Why it can be dead if it is a part of public API?

shamilsan (commented):
It is extremely copy-paste, I am not sure it is acceptable to support three "backends" instead of two 😞

gtest/src/wasm_executor.rs
```diff
@@ -0,0 +1,349 @@
...
+
+    fn set_pages<T: Ext>(
+        mut store: &mut Store<StoreData<T>>,
+        memory: &mut WasmtimeMemory,
+        pages: &BTreeMap<PageNumber, Box<PageBuf>>,
+    ) -> Result<(), String> {
+        let memory_size = WasmPageNumber(memory.size(&mut store) as u32);
+        for (page_number, buffer) in pages {
+            if memory_size <= page_number.to_wasm_page() {
+                panic!("Memory size {:?} less then {:?}", memory_size, page_number);
```
shamilsan, line 249:
```suggestion
                panic!("Memory size {:?} less than {:?}", memory_size, page_number);
```

gtest/src/wasm_executor.rs
```diff
@@ -0,0 +1,349 @@
...
+
+    fn get_function(&mut self, function_name: &str) -> Func {
+        self.instance
+            .get_func(&mut self.store, function_name)
+            .expect("No function with such name")
+    }
+
+    fn read_result(&mut self, ptr_to_result_data: i32) -> Vec<u8> {
+        let mut ptr_to_result_buffer: Vec<u8> = Vec::new();
+        let buffer_size = ptr_to_result_data.to_be_bytes().len();
```
shamilsan, line 206: You can use `mem::size_of` or `mem::size_of_val` here I guess.

gtest/src/wasm_executor.rs
```diff
@@ -0,0 +1,349 @@
...
+                &mut result_len_buffer,
+            )
+            .expect("Failed to read data length");
+
+        let decoded_ptr_to_result =
+            i32::from_ne_bytes(ptr_to_result_buffer.try_into().unwrap()) as usize;
+        let result_len = result_len_buffer[0] as usize;
+
+        let mut results = Vec::new();
+        results.resize(result_len, 0u8);
```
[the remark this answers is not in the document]
shamilsan: As an option (using arrays instead of Vecs allows minimize usage of the heap).

```rust
fn read_result(&mut self, ptr_to_result_data: i32) -> Vec<u8> {
    let offset = ptr_to_result_data as usize;
    let mut ptr = [0_u8; mem::size_of::<i32>()];

    self.memory
        .read(&self.store, offset, &mut ptr)
        .expect("Failed to read data ptr");

    let mut len = [0_u8; mem::size_of::<i32>()];

    self.memory
        .read(&self.store, offset + ptr.len(), &mut len)
        .expect("Failed to read data length");

    let ptr = i32::from_ne_bytes(ptr) as usize;
    let len = i32::from_ne_bytes(len) as usize;

    let mut result = vec![0; len];

    self.memory
        .read(&self.store, ptr, &mut result)
        .expect("Failed to read result");

    result
}
```

gtest/src/wasm_executor.rs
shamilsan: Yes, we can use a shared function that creates this `BTreeMap` of functions.

gtest/src/wasm_executor.rs
shamilsan: See below: https://github.com/gear-tech/gear/pull/944#discussion_r892896020

gtest/src/manager.rs
dustyNdumb: Compiler don't let me to do that because we are borrowing it later to create wasm_executor instance

gtest/src/wasm_executor.rs
dustyNdumb: I add them to disable warnings about unused code. I will delete it when pr will be on the final stage

gtest/src/manager.rs
shamilsan: You are confusing moving (changing ownership) and borrowing.

```rust
        if let Some(message) = &message_option {
            if message.source() != source {
                panic!("Source id in message: {:?} is not equal to source id in function arguments: {:?}", message.source(), source);
            }
        }
```

core-backend/wasmtime/src/funcs_tree_builder.rs
```diff
@@ -0,0 +1,61 @@
...
+use gear_backend_common::{error_processor::IntoExtError, AsTerminationReason, IntoExtInfo};
+use gear_core::env::Ext;
+use wasmtime::{Func, Memory, Store};
+
+pub fn get_funcs_tree<E>(store: &mut Store<StoreData<E>>, memory: Memory) -> BTreeMap<&str, Func>
+where
+    E: Ext + IntoExtInfo + 'static,
+    E::Error: AsTerminationReason + IntoExtError,
+{
+    let mut funcs = BTreeMap::<&str, Func>::new();
```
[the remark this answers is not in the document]
dustyNdumb: I'm afraid we can't use array because now we have such thing as ```forbidden_funcs``` in func tree which are not const size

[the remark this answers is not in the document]
shamilsan: We apply `forbidden_funcs` to the `BTreeMap` later. 

```rust
let funcs: BTreeMap<_,_> = [(..., ...)].into();
for name in forbidden_funcs {
    funcs.insert(name, Funcs::forbidden(&mut store));
}
```

But it is optional, you may leave it as is.

core-backend/wasmtime/src/env.rs
```diff
@@ -20,6 +20,7 @@
 
 use core::fmt;
 
+use crate::funcs_tree_builder::get_funcs_tree;
```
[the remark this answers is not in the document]
breathx: Agree here `mod func_tree` with just `get()`seems more accurate.

core-backend/wasmtime/src/funcs_tree_builder.rs
```diff
@@ -0,0 +1,61 @@
+use crate::{env::StoreData, funcs::FuncsHandler};
```
breathx, line 1: But in general, I don't see any reason to separate this logic into separate logic for both backends

gtest/Cargo.toml
```diff
@@ -17,3 +19,4 @@ colored = "2.0.0"
 env_logger = "0.9.0"
 path-clean = "0.1.0"
 wasm-instrument = { version = "0.1" }
+wasmtime = { version = "0.35.1", default-features = false}
```
breathx, line 22: we should use similar package as inside wasmtime backend (I'm about features here, versions are the the same).

gtest/src/empty_ext.rs
```diff
@@ -0,0 +1,179 @@
+use codec::Encode;
```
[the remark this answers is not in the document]
breathx: @shamilsan previously, I was talking about using the same Ext as in the real production. The only difference is the builder itself. We will use the same ext, but place forbidden call while adding sys call on backend in builder for all sys calls, except `gr_size` and `gr_msg`, `gr_log` for calling `meta_state`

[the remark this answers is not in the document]
breathx: So this module and struct is useless for me

gtest/src/lib.rs
```diff
@@ -1,8 +1,12 @@
+extern crate core;
+
```
breathx, line 2:
```suggestion
```
Isn't that extern useless due to the fact that the core is always appear?

core/src/message/context.rs
```diff
@@ -61,7 +61,7 @@ impl Default for ContextSettings {
 /// Context outcome.
 ///
 /// Contains all sendings and wakes that should be done after execution.
-#[derive(Default, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Decode, Encode, TypeInfo)]
```
breathx, line 64: This `Clone` impl and one for `MessageContext` seems excess for me, because these structs contains too much data for cloning. I want to force further development to work with it by ref by removing ability to clone

gtest/src/manager.rs
```diff
@@ -446,6 +450,61 @@ impl ExtManager {
 
         core_processor::handle_journal(journal, self);
     }
+
+    #[allow(dead_code)]
```
breathx, line 454: I prefer allowing `unused` for all such places, due to logic use cases. Some minimal docs would be also good.

gtest/src/manager.rs
breathx: Please, resolve conversation only once commit pushed, if agreed with reviewer.

gtest/src/manager.rs
```diff
@@ -446,6 +450,61 @@ impl ExtManager {
...
+    pub(crate) fn execute_custom_function(
```
breathx, line 455: Yeah, I see that you created universal functionality for calling any custom function, BUT:
for real production we will apply such changes directly in main executors

so for gtest we should provide only `call_meta(name)`
where name is fn with no-void return type

also executing "name" should never affect storage: there is no need to handle journal after this.

breathx (changes requested):
Yeah, that's probably strange, but I want you to a bit cut the functionality you added, apply restrictions for return type for meta state functions and reply copy-paste with using real ext and newly created forbidden sys calls functionality.

core-backend/wasmtime/src/funcs_tree.rs
```diff
@@ -0,0 +1,75 @@
...
+    funcs.insert("gr_exit", FuncsHandler::exit(store, memory));
+    funcs.insert("gr_origin", FuncsHandler::origin(store, memory));
+    funcs.insert("gr_msg_id", FuncsHandler::msg_id(store, memory));
+    funcs.insert("gr_program_id", FuncsHandler::program_id(store, memory));
+    funcs.insert("gr_read", FuncsHandler::read(store, memory));
+    funcs.insert("gr_reply", FuncsHandler::reply(store, memory));
+    funcs.insert("gr_reply_wgas", FuncsHandler::reply_wgas(store, memory));
+    funcs.insert("gr_reply_commit", FuncsHandler::reply_commit(store, memory));
+    funcs.insert(
+        "gr_reply_commit_wgas",
```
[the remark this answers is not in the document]
dustyNdumb: To be honest I didn't really like this option because choosing "array way" means that forbidden funcs will be inserted later in some other places and that will lose of single responsibility principle 
Are you sure it is a better way?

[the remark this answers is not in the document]
dustyNdumb: oh, I get it
thanks!

gtest/src/wasm_executor.rs
```diff
@@ -0,0 +1,352 @@
...
+            0,
+            None,
+        );
+
+        system.0.borrow_mut().call_meta(
+            200.into(),
+            &program.id,
+            Some(meta_state_message),
+            meta_state_function_name,
+        );
```
breathx, line 316: this should be also available directly from `Program`. Especially with `meta_state` only function

like `program.call_meta(arg)`

gtest/src/wasm_executor.rs
```diff
@@ -0,0 +1,352 @@
...
+        let system = System::default();
+        let program = Program::from_file(
+            &system,
+            "../target/wasm32-unknown-unknown/release/demo_meta.wasm",
+        );
+
+        let meta_state_message = IncomingMessage::new(
+            system.0.borrow_mut().fetch_inc_message_nonce().into(),
+            user_id,
+            Option::<Id>::encode(&Some(Id {
```
breathx, line 218: there is `.encode` method

gtest/src/wasm_executor.rs
```diff
@@ -0,0 +1,352 @@
...
+        Self {
+            instance,
+            store,
+            memory,
+        }
+    }
+
+    /// Executes non-void function by provided name.
+    /// Panics if no function with such name was found or function was void
+    #[allow(dead_code)]
```
breathx, line 83: remove all dead codes from everywhere, please
if any related fn will be public, this won't trigger clippy
provide `Program::call_meta` into public interface

core-processor/src/ext.rs
```diff
@@ -603,3 +605,40 @@ impl EnvExt for Ext {
         &self.forbidden_funcs
     }
 }
+
```
breathx, line 608: I'm really versus any default constructors for ext, due to the importance and uniqueness of the inner components as gas allowance counter etc.

This codebase will reduce in future, once many of this params got united into some general structs

gtest/src/manager.rs
```diff
@@ -446,6 +450,49 @@ impl ExtManager {
 
         core_processor::handle_journal(journal, self);
     }
+
+    /// Call non-void meta function from an actor stored in manager.
+    /// Warning! This is a static call that doesn't change actors pages data.
+    #[allow(dead_code)]
+    pub(crate) fn call_meta(
+        &mut self,
+        source: ProgramId,
```
breathx, line 459: we don't need source here
this call is forbidden for calling meta state

breathx (changes requested):
Cool
The last things to do is to resolve given comments: especially provide public api!

core-processor/src/ext.rs
shamilsan: A good practice is to implement/derive three basic traits for public structs (I call it DDC-rule): `Default`, `Debug`, and `Clone` (see "Rust for Rustaceans" book). But if you promise to refactor this part I am ready to turn a blind eye to that 😄

gtest/src/wasm_executor.rs
```diff
@@ -0,0 +1,352 @@
...
+pub(crate) struct WasmExecutor {
+    instance: Instance,
+    store: Store<StoreData<Ext>>,
+    memory: WasmtimeMemory,
+}
+
+impl WasmExecutor {
+    /// Creates a WasmExecutor instance from a program.
+    /// Also uses provided memory pages for future execution
+    #[allow(dead_code)]
```
shamilsan, line 27: This function is called from `get_executor()`. Why is it possible to be dead?

gtest/src/wasm_executor.rs
```diff
@@ -0,0 +1,311 @@
...
+    program::Program,
+};
+use std::{collections::BTreeMap, mem};
+use wasmtime::{
+    Config, Engine, Extern, Func, Instance, Memory as WasmtimeMemory, MemoryType, Module, Store,
+    Val,
+};
+
+/// Binary meta-functions executor for testing purposes
+pub struct WasmExecutor {
```
breathx, line 18: no need this to be public

gtest/src/wasm_executor.rs
```diff
@@ -0,0 +1,311 @@
...
+    memory: WasmtimeMemory,
+}
+
+impl WasmExecutor {
+    /// Creates a WasmExecutor instance from a program.
+    /// Also uses provided memory pages for future execution
+    pub fn new(
+        program: &Program,
+        memory_pages: &BTreeMap<PageNumber, Box<PageBuf>>,
+        message: Option<IncomingMessage>,
```
breathx, line 30: this message is excess, it's just the executor with preparation

gtest/src/wasm_executor.rs
```diff
@@ -0,0 +1,311 @@
...
+        Self {
+            instance,
+            store,
+            memory,
+        }
+    }
+
+    /// Executes non-void function by provided name.
+    /// Panics if no function with such name was found or function was void
+    pub fn execute(&mut self, function_name: &str) -> Vec<u8> {
```
breathx, line 81: here we may throw payload of the "message" and form this default with given payload inside the method

gtest/src/wasm_executor.rs
```diff
@@ -0,0 +1,311 @@
...
+            .call(&mut self.store, &[], &mut prt_to_result_array)
+            .expect("Failed call");
+
+        match prt_to_result_array[0] {
+            Val::I32(ptr_to_result) => self.read_result(ptr_to_result),
+            _ => panic!("{}", "Got wrong type"),
+        }
+    }
+
+    fn build_ext(program: &Program, message: Option<IncomingMessage>) -> Ext {
```
breathx, line 95: useless message

gtest/src/program.rs
```diff
@@ -248,6 +248,12 @@ impl<'a> Program<'a> {
     pub fn id(&self) -> ProgramId {
         self.id
     }
+
+    pub fn call_meta(&self, message: Option<IncomingMessage>, function_name: &str) -> Vec<u8> {
```
breathx, line 252: public call meta should call "meta_state" function, no need to give custom variants at the moment

breathx: and instead of the message, paste payload
copy-paste and reuse method as payload: T: AsRef<[u8]> -> Vec<u8> and second one with payload:  E: Encode -> D: Decode

gtest/src/wasm_executor.rs
dustyNdumb: in general as I understood you wanted to change message argument to payload (no problem with that) and to move it from executor constructor to execution method. I am really confused about the second part. As I see the building levels in constructor: payload -> message -> messageContext -> ext -> store -> executor
Payload is injected on the first step and  couldn't be changed later so for me there is no good way to relocate message (payload) from constructor to execution method

breathx: This method is for only one use, no need to change payload in current impl

gtest/src/manager.rs
```diff
@@ -446,6 +449,36 @@ impl ExtManager {
...
+        &mut self,
+        program_id: &ProgramId,
+        payload: Payload,
+        function_name: &str,
+    ) -> Vec<u8> {
+        let mut executor = self.get_executor(program_id, payload);
+        executor.execute(function_name)
+    }
+
+    fn get_executor(&mut self, program_id: &ProgramId, payload: Payload) -> WasmExecutor {
```
breathx, line 465: payload may be Option as it was before

gtest/src/manager.rs
```diff
@@ -446,6 +449,36 @@ impl ExtManager {
 
         core_processor::handle_journal(journal, self);
     }
+
+    /// Call non-void meta function from an actor stored in manager.
+    /// Warning! This is a static call that doesn't change actors pages data.
+    pub(crate) fn call_meta(
+        &mut self,
+        program_id: &ProgramId,
+        payload: Payload,
```
breathx, line 458: here option also

gtest/src/program.rs
```diff
@@ -248,6 +248,16 @@ impl<'a> Program<'a> {
     pub fn id(&self) -> ProgramId {
         self.id
     }
+
+    pub fn meta_state(&self, payload: impl Encode) -> Vec<u8> {
```
breathx, line 252:
```suggestion
    pub fn meta_state<E: Encode, D: Decode>(&self, payload: E) -> D {
```

gtest/src/wasm_executor.rs
```diff
@@ -0,0 +1,286 @@
...
+    memory: WasmtimeMemory,
+}
+
+impl WasmExecutor {
+    /// Creates a WasmExecutor instance from a program.
+    /// Also uses provided memory pages for future execution
+    pub(crate) fn new(
+        program: &Program,
+        memory_pages: &BTreeMap<PageNumber, Box<PageBuf>>,
+        payload: Payload,
```
breathx, line 29: opt

breathx (approved):
Finally, last comments 👍

gtest/src/manager.rs
breathx: for `Program` might be call_meta_empty() and call_meta(arg) in both variants bytes and encodable

breathx (approved):

PR #956, opened 2022-05-23
Add comparison of execution times for pallet tests and node runtime tests

### LINKED ISSUE #900: Performance benchmarks
### Problem to Solve

At the moment we don't track the performance regressions or improvements, so we miss the moment while our new feature or changes are useful or not.

### Possible Solution

Provide separate benchmark tests (with time monitoring) on the node and pallet tests.

### Notes

_No response_

gshep:
Resolves #900 

- enable output of JUnit XML for nextest in pallet tests
- add execution time measurement in gear-runtime-test-cli
- add output of similiar JUnit XML in gear-runtime-test-cli
- add utility to compare execution times for provided junit.xml
- adjust CI to collect data regularly
- add script to collect measurements locally and compare results
- add bot commenting PR with compare results

@gear-tech/dev


utils/regression-analysis/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+    let mut statistics: BTreeMap<String, BTreeMap<String, Vec<u64>>> = BTreeMap::default();
+    for entry in fs::read_dir(data_folder_path).unwrap() {
+        let executions = build_tree(disable_filter, &entry.unwrap().path());
+        executions.iter().for_each(|(key, times)| {
+            if !statistics.contains_key(key) {
+                statistics.insert(key.clone(), Default::default());
+            }
+
+            let previous_times = statistics.get_mut(key).unwrap();
+            times.iter().for_each(|(key, &time)| {
```
shamilsan, line 127: Also:
```suggestion
            for (key, time) in times {
```

utils/regression-analysis/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+                let time = (1_000_000_000.0 * time) as u64;
+
+                if let Some(time_vec) = previous_times.get_mut(key) {
+                    let i = match time_vec.binary_search(&time) {
```
shamilsan, line 131: [Result::into_ok_or_err()](https://doc.rust-lang.org/std/result/enum.Result.html#method.into_ok_or_err) is unstable unfortunately 😄

shamilsan (approved):
Awesome work! 🔥

utils/regression-analysis/src/main.rs
gshep: yep )

.github/workflows/CI.yaml
```diff
@@ -169,6 +169,19 @@ jobs:
         env:
           RUSTFLAGS: -Cinstrument-coverage
 
+      - name: "Test (regressions): pallets"
```
breathx, line 172:
```suggestion
      - name: "Test (performance): pallets"
```
Why only regressions? 😢 
Here and in other places.

.github/workflows/CI.yaml
```diff
@@ -169,6 +169,19 @@ jobs:
...
+        if: ${{ github.ref != 'refs/heads/master' }}
+        id: test-regressions-pallets
+        continue-on-error: true
```
breathx, line 175: why do we need to continue on errors? Better to save time

scripts/performance-regressions.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env sh
```
breathx, line 1: I prefer to use absolute path in scripts `$ROOT_DIR/target` ...

scripts/src/test.sh
```diff
@@ -27,7 +27,7 @@ EOF
 }
 
 workspace_test() {
-  cargo nextest run --workspace "$@"
+  cargo nextest run --workspace "$@" --profile ci
```
breathx, line 30: What does this profile mean?

scripts/src/test.sh
```diff
@@ -58,18 +58,20 @@ gtest() {
...
 # $1 - ROOT DIR
-# $2 - yamls list (optional)
+# $2 - TARGET DIR
+# $3 - yamls list (optional)
 rtest() {
   ROOT_DIR="$1"
+  TARGET_DIR="$2"
 
-  YAMLS=$(parse_yamls_list "$2")
+  YAMLS=$(parse_yamls_list "$3")
```
breathx, line 67: That's already solved in more correct way, merge master, please

utils/gear-runtime-test-cli/src/command.rs
```diff
@@ -69,20 +72,63 @@ impl GearRuntimeTestCmd {
...
-        });
+        let (executions, times) = tests
+            .par_iter()
+            .map(|test| {
+                let (fixtures, times) = test
+                    .fixtures
+                    .par_iter()
+                    .map(|fixture| {
+                        new_test_ext().execute_with(|| {
+                            gear_common::reset_storage();
```
breathx, line 83: btw, why do we need storage reset when we run with new test ext?

utils/regression-analysis/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "regression-analysis"
+version = "0.1.0"
+authors = ["Gear Technologies"]
+edition = "2018"
+license = "GPL-3.0"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
breathx, line 8:
```suggestion
```

utils/regression-analysis/src/junit_tree.rs
```diff
@@ -0,0 +1,47 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use std::{collections::BTreeMap, str::FromStr};
+
```
breathx, line 20: keep uses together, please

utils/regression-analysis/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+
+use std::{
+    collections::BTreeMap,
+    fs,
+    path::{Path, PathBuf},
+    str::FromStr,
+};
+
+use clap::{Parser, Subcommand};
+
```
breathx, line 27: same about uses

utils/regression-common/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "regression-common"
+version = "0.1.0"
+authors = ["Gear Technologies"]
+edition = "2018"
+license = "GPL-3.0"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
breathx, line 8:
```suggestion
```

scripts/gear.sh
```diff
@@ -308,7 +308,7 @@ case "$COMMAND" in
 
       rtest)
         header "Running node runtime testsuite"
-        rtest "$ROOT_DIR" "$@"; ;;
+        rtest "$ROOT_DIR" "$TARGET_DIR" "$@"; ;;
```
breathx, line 311: will be outdated once rebased master

utils/regression-analysis/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+        if disable_filter {
+            return true;
+        }
+
+        PALLET_NAMES.iter().any(|&name| name == pallet_name)
+    };
+
+    let junit_xml = std::fs::read_to_string(path).unwrap();
+    let test_suites: TestSuites = from_str(&junit_xml).unwrap();
+    let total_time: BTreeMap<_, _> = [(
```
breathx, line 90:
```suggestion
    let total_time = [(
```

Isn't it may be skipped?

utils/regression-analysis/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+        .iter()
+        .filter_map(|(key, tests)| {
+            statistics.get(key).map(|test_times| {
+                let test_stats = tests
+                    .iter()
+                    .filter_map(|(key, &time)| {
+                        test_times.get(key).map(|times| {
+                            let len = times.len();
+                            let len_remainder = len % 2;
+                            let quartile_lower = median(&times[..len / 2]);
```
breathx, line 165: oh, I didn't find code where `times` sorted could you provide that `.sort()` explicitly here, or that median finding is invalid.

I think it might be better to also add comparison with the average, not only with median

breathx (changes requested):
Magnificent work! But I am really worried about sorting arrays for calculations, and the average value. Provide it and that would be perfect!

scripts/src/test.sh
gshep: It is already merged. `TARGET_DIR` is required there to put resulting `junit.xml` so arguments count increased

scripts/gear.sh
gshep: I'm afraid no since `TARGET_DIR` is required. Explained to `test.sh` above

utils/regression-analysis/src/main.rs
gshep: Good catch! The array is sorted before serialization but yeah, you're right that after deserialization order may be wrong.

However I disagree to compute an average: the characteristic is not robust regarding to outliers.

utils/regression-analysis/src/main.rs
gshep: yep

utils/gear-runtime-test-cli/src/command.rs
gshep: If I'm not mistaken to reset old data of the previous executions since it is written with `spi_io` yet

scripts/src/test.sh
gshep: This is just profile for nextest configured in `.config/nextest.toml`. Since we're gonna run it mostly on CI it's named so.

.github/workflows/CI.yaml
gshep: ok

scripts/src/test.sh
breathx: > 

But we run it manually also. With each `make pre-commit` call etc

utils/gear-runtime-test-cli/src/command.rs
breathx: Strange thing, due to the fact of clear in-memory env in new test ext.
Ok, just run into this and decided to ask. Not related to this PR

utils/regression-analysis/src/main.rs
breathx: > However I disagree to compute an average: the characteristic is not robust regarding to outliers.

Yeah, but one more dimension wouldn't be excess, especially for such careful and lightweight task.
Median doesn't show us all situation also, but makes us closer to truth (e.g. [2,3,3,3,6,7] median eqs 3, while average eqs 4, and for this case average seems more realistic)

breathx (approved):
I'll left average comparison adding up to you.
Great 👍

PR #968, opened 2022-05-25
fix wasmtime backend

grishasobol:
fix wasmtime backend:
Later we unprotect pages after memory deallocation (memory move causes memory deallocation and we cannot touch pages after that)

Add dependency from memory in lazy-pages to avoid this types of bugs in future

### DIFF
--- a/core-backend/wasmtime/src/env.rs
+++ b/core-backend/wasmtime/src/env.rs
@@ -27,7 +27,7 @@ use alloc::{
     vec::Vec,
 };
 use gear_backend_common::{
-    BackendError, BackendReport, Environment, ExtInfo, HostPointer, IntoExtInfo, TerminationReason,
+    BackendError, BackendReport, Environment, HostPointer, IntoExtInfo, TerminationReason,
 };
 use gear_core::{
     env::{ClonedExtCarrier, Ext, ExtCarrier},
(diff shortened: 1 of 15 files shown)


breathx:
First of all, how does `wasmtime` affect `wasmer` issue, or it's just wrong PR name?
Secondly, could you explain what's the difference of using method or copy-pasting code? Some tests would be great too

gshep:
> First of all, how does `wasmtime` affect `wasmer` issue, or it's just wrong PR name? Secondly, could you explain what's the difference of using method or copy-pasting code? Some tests would be great too

If I understand correctly then the issue is that when `self` is moved into the method and destructured all items after that are deallocated (for example `store` & `memory`).

I would suggest to return these items as the method result so their lifetime could be extended to avoid copypasting:
```rust
fn prepare_post_execution_data(
        self,
    ) -> Result<(ExtInfo, HostPointer, StoreType, MemroyType), BackendError<WasmtimeEnvironmentError>> {
         /* ... */
            .map(|info| (info, wasm_memory_addr, store, memory))
    }
```

and then:
```rust
let (ext, wasm_addr, _store, _memory) = object.prepare_post_execution_data()?;
```

core-backend/wasmtime/src/env.rs
```diff
@@ -334,7 +324,24 @@ where
...
+            .into_inner()
+            .into_ext_info(|offset: usize, buffer: &mut [u8]| {
+                memory.read(&mut store, offset, buffer)
+            })
+            .map_err(|(reason, gas_amount)| BackendError {
+                reason: WasmtimeEnvironmentError::MemoryAccess(reason),
+                description: None,
+                gas_amount,
+            })
+            .map(|info| (info, wasm_memory_addr))?;
```
gshep, line 344: I don't know what the way you're gonna go however the map in this case is unnecessary.

grishasobol:
> First of all, how does `wasmtime` affect `wasmer` issue, or it's just wrong PR name? Secondly, could you explain what's the difference of using method or copy-pasting code? Some tests would be great too

Yes it's wrong name: wasmtime must be

grishasobol:
> First of all, how does `wasmtime` affect `wasmer` issue, or it's just wrong PR name? Secondly, could you explain what's the difference of using method or copy-pasting code? Some tests would be great too

We move memory and then drop memory before. Now memory live until the end of function

examples/ping-gas/src/lib.rs
```diff
@@ -1,11 +1,24 @@
...
+//     fn lol();
+// }
+
+// struct A;
+// impl T for A {
+//     fn lol() {};
+// }
+
+// environmental::environmental!(ext: trait T);
+
```
SabaunT, line 13: ?

examples/ping-gas/src/lib.rs
```diff
@@ -1,11 +1,24 @@
...
 use gstd::{debug, msg, prelude::*};
 
 #[no_mangle]
 pub unsafe extern "C" fn handle() {
     debug!("Hello from ping handle");
 
+    // let mVut a = A;
```
SabaunT, line 20: same

SabaunT (changes requested):

gshep (approved):

gshep (approved):

SabaunT (approved):
Discussed

Issue #642, opened 2022-02-09
Set minimal gas amount for mailbox messages


breathx:
### Problem to Solve

At the moment, we have no fees for being in mailbox (gas also cuts off before message inserted there), but once we got that feature, messages with 0 gas send to mailbox will not be able to be replied on and will not stay in the mailbox for future.

### Possible Solution

Set the minimal gas count for message to be inserted in mailbox. This gas will be consuming as rent for storing and message will be available for reply. Once gas ends, message removes.
Messages with gas limit less than that minimum will not be added in mailbox, but will be seen in logs.

PR #971, opened 2022-05-26
Clean redundant gas value when exit

clearloop:
## Changes

- [x] process note `MessageConsumed` before `ExitDispatched`

## References

- #975


pallets/gear/src/manager.rs
```diff
@@ -268,10 +268,16 @@ where
         Pallet::<T>::deposit_event(event);
     }
 
+    fn gas_recovered(&mut self, message_id: MessageId) {
+        let message_id = message_id.into_origin();
+        log::debug!("Recovered gas from: {:?}", message_id);
+        GasTree::<T>::remove(message_id);
+    }
```
clearloop, line 275: Not sure if there is any side effects for just removing the node

pallets/gear/src/manager.rs
```diff
@@ -268,10 +268,19 @@ where
         Pallet::<T>::deposit_event(event);
     }
 
+    fn gas_recovered(&mut self, message_id: MessageId) {
+        let message_id = message_id.into_origin();
+        if let Ok(Some(gas_left)) = GasPallet::<T>::get_limit(message_id) {
+            log::debug!("Recovered: {:?} from: {:?}", gas_left, message_id);
+            GasPallet::<T>::decrease_gas_allowance(gas_left);
+            GasTree::<T>::remove(message_id);
```
clearloop, line 276: plz review this, not sure if there is any side effects for just removing the node

SabaunT: Removing gas node from the tree must not be available. If there are unintended nodes left, it's an algorithmic error that we must correct. If I'm not mistaken, here could be improper side effects if gas less messages were sent in the program, before `exit` was called.

SabaunT (changes requested):
Wow, well done with first PR! :+1: You  figured it out quickly, that message consume note should be used.

pallets/gear/src/manager.rs
clearloop: > Generally, I think there is no need in a new journal note

`GasBurn` also have the ability to clean gas but it will run `repatriate_reserved` also which is not expected, what about adding a `message_id` in `ExitDispatch`, and clean the gas value inside `handler.exit_dispatch()` ?

pallets/gear/src/manager.rs
```diff
@@ -315,8 +315,27 @@ where
...
+        &mut self,
+        id_exited: ProgramId,
+        message_id: MessageId,
+        value_destination: ProgramId,
+    ) {
         let program_id = id_exited.into_origin();
+        log::debug!("Exit program {:?}", &program_id);
+
+        let message_id = message_id.into_origin();
+        if let Ok(Some(gas_left)) = T::GasHandler::get_limit(message_id) {
```
clearloop, line 328: if we pass `gas_amount.left()` into this function in `processor.rs` we'll not need to run this 

```
T::GasHandler::get_limit(message_id)
```

again

core-processor/src/processor.rs
```diff
@@ -226,6 +226,7 @@ fn process_success(
 
     match kind {
         Exit(value_destination) => {
+            journal.push(JournalNote::MessageConsumed(message_id));
```
breathx, line 229: I would refactor this matching like this:
```rust
if let Wait = kind {
    journal.push(JournalNote::WaitDispatch(
        dispatch.into_stored(program_id, context_store),
    ));
} else {
    if let Exit(value_destination) = kind {
        journal.push(JournalNote::ExitDispatch {
            id_exited: program_id,
            value_destination,
        });
    }

    let outcome = match dispatch.kind() {
        DispatchKind::Init => DispatchOutcome::InitSuccess {
            message_id,
            origin,
            program_id,
        },
        _ => DispatchOutcome::Success(message_id),
    };

    journal.push(JournalNote::MessageDispatched(outcome));
    journal.push(JournalNote::MessageConsumed(message_id));
}
```

Cause MessageDispatched should also be deposited.
Or split it by adding one more DispatchOutcome to show that the message was exited, so program is no more available.

pallets/gear/src/tests.rs
```diff
@@ -2422,6 +2422,64 @@ fn exit_handle() {
...
+            WASM_BINARY.to_vec(),
+            DEFAULT_SALT.to_vec(),
+            EMPTY_PAYLOAD.to_vec(),
+            10_000_000,
+            0,
+        ));
+
+        run_to_block(2, None);
+
+        let gas_limit = 50_000_000_000;
```
breathx, line 2443: From where did you got this limit?

breathx: I don't like explicitly specified constants for limits, you know why

core-processor/src/processor.rs
breathx: Yeah, the second variant with propagating new event seems better!

breathx (changes requested):
Resolve the comment about `MessageDispatched` note and it would be great fix!
Well done 👍

core-processor/src/processor.rs
clearloop: > Yeah, the second variant with propagating new event seems better!

make sense, btw, should we `remove_waiting_message` in `exit_dispatch` also? 

https://github.com/gear-tech/gear/blob/3981f45830a519b68fce27be8767823f2e3497d5/pallets/gear/src/manager.rs#L318

exit_dispatch now only

- set_program_terminated_status
- remove_program_waitlist

pallets/gear/src/manager.rs
```diff
@@ -158,6 +158,19 @@ where
 {
     fn message_dispatched(&mut self, outcome: CoreDispatchOutcome) {
         let event = match outcome {
+            CoreDispatchOutcome::Exit {
```
clearloop, line 161: Didn't add any context for this because we already have `exit_dispatch`

SabaunT (approved):
Looks good, looking forward to merging it.

breathx (changes requested):
https://github.com/gear-tech/gear/pull/971#discussion_r884657632

Agree here. Need to remove waiting init ids and their values in waitlist as well and provide test for that where exit called in init fn. This may be done in separate PR, but I don't see any trouble to append it in this one

breathx:
Despite my feature-request, great work!

clearloop:
> [#971 (comment)](https://github.com/gear-tech/gear/pull/971#discussion_r884657632)
> 
> Agree here. Need to remove waiting init ids and their values in waitlist as well and provide test for that where exit called in init fn. This may be done in separate PR, but I don't see any trouble to append it in this one

thx! this is really helpful for me to understand https://github.com/gear-tech/gear/issues/642

examples/binaries/init-wait/src/code.rs
```diff
@@ -32,6 +32,10 @@ pub unsafe extern "C" fn init() {
             msg::send(msg::source(), b"PING", 0).unwrap();
             STATE = State::WaitForReply;
             exec::wait();
+
+            if msg::load_bytes() == b"EXIT".to_vec() {
+                exec::exit(msg::source())
+            }
```
clearloop, line 38: reused this program

breathx: уeah, no problem reusing it

pallets/gear/src/manager.rs
```diff
@@ -325,6 +338,7 @@ where
                 .unwrap_or_else(|e| unreachable!("Message queue corrupted! {:?}", e));
         }
 
+        let _ = common::waiting_init_take_messages(id_exited);
```
breathx, line 341: That won't work.
`waiting_init_take_messages` goes directly to sp_io to already removed storage. Messages from waitlist removes two lines above, but there is still storage for ProgramId and Vec<MessageId> mapping "waiting_init" directly in sp_io.
we need just to clear that prefix - provide such method in common
like that

`# gear-common/src/lib.rs`
```rust
pub fn clear_waiting_init_ids(prog_id: ProgramId) {
    let key = waiting_init_prefix(prog_id);
    sp_io::storage::clear(&key);
}
```

examples/binaries/init-wait/src/code.rs
breathx: but this code is unreachable :(
once we ran into wait wout our async loop it won't be executed further
and after line 33 we change the "state" so we will not have ability to reach that code

pallets/gear/src/tests.rs
```diff
@@ -2421,6 +2421,133 @@ fn exit_handle() {
...
+        );
+
+        // reserved balance after execution is zero
+        let reserved_after_execution = BalancesPallet::<Test>::reserved_balance(USER_1);
+        assert!(reserved_after_execution.is_zero());
+    })
+}
+
+#[test]
+fn init_wait_reply_exit() {
```
breathx, line 2481: Seems that test should work another way

examples/binaries/init-wait/src/code.rs
clearloop: so what about doing this in `handle_reply`?

pallets/gear/src/tests.rs
```diff
@@ -2424,6 +2424,133 @@ fn exit_handle() {
...
+            EMPTY_PAYLOAD.to_vec(),
+            Vec::new(),
+            2_000_000_000u64,
+            0u128
+        ));
+        let pid = get_last_program_id();
+
+        // block 2
+        //
+        // - send exit message to the program
```
breathx, line 2503:
```suggestion
```

breathx (approved):
Awesome 👍

breathx (approved):

SabaunT (approved):

PR #975, opened 2022-05-26
fix(gas): Remove node key instead of origin id in function consume

### LINKED ISSUE #977: Refactor the algo of getting the gas spent in `get_gas_spent`
### File Location(s)

### Proposal

clearloop:
Resolves #977

## Summary

`fn consume` removes `id` insteand of `key` by accident

## Changes

- [x] fix(gas): Remove node key instead of origin id in function consume

## References

- #971
- #977
- #981

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1432,6 +1432,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "demo-waiting-proxy"
+version = "0.1.0"
+dependencies = [
+ "gear-wasm-builder",
+ "gstd",
+]
+
+[[package]]
 name = "derive_more"
 version = "0.99.17"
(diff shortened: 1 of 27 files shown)


clearloop:
hmmm, looks like this issue related to another new error https://github.com/gear-tech/gear/runs/6611638164?check_suite_focus=true#step:16:598

breathx:
Great find!

Now need to find out the reason of tests failing

breathx:
Great find!

Now need to find out the reason of tests failing

clearloop:
The CI broken because of https://github.com/gear-tech/gear/pull/976#issuecomment-1139916801

clearloop:
#984

PR #976, opened 2022-05-26
Fix value sending in async scenarios

### LINKED ISSUE #974: Value charged multiple times in a cascade of async calls
If we have a cascade of contracts calls:
```
User                  Contract A                                            Contract B
  | -- msg_with_value --> |                                                     |
                          |---- msg::send_bytes_and_wait_for_reply(payload) --->|
                                                                                |
                          |<------------------ reply(output) -------------------|
                          |
  |<------- reply --------|
```
then the value passed to the `Contract A` will be attempted to be transferred every time the contract gets to execute a message (on the first entry, upon being woken up from the wait list etc.) - a journal note `SendValue` is generated and processed by the journal handler.
As a result, if there is enough value reserved for the original message (to pay for gas, not value) then this will go unnoticed except that the user will be overcharged. In the worst case we can fall through into an `unreachable!` branch and panic.

To reproduce:
1. Pull a custom branch https://github.com/gear-tech/gear/blob/ek-expose-bug/pallets/gear/src/tests.rs#L3149.
2. Run the test `cascading_messages_with_value_do_not_overcharge`.

breathx:
Resolves #974.

See `core-processor/src/processor.rs`

Thanks to @ekovalev for bug revealing and providing test for it.

- Bug fixed.
- Docs provided.
- Test updated (previous values asserted with inversion).

@gear-tech/dev

### DIFF
--- a/pallets/gear/src/tests.rs
+++ b/pallets/gear/src/tests.rs
@@ -31,6 +31,7 @@ use demo_compose::WASM_BINARY as COMPOSE_WASM_BINARY;
 use demo_distributor::{Request, WASM_BINARY};
 use demo_mul_by_const::WASM_BINARY as MUL_CONST_WASM_BINARY;
 use demo_program_factory::{CreateProgram, WASM_BINARY as PROGRAM_FACTORY_WASM_BINARY};
+use demo_waiting_proxy::WASM_BINARY as WAITING_PROXY_WASM_BINARY;
 use frame_support::{assert_noop, assert_ok};
 use frame_system::Pallet as SystemPallet;
 use gear_core::{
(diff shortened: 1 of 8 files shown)


pallets/gear/src/tests.rs
```diff
@@ -3144,6 +3145,77 @@ fn test_reply_to_terminated_program() {
...
+        // The sender's account has enough funds for both gas and `value`, therefore
+        // expecting the message to be processed successfully.
+        // Expected outcome: the sender's balance has decreased by the (gas + `value`),
+        // that is by 2,302,360,260.
+        //
+        // TODO: instead, after the proxy message has re-entered having been woken up
+        // it attempts to do value transfer again.
+        // The `repatriate_reserved()` in this case still succeeds even though the actual
+        // amount repatriated is less than requested 10_000_000 => must thrown here!!!
+        // However, the sender's balance ends up overcharged.
```
SabaunT, line 3188: seems that behavior is fixed? Maybe should state that and remove **TODO**

pallets/gear/src/tests.rs
```diff
@@ -3144,6 +3145,77 @@ fn test_reply_to_terminated_program() {
...
+
+        assert_ok!(Gear::send_message(
+            Origin::signed(USER_1),
+            wrapper_id,
+            100_u64.to_le_bytes().to_vec(),
+            gas_reserved,
+            value,
+        ));
+
+        run_to_block(3, None);
```
SabaunT, line 3203: shouldn't balance checks be performed before and after that call?

SabaunT (approved):
nice catch @ekovalev!
nice job @breathx

clearloop:
Anybody run `cargo test` under `pallets/gear` after this commit? got `cascading_messages_with_value_do_not_overcharge` failed on my machine:

```
failures:

---- tests::cascading_messages_with_value_do_not_overcharge stdout ----
thread 'tests::cascading_messages_with_value_do_not_overcharge' panicked at 'assertion failed: `(left == right)`
  left: `497354549734`,
 right: `497336081914`', pallets/gear/src/tests.rs:3205:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Issue #915, opened 2022-05-05
Selective yamls run


breathx:
### Problem

Makefile doesn't handle "yamls" for spec testing enough. It exactly build only needed demos, but runs all yamls instead of specified ones.

### Steps

`make gtest yamls="gear-test/spec/test_ping.yaml"`

### Possible Solution

Recheck `gtest_debug()` function in `scripts/src/test.sh`

PR #967, opened 2022-05-24
Introduce `Waitlist` type for `Messenger`

breathx:
Major changes:
- Waitlist introduced for `Messenger` trait.
- New types introduced.
- Docs provided.
- `IterableDoubleMap` renamed to `IterableByKeyMap`.

Minor changes:
- Replaced H256 with proper types, if available.
- Usage pallet kept this previous logic for now #969.
- Added script for build gear-test crate.

PR #980, opened 2022-05-27
fix problems in Makefile and scripts

grishasobol:
fix deps in Makefile
add ability to run special yamls in rtest

fix issue #915

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,11 +18,8 @@ clean:
 .PHONY: clean-examples
 clean-examples:
 	@ rm -rf ./target/wasm32-unknown-unknown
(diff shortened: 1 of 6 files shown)


breathx:
Does it fix #915 ?

breathx (approved):
Yeah, good job!
But I am confused should we rebuild gear-test all of the time we need to run spec testing. Using already build binary decreases `make pre-commit` time. I don't like also that you didn't provide documentation for ./gear.sh build gear-test.
Same thing was added in #967, could we merge that one, merging into your PR afterwards?

PR #981, opened 2022-05-27
refactor(gear): estimate gas only with notes that have spent gas

### LINKED ISSUE #977: Refactor the algo of getting the gas spent in `get_gas_spent`

clearloop:
Resolves #977 

- [x] estimate gas only with notes that have spent gas
  - [x] removed `MessageConsumed` because it doesn't spend any gas on current node
  - [x] added `WaitDispatch` because it spends gas also

## References

- #975 
- #971

### DIFF
--- a/pallets/gear/src/lib.rs
+++ b/pallets/gear/src/lib.rs
@@ -562,7 +562,29 @@ pub mod pallet {
                     )
                 };
 
-                core_processor::handle_journal(journal.clone(), &mut ext_manager);
+                // TODO: Check whether we charge gas fee for submitting code after #646
+                let notes = journal
+                    .into_iter()
+                    .filter_map(|note| -> Option<Result<JournalNote, _>> {
+                        match note {
+                            JournalNote::GasBurned { .. }
+                            | JournalNote::WakeMessage { .. }
+                            | JournalNote::SendDispatch { .. }
+                            | JournalNote::WaitDispatch(..) => Some(Ok(note)),
+                            JournalNote::MessageDispatched(CoreDispatchOutcome::MessageTrap {
+                                trap,
+                                ..
+                            }) => Some(Err(format!(
+                                "Program terminated with a trap: {}",
+                                trap.unwrap_or_else(|| "No reason".to_string())
+                            )
+                            .into_bytes())),
+                            _ => None,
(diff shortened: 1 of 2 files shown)


pallets/gear/src/lib.rs
```diff
@@ -562,7 +562,29 @@ pub mod pallet {
...
-                core_processor::handle_journal(journal.clone(), &mut ext_manager);
+                // TODO: Check whether we charge gas fee for submitting code after #646
+                let notes = journal
+                    .into_iter()
+                    .filter_map(|note| -> Option<Result<JournalNote, _>> {
+                        match note {
+                            JournalNote::GasBurned { .. }
+                            | JournalNote::WakeMessage { .. }
+                            | JournalNote::SendDispatch { .. }
+                            | JournalNote::WaitDispatch(..) => Some(Ok(note)),
```
clearloop, line 573: please review these, removed `MessageConsumed` because it doesn't spend any gas on current node

ekovalev:
I guess this also resolves #982 that I had created before I noticed this PR.

SabaunT:
Personally I suggest moving works here to #975, because these changes are closely related and the refactoring is due to found bug in the latter PR. Also it would be easier to review it.

pallets/gear/src/lib.rs
```diff
@@ -562,7 +562,29 @@ pub mod pallet {
                     )
                 };
 
-                core_processor::handle_journal(journal.clone(), &mut ext_manager);
+                // TODO: Check whether we charge gas fee for submitting code after #646
+                let notes = journal
+                    .into_iter()
+                    .filter_map(|note| -> Option<Result<JournalNote, _>> {
+                        match note {
+                            JournalNote::GasBurned { .. }
```
SabaunT, line 570: Seems because of that note you will have to refactor `get_gas_spent` another time in #975, because later value of  `remaining_gas` variable will be None (in case we had a simple execution, when program doesn't send any message).

clearloop:
> Personally I suggest moving works here to #975, because these changes are closely related and the refactoring is due to found bug in the latter PR. Also it would be easier to review it.

agree, I can just merge this branch into #975

PR #983, opened 2022-05-28
Make `gr_create_program_wgas` fallible

### LINKED ISSUE #929: Make `create_program` sys-call fallible
### Problem to Solve

-

### Possible Solution

-

### Notes

_No response_

ark0f:
Resolves #929


SabaunT:
Looking forward to merging #937 before merging this one

pallets/gear/src/tests.rs
```diff
@@ -2889,7 +2889,6 @@ fn test_two_contracts_composition_works() {
 //
 // Note: on manager level message will not be included to the [queue](https://github.com/gear-tech/gear/blob/master/pallets/gear/src/manager.rs#L351-L364)
 // But it's is not preferable to enter that `if` clause.
-// todo # 929 After create_program sys-call becomes fallible, tests must be changed
 #[test]
 fn test_create_program_with_value_lt_ed() {
```
SabaunT, line 2893: I think `test_create_program_with_value_lt_ed` and `test_create_program_with_exceeding_value` should result in different storage mutations after adding the fallible feature. For example, **no** mails should be sent to user after unsuccessful calls to `create_program_with_gas`, but these tests base on that behaviour

SabaunT (changes requested):
Looking forward to merging https://github.com/gear-tech/gear/pull/937 before merging this one.
Everything is nice!

pallets/gear/src/tests.rs
ark0f: So what do you want me to do?

SabaunT: Would be nice if you test the fallible feature on one of tests I stated in the last comment. That requires: handling `Result` of the call in the program (not `unwrap`ing), checking whether mail with error code wasn't sent to the original message sender (so must have logically opposite to current mail existence check). Was it clear?

examples/binaries/init-with-value/src/lib.rs
```diff
@@ -60,20 +90,41 @@ mod wasm {
...
+                            if let Some(expected_error) = expected_error {
+                                assert_eq!(
+                                    err,
+                                    ContractError::Ext(ExtError::Message(
+                                        expected_error.into_message_error()
+                                    ))
+                                )
+                            }
+                        }
+                    }
```
SabaunT, line 121: Although idea is quite original, but I suggest the next thing. that seems a bit easier: 
1. on error send a reply to user with an actual `ContractError::Ext(...)` (or could be just `MessageError`, or it's string format) error encoded as a payload.
2. check in the pallet test whether the message with expected error payload exists.

One more reason for such changes is to avoid dirty conversions from `ExpectedError` to `MessageError` using hard-coded values, which can cause problems in the future

SabaunT (approved):
Nice job, approving that! :+1: But please resolve first minor things stated in the comments

examples/binaries/init-with-value/src/lib.rs
ark0f: What do you think about current realisation? Error is created on pallet side and checked on contract's one

gcore/src/prog.rs
```diff
@@ -107,7 +110,7 @@ pub fn create_program_with_gas(
     payload: &[u8],
     gas_limit: u64,
     value: u128,
-) -> ActorId {
+) -> Result<ActorId, ExtError> {
```
breathx, line 113: I would provide custom `Result<T>` for `gcore` also, replacing other explicit Result<T, ExtError> afterwards

breathx (approved):
Merging actual master will break your test correctives a bit

Well done

PR #984, opened 2022-05-28
Fix get gas spent and update cascade test

### LINKED ISSUE #977: Refactor the algo of getting the gas spent in `get_gas_spent`

breathx:
- Checkouted from #975 so fixes #977 and contains refactored and right calculation in `get_gas_spent`.
- Provided similar to `get_gas_spent` function `get_gas_burned` which adds up only gas which will never returned.
- Cascade test now depend on this functions return values, not on constants.

@gear-tech/dev

### DIFF
--- a/examples/binaries/mul-by-const/src/lib.rs
+++ b/examples/binaries/mul-by-const/src/lib.rs
@@ -18,7 +18,7 @@
 
 // This contract recursively composes itself with another contract (the other contract
 // being applied to the input data first): `c(f) = (c(f) . f) x`.
-// Every call to the auto_composer contract incremets the internal `ITER` counter.
+// Every call to the auto_composer contract increments the internal `ITER` counter.
 // As soon as the counter reaches the `MAX_ITER`, the recursion stops.
 // Effectively, this procedure executes a composition of `MAX_ITER` contracts `f`
 // where the output of the previous call is fed to the input of the next call.
--- a/pallets/gear/src/lib.rs
+++ b/pallets/gear/src/lib.rs
@@ -571,10 +571,8 @@ pub mod pallet {
                             b"Internal error: unable to get gas limit after execution".to_vec()
                         })?
                     {
-                        let gas_spent = initial_gas.saturating_sub(remaining_gas);
-                        if gas_spent > max_gas_spent {
-                            max_gas_spent = gas_spent;
-                        }
+                        max_gas_spent =
+                            max_gas_spent.max(initial_gas.saturating_sub(remaining_gas));
                     };
 
                     if let JournalNote::MessageDispatched(CoreDispatchOutcome::MessageTrap {
--- a/pallets/gear/src/lib.rs
+++ b/pallets/gear/src/lib.rs
@@ -562,215 +562,36 @@ pub mod pallet {
                     )
                 };
 
-                 // TODO: Check whether we charge gas fee for submitting code after #646
-                 for note in journal.into_iter() {
+                // TODO: Check whether we charge gas fee for submitting code after #646
+                for note in journal.into_iter() {
                     core_processor::handle_journal(vec![note.clone()], &mut ext_manager);
 
-                    if let Some(remaining_gas) = T::GasHandler::get_limit(root_message_id)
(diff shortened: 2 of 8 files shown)


examples/binaries/mul-by-const/src/lib.rs
```diff
@@ -18,7 +18,7 @@
 
 // This contract recursively composes itself with another contract (the other contract
 // being applied to the input data first): `c(f) = (c(f) . f) x`.
-// Every call to the auto_composer contract incremets the internal `ITER` counter.
+// Every call to the auto_composer contract increments the internal `ITER` counter.
```
ekovalev, line 21: Looks like a description of some other contract (`demo-compose` most likely). Anyway, I will revisit it later and put a proper description

pallets/gear/src/lib.rs
```diff
@@ -562,39 +569,30 @@ pub mod pallet {
                     )
                 };
 
-                core_processor::handle_journal(journal.clone(), &mut ext_manager);
+                // TODO: Check whether we charge gas fee for submitting code after #646
+                for note in journal.into_iter() {
+                    core_processor::handle_journal(vec![note.clone()], &mut ext_manager);
```
clearloop, line 574: why we still need to handle every journal on this? looks like only
- GasBurned
- WakeMessage
- SendDispatch
- WaitDispatch

spend gas

examples/binaries/mul-by-const/src/lib.rs
breathx: Yeah, seems like this :S
I don't actually know how it appeared, but yeah, better to resolve in separate PR

pallets/gear/src/lib.rs
breathx: We should handle all journal to apply changes and provide messages.

Lets go through the `get_gas_spent` idea:
It's an RPC call to get the amount of gas needed to success execution.
It works like this:
- We set for the initial message maximum of available gas limit: `T::BlockGasLimit::get()`
- We clean current message queue (It has no effect for storage in RPC-call processing due to overlay, but applies changes on storage inside test, be careful with it)
- We insert initial message into the queue
- We run our queue processing till the moment where there is no messages left
- On each iteration we parse ALL created journal (its notes) to apply proper storage changes, emulating real execution
- To don't be chained with concrete notes (avoiding future collisions, cause we may forget about `get_gas_spent`), we go through all of them in custom parsing - after each note processed we query current gas for initial message to find maximal diff from the initial one - this would be the resulting value we should return to user.

`get_gas_burned` - works the same, but we don't calculate minimal amount of gas needed for success execution (that maximal diff) - we just add up all gas burned to get the diff of the user balance after all executions, cause only this gas won't be returned on consume of the user message.

This stuff needs optimisation and code beautifying, but atm this is solution for issues we ran into.

pallets/gear/src/lib.rs
```diff
@@ -562,39 +569,30 @@ pub mod pallet {
                     )
                 };
 
-                core_processor::handle_journal(journal.clone(), &mut ext_manager);
+                // TODO: Check whether we charge gas fee for submitting code after #646
+                for note in journal.into_iter() {
```
shamilsan, line 573: No need to explicitly call `.into_iter()`.
```suggestion
                for note in journal {
```

pallets/gear/src/mock.rs
```diff
@@ -315,3 +322,182 @@ pub fn calc_handle_gas_spent(source: H256, dest: H256, payload: Vec<u8>) -> (u64
...
+                ProgramId::from_origin(source),
+                actor_id,
+                u64::MAX,
+                T::OutgoingLimit::get(),
+                schedule.host_fn_weights.clone().into_core(),
+            )
+        };
+
+        // TODO: Check whether we charge gas fee for submitting code after #646
+        for note in journal.iter() {
```
shamilsan, line 481: I'd prefer to not call `.iter()` explicitly.
```suggestion
        for note in &journal {
```

shamilsan (approved):
Looks great to me 👍

PR #997, opened 2022-06-01
add checks for wasm memory buffer begin addr

grishasobol:
1) Solve problem, that if memory size == 0 then wasm executors  can have garbage in pointer to memory buffer.
2) Some refactoring in lazy pages errors

### DIFF
--- a/common/src/lazy_pages.rs
+++ b/common/src/lazy_pages.rs
@@ -23,7 +23,7 @@ use gear_core::{
     ids::ProgramId,
     memory::{HostPointer, Memory, PageBuf, PageNumber},
 };
-use gear_runtime_interface::{gear_ri, GetReleasedPageError, MprotectError};
+use gear_runtime_interface::{gear_ri, RIError};
 use sp_std::{
     collections::{btree_map::BTreeMap, btree_set::BTreeSet},
     vec::Vec,
@@ -31,40 +31,39 @@ use sp_std::{
 
 #[derive(Debug, Clone, PartialEq, Eq, derive_more::Display)]
 pub enum Error {
-    #[display(fmt = "{}", _0)]
-    Mprotect(MprotectError),
-    #[display(fmt = "{}", _0)]
-    GetReleasedPage(GetReleasedPageError),
-    #[display(fmt = "Cannot convert vec to page data")]
-    VecToPageData,
-    #[display(fmt = "Cannot find page data in storage")]
-    NoPageDataInStorage,
-    #[display(fmt = "Released page {:?} has initial data", _0)]
-    ReleasedPageHasData(PageNumber),
+    #[display(fmt = "RUNTIME INTERFACE ERROR: {}", _0)]
+    RIError(RIError),
+    #[display(fmt = "RUNTIME INTERFACE ERROR: {:?} has no released data", _0)]
+    ReleasedPageHasNoData(PageNumber),
+    #[display(fmt = "RUNTIME ERROR: released page {:?} has initial data", _0)]
+    ReleasedPageHasInitialData(PageNumber),
+    #[display(fmt = "RUNTIME ERROR: wasm memory buffer is undefined")]
+    WasmMemBufferIsUndefined,
 }
 
-impl From<MprotectError> for Error {
-    fn from(err: MprotectError) -> Self {
-        Self::Mprotect(err)
-    }
-}
-
-impl From<GetReleasedPageError> for Error {
-    fn from(err: GetReleasedPageError) -> Self {
-        Self::GetReleasedPage(err)
+impl From<RIError> for Error {
+    fn from(err: RIError) -> Self {
+        Self::RIError(err)
     }
 }
 
 fn mprotect_lazy_pages(mem: &dyn Memory, protect: bool) -> Result<(), Error> {
-    let wasm_mem_addr = mem.get_buffer_host_addr();
-    gear_ri::mprotect_lazy_pages(wasm_mem_addr, protect).map_err(Into::into)
+    let wasm_mem_addr = match mem.get_buffer_host_addr() {
+        None => return Ok(()),
+        Some(addr) => addr,
+    };
+    gear_ri::mprotect_lazy_pages(wasm_mem_addr, protect)
+        .map_err(Into::into)
+        .map_err(|e| {
+            log::error!("{} (it's better to stop node now)", e);
+            e
+        })
 }
 
 /// Try to enable and initialize lazy pages env
 pub fn try_to_enable_lazy_pages() -> bool {
     if !gear_ri::init_lazy_pages() {
-        // TODO: lazy-pages must be disabled in validators in relay-chain,
-        // but it can be fixed in future only.
+        // TODO: lazy-pages must be disabled in validators in relay-chain.
         log::debug!("lazy-pages: disabled or unsupported");
         false
     } else {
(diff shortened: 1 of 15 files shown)


common/src/lazy_pages.rs
```diff
@@ -23,48 +23,47 @@ use gear_core::{
...
-    Mprotect(MprotectError),
-    #[display(fmt = "{}", _0)]
-    GetReleasedPage(GetReleasedPageError),
-    #[display(fmt = "Cannot convert vec to page data")]
-    VecToPageData,
-    #[display(fmt = "Cannot find page data in storage")]
-    NoPageDataInStorage,
-    #[display(fmt = "Released page {:?} has initial data", _0)]
-    ReleasedPageHasData(PageNumber),
+    #[display(fmt = "RUNTIME INTERFACE ERROR: {}", _0)]
```
ark0f, line 34: Why caps?

common/src/lazy_pages.rs
```diff
@@ -23,48 +23,47 @@ use gear_core::{
...
+    RIError(RIError),
+    #[display(fmt = "RUNTIME INTERFACE ERROR: {:?} has no released data", _0)]
```
ark0f, line 36: If this variant is runtime interface error too then maybe it's better to move the variant into the error?

common/src/lazy_pages.rs
```diff
@@ -23,48 +23,47 @@ use gear_core::{
...
+    ReleasedPageHasNoData(PageNumber),
+    #[display(fmt = "RUNTIME ERROR: released page {:?} has initial data", _0)]
```
ark0f, line 38: What is `RUNTIME ERROR:` prefix for?

common/src/lazy_pages.rs
grishasobol: Because this is not common error - this error can lead to problems in consensus or UB.

common/src/lazy_pages.rs
grishasobol: No first error contains error from RI second not

common/src/lazy_pages.rs
grishasobol: This error is also not common  - means that. we have some bugs in runtime

common/src/lazy_pages.rs
```diff
@@ -88,7 +87,21 @@ pub fn protect_pages_and_init_info(
...
 
-    gear_ri::set_wasm_mem_begin_addr(mem.get_buffer_host_addr());
+    let addr = match mem.get_buffer_host_addr() {
+        None => {
+            return if !lazy_pages.is_empty() {
+                // In this case wasm buffer cannot be undefined
+                Err(Error::WasmMemBufferIsUndefined)
+            } else {
+                Ok(())
+            };
```
SabaunT, line 97:
```suggestion
            let pages_are_empty = lazy_pages.is_empty();
            // if pages aren't empty, then buffer can't be defined
            return pages_are_empty.then(|| ()).ok_or(Error::WasmMemBufferIsUndefined);
           };
```

common/src/lazy_pages.rs
SabaunT, line 34: I'm not sure how to do that, but maybe there are ways to make the  "RUNTIME INTERFACE ERROR" sub-string common for all variants of errors? The one that comes up to my mind is to make it the next way: you separate in this module `Error` and `RuntimeErrorInner`. The latter one has all the variants of the current `Error` version. The new `Error` version has only one - `RIError(RuntimeErrorInner)`, for which the next attribute will be defined: `#[display(fmt = "RUNTIME INTERFACE ERROR: {:?}", _0)]. Maybe there are some easy ways. dunno.

common/src/lazy_pages.rs
```diff
@@ -23,48 +23,47 @@ use gear_core::{
...
-    let wasm_mem_addr = mem.get_buffer_host_addr();
-    gear_ri::mprotect_lazy_pages(wasm_mem_addr, protect).map_err(Into::into)
+    let wasm_mem_addr = match mem.get_buffer_host_addr() {
+        None => return Ok(()),
+        Some(addr) => addr,
+    };
+    gear_ri::mprotect_lazy_pages(wasm_mem_addr, protect)
+        .map_err(Into::into)
+        .map_err(|e| {
+            log::error!("{} (it's better to stop node now)", e);
```
SabaunT, line 58: just for clarification: this message means, that after getting error node can be up. Is that true? What if validator don't stop the node? Shall we stop it for validator?

SabaunT, line 58: just for clarification: this message means, that after getting error node can be up. Is that true? What if validator don't stop the node? Shall we stop it for validator?

SabaunT (changes requested):
Interim review

common/src/lazy_pages.rs
ark0f: Maybe you merge `ReleasedPagesHasInitialData` and `WasmMemBufferIsUndefined` into 1 error?

common/src/lazy_pages.rs
ark0f: But you still writes `RUNTIME INTERFACE ERROR` for both of them

common/src/lazy_pages.rs
ark0f: I agree with ISab

common/src/lazy_pages.rs
grishasobol: Yes whats the problem ?

common/src/lazy_pages.rs
grishasobol: No it's different errors

common/src/lazy_pages.rs
ark0f: It looks like they are both from RI then (semantically)

common/src/lazy_pages.rs
grishasobol: Yes we can but It can be unsafe - because we cannot guaranty that this won't happen for all validators, so in the case all nodes will fall down - and we do not wanna this

common/src/lazy_pages.rs
grishasobol: I think this suggestion is much less understandable

common/src/lazy_pages.rs
grishasobol: I don't see any reasons to do this . What the reason to make one more Error type? I've just marked errors which I want to mark that the problem is connected with using RI.

common/src/lazy_pages.rs
grishasobol: No one error is wrapper for RI error, second is error which we consider like RI error

common/src/lazy_pages.rs
SabaunT: resolved after discussion

SabaunT (approved):

core/src/memory.rs
```diff
@@ -253,7 +257,18 @@ pub trait Memory {
     fn data_size(&self) -> usize;
 
     /// Returns native addr of wasm memory buffer in wasm executor
-    fn get_buffer_host_addr(&self) -> HostPointer;
+    fn get_buffer_host_addr(&self) -> Option<HostPointer> {
+        if self.size() == 0.into() {
+            None
+        } else {
+            unsafe { Some(self.get_buffer_host_addr_unsafe()) }
```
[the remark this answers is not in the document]
grishasobol: Ok done

PR #999, opened 2022-06-01
feat(demo): execution over blocks

### LINKED ISSUE #478: Transaction execution over multiple blocks
### Problem to Solve

@NikVolf  suggested

There is a task with heavy computations so it can't be executed within one `handle`. The idea is that program sends messages to itself with the remaining gas and continues computations.

### Possible Solution

The main handle logic should look like this:
```rust
for item in items { 
  results.push(send_and_wait_for_reply(exec::program_id(), item, gas_available())
}
reply(results)
```

The program just sends message to itself, performs single computation and replies with the result.

### Notes

_No response_

clearloop:
Resolves #478.

- introduce `demo-pow`
-

@reviewer-or-team


examples/binaries/pow/src/code.rs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        pkg = msg::send_and_wait_for_reply::<Package, Package>(
+            exec::program_id(),
+            pkg.calc(),
+            exec::gas_available().into(),
+        )
+        .expect("send message failed")
+        .await
+        .expect("get reply failed");
```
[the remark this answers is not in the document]
gshep: Yep, there is a logic mistake: handle will not execute ever while init doesn't get finished. So all messages sent there will fall into the wait list.

examples/binaries/pow/src/code.rs
```diff
@@ -0,0 +1,88 @@
...
+                return self;
+            } else if self.ptr == self.pow.exponent {
+                self.done = true;
+                return self;
+            }
+
+            self.ptr += 1;
+            self.result = self.pow.base.saturating_mul(self.result);
+            self
+        }
```
gshep, line 75: I'm afraid this computation is not heavy enough. I discussed the task a couple of month ago with @NikVolf and he suggested to use VDF as the operation.

examples/binaries/pow/src/code.rs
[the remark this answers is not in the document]
clearloop: thx! the recursive function works after moving this logic into `fn handle`

examples/binaries/pow/src/code.rs
clearloop: It might be pretty complex if so, I'll do some research on it

pallets/gear/src/tests.rs
```diff
@@ -3433,6 +3433,62 @@ fn cascading_messages_with_value_do_not_overcharge() {
...
+            .encode(),
+            block_gas_limit,
+            block_gas_limit.into(),
+        ));
+
+        run_to_block(3, Some(500_000));
+        assert_eq!(get_last_mail(USER_1), None);
+
+        run_to_block(4, Some(500_000));
+        assert_eq!(get_last_mail(USER_1), None);
```
clearloop, line 3474: use this `remaining_weight` to confirm executing over blocks

clearloop:
> B calculates it and checks gas left on each iteration (needs to be optimised)

looks like we don't have the interface getting gas_limit inside a program?

examples/binaries/calc-hash/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+                return true;
+            }
+
+            if sha2_256(p) != path[next] {
+                return false;
+            }
+        }
+
+        false
+    }
```
clearloop, line 40: thx to Georgy, now using `sha2_256` to mock the situation

examples/binaries/calc-hash/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+codec = { package = "parity-scale-codec", version = "3.1.2", default-features = false, features = ["derive"]}
+sha2 = { version = "0.10.2", default-features = false }
+
+[build-dependencies]
+gear-wasm-builder = { path = "../../../utils/wasm-builder" }
+
+[features]
+std = [
+  "sha2/std",
+  "codec/std",
```
breathx, line 19: For our binaries `std` feature are not familiar one.
I don't think there are some reason to increase binary size by including other features

examples/binaries/calc-hash/aggregator/build.rs
```diff
@@ -0,0 +1,23 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+fn main() {
+    println!("cargo:rerun-if-changed=src/code.rs");
+
```
breathx, line 21:
```suggestion
```

examples/binaries/calc-hash/aggregator/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+codec = { package = "parity-scale-codec", version = "3.1.2", default-features = false, features = ["derive"]}
+shared = { path = "../", package = "demo-calc-hash", default-features = false }
+
+[build-dependencies]
+gear-wasm-builder = { path = "../../../../utils/wasm-builder" }
+
+[features]
+std = [
+  "shared/std",
+  "codec/std",
```
breathx, line 22: same here

examples/binaries/calc-hash/aggregator/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+]
+default = [ "std" ]
```
breathx, line 24: why do we need this by default?

examples/binaries/calc-hash/calculator/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+gstd = { path = "../../../../gstd", features = ["debug"] }
+
+# encoding
+codec = { package = "parity-scale-codec", version = "3.1.2", default-features = false, features = ["derive"]}
+shared = { path = "..", package = "demo-calc-hash", default-features = false }
+
+[build-dependencies]
+gear-wasm-builder = { path = "../../../../utils/wasm-builder" }
+
+[features]
```
breathx, line 19: same as above

examples/binaries/calc-hash/calculator/build.rs
```diff
@@ -0,0 +1,23 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+fn main() {
+    println!("cargo:rerun-if-changed=src/code.rs");
```
breathx, line 20:
```suggestion
```

pallets/gear/src/tests.rs
```diff
@@ -3703,6 +3705,92 @@ fn cascading_messages_with_value_do_not_overcharge() {
...
+            block_gas_limit,
+            0,
+        ));
+
+        run_to_block(2, None);
+        assert_eq!(maybe_last_mail(USER_1), None);
+
+        // loop calculation
+        let path: Vec<[u8; 32]>;
+        loop {
```
breathx, line 3772: I prefer `while maybe_last_mail.is_none`

Some counters and debug outputs to show that this runs multiple times would also be informative

pallets/gear/src/tests.rs
```diff
@@ -3703,6 +3705,92 @@ fn cascading_messages_with_value_do_not_overcharge() {
...
 
+#[test]
+fn execution_over_blocks() {
+    init_logger();
+    new_test_ext().execute_with(|| {
+        use demo_calc_hash::{Method, Package};
+        use demo_calc_hash_aggregator::WASM_BINARY;
+        use demo_calc_hash_calculator::WASM_BINARY as CALC_WASM_BINARY;
+        let block_gas_limit = BlockGasLimitOf::<Test>::get();
+        let max_gas_spent_per_hash = 30_000_000_000u64;
```
breathx, line 3716: probably better to pre-calculate it, instead of explicit specifying?

pallets/gear/src/tests.rs
```diff
@@ -3703,6 +3705,92 @@ fn cascading_messages_with_value_do_not_overcharge() {
...
+        // trigger calculation
+        assert_ok!(Gear::send_message(
+            Origin::signed(USER_1),
+            aggregator,
+            Method::Start(Package {
+                paths: vec![[0; 32]],
+                expected,
+            })
+            .encode(),
+            block_gas_limit,
```
breathx, line 3763: I would send here some small value to show that start fn is cheap, cause only calculations are heavy and cost a lot.

pallets/gear/src/tests.rs
```diff
@@ -3703,6 +3705,92 @@ fn cascading_messages_with_value_do_not_overcharge() {
...
+                0,
+            ));
+
+            run_to_next_block(None);
+
+            if let Some(mail) = maybe_last_mail(USER_1) {
+                path = Vec::<[u8; 32]>::decode(&mut mail.payload()).expect("get response failed");
+                assert!(Package::verify(&path));
+                break;
+            }
```
breathx, line 3787: btw, we should expect reply on start message, not any other one

examples/binaries/calc-hash/src/lib.rs
```diff
@@ -0,0 +1,60 @@
+#![cfg_attr(not(feature = "std"), no_std)]
```
breathx, line 1: this one should be always no_std, due to representing just some common, shared entities

examples/binaries/calc-hash/src/lib.rs
```diff
@@ -0,0 +1,60 @@
...
+#[derive(Clone, Debug, Encode, Decode)]
+pub struct Package {
+    /// The paths of the calculation
+    pub paths: Vec<[u8; 32]>,
+    /// Expected result.
+    pub expected: [u8; 32],
+}
+
+impl Package {
+    /// verify path
```
breathx, line 25: Please, follow common doc styling
```suggestion
    /// Path verification.
```

examples/binaries/calc-hash/calculator/src/code.rs
```diff
@@ -0,0 +1,11 @@
+use gstd::{msg, prelude::*};
+use shared::Package;
+
+#[gstd::async_main]
+async fn main() {
+    let mut pkg: Package = msg::load().expect("invalid pow args");
```
breathx, line 6: that's not pow already

examples/binaries/calc-hash/calculator/src/code.rs
```diff
@@ -0,0 +1,11 @@
+use gstd::{msg, prelude::*};
+use shared::Package;
+
+#[gstd::async_main]
```
breathx, line 4: no reason to be async - it's just overhead

examples/binaries/calc-hash/calculator/src/code.rs
```diff
@@ -0,0 +1,11 @@
+use gstd::{msg, prelude::*};
```
breathx, line 1: btw, this demo may be a part of aggregator itself (special command from source == self id)
or may contain more that one calculation iteration to reduce amount of messages sendings

examples/binaries/calc-hash/aggregator/src/code.rs
```diff
@@ -0,0 +1,59 @@
+use gstd::{exec, msg, ActorId, Decode, MessageId, Vec};
+use shared::{Method, Package};
+
+#[no_mangle]
+pub unsafe extern "C" fn init() {
+    (state::THRESHOLD, state::CALCULATOR) =
+        msg::load::<(u64, ActorId)>().expect("invalid calculator address");
```
breathx, line 7:
```suggestion
        msg::load::<(u64, ActorId)>().expect("Invalid initial data");
```

There is no guarantee that calculator address was invalid, if load failed

breathx, line 7:
```suggestion
        msg::load().expect("invalid calculator address");
```
can't we avoid type specification?

examples/binaries/calc-hash/aggregator/src/code.rs
```diff
@@ -0,0 +1,59 @@
...
+mod state {
+    use gstd::{ActorId, Vec};
+    use shared::Package;
+
+    pub static mut CALCULATOR: ActorId = ActorId::new([0; 32]);
+    pub static mut THRESHOLD: u64 = 0;
+    pub static mut STATUS: Package = Package {
+        paths: Vec::new(),
+        expected: [0; 32],
+    };
```
breathx, line 58: for initial data Option here seems more correct.

breathx (changes requested):
In general, lgtm, just some refactoring stuff in comments

examples/binaries/calc-hash/aggregator/Cargo.toml
clearloop: can not build WASM_BINARY without this, each of examples/binaries/* have this

examples/binaries/calc-hash/over-blocks/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+# encoding
+codec = { package = "parity-scale-codec", version = "3.1.2", default-features = false, features = ["derive"]}
+shared = { path = "../", package = "demo-calc-hash" }
+
+[build-dependencies]
+gear-wasm-builder = { path = "../../../../utils/wasm-builder" }
+
+[features]
+std = []
+default = ["std"]
```
clearloop, line 21: Can not go without this `features` declaration if wanna to test these programs inside `pallet/gear/tests.rs`, each of the testing programs have this.

clearloop (commented):
Tried to resolve all request changes, but still have no ideas about the following 3 ))

pallets/gear/src/tests.rs
```diff
@@ -3703,6 +3705,176 @@ fn cascading_messages_with_value_do_not_overcharge() {
...
+            WASM_BINARY.to_vec(),
+            DEFAULT_SALT.to_vec(),
+            // Hardly can estimate this since this is
+            //
+            // the_max_cost_of_one_sha2_256
+            // + reply the result to program
+            // + reply the result to user
+            //
+            // unless we write a interface in the program to calculate this only.
+            30_000_000_000_u64.encode(),
```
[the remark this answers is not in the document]
breathx: we can do program, which creates one calculation and run calculate_gas on it, adding some small constant for messaging afterward

or just calculate_gas for single call of this program with calculate command

pallets/gear/src/tests.rs
```diff
@@ -4136,6 +4308,10 @@ mod utils {
             .expect("can't find message send event")
     }
 
+    pub(super) fn maybe_last_mail(account: AccountId) -> Option<StoredMessage> {
+        MailboxOf::<Test>::iter_key(account).last()
```
breathx, line 4312: order isn't guaranteed in mailbox
so the best solution to get last mail is to iter events and find the latest UserMessageSent to given account_id

examples/binaries/calc-hash/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    /// Calculate the next path.
+    pub fn calc(mut self) -> Self {
+        self.paths.push(sha2_256(&self.ptr()));
+        self
+    }
+
+    /// Check if the calculation is finished.
+    pub fn finished(&self) -> bool {
+        self.ptr() == self.expected
```
breathx, line 55: it's not correct at all to check expectations in some calculations. here should be counter checker (how much times we did calculation)

examples/binaries/calc-hash/over-blocks/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+
+#[cfg(not(feature = "std"))]
+mod wasm {
+    include! {"./code.rs"}
+}
+
+/// Program methods.
+#[derive(Debug, Encode, Decode)]
+pub enum Method {
+    Start(PackageWithId),
```
breathx, line 40: better to receive here the id, initial data and amount of hashes we should do

examples/binaries/calc-hash/over-blocks/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+    Refuel(PackageId),
+    Calculate(Package),
```
breathx, line 42: we don't need to waste out computationals for encoding and decoding package structure - they are all in memory always, we just need here package id

examples/binaries/calc-hash/over-blocks/src/code.rs
```diff
@@ -0,0 +1,71 @@
...
+unsafe extern "C" fn init() {
+    state::THRESHOLD = Some(msg::load().expect("Invalid threshold."));
+}
+
+#[gstd::async_main]
+async fn main() {
+    let method = msg::load::<Method>().expect("Invalid contract method.");
+
+    unsafe {
+        let threshold = state::THRESHOLD.expect("Threshold has not been set.");
```
breathx, line 23: (Minor comment as best practice) I'd place this to the start of main just to clarify that if we run into some critical issue and don't have threshold set, we won't waste gas for decoding payload

examples/binaries/calc-hash/over-blocks/src/code.rs
```diff
@@ -0,0 +1,71 @@
...
+                    pkg = pkg.calc();
+
+                    if pkg.finished() {
+                        break;
+                    }
+                }
+
+                let _ = msg::reply(pkg, 0).expect("Send reply failed.");
+            }
+        }
```
breathx, line 41: (always remember that the algorithmic language is our everything xD)
this stuff in words should look like this:

STORING DATA:
In memory, you keep mapping PackageId -> (MessageId, Package), where message id is initial one, which started calculations.

INPUT and logic:
Method::Start:
```rust
if !registry.contains(id) {
    registry.insert(id, (msg_id, pkg))
}

let (_, pkg_re)f = registry.get(id);

if pkg_ref.finished() {
    msg::reply(pkg_ref.result())
} else {
    wait()
}
```

Method::Refuel:
```rust
let (&message_id, pkg_ref) = registry.get(id).expect("Wrong id");
while can_calculate() {
   pkg_ref.calc();
   
   if pkg_ref.finished() {
       wake(message_id);
       return
   }
}

//optionally
msg::reply("user, give me more gas");
```

Method::Calculate seems excess in such arch for me

breathx (changes requested):

examples/binaries/calc-hash/over-blocks/src/code.rs
clearloop: > Method::Start:
>
> ```rust
> if !registry.contains(id) {
>     registry.insert(id, (msg_id, pkg))
> }
> 
> let (_, pkg_re)f = registry.get(id);
> 
> if pkg_ref.finished() {
>     msg::reply(pkg_ref.result())
> } else {
>     wait()
> }
> ```

As I saw in logs, the logic before `wait` will be executed before wake, for the logic of this snippet, do you mean the logic of wait will execute for twice? one before `wait` one after `wake`?

breathx: > > Method::Start:
> > ```rust
> > if !registry.contains(id) {
> >     registry.insert(id, (msg_id, pkg))
> > }
> > 
> > let (_, pkg_re)f = registry.get(id);
> > 
> > if pkg_ref.finished() {
> >     msg::reply(pkg_ref.result())
> > } else {
> >     wait()
> > }
> > ```
> 
> As I saw in logs, the logic before `wait` will be executed before wake, for the logic of this snippet, do you mean the logic before wait will execute for twice? one before `wait` one after `wake`?

yeah, because there is no async context. with such code whole demo may be sync

breathx: if pre-wait logic will be executed after wake, that won't break anything

pallets/gear/src/tests.rs
```diff
@@ -3706,6 +3708,234 @@ fn cascading_messages_with_value_do_not_overcharge() {
...
+        let init_gas = Gear::calculate_gas_info(
+            USER_1.into_origin(),
+            HandleKind::Init(WASM_BINARY.to_vec()),
+            EMPTY_PAYLOAD.to_vec(),
+            0,
+            true,
+        )
+        .expect("Failed to get gas spent");
+
+        // deploy demo-calc-hash-over-blocks
```
breathx, line 3740: seems that this is uploading of single-block prog

breathx (approved):

examples/binaries/calc-hash/over-blocks/src/code.rs
```diff
@@ -0,0 +1,115 @@
...
+                    msg::reply(pkg.result(), 0).expect("send reply failed");
+                } else {
+                    exec::wait();
+                }
+            }
+            // Proxy the `Calculate` method for mocking aggregator && calculator.
+            Method::Refuel(id) => {
+                msg::send(exec::program_id(), Method::Calculate(id), 0)
+                    .expect("Send message failed.");
+            }
```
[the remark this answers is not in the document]
clearloop: > still actual for the method

oh forgot to delete this method, now I'm using `maybe_last_message` in the tests

pallets/gear/src/tests.rs
```diff
@@ -4139,6 +4373,29 @@ mod utils {
             .expect("can't find message send event")
     }
 
+    pub(super) fn maybe_last_message(account: AccountId) -> Option<StoredMessage> {
+        SystemPallet::<Test>::events()
+            .iter()
+            .rev()
+            .filter_map(|r| {
+                if let MockEvent::Gear(e) = r.event.clone() {
+                    Some(e)
```
shamilsan, line 4382: Better to clone inside if it is possible.
```suggestion
                if let MockEvent::Gear(e) = r.event {
                    Some(e.clone())
```

examples/binaries/calc-hash/over-blocks/src/code.rs
```diff
@@ -0,0 +1,120 @@
...
+use types::Package;
+
+#[no_mangle]
+unsafe extern "C" fn init() {
+    state::THRESHOLD = Some(msg::load().expect("Invalid threshold."));
+}
+
+#[no_mangle]
+unsafe extern "C" fn handle() {
+    unsafe {
```
shamilsan, line 13: Have I missed it, why `unsafe` block inside an unsafe function?

examples/binaries/calc-hash/over-blocks/src/code.rs
```diff
@@ -0,0 +1,120 @@
...
+        /// New package.
+        pub fn new(expected: u128, src: [u8; 32]) -> Self {
+            Self {
+                expected,
+                message_id: msg::id(),
+                package: shared::Package::new(src),
+            }
+        }
+
+        /// Deref `Pacakge::calc`
```
shamilsan, line 98: Etc.
```suggestion
        /// Deref `Package::calc`
```

examples/binaries/calc-hash/over-blocks/src/code.rs
```diff
@@ -0,0 +1,120 @@
...
+                    .expect("Send message failed.");
+            }
+            Method::Calculate(id) => {
+                if msg::source() != exec::program_id() {
+                    panic!(
+                        "Invalid caller, this is a private method reserved for the program itself."
+                    );
+                }
+
+                let mut pkg = state::REGISTRY
```
shamilsan, line 43: I don't see where you mutate `pkg`.
```suggestion
                let pkg = state::REGISTRY
```

pallets/gear/src/tests.rs
```diff
@@ -3706,6 +3708,234 @@ fn cascading_messages_with_value_do_not_overcharge() {
...
+            Origin::signed(USER_1),
+            WASM_BINARY.to_vec(),
+            DEFAULT_SALT.to_vec(),
+            EMPTY_PAYLOAD.to_vec(),
+            5_000_000_000,
+            0,
+        ));
+        let in_one_block = get_last_program_id();
+
+        assert!(program_exists(in_one_block.into_origin()));
```
shamilsan, line 3851: Importing functions is not Rust idiomatic (https://doc.rust-lang.org/stable/book/ch07-04-bringing-paths-into-scope-with-the-use-keyword.html)
```suggestion
        assert!(common::program_exists(in_one_block.into_origin()));
```

examples/binaries/calc-hash/over-blocks/src/code.rs
clearloop: oh sry, converted from async fn

examples/binaries/calc-hash/over-blocks/src/code.rs
clearloop: `pkg.calc()` is under `&mut self`

pallets/gear/src/tests.rs
clearloop: would be like

```rust
  pub(super) fn maybe_last_message(account: AccountId) -> Option<StoredMessage> {                                                                                                                                                                                             
        SystemPallet::<Test>::events()                                                                                                                                                                                                                                          
            .into_iter()                                                                                                                                                                                                                                                        
            .rev()                                                                                                                                                                                                                                                              
            .find_map(|e| match e.event {                                                                                                                                                                                                                                       
                MockEvent::Gear(Event::UserMessageSent { message, .. }) => {                                                                                                                                                                                                    
                    if message.destination() == account.into() {                                                                                                                                                                                                                
                        Some(message)                                                                                                                                                                                                                                           
                    } else {                                                                                                                                                                                                                                                    
                        None                                                                                                                                                                                                                                                    
                    }                                                                                                                                                                                                                                                           
                }                                                                                                                                                                                                                                                               
                _ => None,                                                                                                                                                                                                                                                      
            })                                                                                                                                                                                                                                                                  
    }     
```

examples/binaries/calc-hash/over-blocks/src/code.rs
```diff
@@ -0,0 +1,115 @@
...
+                package: shared::Package::new(src),
+            }
+        }
+
+        /// Deref `Package::calc`
+        pub fn calc(&mut self) {
+            self.package.calc();
+        }
+
+        /// Deref `Pacakge::finished`
```
shamilsan, line 98: Please check all `Pacakge`s 😄 
```suggestion
        /// Deref `Package::finished`
```

shamilsan (approved):

examples/binaries/calc-hash/over-blocks/src/code.rs
clearloop: 🥲

clearloop:
```
 3:40AM pwd
/home/clearloop/gear
 3:41AM grep -rn Pacakge **/*.rs
 3:41AM grep -rn pacakge **/*.rs
 3:41AM 
```

PR #1000, opened 2022-06-01
New events

### LINKED ISSUE #793: Rethink pallet events

breathx:
Resolves #793.

TODO:
- [x] update gear js runtime upgrade test

**Release notes:** _to be written..._


common/src/event.rs
```diff
@@ -0,0 +1,203 @@
...
+/// Type of changes applied to program in storage.
+#[derive(Debug, Encode, Decode, Clone, PartialEq, Eq, TypeInfo)]
+pub enum ProgramChangeKind<BlockNumber> {
+    /// Program become active and ready for interaction.
+    ///
+    /// Appear when new program created, paused program was resumed
+    /// or expiration block number updated.
+    ///
+    /// Expiration block number presents block number when this program become
+    /// paused due to losing ability to pay rent for holding.
```
NikVolf, line 181:
```suggestion
    /// Active status achieved.
    ///
    /// Occurs when new program created, paused program was resumed
    /// or expiration block number updated.
    ///
    /// Expiration block number presents block number when this program become
    /// paused due to losing ability to pay rent for holding.
```

common/src/event.rs
```diff
@@ -0,0 +1,203 @@
...
+    Active { expiration: BlockNumber },
+
+    // TODO: consider about addition expiration block number
+    // to `Self::Inactive` variant for cleaning storage and
+    // adding ability to take this id for other actors.
+    /// Program become inactive forever due to `gr_exit` call.
+    Inactive,
+
+    /// Program paused, so is no longer available for interaction, but can be
+    /// resumed by paying rent and giving whole data related to it.
```
NikVolf, line 191:
```suggestion
    /// Paused status.
    ///
    /// Program is no longer available for interaction, but can be
    /// resumed by paying rent and giving whole data related to it.
```

common/src/event.rs
```diff
@@ -0,0 +1,203 @@
...
+    /// Program become active and ready for interaction.
+    ///
+    /// Appear when new program created, paused program was resumed
+    /// or expiration block number updated.
+    ///
+    /// Expiration block number presents block number when this program become
+    /// paused due to losing ability to pay rent for holding.
+    Active { expiration: BlockNumber },
+
+    // TODO: consider about addition expiration block number
```
NikVolf, line 184: TODO should have corresponding issue created

pallets/gear/src/manager.rs
```diff
@@ -85,14 +83,47 @@ pub enum HandleKind {
     Reply(H256, ExitCode),
 }
 
+pub struct ExtManager<T: Config> {
+    // Ids checked that they are users.
```
shamilsan, line 87: Are these comments intended to be doc-comments?
```suggestion
    /// Ids checked that they are users.
```

pallets/gear/src/manager.rs
```diff
@@ -85,14 +83,47 @@ pub enum HandleKind {
...
+impl<T: Config> From<ExtManager<T>> for QueuePostProcessingData {
+    fn from(ext_manager: ExtManager<T>) -> Self {
+        Self {
+            dispatch_statuses: ext_manager.dispatch_statuses,
+            state_changes: ext_manager.state_changes,
+        }
+    }
+}
+
 impl<T: Config> Default for ExtManager<T>
```
shamilsan, line 116: Can we just derive `Default`?

common/src/event.rs
```diff
@@ -0,0 +1,203 @@
...
+
+// Empty implementation for `()` to skip requirements.
+impl SystemReason for () {}
+
+/// Composite reason type for any action happened on chain.
+#[derive(Debug, Encode, Decode, Clone, PartialEq, Eq, TypeInfo)]
+pub enum Reason<R: RuntimeReason, S: SystemReason> {
+    /// Runtime reason variant.
+    ///
+    /// This means that actor manually forced some action,
```
ekovalev, line 82: actor **manually** forced => actor **explicitly** forced ?

common/src/event.rs
```diff
@@ -0,0 +1,203 @@
...
+    /// process received message yet.
+    ProgramIsNotInitialized,
+}
+
+/// Composite reason for messages waiting.
+pub type MessageWaitedReason = Reason<MessageWaitedRuntimeReason, MessageWaitedSystemReason>;
+
+/// Runtime reason for messages waking.
+#[derive(Debug, Encode, Decode, Clone, PartialEq, Eq, TypeInfo, RuntimeReason)]
+pub enum MessageWakenRuntimeReason {
```
ekovalev, line 112: `MessageWoken` or `MessageAwakened` maybe?

pallets/gas/src/lib.rs
```diff
@@ -120,6 +120,19 @@ impl ValueNode {
             }
         }
     }
+
+    /// Returns the MessageId (as Origin) of the external node of current value tree.
+    /// If some node along the upstream path is missing, returns an error (tree is invalidated).
+    pub fn root_id<T: Config>(&self) -> Result<H256, DispatchError> {
```
ekovalev, line 126: How is this different from the `root_origin()`? Or is it intended to replace the old one?

pallets/gas/src/lib.rs
```diff
@@ -317,6 +330,16 @@ where
         })
     }
 
+    fn get_origin_id(key: H256) -> Result<Option<H256>, DispatchError> {
```
ekovalev, line 333: Is this a replacement for the old `get_origin()` function?

pallets/gear/src/manager.rs
breathx: I don't think we need to doc comment private fields anywhere, so it's just a comments, but already adapted it, also added above ExtManager itself

common/src/event.rs
breathx: Renamed to `MessageWoken`

pallets/gear/src/manager.rs
```diff
@@ -156,120 +209,104 @@ impl<T: Config> JournalHandler for ExtManager<T>
...
+            common::waiting_init_take_messages(p_id)
+                .into_iter()
+                .for_each(|m_id| {
+                    if let Ok((m, _)) = WaitlistOf::<T>::remove(p_id, m_id) {
+                        Pallet::<T>::deposit_event(Event::<T>::MessageWoken {
+                            id: m_id,
+                            reason: MessageWokenSystemReason::ProgramGotInitialized.into_reason(),
+                        });
+
+                        // TODO: update gas limit in ValueTree here.
```
gshep, line 230: Better to specify also task number for TOOD

pallets/gear/src/manager.rs
```diff
@@ -101,6 +133,27 @@ impl<T: Config> ExtManager<T>
...
+            self.programs.insert(*id);
+            true
+        } else {
+            self.users.insert(*id);
+            false
+        }
+    }
+
+    /// Check if id is not program.
+    pub fn is_user(&mut self, id: &ProgramId) -> bool {
```
gshep, line 153: Frankly speaking this signature and `is_program` are looking weird

pallets/gear/src/lib.rs
```diff
@@ -874,10 +962,23 @@ pub mod pallet {
                 }
             }
 
+            let post_data: QueuePostProcessingData = ext_manager.into();
+
+            // TODO: consider about merging this data into separate event or `Event::MessagesDispatched`
+            for id in post_data.state_changes {
```
gshep, line 968: This is the place I do care about: if there would be processed 100 programs then about 100 events would be emitted. So  I agree with what said in the comment: merge all data in the one event.

common/src/event.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// Program become inactive forever due to `gr_exit` call.
+    Inactive,
+
+    // TODO: consider about addition expiration block number (issue #1014).
+    /// Paused status.
+    ///
+    /// Program is no longer available for interaction, but can be
+    /// resumed by paying rent and giving whole data related to it.
+    Paused {
+        /// Code hash them program relates to.
```
gshep, line 194:
```suggestion
        /// Code hash the program relates to.
```

gshep (approved):
Great work! There are several remarks not so important though.

common/src/lib.rs
```diff
@@ -191,6 +195,12 @@ pub trait ValueTree {
     /// the `key` belongs to a subtree originating at such "orphan" node.
     fn get_origin(key: Self::Key) -> Result<Option<Self::ExternalOrigin>, Self::Error>;
 
+    /// The id of external node for a key, if the latter exists, `None` otherwise.
+    ///
+    /// Error occurs if the tree is invalidated (has "orphan" nodes), and the node identified by
+    /// the `key` belongs to a subtree originating at such "orphan" node.
+    fn get_origin_key(key: Self::Key) -> Result<Option<Self::ExternalOrigin>, Self::Error>;
```
ekovalev, line 202: The signature of this function is identical to the one above - it's natural to think they are doing the same thing.
If this function returns the key of a value tree root node then it should return `Self::Key`, should it not?
The `Self::ExternalOrigin` type is supposed to identify the external owner of the whole tree - that is some actor in actor space. As opposed to the `Self::Key` type which defines the key of a node (including the root node).
The fact they are both represented in the `pallet-gas` as `H256` is a bit confusing though.

pallets/gas/src/lib.rs
```diff
@@ -120,6 +120,19 @@ impl ValueNode {
             }
         }
     }
+
+    /// Returns the MessageId (as Origin) of the external node of current value tree.
```
ekovalev, line 124: > the MessageId (as Origin) of the external node of current value tree

doesn't make sense in my opinion. `MessageId` can neither be an `Origin`, nor the external node of a value tree has to be a part of the same context as messages and be linked to any message by id.

Why not just something like "the key of the root node of a value tree"?

common/src/lib.rs
breathx: Yeah, that's true, refactored

pallets/gear/src/manager.rs
breathx: any suggestions?

pallets/gear/src/lib.rs
breathx: `CodeChanged` can't be interpreted as `CodeChanged` due to the fact of processing especially in extrinsic and a bit cases in process_queue in future. Same here about initialising programs - they took place in extrinsics and in mq processing also. What's the best option how to solve this? I really would like to open issue and consider in future

The only easy option I see is to separate life-cycle dependent stuff about programs from their state changes data. How do you feel about it?

pallets/gear/src/manager.rs
gshep: Maybe just rename to something like `check_if_user_and_add`/`if_user_remember`? Usually functions named `is_something` don't have to accept `&mut self`. Anyway these two are used there only so remark is not important

shamilsan (approved):

PR #1001, opened 2022-06-02
feat(wasm-builder): check crate-type while parsing manifest

### LINKED ISSUE #985: Validate `crate-type` in `gear_wasm_builder`
### Problem to Solve

Thanks for explanation for @shamilsan 

> We can't use crate-type = ["cdylib"] with custom builders

> The root of problem is that we can't call cargo from a build script due to the locking. So the workaround is to create a temporary crate that imports our as a lib. Cargo can't import executables (bin, cdylib etc), but libs only (rlib).

### Possible Solution

Add check for `crate-type` in `gear_wasm_builder`, which will fail build with **our** proper explanation, if binary type set.

clearloop:
Resolves #985 

- check `[lib.crate-type]` in `wasm-builder`
-
-

@reviewer-or-team

### DIFF
--- a/utils/wasm-builder/src/crate_info.rs
+++ b/utils/wasm-builder/src/crate_info.rs
@@ -48,7 +48,7 @@ impl CrateInfo {
             .context("unable to invoke `cargo metadata`")?;
 
         let root_package = Self::root_package(&metadata)
-            .ok_or(BuilderError::RootPackageNotFound.into())
+            .ok_or_else(|| BuilderError::RootPackageNotFound.into())
             .and_then(Self::check)?;
         let name = root_package.name.clone();
         let snake_case_name = name.replace('-', "_");
--- a/utils/wasm-builder/src/crate_info.rs
+++ b/utils/wasm-builder/src/crate_info.rs
@@ -34,36 +34,6 @@ pub struct CrateInfo {
 }
 
 impl CrateInfo {
-    /// check package
-    ///
-    /// - if crate-type includes "lib" or "rlib":
-    fn check(pkg: &Package) -> StdResult<&Package, BuilderError> {
-        // cargo can't import executables (bin, cdylib etc), but libs
-        // only (rlib).
-        //
-        // if no `[lib]` table, the `crate_types` will be [ "lib" ]
-        // by default, we can not detect if this is "rlib" because it
-        // is the "compiler recommended" style of library.
-        //
-        // see also https://doc.rust-lang.org/reference/linkage.html
-        let lib_s = "lib".to_string();
-        let rlib_s = "rlib".to_string();
-        let _ = pkg
-            .targets
-            .iter()
-            .find(|target| {
-                target.name.eq(&target.name)
-                    && target
-                        .crate_types
-                        .iter()
-                        .find(|ty| **ty == lib_s || **ty == rlib_s)
-                        .is_some()
-            })
-            .ok_or(BuilderError::InvalidCrateType)?;
-
-        Ok(pkg)
-    }
-
     /// Create a new `CrateInfo` from a path to the `Cargo.toml`.
     pub fn from_manifest(manifest_path: &Path) -> Result<Self> {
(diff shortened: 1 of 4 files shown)


utils/wasm-builder/src/crate_info.rs
```diff
@@ -18,7 +18,7 @@
 
 use anyhow::{Context, Result};
 use cargo_metadata::{Metadata, MetadataCommand, Package};
-use std::path::Path;
+use std::{path::Path, result::Result as StdResult};
```
shamilsan, line 21: Why do we need `std::Result` here instead of `anyhow::Result`?

utils/wasm-builder/src/crate_info.rs
```diff
@@ -67,4 +68,33 @@ impl CrateInfo {
             .iter()
             .find(|package| package.id == *root_id)
     }
+
+    /// check package
+    ///
+    /// - if crate-type contains "lib" or "rlib":
+    fn check(pkg: &Package) -> StdResult<&Package, BuilderError> {
```
shamilsan, line 75:
```suggestion
    fn check(pkg: &Package) -> Result<&Package> {
```

utils/wasm-builder/src/crate_info.rs
```diff
@@ -47,8 +47,9 @@ impl CrateInfo {
             .exec()
             .context("unable to invoke `cargo metadata`")?;
 
-        let root_package =
-            Self::root_package(&metadata).ok_or(BuilderError::RootPackageNotFound)?;
+        let root_package = Self::root_package(&metadata)
+            .ok_or(BuilderError::RootPackageNotFound)
```
shamilsan, line 51:
```suggestion
            .ok_or(BuilderError::RootPackageNotFound.into())
```

utils/wasm-builder/src/crate_info.rs
```diff
@@ -67,4 +68,33 @@ impl CrateInfo {
...
+        // is the "compiler recommended" style of library.
+        //
+        // see also https://doc.rust-lang.org/reference/linkage.html
+        let lib_s = "lib".to_string();
+        let rlib_s = "rlib".to_string();
+        let _ = pkg
+            .targets
+            .iter()
+            .find(|target| {
+                target.name.eq(&target.name)
```
shamilsan, line 90: Do we compare `target.name` with itself?

utils/wasm-builder/src/crate_info.rs
```diff
@@ -67,4 +68,33 @@ impl CrateInfo {
...
+        // cargo can't import executables (bin, cdylib etc), but libs
+        // only (rlib).
+        //
+        // if no `[lib]` table, the `crate_types` will be [ "lib" ]
+        // by default, we can not detect if this is "rlib" because it
+        // is the "compiler recommended" style of library.
+        //
+        // see also https://doc.rust-lang.org/reference/linkage.html
+        let lib_s = "lib".to_string();
+        let rlib_s = "rlib".to_string();
```
shamilsan, line 85: We can compare with string slices too.
```suggestion
        let lib_s = "lib";
        let rlib_s = "rlib";
```

utils/wasm-builder/src/crate_info.rs
```diff
@@ -67,4 +68,33 @@ impl CrateInfo {
...
+        let _ = pkg
+            .targets
+            .iter()
+            .find(|target| {
+                target.name.eq(&target.name)
+                    && target
+                        .crate_types
+                        .iter()
+                        .any(|ty| **ty == lib_s || **ty == rlib_s)
```
shamilsan, line 94:
```suggestion
                        .any(|ty| ty == lib_s || ty == rlib_s)
```

clearloop: great, used to use `contains()` and could not compare &String with &str

shamilsan (approved):

Issue #1018, opened 2022-06-06
get gas spent cannot work with send value


grishasobol:
### Problem

Get gas spent failed with panic:
```
qT: 1: Unable to get gas spent.: Application, Execution, RuntimePanicked, internal error: entered unreachable code: All requested value for unreserve must be freed. For more info, see module docs.
```

### Steps

Thru gear-idea:

Everywhere in payload choose manual-input

1) Upload nft
payload:
```
{
    "name": "Text",
    "symbol": "Text",
    "baseUri": "Text",
    "royalties": null
}
```

2) Upload marketplace
payload:
```
{
    "ownerId": "YourID",
    "treasuryId": "YourID",
    "treasuryFee": "1"
}
```

`YourId` - it's id of your polka dot account processed thru https://ss58.org/  Address -> Key

3) Send message in market place:
payload:
```
{"AddMarketData": {"nftContractId": "YourID", "ftContractId": null, "tokenId": 1, "price": "1000"}}
```

Set `value = 5000` and `push get_gas_spent`

error:
```
qT: 1: Unable to get gas spent.: Application, Execution, RuntimePanicked, internal error: entered unreachable code: All requested value for unreserve must be freed. For more info, see module docs.
```

breathx:
Duplicate of #982

Issue #1005, opened 2022-06-02
Isolate `get_gas_spent` for tests


breathx:
### Problem to Solve

`get_gas_spent` affects storage. For RPC call its okay, due to removal afterwards from overlay. For usage this method in tests, that's the problem, cause balances and program state is important here.

Relates to #982 also

### Possible Solution

Provide ability to clone current storage state in closure for execution of test externalities.

```rust
let ext = new_test_externalities();

ext.execute_with(|| {
    // blah-blah
    
    // clone/ref/copy of currently executing ext runs some closure and returns value
    let value_from_copy = ext_clone.execute_with(|| { get_gas_spent( .. ) })

    // blah-blah
})
```

Issue #982, opened 2022-05-28
Make sending value be safety for `get_gas_spent`


ekovalev:
Currently `get_gas_spent` function in `pallet-gear` accepts `value` as an input argument.
This yields panic in case one provides a `value` that is non-zero: the `SendValue` journal note is handled by the journal handler as if the respective amount had been reserved on the user's account while it hadn't been.

Possible solution: change the API by removing the `value` from the input arguments and always supply 0 into the fake dispatch; alternatively, skip the `SendValue` note handling when processing the dispatch outcome.

breathx:
Why can't we just emulate it? In the start of `get_gas_spent` execution, give the `source` required amount of gas and value in balances pallet, properly reserve it and continue processing

ekovalev:
> Why can't we just emulate it? In the start of `get_gas_spent` execution, give the `source` required amount of gas and value in balances pallet, properly reserve it and continue processing

Well, it doesn't have much sense because the estimated gas doesn't depend on the `value`. If so why bother with unnecessary shenanigans?

Anyway, I'm fine with minting the `value` amount to the user on whose behalf the dispatch is created, reserving it and burning before the function returns. Anything, as long as it doesn't crash

breathx:
> Well, it doesn't make much sense because the estimated gas doesn't depend on the `value`. If so why bother with unnecessary shenanigans?

Nah, you miss cases when contracts logic depend on amount of value given to it

grishasobol:
Variant to reproduce problem. Wasm files are in attach

Get gas spent failed with panic:
```
qT: 1: Unable to get gas spent.: Application, Execution, RuntimePanicked, internal error: entered unreachable code: All requested value for unreserve must be freed. For more info, see module docs.
```

### Steps

Thru gear-idea:

Everywhere in payload choose manual-input

1) Upload nft
payload:
```
{
    "name": "Text",
    "symbol": "Text",
    "baseUri": "Text",
    "royalties": null
}
```

2) Upload marketplace
payload:
```
{
    "ownerId": "YourID",
    "treasuryId": "YourID",
    "treasuryFee": "1"
}
```

`YourId` - it's id of your polka dot account processed thru https://ss58.org/  Address -> Key

3) Send message in market place:
payload:
```
{"AddMarketData": {"nftContractId": "YourID", "ftContractId": null, "tokenId": 1, "price": "1000"}}
```

Set `value = 5000` and `push get_gas_spent`

error:
```
[...]

PR #1019, opened 2022-06-06
get gas spent fast fix

grishasobol:
fast fix to be able use get_gas_spent for contracts which send value


grishasobol:
Right fix must be made in #1018

breathx (changes requested):
This "fix" totally breaks the logic of rpc call and contracts interaction inside of it.
Issue #982 explains why does it happen and contains ready-for-implement approach for resolving, so resolve this one, please. It doesn't require more lines changed than in this fix.
Afterward #1005 resolves theoretical problems with tests and `get_gas_spent`.

grishasobol:
Closed because we decide to make right complete solution

PR #1030, opened 2022-06-08
Remove clippy warnings from `gstd`

### LINKED ISSUE #1012: remove warnings when use gstd without debug feature
### File Location(s)

_No response_

### Proposal

You can try this:
remove from some examples "debug" feature in gstd deps and make
```
make examples
```
You will see something like this:
```
warning: unused variable: `panic_info`
  --> /Users/gsobol/src/gear/gstd/src/common/handlers.rs:46:14
   |
46 | pub fn panic(panic_info: &core::panic::PanicInfo) -> ! {
   |              ^^^^^^^^^^ help: if this is intentional, prefix it with an underscore: `_panic_info`
   |
   = note: `#[warn(unused_variables)]` on by default

warning: `gstd` (lib) generated 1 warning
warning: unused import: `exec`
 --> sum/src/lib.rs:4:19
  |
4 | use gstd::{debug, exec, msg, prelude::*, ActorId};
  |                   ^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused variable: `log`
  --> sum/src/lib.rs:42:9
   |
42 |     for log in MESSAGE_LOG.iter() {
   |         ^^^ help: if this is intentional, prefix it with an underscore: `_log`
   |
```

Also some tests from examples/binaries cannot be built without "debug" feature in gstd. I don't think it's right behaviour

breathx:
Resolves #1012.

The issue was in unused vars without `debug` feature due to different macros and handlers implementations.

### DIFF
--- a/examples/binaries/compose/src/lib.rs
+++ b/examples/binaries/compose/src/lib.rs
@@ -18,7 +18,7 @@
 
 // This contract recursively composes itself with another contract (the other contract
 // being applied to the input data first): `c(f) = (c(f) . f) x`.
-// Every call to the auto_composer contract incremets the internal `ITER` counter.
+// Every call to the auto_composer contract increments the internal `ITER` counter.
 // As soon as the counter reaches the `MAX_ITER`, the recursion stops.
 // Effectively, this procedure executes a composition of `MAX_ITER` contracts `f`
 // where the output of the previous call is fed to the input of the next call.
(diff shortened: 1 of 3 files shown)


shamilsan (approved):
Great if these `let _ `s will be optimized by the compiler 😄

breathx:
> Great if these `let _ `s will be optimized by the compiler 😄

Seems so. This doesn't even move value.
[Playground to reproduce](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4b596a262f310c6d6792057f1d73b4e0)

Note: replacing `let _ = a;` with `let b = a;` will cause error.

PR #1031, opened 2022-06-08
Fix pages update for mode without lazy-pages feature

grishasobol:
Currently pages update in storage behaviour is different for node builded with lazy-pages and without. It's wrong behaviour. This patch fix it in executor function: prepare_memory.
Also make new test in pallet-gear-debug which checks this case.
Also make refactoring in core-processor and core-backend
Add pallet tests with lazy-pages feature in CI

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -164,8 +164,12 @@ jobs:
       - name: "Build: Split examples by .opt and .meta"
         run: ./scripts/gear.sh build examples-proc
 
-      - name: "Test: Gear pallet tests"
-        run: ./scripts/gear.sh test pallet --release
+      - name: "Test: Gear pallet tests with lazy pages"
+        run: |
+          cargo test -p pallet-gear --features=lazy-pages --release
+          cargo test -p pallet-gear-debug --features=lazy-pages --release
+          cargo test -p pallet-gear-payment --features=lazy-pages --release
+          cargo test -p pallet-usage --features=lazy-pages --release
 
       - name: "Test: Gear workspace"
         run: ./scripts/gear.sh test gear --release
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -195,7 +195,7 @@ jobs:
       - name: "Test: Process node runtime testsuite"
         run: ./scripts/gear.sh test rtest
 
-      # TODO: make also rtest run for node without lazy-pages feature
+      # TODO: make also rtest run for node without lazy-pages feature (issue #1023)
 
       - name: "Test (performance): node runtime testsuite"
         if: ${{ github.ref != 'refs/heads/master' }}
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -164,6 +164,9 @@ jobs:
       - name: "Build: Split examples by .opt and .meta"
         run: ./scripts/gear.sh build examples-proc
 
+      - name: "Test: Gear pallet tests"
+        run: ./scripts/gear.sh test pallet --release
+
       - name: "Test: Gear workspace"
         run: ./scripts/gear.sh test gear --release
         env:
@@ -191,6 +194,8 @@ jobs:
 
       - name: "Test: Process node runtime testsuite"
         run: ./scripts/gear.sh test rtest
+
+      # TODO: make also rtest run for node without lazy-pages feature
 
       - name: "Test (performance): node runtime testsuite"
         if: ${{ github.ref != 'refs/heads/master' }}
--- a/core-processor/src/common.rs
+++ b/core-processor/src/common.rs
@@ -330,6 +330,9 @@ pub enum ExecutionErrorReason {
     /// Cannot read initial memory data from wasm memory.
     #[display(fmt = "Cannot read data for {:?}: {}", _0, _1)]
     InitialMemoryReadFailed(PageNumber, MemoryError),
+    /// Cannot write initial data to wasm memory.
+    #[display(fmt = "Cannot write intial data for {:?}: {}", _0, _1)]
+    InitialDataWriteFailed(PageNumber, MemoryError),
 }
 
 /// Executable actor.
--- a/core-processor/src/common.rs
+++ b/core-processor/src/common.rs
@@ -288,9 +288,6 @@ pub enum ExecutionErrorReason {
     /// Backend error
     #[display(fmt = "{}", _0)]
     Backend(String),
-    /// Processor error
-    #[display(fmt = "{}", _0)]
-    Processor(String),
     /// Ext error
     #[display(fmt = "{}", _0)]
     Ext(String),
@@ -327,6 +324,12 @@ pub enum ExecutionErrorReason {
     /// Page with data is not allocated for program
     #[display(fmt = "{:?} is not allocated for program", _0)]
     PageIsNotAllocated(PageNumber),
+    /// Lazy pages init failed for current program.
+    #[display(fmt = "Cannot init lazy pages for program: {}", _0)]
+    LazyPagesInitFailed(String),
+    /// Cannot read initial memory data from wasm memory.
+    #[display(fmt = "Cannot read data for {:?}: {}", _0, _1)]
+    InitialMemoryReadFailed(PageNumber, MemoryError),
 }
 
 /// Executable actor.
--- a/core-processor/src/common.rs
+++ b/core-processor/src/common.rs
@@ -318,9 +318,6 @@ pub enum ExecutionErrorReason {
     /// Changed page has no data in initial pages
     #[display(fmt = "Changed page has no data in initial pages")]
     PageNoData,
-    /// Ext works with lazy pages, but lazy pages env is not enabled
-    #[display(fmt = "Ext works with lazy pages, but lazy pages env is not enabled")]
-    LazyPagesInconsistentState,
     /// Page with data is not allocated for program
     #[display(fmt = "{:?} is not allocated for program", _0)]
     PageIsNotAllocated(PageNumber),
--- a/core-processor/src/executor.rs
+++ b/core-processor/src/executor.rs
@@ -29,7 +29,7 @@ use gear_backend_common::{BackendReport, Environment, IntoExtInfo, TerminationRe
 use gear_core::{
     env::Ext as EnvExt,
     gas::{ChargeResult, GasAllowanceCounter, GasCounter, ValueCounter},
-    memory::{AllocationsContext, WasmPageNumber},
+    memory::{AllocationsContext, PageBuf, WasmPageNumber},
     message::{ContextSettings, IncomingDispatch, MessageContext},
 };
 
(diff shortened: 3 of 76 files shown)


core-processor/src/common.rs
```diff
@@ -321,12 +318,18 @@ pub enum ExecutionErrorReason {
...
     PageIsNotAllocated(PageNumber),
+    /// Lazy pages init failed for current program.
+    #[display(fmt = "Cannot init lazy pages for program: {}", _0)]
+    LazyPagesInitFailed(String),
+    /// Cannot read initial memory data from wasm memory.
+    #[display(fmt = "Cannot read data for {:?}: {}", _0, _1)]
+    InitialMemoryReadFailed(PageNumber, MemoryError),
+    /// Cannot write initial data to wasm memory.
+    #[display(fmt = "Cannot write intial data for {:?}: {}", _0, _1)]
+    InitialDataWriteFailed(PageNumber, MemoryError),
```
breathx, line 332: Can't we wrap these and similar variants with some general option of `ExecutionErrorReason`
Like `ExecutionErrorReason::Module( **these vars** )`

breathx: I just want to reduce amount of variants in `ExecutionErrorReason`

pallets/gear-debug/src/lib.rs
```diff
@@ -98,7 +98,20 @@ pub mod pallet {
...
+                            // Prints only bytes which is not null
+                            let mut no_zero_bytes = BTreeMap::new();
+                            for (byte_idx, byte_value) in data.iter().enumerate() {
+                                if *byte_value != 0 {
+                                    no_zero_bytes.insert(byte_idx, *byte_value);
+                                }
+                            }
+                            (*page, no_zero_bytes)
+                        })
+                        .collect::<BTreeMap<PageNumber, BTreeMap<usize, u8>>>(),
```
breathx, line 114: 1. better to replace back usize with page_number to improve readability in logs
2. why do we need this spam with non-zero bytes?
3. it's easier to call `data.iter().filter( **check** ).collect()`

pallets/gear-debug/src/tests.rs
```diff
@@ -300,3 +282,227 @@ fn get_last_message_id() -> MessageId {
...
+                        persistent_pages: persistent_pages.clone(),
+                        code_hash: generate_code_hash(&code),
+                    }),
+                }],
+            })
+            .into(),
+        );
+
+        PalletGear::<Test>::send_message(origin, program_id, vec![], 10_000_000_u64, 0_u128)
+            .expect("Failed to send message");
```
breathx, line 480: it's better to call any extrinsic under macros:
`assert_ok!` for this case, and `assert_noop!`, `assert_err!` for others.

pallets/gear-debug/src/tests.rs
```diff
@@ -300,3 +282,227 @@ fn get_last_message_id() -> MessageId {
...
+
+        run_to_block(3, None); // no message will get processed
```
breathx, line 482: why no message will be processed?

pallets/gear-debug/src/tests.rs
```diff
@@ -300,3 +282,227 @@ fn get_last_message_id() -> MessageId {
...
+        // Enable debug-mode
+        DebugMode::<Test>::put(true);
+
+        run_to_block(2, None);
+
+        Pallet::<Test>::do_snapshot();
+
+        let static_pages = WasmPageNumber(8);
+
+        let page1_addr = 0x10000;
```
breathx, line 444: damn, provide this values at the start of the test and use them inside wat, please
as I remember, formatting raw string forks well

core-processor/src/executor.rs
```diff
@@ -255,7 +345,7 @@ pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environ
         }
     };
 
-    log::trace!("term reason = {:?}", termination);
+    log::debug!("term reason = {:?}", termination);
```
breathx, line 348:
```suggestion
    log::trace!("term reason = {:?}", termination);
```

Termination reason as clear on upper layers

core-processor/src/executor.rs
```diff
@@ -197,39 +299,27 @@ pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environ
         settings.forbidden_funcs,
     );
 
-    let mut env = E::new(ext, program.raw_code(), &pages_data, mem_size).map_err(|err| {
-        log::error!("Setup instance err = {}", err);
+    let mut env = E::new(ext, program.raw_code(), mem_size).map_err(|err| {
+        log::debug!("Setup instance err = {}", err);
```
breathx, line 303:
```suggestion
        log::error!("Setup instance err = {}", err);
```
Error should be error

core-processor/src/executor.rs
```diff
@@ -24,102 +24,60 @@ use crate::{
...
+};
 use gear_backend_common::{BackendReport, Environment, IntoExtInfo, TerminationReason};
 use gear_core::{
     env::Ext as EnvExt,
     gas::{ChargeResult, GasAllowanceCounter, GasCounter, ValueCounter},
-    memory::{AllocationsContext, WasmPageNumber},
+    ids::ProgramId,
+    memory::{AllocationsContext, Memory, PageBuf, PageNumber, WasmPageNumber},
     message::{ContextSettings, IncomingDispatch, MessageContext},
 };
```
breathx, line 40: That's really strange, that refactoring increases amount of lines by 20%.
But ok, it seems ok for that moment, due to need of final refactoring of core-processor, once all api and use cases would be clear.

breathx (changes requested):

core-processor/src/executor.rs
grishasobol: It's not error for runtime, it's instantiation error which actually caused by wrong contract. I think we should use log::error for cases which happens unexpectedly for us.

core-processor/src/executor.rs
grishasobol: Where?

pallets/gear-debug/src/tests.rs
grishasobol: Looks good for me, because
1) it's just a test
2) it's better to view values as is inside wasm

pallets/gear-debug/src/lib.rs
grishasobol: 1. No
2. Because in most cases data in wasm page is zero,  so in most cases this will looks good.
3. May be I will check it

pallets/gear-debug/src/tests.rs
grishasobol: What's the difference ?

core-processor/src/common.rs
grishasobol: I think it's not bad to have a lot of error cases. May be you can produce propose which will be informative as this and contains less errors in enum?

pallets/gear-debug/Cargo.toml
```diff
@@ -21,6 +21,7 @@ log = { version = "0.4.17", default-features = false }
 # Internal deps
 common = { package = "gear-common", path = "../../common", default-features = false }
 gear-core = { path = "../../core" }
+pallet-gear = { path = "../gear", default-features = false }
```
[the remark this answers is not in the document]
grishasobol: this is contrary to reality

.github/workflows/CI.yaml
```diff
@@ -164,6 +164,9 @@ jobs:
       - name: "Build: Split examples by .opt and .meta"
         run: ./scripts/gear.sh build examples-proc
 
+      - name: "Test: Gear pallet tests"
```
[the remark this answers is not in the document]
grishasobol: Yes I think I must do it

pallets/gear-debug/src/tests.rs
grishasobol: Will remove this

core-processor/src/executor.rs
grishasobol: That's really strange to evaluate refactoring by line numbers

scripts/src/test.sh
```diff
@@ -77,6 +77,16 @@ rtest() {
...
+  cargo test -p pallet-gear --features=lazy-pages "$@"
+  cargo test -p pallet-gear-debug "$@"
+  cargo test -p pallet-gear-debug --features=lazy-pages "$@"
+  cargo test -p pallet-gear-payment "$@"
+  cargo test -p pallet-gear-payment --features=lazy-pages "$@"
+  cargo test -p pallet-usage "$@"
+  cargo test -p pallet-usage --features=lazy-pages "$@"
+  cargo test -p pallet-gear-messenger "$@"
+  cargo test -p pallet-gear-program "$@"
+  cargo test -p pallet-gas "$@"
```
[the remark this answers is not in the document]
grishasobol: Currently we cannot use it because:
```
cargo check -p pallet-gear --features=runtime-benchmarks --no-default-feature
```
failed with errors

pallets/gear-debug/src/tests.rs
gshep: `assert_noop` checks that the storage didn't changed while extrinsic failed. Others are just more idiomatic to use in substrate tests.

scripts/src/test.sh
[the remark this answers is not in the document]
grishasobol: Currently remove lines with lazy-pages feature and run them as is in CI

.github/workflows/CI.yaml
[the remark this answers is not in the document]
grishasobol: Done

pallets/gear-debug/src/tests.rs
grishasobol: Done

pallets/gear-debug/src/lib.rs
grishasobol: 3th is done

pallets/gear-debug/src/tests.rs
grishasobol: Ok change to assert_ok!

core-processor/src/common.rs
breathx: I'm talking about only this enum `ExecutionErrorReason`
I suggest to provide something like
```rust
enum MemoryStateError {
/// Cannot read initial memory data from wasm memory.
     #[display(fmt = "Cannot read data for {:?}: {}", _0, _1)]
     InitialMemoryReadFailed(PageNumber, MemoryError),
     /// Cannot write initial data to wasm memory.
     #[display(fmt = "Cannot write intial data for {:?}: {}", _0, _1)]
     InitialDataWriteFailed(PageNumber, MemoryError),
}
```
and reuse it in `ExecutionErrorReason`:

```rust
enum ExecutionErrorReason {
  // .. //
  MemoryState(MemoryStateError),
}
```

breathx: And I totally agree with many error variants to make debugging and throwing errors as concrete and clear as it could be

core-processor/src/executor.rs
breathx: It throws as DispatchResultKind

core-processor/src/executor.rs
breathx: agree

core-processor/src/executor.rs
breathx: Just friendly notice.
I mean that in future some good trait for executor should be present

pallets/gear-debug/src/tests.rs
breathx: nevermind

breathx (approved):

core-processor/src/executor.rs
grishasobol: But it's better to show execution term reason in logs

PR #1043, opened 2022-06-09
Temporary fix for sending value in get_gas_spent

gshep:
Temporary fix before #982 

@gear-tech/dev


shamilsan (commented):
We've discussed minting value if it is not sufficient. Will we do it later?

gshep:
> We've discussed minting value if it is not sufficient. Will we do it later?

yep, in #982

shamilsan:
> yep, in #982

Should we write some test here or we don't have time for it?

PR #1044, opened 2022-06-10
Deposit funds to source in get_gas_spent

### LINKED ISSUE #982: Make sending value be safety for `get_gas_spent`

gshep:
Resolves #982 

- use extrinsics in `get_gas_spent`
- mint maximal balance to the source

@gear-tech/dev

### DIFF
--- a/pallets/gear/src/lib.rs
+++ b/pallets/gear/src/lib.rs
@@ -59,7 +59,7 @@ use gear_core::{
 };
 use pallet_gas::Pallet as GasPallet;
 use primitive_types::H256;
-use sp_runtime::traits::{UniqueSaturatedInto, Zero};
+use sp_runtime::traits::{Saturating, UniqueSaturatedInto, Zero};
 use sp_std::{
     collections::{btree_map::BTreeMap, btree_set::BTreeSet},
     convert::TryInto,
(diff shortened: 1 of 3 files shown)


shamilsan (approved):
Looks great to me but need to comprehensively test it including https://idea.gear-tech.io

gshep:
> Looks great to me but need to comprehensively test it including https://idea.gear-tech.io

Thanks for hint that there is a script! Looks like works as it should:
```
$ ./scripts/calc-gas-spent.sh 
Init message:
{"jsonrpc":"2.0","result":1700000000,"id":1}

Handle message:
{"jsonrpc":"2.0","result":359927534,"id":2}

Reply message:
{"jsonrpc":"2.0","result":340000000,"id":3}
```

breathx (commented):
rpc calls and get_gas_burned should be also updated with double call of get_gas_spent impl
renaming would be also useful

PR #1046, opened 2022-06-10
Make `gcore` and `gstd` docs lighten

### LINKED ISSUE #1029: Optimize documentation
### File Location(s)

All of your user libs:
- `galloc/*`
- `gcore/*`
- `gstd/*`
- `gtest/*`

### Proposal

We have a lot of pretty similar functions, for example, `gstd::msg::send` and `gstd::msg::send_bytes`. 
To avoid documentation copy-pasting, decided to ref other function doc.

#### Correct:
```rust
/// Does something special!
pub fn a() {
    // ...
}

/// Same as `[a]`, but with argument.
pub fn b(arg: u64) {
    // ...
}
```

#### Incorrect:
```rust
/// Does something special!
pub fn a() {
    // ...
}

/// Does something special!
/// 
/// With argument!
pub fn b(arg: u64) {
    // ...
}
```

ukint-vs:
Resolves #1029 .

Remove duplicates.

@gear-tech/dev

### DIFF
--- a/gcore/src/msg.rs
+++ b/gcore/src/msg.rs
@@ -213,6 +213,10 @@ pub fn reply(payload: &[u8], value: u128) -> Result<MessageId> {
 ///     msg::reply_with_gas(b"PING", 0, 0).unwrap();
 /// }
 /// ```
+///
+/// # See also
+///
+/// [`reply_push`] function allows to form a reply message in parts.
 pub fn reply_with_gas(payload: &[u8], gas_limit: u64, value: u128) -> Result<MessageId> {
     unsafe {
         let mut message_id = MessageId::default();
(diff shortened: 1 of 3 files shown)


gcore/src/msg.rs
```diff
@@ -230,10 +213,6 @@ pub fn reply(payload: &[u8], value: u128) -> Result<MessageId> {
 ///     msg::reply_with_gas(b"PING", 0, 0).unwrap();
 /// }
 /// ```
-///
-/// # See also
-///
-/// [`reply_push`] function allows to form a reply message in parts.
```
shamilsan, line 236: I like this `See also` section. Maybe it is worth keeping it?

breathx (approved):
But as I said before, better to somehow ref docs about how gasless and gasfull sending works under the hood

PR #1047, opened 2022-06-10
Refactor lazy pages

ark0f:
Also we have Windows support now

@gear-tech/dev


common/src/lazy_pages.rs
```diff
@@ -142,7 +142,7 @@ pub fn protect_lazy_pages_and_update_wasm_mem_addr(
     let new_mem_addr = mem.get_buffer_host_addr();
     if new_mem_addr != old_mem_addr {
         log::debug!(
-            "backend executor has changed wasm mem buff: from {:#x?} to {:#x?}",
```
grishasobol, line 145: Why?

grishasobol (changes requested):

common/src/lazy_pages.rs
ark0f: 'Cause it will look like
```
backend executor has changed wasm mem buff: from Some(
    0xabc
) to Some(
    0xqwe
)
```

grishasobol: And now: 
```
backend executor has changed wasm mem buff: from Some(
abc
) to Some(
qwe
)
```
?

ark0f: And now it's:
```
backend executor has changed wasm mem buff: from Some(0xabc) to Some(0xqwe)
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=4f2b3998339b3142e51817f76e1a1671

grishasobol: Which bad because if number will be like this: 1234  it's completely misunderstandable wether number is hex or decimal

ark0f: Fixed

grishasobol (approved):

grishasobol (approved):

PR #1050, opened 2022-06-13
Enable loki and add archive node

kvs96:


breathx:
Some comments "why and how" would be useful for any changes appear in scripts/docker files/deployment.

PR #1051, opened 2022-06-13
Refactor get_gas_spent

### LINKED ISSUE #1049: Refactor get_gas_spent to calculate_gas
### File Location(s)

_No response_

### Proposal

`fn calculate_gas(...) -> (gas_spent: u64, required_balance: u64)`

gshep:
Resolves #1049

- rename to calculate_gas_info
- use extrinsics in the function
- remove redundant code
- update rpc calls

@gear-tech/dev 

**Release Notes**: `get_gas_spent` was renamed to `calculate_gas_info` to render its new logic: more details about required gas for execution are now provided by the function. New function argument added that controls whether execution of secondary programs may result in a trap. The function calculates gas amount with greater precision.


pallets/gear/src/lib.rs
```diff
@@ -101,6 +102,14 @@ impl DebugInfo for () {
     }
 }
 
+#[derive(Clone, Debug, Decode, Encode, PartialEq, Eq, scale_info::TypeInfo)]
+#[cfg_attr(feature = "std", derive(serde::Deserialize, serde::Serialize))]
+pub struct GasInfo {
+    pub spent: u64,
+    pub to_send: u64,
+    pub burnt: u64,
```
shamilsan, line 110: Let's think about names.
- `spent` is the gas limit enough to process the message
- `to_send` is the gas amount to be sent to other actors
- `burnt` is the gas amount that will be deducted from the sender's account for the message processing

So I'd rename `spent` to `min_limit` or similar, what do you think?

pallets/gear/src/lib.rs
```diff
@@ -494,42 +503,78 @@ pub mod pallet {
...
+            let GasInfo { spent, to_send, .. } = Self::calculate_gas_info_impl(
+                source,
+                kind.clone(),
+                initial_gas,
+                payload.clone(),
+                value,
+            )?;
+
+            // TODO: adding `to_send` until #642 implemented
+            Self::calculate_gas_info_impl(source, kind, spent + to_send, payload, value).map(
```
shamilsan, line 522: How will we distinguish the first call error from the second one?

pallets/gear/src/lib.rs
gshep: Yes, it makes sense

pallets/gear/src/lib.rs
gshep: Frankly speaking I don't think we need to distinguish them: any error means that required gas cannot be determined

shamilsan (approved):
Awesome work 🔥 
We've discussed returning an error when trying to send gas to the user mailbox but as I see you calc it anyway.

pallets/gear/src/lib.rs
shamilsan: Ok, let's collect feedback from users.

shamilsan:
Also, it is important to merge this PR after the IDEA frontend is ready (let me add @osipov-mit to reviewers).

pallets/gear/src/lib.rs
```diff
@@ -494,42 +503,78 @@ pub mod pallet {
...
+                        spent,
+                        to_send,
+                        burnt,
+                    },
+                )
+            })
         }
 
-        fn get_gas_spent_impl(
+        fn calculate_gas_info_impl(
```
SabaunT, line 561: Problems can arise when testing 2 sequential `submit_program` calls during pallet tests. Discussed

SabaunT (changes requested):
Interim review

pallets/gear/src/lib.rs
```diff
@@ -101,6 +102,14 @@ impl DebugInfo for () {
     }
 }
 
+#[derive(Clone, Debug, Decode, Encode, PartialEq, Eq, scale_info::TypeInfo)]
+#[cfg_attr(feature = "std", derive(serde::Deserialize, serde::Serialize))]
+pub struct GasInfo {
+    pub spent: u64,
+    pub to_send: u64,
+    pub burnt: u64,
+}
```
breathx, line 111:
```suggestion
#[derive(Clone, Debug, Decode, Encode, PartialEq, Eq, scale_info::TypeInfo)]
#[cfg_attr(feature = "std", derive(serde::Deserialize, serde::Serialize))]
pub struct GasInfo {
    pub spent: u64,
    pub to_send: u64,
    pub burned: u64,
}

impl GasInfo {
    pub fn balance_diff(&self) -> u64 {
        self.to_send.saturating_add(self.burned)
    }
}
```

breathx: I also missed the fact that exactly gas can't be sent to anybody - we can paste here gas_reserved (which will be used for charging rent for anything) and VALUE to be sent, which is u128, so maybe need to consider about renaming GasInfo into some EconomicInfo/BalanceInfo

breathx: `GasInfo` is lightweight enough for me to be also `Copy`

breathx: Some docs are also need

pallets/gear/src/lib.rs
```diff
@@ -494,42 +503,78 @@ pub mod pallet {
...
+            let initial_gas = <T as pallet_gas::Config>::BlockGasLimit::get();
+            let GasInfo { spent, to_send, .. } = Self::calculate_gas_info_impl(
+                source,
+                kind.clone(),
+                initial_gas,
+                payload.clone(),
+                value,
+            )?;
+
+            // TODO: adding `to_send` until #642 implemented
```
breathx, line 521: this seems not be used at all rn

breathx: that's bad workaround for me

pallets/gear/src/lib.rs
```diff
@@ -645,21 +692,37 @@ pub mod pallet {
...
+                        JournalNote::GasBurned { amount, .. } => {
+                            burnt = burnt.saturating_add(amount);
+                        }
+
+                        JournalNote::MessageDispatched {
+                            outcome: CoreDispatchOutcome::MessageTrap { trap, .. },
+                            ..
+                        } => {
+                            return Err(format!(
+                                "Program terminated with a trap: {}",
```
breathx, line 710: Safety flag for me should indicate do we allow other programs panic

For any case panic in currently checking contract means error and should be so
But for many other cases if our contract handles other contracts panic, this is the correct behaviour.

With safety flag we will or won't fail this fn if panics occurred not in main calculating program

breathx (changes requested):
Yeah, great work!

I just want to clarify moment with return types and append useful feature to rpc call within the PR

This PR also **MUST NOT** be merged, until @osipov-mit approve and create similar changes on JS side, once core team approved it.

breathx:
Release notes to be written also, exactly

pallets/gear/src/tests.rs
```diff
@@ -62,8 +62,16 @@ fn unstoppable_block_execution_works() {
 
         run_to_block(2, None);
 
-        let (expected_burned_gas, _) =
-            calc_handle_gas_spent(USER_1.into_origin(), program_id, EMPTY_PAYLOAD.to_vec());
+        let GasInfo {
```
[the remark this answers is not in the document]
gshep: discussed

pallets/gear/src/lib.rs
gshep: discussed

pallets/gear/src/lib.rs
```diff
@@ -110,7 +110,7 @@ pub struct GasInfo {
     /// Represents minimum gas limit required for execution.
     pub min_limit: u64,
     /// Gas amount that was send to other programs during exection.
```
breathx, line 112: ... that we reserve for some other on-chain interactions

pallets/gear/src/lib.rs
```diff
@@ -563,6 +662,8 @@ pub mod pallet {
                 <T as Config>::Currency::minimum_balance().unique_saturated_into();
 
             let mut max_gas_spent = 0;
+            let mut gas_to_send = 0;
```
breathx, line 665: better to keep naming as in other places
```suggestion
            let mut reserved = 0;
```

breathx: this can be calculated at the end of the processing only once

pallets/gear/src/tests.rs
```diff
@@ -801,7 +810,9 @@ fn lazy_pages() {
     });
 }
 
+// TODO: ignored until #642 implemented
```
breathx, line 813: Do we really can't avoid it change change the demo a bit?
Or you some proxy contract?

breathx (approved):
Glad to see such improvements 👍

pallets/gear/src/tests.rs
gshep: Programs in both tests are sending message with value to a user:
`(import "env" "gr_send_wgas" (func $send (param i32 i32 i32 i64 i32 i32) (result i32)))`

and as I suppose it is a required part of test logic.

osipov-mit (approved):

SabaunT (approved):

breathx (approved):
Great

PR #1053, opened 2022-06-13
add async recursion test and example

grishasobol:
add example which recursively uses async send and reply

### DIFF
--- a/examples/async-recursion/src/lib.rs
+++ b/examples/async-recursion/src/lib.rs
@@ -22,8 +22,8 @@ pub unsafe extern "C" fn init() {
     .expect("Unable to create ActorId");
 }
 
-/// Send message "PING" and waits for reply, then recursevelly
-/// repeats with `val` decreased by reply len, until val > 0.
+/// Send message "PING" and wait for a reply, then recursively
+/// repeat with `val` decreased by reply len while `val` > reply len.
 #[async_recursion]
 async fn rec_func(val: u32) {
     let reply = msg::send_bytes_and_wait_for_reply(unsafe { DEST }, b"PING", 0)
--- a/examples/async-recursion/src/lib.rs
+++ b/examples/async-recursion/src/lib.rs
@@ -25,7 +25,7 @@ pub unsafe extern "C" fn init() {
 /// Send message "PING" and wait for a reply, then recursively
 /// repeat with `val` decreased by reply len while `val` > reply len.
 #[async_recursion]
-async fn rec_func(val: u32) {
+async fn rec_func(val: usize) {
     let reply = msg::send_bytes_and_wait_for_reply(unsafe { DEST }, b"PING", 0)
         .expect("Error in sending message")
         .await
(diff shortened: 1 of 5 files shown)


examples/async-recursion/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+        .collect()
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn init() {
+    let input = String::from_utf8(msg::load_bytes()).expect("Invalid message: should be utf-8");
+    DEST = ActorId::from_slice(
+        &decode_hex(&input).expect("INTIALIZATION FAILED: INVALID PROGRAM ID"),
+    )
+    .expect("Unable to create ActorId");
```
gshep, line 22: ActorId supports codec so the code can be more type-safe but `decode_hex` can be removed (which is already copypasted around):
```suggestion
gstd::metadata! {
    title: "demo async recursion",
    init:
        input: ActorId,
}

#[no_mangle]
pub unsafe extern "C" fn init() {
    DEST = msg::load().expect("Failed to decode `ActorId`");
```

examples/async-recursion/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+async fn rec_func(val: u32) {
+    let reply = msg::send_bytes_and_wait_for_reply(unsafe { DEST }, b"PING", 0)
+        .expect("Error in sending message")
+        .await
+        .expect("Error in async message processing");
+
+    msg::send_bytes(msg::source(), format!("Hello, val = {}", val), 0).unwrap();
+
+    if val - reply.len() as u32 > 0 {
+        rec_func(val - reply.len() as u32).await;
```
gshep, line 37: Another tiny remark is that `value` could have `usize` type so there is no need to cast it to u32:
```rust
if val > reply.len() {
    rec_func(val - reply.len()).await;
```

gshep (approved):
Looks good! :+1:

examples/async-recursion/src/lib.rs
```diff
@@ -0,0 +1,44 @@
...
+pub unsafe extern "C" fn init() {
+    let input = String::from_utf8(msg::load_bytes()).expect("Invalid message: should be utf-8");
+    DEST = ActorId::from_slice(
+        &decode_hex(&input).expect("INTIALIZATION FAILED: INVALID PROGRAM ID"),
+    )
+    .expect("Unable to create ActorId");
+}
+
+/// Send message "PING" and waits for reply, then recursevelly
+/// repeats with `val` decreased by reply len, until val > 0.
```
shamilsan, line 26:
```suggestion
/// Send message "PING" and wait for a reply, then recursively
/// repeat with `val` decreased by reply len while `val` > reply len.
```

examples/async-recursion/src/lib.rs
grishasobol: Done

examples/async-recursion/src/lib.rs
grishasobol: Unfortunately this  solution is not working for our test system, because rtest and gtest do not encode ActorId, they just send it as utf-8.

gshep: `kind: bytes` may be used with provided hex-encoded byte-string '0xABCDEF...789'

PR #1063, opened 2022-06-15
feat(gear): set minimal gas amount for mailbox messages

### LINKED ISSUE #642: Set minimal gas amount for mailbox messages

clearloop:
Resolves #642

- [x] Add minimal gas amount constant
- [x] On sent to mailbox, compare gas limit with this value, 
  - [x] if enough - insert it to mb and deposit event
  - [x] else - just deposit event
- [x] create ValueNode for mailbox messages

@gear-tech/dev


pallets/gear/src/lib.rs
```diff
@@ -184,6 +184,15 @@ pub mod pallet {
...
         type CodeStorage: CodeStorage;
 
+        /// The minimal gas count for message to be inserted in mailbox.
+        ///
+        /// This gas will be consuming as rent for storing and message will be available
+        /// for reply, once gas ends, message removes.
+        ///
+        /// Messages with gas limit less than that minimum will not be added in mailbox,
+        /// but will be seen in logs.
+        type MessageRent: Get<u64>;
```
breathx, line 194: I'd rename this into something like "MailboxThreshold"

pallets/gear/src/lib.rs
```diff
@@ -1436,14 +1443,14 @@ pub mod pallet {
...
-                // and no gas tree needs to be created
-                let origin = who.into_origin();
                 let message = message.into_stored(ProgramId::from_origin(origin));
 
-                // TODO: update logic of insertion into mailbox following new
-                // flow and deposit appropriate event (issue #1010).
-                MailboxOf::<T>::insert(message.clone())?;
+                if gas_limit >= T::MessageRent::get() {
+                    // TODO: update logic of insertion into mailbox following new
+                    // flow and deposit appropriate event (issue #1010).
```
breathx, line 1450: If we insert message in mailbox, than we need to deposit this event:
```rust
Pallet::<T>::deposit_event(Event::UserMessageSent {
    message,
    expiration: Some(T::BlockNumber::zero()),
});
```
T::BlockNumber::zero() will be replaced with the real one soon

if we don't insert message into mailbox, but for showing it in events, we should deposit:
```rust
Pallet::<T>::deposit_event(Event::UserMessageSent {
    message,
    expiration: None,
});
```

Apply this logic everywhere, please

pallets/gear/src/lib.rs
```diff
@@ -1436,14 +1443,14 @@ pub mod pallet {
...
+                    MailboxOf::<T>::insert(message.clone())?;
+                    let _ = T::GasHandler::create(origin, message_id.into_origin(), gas_limit);
```
breathx, line 1452: I want you to provide separate `ValueNode` type, instead of using `External` one. It may be contructed with other `ValueNode::cut` (this is the new one from my imagination), as it does in `split`.
This new type should be able to call only `spend` and `consume` (other variants should return Err), contains origin
Tests should be provided for this value nodes also.
They must be just separated nodes which can only depend on other existing gas in upper tree.

pallets/gear/src/lib.rs
breathx: Also for removing everywhere.
If you want, you can separate this cutting logic into different PR and then come back to this one.

breathx (changes requested):

pallets/gear/src/lib.rs
clearloop: 1. only `spend` and `consume` means without `split_xx` ( for mutated calls only support `spend` and `consume` )?  
2. what about `ValType::ReservedExternal`?
3. should I provide `ValueNode::cut` in this pr?
4. for a new `ValType` instead of `External`, I may provide a new method for creating this `ValueNode`, mb `T::GasHandler::reserve`

breathx: > * only `spend` and `consume` means without `split_xx` ( for mutated calls only support `spend` and `consume` )?

Yeah, other mutating methods should always return `Err(_)` with probably something like `Self::error::forbidden()`

> * what about `ValType::ReservedExternal`?

Considering stuff. See the question no.4 answer.

> * should I provide `ValueNode::cut` in this pr?

As I answered in DM, probably the best way is to represent such node in separate PR, review && merge it, and afterward continue this one.

> * for a new `ValType` instead of `External`, I may provide a new method for creating this `ValueNode`, mb `T::GasHandler::reserve`

What if message to mailbox sent without gas specification? You need to depend on some node with value.

pallets/gear/src/tests.rs
```diff
@@ -1031,8 +1119,9 @@ fn mailbox_works() {
         // caution: runs to block 2
         let reply_to_id = setup_mailbox_test_state(USER_1);
 
-        // Ensure that all the gas has been returned to the sender upon messages processing
-        assert_eq!(BalancesPallet::<Test>::reserved_balance(USER_1), 0);
+        // this 10000000 is from the gas_limit parameter of the send instruction
```
breathx, line 1122: separate this constant into some variable, please

breathx (approved):
In general, seems nice
Gj! 👍

breathx (changes requested):
Ohh, totally forgot about restrictions inside the externalities.
We need to forbid sending gas full messages, which gas limit >0 but <ms_threshold, as we do for value and ED

pallets/gear/src/manager.rs
```diff
@@ -489,23 +489,43 @@ where
...
-                });
+            if let Some(gas_limit) = gas_limit {
+                if gas_limit > mailbox_threshold {
+                    append_message_to_mailbox(gas_limit);
+                }
             } else {
-                log::error!("Error occurred in mailbox insertion")
+                let gas_limit = T::GasHandler::get_limit(message_id)
+                    .unwrap_or(None)
+                    .unwrap_or(0);
```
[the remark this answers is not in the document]
clearloop: 517 is `Result`, and 518 is `Option`

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -586,6 +594,7 @@ benchmarks! {
                     outgoing_limit,
```
breathx, line 594: btw, you're right that changing this incoming params hurtful, couldn't it be wrapped into some other function, which will change only required params, cause most of them unchangeable from bench to bench

breathx (approved):

PR #1066, opened 2022-06-16
Paste error explanation into error reply

### LINKED ISSUE #970: Paste error explanation into error replies
### Problem to Solve

Remove excess events, related to errors in program's execution. Provide explanation not only for console logs, but also for error reply message payload. Limit that value to avoid abusing by panicing.

breathx:
Resolves #970.

**Release notes**: now error replies contains error explanation in payload, so the contracts and user side can detect which error occurred. For depositing events and mailbox propagates string explanation, built on present display formatter, for cross-program-invocation, encoded bytes sent, see more updates on it in #1084.


core-backend/common/src/lib.rs
```diff
@@ -41,7 +42,26 @@ use gear_core::{
...
+pub struct TrimmedString(String);
+
+// Amount of characters allowed to be thrown as string explanation of the error.
+const TRIMMING_LEN: usize = 1024;
+
+impl<T: Into<String>> From<T> for TrimmedString {
+    fn from(other: T) -> Self {
+        let mut string = other.into();
+
+        if string.len() >= TRIMMING_LEN {
```
shamilsan, line 55: Not UTF-8 safe I guess as `String::len()` returns bytes length not chars count. A truncated string may become invalid.

core/src/message/reply.rs
```diff
@@ -169,9 +169,12 @@ impl ReplyPacket {
         }
     }
 
+    // TODO: consider using here impl `CoreError` and provide `AsExitCode`
```
SabaunT, line 172: any issue for that?

breathx: Appended ref to #1083

core-backend/common/src/lib.rs
breathx: Fixed it in the most safety, canonical and optimised way, which I found for that case.
Re-review, please

breathx:
One more related issue opened. It's about extended capabilities of this PR: #1084.

core-backend/common/src/utils.rs
```diff
@@ -0,0 +1,117 @@
...
+pub(crate) fn smart_truncate(s: &mut String, max_bytes: usize) {
+    let mut last_byte = max_bytes;
+
+    if s.len() > last_byte {
+        while !s.is_char_boundary(last_byte) && last_byte != 0 {
+            last_byte = last_byte.saturating_sub(1);
+        }
+
+        s.truncate(last_byte);
+    }
```
gshep, line 12: there is an easier alternative:
```rust
while s.len() > max_bytes {
    s.pop();
}
```

How do you look at it?

breathx: what if s.len is 100k bytes?

breathx: current solution has LAST_CHARACTER_BYTES_AMOUNT iterations in the worst case

core-backend/common/src/utils.rs
```diff
@@ -0,0 +1,117 @@
+use alloc::string::String;
+
+pub(crate) fn smart_truncate(s: &mut String, max_bytes: usize) {
+    let mut last_byte = max_bytes;
+
+    if s.len() > last_byte {
+        while !s.is_char_boundary(last_byte) && last_byte != 0 {
+            last_byte = last_byte.saturating_sub(1);
```
shamilsan, line 8: We may omit the check against zero if we use `saturating_sub`. Or use just `-=`;

```rust
assert_eq!(0_usize.saturating_sub(1), 0);

Comment (breathx) [reply core-backend/common/src/utils.rs:8]: yeah, sure, but if we replace this check, we **theoretically** can run into infinity loop here: `while !s.is_char_boundary(last_byte)`, but this is unreachable, cause `assert!(any_string.is_char_boundary(0))`

this is really cheap check, so we can just left it as it is, isn't it?

File: core-backend/common/Cargo.toml [hunk 5]
```diff
@@ -10,3 +10,5 @@ gear-core = { path = "../../core" }
 gear-core-errors = { path = "../../core-errors", features = ["codec"] }
 log = { version = "0.4.17", default-features = false }
 derive_more = "0.99.17"
+codec = { package = "parity-scale-codec", version = "3.1.2", features = ["derive", "full"], default-features = false }
```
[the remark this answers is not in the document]
breathx: `Encode` implemented to `String` only with that feature

core-backend/common/src/utils.rs
breathx: nvm, will remove it after master merge
no reason to keep excess checks, you're right

shamilsan (approved):

pallets/usage/src/lib.rs
```diff
@@ -406,14 +407,27 @@ pub mod pallet {
...
+                                let message = match dispatch.exit_code() {
+                                    Some(0) | None => dispatch.into_parts().1,
+                                    _ => {
+                                        let message = dispatch.into_parts().1;
+                                        message
+                                            .clone()
+                                            .with_string_payload::<ExecutionErrorReason>()
+                                            .unwrap_or(message)
+                                    }
+                                };
```
gshep, line 420: Maybe it is worth to extract the part to standalone function and place it somewhere in core/common?
```rust
fn to_message_with_string_payload_if_any(dispatch: Dispatch) -> Message {
    let exit_code = dispatch.exit_code();
    let (_, message, ..) = dispatch.into_parts();
    match exit_code {
        Some(0) | None => message,
        _ => message
                .clone()
                .with_string_payload::<ExecutionErrorReason>()
                .unwrap_or(message)
    }
}
```

gshep (approved):
Looks good to me! :gear:

osipov-mit (approved):

pallets/usage/src/lib.rs
breathx: Discussed. 
I don't want to provide such api for core, especially due to the fact that that in #1083 we will use concrete type in error creation, instead of just bytes, so converting payload to string will be infallible

PR #1081, opened 2022-06-18
A little backend cleanup

### LINKED ISSUE #1067: Design and clean up error types linking
### Proposal

We have a lot of error types and reasons them to occur.

Need to somehow unite them by providing soft api for converting from lower level to upper one. 

Remove duplicates (`TerminationReason` and `TerminationReasonKind`) and remove from everywhere Option<trap_reason>, cause it should be always known for now (opt is legit for initial state and optional storing while running backend, any further transformations should be concrete)

Remove `description` field from exts, etc.

Waiting for any other thoughts on how to implement.

ark0f:
Resolves #1067.


core-backend/sandbox/src/env.rs
```diff
@@ -46,16 +46,18 @@ use sp_sandbox::{
...
     #[display(fmt = "Unable to set module memory data")]
     SetModuleMemoryData,
     #[display(fmt = "Unable to save static pages initial data")]
     SaveStaticPagesInitialData,
-    #[display(fmt = "Failed to create env memory")]
-    CreateEnvMemory,
+    #[display(fmt = "Failed to create env memory: {:?}", _0)]
+    CreateEnvMemory(sp_sandbox::Error),
+    #[display(fmt = "No trap explanation")]
+    NoTrapExplanation,
```
breathx, line 60: There should be no cases without trap explanation. Our backend field keeps zero only for cases of `unreachable` section in wasm. This should be checked somehow and research if there are similar-behaviour reasons

core-backend/sandbox/src/env.rs
```diff
@@ -73,7 +75,6 @@ pub(crate) struct Runtime<E: Ext> {
     pub ext: ExtCarrier<E>,
     pub memory: MemoryWrap,
     pub trap: Option<FuncError<E::Error>>,
```
breathx, line 77: how do you feel to keep here not the Option with error, but the "actual TerminationReason"? It starts with `Success` and changes to other variants in appropriate cases? could we do something like that?

core-backend/sandbox/src/funcs.rs
```diff
@@ -92,14 +92,10 @@ pub enum FuncError<E> {
     DebugString(FromUtf8Error),
     #[display(fmt = "`gr_error` expects error occurred earlier")]
     SyscallErrorExpected,
+    #[display(fmt = "Terminated: {:?}", _0)]
+    Terminated(TerminationReason),
     #[display(fmt = "`gr_exit` has been called")]
     Exit,
```
breathx, line 98: Why doesn't `Exit` case included in `Terminated`?

core-backend/wasmtime/Cargo.toml
```diff
@@ -11,6 +11,7 @@ gear-core-errors = { path = "../../core-errors", features = ["codec"] }
 gear-backend-common = { path = "../common" }
 
 wasmtime = { version = "0.35.1", default-features = false, features = ["parallel-compilation", "cranelift"] }
+anyhow = "1.0.57"
```
breathx, line 14: should we disable default features?

breathx: there is the `std` one

core-backend/wasmtime/src/env.rs
```diff
@@ -50,20 +49,22 @@ pub struct StoreData<E: Ext> {
...
     #[display(fmt = "Unable to set module memory data")]
     SetModuleMemoryData,
     #[display(fmt = "Unable to save static pages initial data")]
     SaveStaticPagesInitialData,
-    #[display(fmt = "Failed to create env memory")]
-    CreateEnvMemory,
+    #[display(fmt = "Failed to create env memory: {}", _0)]
+    CreateEnvMemory(anyhow::Error),
+    #[display(fmt = "No trap explanation")]
+    NoTrapExplanation,
```
breathx, line 67: same as in sandbox

core-backend/wasmtime/src/funcs.rs
```diff
@@ -829,12 +829,9 @@ where
...
         let func = move |mut caller: Caller<'_, StoreData<E>>| -> Result<(), Trap> {
-            caller.data_mut().termination_reason = Some(TerminationReason::Trap {
-                explanation: Some(TrapExplanation::Other(
-                    "Unable to call a forbidden function".into(),
-                )),
-                description: None,
-            });
+            caller.data_mut().termination_reason = Some(TerminationReason::Trap(
+                TrapExplanation::Other("Unable to call a forbidden function".into()),
+            ));
```
breathx, line 834: Probably better to provide `TerminationReason::Trap(TrapExplanation::ForbiddenFunction)`?


By the way, for `TrapExplanation` `Cow` seems good, to don't allocate until we need to do something with the inner str

breathx (approved):
Looks great 👍 
I really want you to do some research to detect wasm native variants to break the execution, which we should detect to avoid cases with "Unknown error". They all should be transparent!

shamilsan (approved):
Going in the right direction 👍

pallets/gear/src/manager.rs
```diff
@@ -258,15 +258,12 @@ where
...
-                log::trace!("Dispatch outcome trap: {:?}", message_id);
-
-                if let Some(reason) = trap {
-                    log::info!(
-                        "🪤 Program {} terminated with a trap: {}",
-                        program_id.into_origin(),
-                        reason
-                    );
-                };
+                log::trace!(
```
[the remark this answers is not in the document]
ark0f: Oops, missed the fact there are different levels. Fixed

breathx:
> 

Seems that first two cases covered by our platform logic
Can't the third case be detected on the instantiating time?
And the fourth seems to be alone. To this is not actually unknown reason, that's exactly `WasmRuntimeTrap` / unreachable section always

core-backend/sandbox/src/funcs.rs
```diff
@@ -92,23 +93,27 @@ pub enum FuncError<E> {
...
         match self {
             Self::Core(err) => Some(err),
             _ => None,
         }
     }
+
+    pub fn into_termination_reason(self) -> TerminationReason {
+        match self {
+            Self::Terminated(reason) => reason,
+            err => TerminationReason::Trap(TrapExplanation::Other(err.to_string())),
```
breathx, line 114: As I said previously, can't we pass here not string, but the cow? For future encoding, we don't need to allocate always.

core-backend/sandbox/src/env.rs
```diff
@@ -72,8 +72,7 @@ pub struct SandboxEnvironment<E: Ext + IntoExtInfo> {
 pub(crate) struct Runtime<E: Ext> {
     pub ext: ExtCarrier<E>,
     pub memory: MemoryWrap,
-    pub trap: Option<FuncError<E::Error>>,
-    pub termination_reason: Option<TerminationReasonKind>,
+    pub trap: FuncError<E::Error>,
```
breathx, line 75: that's not correct naming so, if it may be success termination.

breathx (approved):

core-backend/sandbox/src/funcs.rs
breathx: The lifecycle of errors would be
created -> thrown to upper layer -> (optionally) printed in `log::debug` -> encoded via `(E: Encode) trap.encode()` -> sent as bytes after encode.

`&'static str` encodes the same as `String`, but need to check it if I'm wrong, so no need in allocating string always .

ark0f: I wondering do we actually need it now as `Other` variant is always built with `String` and is it premature optimisations or not

core-backend/wasmtime/Cargo.toml
breathx: https://github.com/gear-tech/gear/blob/3009d0e2fe986abf11511975e12056d1d0b6f1ee/core-backend/wasmtime/src/lib.rs#L21
Isn't it?

core-backend/sandbox/src/funcs.rs
breathx: Ok, lets left it as it is atm

PR #1085, opened 2022-06-19
Store module exports in `gear_core::code::Code`

ukint-vs:
Store the exports of the wasm module in `gear_core::code::Code` so that they can be reused.
On each execution we deserialize the binary code into a `parity_wasm::Module` which affects the performance in a bad way.

I will perform a storage migration for this change.

@gear-tech/dev

### DIFF
--- a/core/src/code.rs
+++ b/core/src/code.rs
@@ -25,6 +25,31 @@ use parity_wasm::elements::{Internal, Module};
 use scale_info::TypeInfo;
 use wasm_instrument::gas_metering::Rules;
 
+/// Parse function exports from wasm module into [`DispatchKind`].
+fn get_exports(module: &Module) -> Result<BTreeSet<DispatchKind>, CodeError> {
+    let mut exports = BTreeSet::<DispatchKind>::new();
+
+    for entry in module
+        .export_section()
+        .ok_or(CodeError::ExportSectionNotFound)?
+        .entries()
+        .iter()
+    {
+        if let Internal::Function(_) = entry.internal() {
+            if entry.field() == DispatchKind::Init.into_entry() {
+                exports.insert(DispatchKind::Init);
+            } else if entry.field() == DispatchKind::Handle.into_entry() {
+                exports.insert(DispatchKind::Handle);
+            } else if entry.field() == DispatchKind::Reply.into_entry() {
+                exports.insert(DispatchKind::Reply);
+            } else {
+                return Err(CodeError::NonGearExportFnFound);
+            }
+        }
+    }
+    Ok(exports)
+}
+
 /// Instrumentation error.
 #[derive(Debug)]
 pub enum CodeError {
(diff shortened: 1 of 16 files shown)


ukint-vs:
btw wasm export should be required?

examples/exit-code/src/lib.rs
```diff
@@ -4,6 +4,9 @@ use gstd::{msg, prelude::ToString, ActorId};
 
 static mut HOST: ActorId = ActorId::new([0u8; 32]);
 
+#[no_mangle]
+pub unsafe extern "C" fn handle() {}
```
breathx, line 8: need to place comment here.
it was containing the only `handle_reply` previously, what is just for testing and isn't valid at all, cause we can't receive reply on message if we never send it

in tests we send message with source = id of this demo, to check `msg::exit_code()` functionality

core/src/code.rs
```diff
@@ -97,19 +105,47 @@ impl Code {
...
+            Ok(Self {
+                code: instrumented,
+                raw_code,
+                exports,
+                static_pages,
+                instruction_weights_version: version,
+            })
+        } else {
+            Err(CodeError::RequiredExportFnNotFound)
+        }
```
gshep, line 148: If I'm not mistaken then this part of code could be extracted to the new standalone method and called here and in `new_raw` accrodingly.

gshep (approved):
Good job! :gear:

pallets/gear-program/src/migration.rs
```diff
@@ -17,13 +17,114 @@
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
 use crate::{Config, Pallet, Weight};
+use codec::{Decode, Encode};
+use frame_support::{codec, pallet_prelude::*, storage_alias, traits::Get, Identity};
 
 /// Wrapper for all migrations of this pallet, based on `StorageVersion`.
 pub fn migrate<T: Config>() -> Weight {
-    use frame_support::traits::StorageVersion;
+    let version = StorageVersion::get::<Pallet<T>>();
```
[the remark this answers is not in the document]
ukint-vs: instead of a network wipe we need to start doing migrations but I agree that this time we can skip

pallets/gear-program/src/migration.rs
```diff
@@ -19,7 +19,10 @@
 use crate::{Config, Pallet, Weight};
 
 /// Wrapper for all migrations of this pallet, based on `StorageVersion`.
-pub fn migrate<T: Config>() -> Weight {
+pub fn migrate<T: Config>() -> Weight
+where
+    <T as frame_system::Config>::AccountId: common::Origin,
```
breathx, line 24: import origin upper
...
```suggestion
    <T as Config>::AccountId: Origin,
```

pallets/gear-program/src/tests.rs
```diff
@@ -115,7 +115,7 @@ fn pause_program_works() {
 #[test]
 fn pause_program_twice_fails() {
     new_test_ext().execute_with(|| {
-        let raw_code = hex!("0061736d01000000020f0103656e76066d656d6f7279020001").to_vec();
+        let raw_code = hex!("0061736d01000000010401600000020f0103656e76066d656d6f727902000103020100070a010668616e646c6500000a040102000b0019046e616d650203010000060d01000a656e762e6d656d6f7279").to_vec();
```
breathx, line 118: why do the code size increased so much?

breathx (approved):

pallets/gear-program/src/tests.rs
ukint-vs: cuz of import/export section

breathx: shouldn't we provide here just code and append export sections in `code::new()`?

ukint-vs: didn't understand what you mean

PR #1095, opened 2022-06-21
feat(gas): introduce `ValueType::ReservedLocal`

clearloop:
Ref #1063.

- [x] introduce `ValueType::ReservedLocal`
- [x] restrict calling `split_*` for `ValueType::ReservedLocal`
- [x] introduce `ValueNode::cut`
- [x] tests for this

### DIFF
--- a/pallets/gas/src/lib.rs
+++ b/pallets/gas/src/lib.rs
@@ -352,6 +352,11 @@ pub mod pallet {
         ) -> DispatchResult {
             let mut parent = Self::get_node(key).ok_or(Error::<T>::NodeNotFound)?;
 
+            // Check if the parent node is reserved
+            if let ValueType::ReservedLocal { .. } = parent.inner {
+                return Err(Error::<T>::Forbidden.into());
+            }
+
             ensure!(!parent.consumed, Error::<T>::NodeWasConsumed);
             // This also checks if key == new_node_key
             ensure!(
@@ -361,14 +366,9 @@ pub mod pallet {
 
             // Detect inner from `reserve`.
             let inner = if reserve {
-                let id = Self::get_origin_key(key)?.ok_or(Error::<T>::ParentIsLost)?;
+                let id = Self::get_origin(key)?.ok_or(Error::<T>::ParentIsLost)?;
                 ValueType::ReservedLocal { id, value: amount }
             } else {
-                // Check if the value node is reserved
-                if let ValueType::ReservedLocal { .. } = parent.inner {
-                    return Err(Error::<T>::Forbidden.into());
-                }
-
                 parent.spec_refs = parent.spec_refs.saturating_add(1);
 
                 ValueType::SpecifiedLocal {
@@ -452,14 +452,9 @@ where
         Ok(if let Some(node) = Self::get_node(key) {
             // key known, must return the origin, unless corrupted
             let (root, _) = node.root::<T>()?;
-            if let ValueNode {
-                inner: ValueType::External { id, .. },
-                ..
-            } = root
-            {
-                Some(id)
-            } else {
-                unreachable!("Guaranteed by ValueNode::root method");
+            match root.inner {
+                ValueType::External { id, .. } | ValueType::ReservedLocal { id, .. } => Some(id),
+                _ => unreachable!("Guaranteed by ValueNode::root method"),
             }
         } else {
             // key unknown - legitimate result
--- a/pallets/gas/src/lib.rs
+++ b/pallets/gas/src/lib.rs
@@ -38,6 +38,7 @@ mod property_tests;
 #[derive(Clone, Decode, Debug, Encode, MaxEncodedLen, TypeInfo)]
 pub enum ValueType {
     External { id: H256, value: u64 },
+    ReservedExternal { parent: H256, value: u64 },
     SpecifiedLocal { parent: H256, value: u64 },
     UnspecifiedLocal { parent: H256 },
 }
@@ -55,6 +56,7 @@ impl Default for ValueType {
 pub struct ValueNode {
     pub spec_refs: u32,
     pub unspec_refs: u32,
+    pub reserved_refs: u32,
     pub inner: ValueType,
     pub consumed: bool,
 }
@@ -65,6 +67,7 @@ impl ValueNode {
             inner: ValueType::External { id: origin, value },
             spec_refs: 0,
             unspec_refs: 0,
+            reserved_refs: 0,
             consumed: false,
         }
     }
@@ -72,6 +75,7 @@ impl ValueNode {
     pub fn inner_value(&self) -> Option<u64> {
         match self.inner {
             ValueType::External { value, .. } => Some(value),
+            ValueType::ReservedExternal { value, .. } => Some(value),
             ValueType::SpecifiedLocal { value, .. } => Some(value),
             ValueType::UnspecifiedLocal { .. } => None,
         }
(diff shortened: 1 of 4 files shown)


pallets/gas/src/lib.rs
```diff
@@ -327,6 +340,83 @@ pub mod pallet {
 
             Ok(None)
         }
+
+        pub(super) fn create_from_with_value(
+            key: H256,
+            new_node_key: H256,
+            amount: u64,
+            reserve: bool,
```
clearloop, line 348: copyed from the former `split_with_value`, but with this `reserve` param to detect which kind of node will be created

pallets/gas/src/lib.rs
```diff
@@ -55,6 +56,7 @@ impl Default for ValueType {
 pub struct ValueNode {
     pub spec_refs: u32,
     pub unspec_refs: u32,
+    pub reserved_refs: u32,
```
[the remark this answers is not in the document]
clearloop: > what's the need of having refs for ReservedLocal nodes when they aren't used to return any value to upstream node (like SpecifiedLocal does), nor must be used to block parent node consumption?

yeah I'm thinking of this at the current point, if the value we dispatch to this node is just the multiple of MessageThreshold, there should be no reminders after consuming.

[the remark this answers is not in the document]
SabaunT: I'm not sure, but semantics of `ReservedLocal`, which is a gas node type for mailbox messages, seems to be the following: 
1. **It can be created** by calling `cut` on any node of the gas tree.
2. **No splits** on it are possible, because mailbox message isn't executed in the context of the program, so it can generate new messages, therefore, new gas nodes. When message is in mailbox - it's done, can be only read and it's value can be claimed.
3. **No spends** on it are possible for reason state earlier - you can't partially read message from mailbox or partially claim value from it, while spend in most cases is about it (partial consumption of the gas in the node).
4. **It can be consumed.** It's consumption mustn't block consumption of other nodes in the tree for reasons stated earlier about spending value and splits.

Therefore, solution can be even simpler:
- just store the node in the storage without being a structural part of the tree;
- such node must know the `origin`'s address to return value back to him;
In storage it can be something like `GasTreeStorage { key: MessageId, value: ReservedLocalNode(origin_address, amount) }`.

Even more radical solution suggested by @ekovalev can be to just have a simple `StorageValue` type storage which stores gas values of mailbox messages.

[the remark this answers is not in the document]
ekovalev: Regardless of the motivation for this particular change, I'm still generally against complicating the gas tree.

The original purpose of it was to represent the **computational** budget a smart contract can have at its disposal to run its code. In my option all other use cases (including rent payment for keeping a message in mailbox) should be handled differently.

**Note**: this is different from paying rent while staying in Wait List because waitlisted messages represent futures in `pending` state, that is the life cycle of such message is not yet over at that point. As opposed to messages in mailbox - those ones are `ready` and can neither spawn new computations nor should be blocking release of reserved funds to the original sender. But it's a conceptual discussion - has nothing to do with current implementation.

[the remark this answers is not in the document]
clearloop: > 3. No spends on it are possible for reason stated earlier - you can't partially read message from mailbox or partially claim value from it, while spend in most cases is about it (partial consumption of the gas in the node).

but we may need to burn the value of it on each thershold?

> The original purpose of it was to represent the computational budget a smart contract can have at its disposal to run its code. In my option all other use cases (including rent payment for keeping a message in mailbox) should be handled differently.

make sense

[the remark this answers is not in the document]
SabaunT: > > 3. No spends on it are possible for reason stated earlier - you can't partially read message from mailbox or partially claim value from it, while spend in most cases is about it (partial consumption of the gas in the node).
> 
> but we may need to burn the value of it on each thershold?

Oh that's right. So `spend` is possible in case we implement current approach to charging a rent for being placed in mailbox. But basing on https://github.com/gear-tech/gear/pull/1095/files#r902406505 and https://github.com/gear-tech/gear/pull/1095/files#r902445764 I believe we can implement it differently

pallets/gas/src/tests.rs
```diff
@@ -46,6 +46,41 @@ fn simple_value_tree() {
     });
 }
 
+#[test]
```
breathx, line 49: please, provide also test, that if we cut N from node A with initial balance M to reserved node B, than value of B is N, and value of A is M - N

breathx (approved):

pallets/gas/src/tests.rs
```diff
@@ -419,32 +420,34 @@ fn limit_vs_origin() {
...
 
-        // Original 1000 less 600 that were `split_with_value`
-        assert_eq!(Gas::get_limit(root_node).unwrap(), Some(400));
+        // Original 1000 less 200 that were `cut` and `split_with_value`
+        assert_eq!(Gas::get_limit(root_node).unwrap(), Some(200));
 
-        // Parent's 400
-        assert_eq!(Gas::get_limit(split_1).unwrap(), Some(400));
+        // Parent's 200
+        assert_eq!(Gas::get_limit(split_1).unwrap(), Some(200));
```
breathx, line 434: could you change initial from 1000 to 1100 to avoid messing up this value (become 300) with the cut one

pallets/gas/src/lib.rs
```diff
@@ -314,7 +326,7 @@ pub mod pallet {
                 GasTree::<T>::remove(node_id);
 
                 match node.inner {
-                    ValueType::External { id, value } => {
+                    ValueType::External { id, value } | ValueType::ReservedLocal { id, value } => {
```
SabaunT, line 329: can we reach `ReservedLocal` nodes in `check_consumed` call?

pallets/gas/src/tests.rs
```diff
@@ -46,6 +46,41 @@ fn simple_value_tree() {
...
+            assert_ok!(Gas::split(root, unspecfied));
+        }
+
+        assert_eq!(Gas::total_supply(), total_supply);
+
+        // consume nodes
+        {
+            assert_ok!(Gas::consume(unspecfied));
+            assert_ok!(Gas::consume(specified));
+            assert_ok!(Gas::consume(root));
```
SabaunT, line 76: please add here that the returned result `is_some` and that root was deleted, so `GasTree` has only one node - the reserved one.

pallets/gas/src/tests.rs
```diff
@@ -46,6 +46,41 @@ fn simple_value_tree() {
     });
 }
 
+#[test]
+fn value_tree_with_all_kinds_of_nodes() {
```
SabaunT, line 50: what about making it more complicated by cutting value from unspecified and specified nodes as well?

SabaunT (approved):
Looks great! :+1:
Please see minor comments

pallets/gas/src/lib.rs
clearloop: good catch, no we can't, as I can see for now, the `check_consumed` only work for parent nodes, but looks like we can not reach `External` also )

pallets/gas/src/tests.rs
clearloop: mb I can provide a new test `can_cut_reserved_nodes` for this?

SabaunT: > mb I can provide a new test `can_cut_reserved_nodes` for this?

> Hope you are talking about only about testing, cause we didn't expect atm that cut node can be gas less

no-no
I mean cutting not only root, but specified and unspecified nodes as well and checking node's balances and such

pallets/gas/src/tests.rs
clearloop: https://github.com/gear-tech/gear/blob/fa9763823e7bb3f1f15f74bb252c2a30525f010f/pallets/gas/src/tests.rs#L50

breathx (approved):

ekovalev (approved):

Issue #1096, opened 2022-06-21
Change program upload logic in `gtest`


ukint-vs:
### File Location(s)

gtest/src/program.rs
gtest/src/manager.rs

### Proposal

Since we reject non-gear exports in `gear_core::code::Code::try_new()` we need to process meta wasm file without using `gear_core::program::Program` and execute it as a regular wasm binary without the core logic.
I see it so the manager has a `meta_binaries: BTreeMap<CodeId, Vec<u8>> ` field and executes it via `wasmtime` with memory from `Program`.

For now we are using `Code::new_raw` which cannot be used for testing purposes.

See https://github.com/gear-tech/gear/pull/1085

breathx:
Need to allow creating programs from `.opt.wasm` and `.meta.wasm` separately, or split it under the hood, if given binary has `.wasm` extension with `wasm-proc`

PR #1103, opened 2022-06-22
gtest: Add possibility to load both `.wasm` and `.opt.wasm`

shamilsan:
One should have the possibility to use `.wasm` instead of `.opt.wasm` in `gtest` to be able to test meta functions too.

- `Program::current/current_with_id` now loads `.wasm` binary instead of `.opt.wasm`.
- Added `Program::current_opt/current_opt_with_id` to load `.opt.wasm` binary.
- Updated `gear-wasm-builder` to write the only base path to the `.binpath` file.

@gear-tech/dev

### DIFF
--- a/gtest/src/program.rs
+++ b/gtest/src/program.rs
@@ -161,13 +161,20 @@ impl<'a> Program<'a> {
         system: &'a System,
         id: I,
     ) -> Self {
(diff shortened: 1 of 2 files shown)


breathx:
Wasn't it better to resolve #1096 here?

PR #1106, opened 2022-06-22
Extend `gtest::Program::meta_state` functionality and make it fallible

### LINKED ISSUE #1089: Extend `gtest::Program::meta_state` functionality and make it fallible
### Problem to Solve

For now `meta_state` functions panics if the associated binary wasm function panics in the executor. Also, it gives no error message about its failure

### Possible Solution

`gtest::manager` doesn't panic even if inner execution does. It implements `RunResult` logic that shows if execution failed and stores its error message. We can use that logic in `gtest::wasm_executor`

### Notes

_No response_

dustyNdumb:
Resolves #1089 

- Make ```meta_state``` failable and show error messages
- Add ```meta_state``` to ```gtest::Program::WasmProgram``` trait

**Release Notes**: ...

@gear-tech/dev


core/src/code.rs
```diff
@@ -163,6 +163,7 @@ impl Code {
         version: u32,
         module: Option<Module>,
         instrument_with_const_rules: bool,
+        is_meta: bool,
```
breathx, line 166: useless field

gtest/src/program.rs
```diff
@@ -206,11 +207,12 @@ impl<'a> Program<'a> {
             .expect("Unable to get root directory of the project")
             .join(path)
             .clean();
-
+        let is_meta = path.to_str().unwrap_or_default().ends_with("meta.wasm");
```
breathx, line 210: such functionality will be resolved in #1096

breathx: no need in this changes within PR of fallibility of meta_state fn

gtest/src/wasm_executor.rs
```diff
@@ -208,24 +238,32 @@ mod meta_tests {
...
-            hex: vec![2u8],
-        }));
+        let result: Vec<Wallet> = program
+            .meta_state(&Some(Id {
+                decimal: 2,
+                hex: vec![2u8],
+            }))
+            .expect("Meta_state failed");
 
         assert_eq!(result.encode(), vec![0]);
```
breathx, line 248: no need in encoding here. compare Vec<Wallet> types

gtest/src/wasm_executor.rs
```diff
@@ -208,24 +238,32 @@ mod meta_tests {
...
 
     #[test]
-    #[should_panic(expected = "Failed call: wasm trap: wasm `unreachable` instruction executed")]
-    fn test_failing_with_empty_payload() {
+    fn test_meta_extension_happy_case() {
         let system = System::default();
         let program = Program::from_file(
             &system,
-            "../target/wasm32-unknown-unknown/release/demo_meta.wasm",
+            "../target/wasm32-unknown-unknown/release/demo_meta.meta.wasm",
```
breathx, line 256: same as above, it's separate PR logic

breathx (changes requested):

gtest/src/program.rs
```diff
@@ -297,23 +299,23 @@ impl<'a> Program<'a> {
         self.id
     }
 
-    pub fn meta_state<E: Encode, D: Decode>(&self, payload: E) -> D {
-        D::decode(&mut self.meta_state_with_bytes(payload.encode()).as_slice())
-            .expect("Failed to decode result")
+    pub fn meta_state<E: Encode, D: Decode>(&self, payload: E) -> Result<D, String> {
```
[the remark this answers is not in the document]
shamilsan: Isn't it better to introduce some error type instead of `String`? It will be simpler to refactor it in the future.

gtest/src/wasm_executor.rs
```diff
@@ -37,58 +42,80 @@ impl WasmExecutor {
...
+            .map_err(|error| self.get_error_message(error))?;
 
         match prt_to_result_array[0] {
             Val::I32(ptr_to_result) => self.read_result(ptr_to_result),
-            _ => panic!("{}", "Got wrong type"),
+            _ => Err("Got wrong type".to_string()),
+        }
+    }
+
+    fn get_error_message(&self, error: Error) -> String {
```
shamilsan, line 104: As far as I remember all our errors implement the `Display` trait so this function looks excessive.

gtest/src/wasm_executor.rs
```diff
@@ -293,6 +332,23 @@ mod meta_tests {
         system
             .0
             .borrow_mut()
-            .call_meta(&program.id, None, void_function_name);
+            .call_meta(&program.id, None, void_function_name)
+            .expect("Meta_state failed");
+    }
+
+    #[test]
+    #[should_panic(
```
shamilsan, line 340: Isn't it better to use `assert_eq` with the function result instead of `should_panic`?

runtime/src/lib.rs
```diff
@@ -123,7 +123,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     // The version of the runtime specification. A full node will not attempt to use its native
     //   runtime in substitute for the on-chain Wasm runtime unless all of `spec_name`,
     //   `spec_version`, and `authoring_version` are the same between Wasm and native.
-    spec_version: 1180,
+    spec_version: 1200,
```
shamilsan, line 126: No needed after you remove useless changes. Anyway, we increment `spec_version` by ten, not more.

shamilsan: Removed.

gtest/src/wasm_executor.rs
shamilsan: Done with `assert!(matches!(...))`.

gtest/src/wasm_executor.rs
shamilsan: Removed this function.

gtest/src/wasm_executor.rs
shamilsan: Removed.

gtest/src/program.rs
shamilsan: Removed.

core/src/code.rs
shamilsan: Removed.

gtest/src/program.rs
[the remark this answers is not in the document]
shamilsan: Introduced `TestError` and `TestResult` types.

core/src/code.rs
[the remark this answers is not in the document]
shamilsan: Removed as obsolete.

gtest/src/wasm_executor.rs
shamilsan: Encoding removed.

gtest/src/error.rs
```diff
@@ -0,0 +1,97 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use anyhow::Error as AnyhowError;
+use codec::Error as CodecError;
+use core_processor::ProcessorError;
+use gear_core::{ids::ProgramId, memory::WasmPageNumber};
+use wasmtime::MemoryAccessError;
+
+pub type TestResult<T> = Result<T, TestError>;
```
ark0f, line 25: I think `Test` prefix is unnecessary as for error

gtest/src/wasm_executor.rs
```diff
@@ -40,58 +58,68 @@ impl WasmExecutor {
...
-        let module = Module::new(&engine, meta_binary).expect("Failed to create module");
+        let module = Module::new(&engine, meta_binary)?;
 
         let mut memory =
-            WasmtimeMemory::new(&mut store, MemoryType::new(program.static_pages().0, None))
-                .expect("Failed to create memory");
+            WasmtimeMemory::new(&mut store, MemoryType::new(program.static_pages().0, None))?;
+
         let funcs = funcs_tree::build(&mut store, memory, None);
         let mut externs = Vec::with_capacity(module.imports().len());
```
ark0f, line 70: BTW, all of this logic can be replaced with `wasmtime::Linker`

gtest/src/error.rs
shamilsan: This was made to disambiguate it with `core::result::Result` (that is used in `pub trait WasmProgram` for example), `anyhow::Result`, etc. I can rename `TestResult` to `Result` but this forces `gtest` users to disambiguate them manually. I don't think it is a good idea.

gtest/src/wasm_executor.rs
shamilsan: Not very familiar with `wasmtime`. Please check this commit to whether I understand you correctly:

https://github.com/gear-tech/gear/pull/1106/commits/8f1374d39e364ca8eb5533a338d97b4536b78ff3

ark0f: Oh, I'm sorry. I thought `Linker` can do something more. You used it correctly

gtest/src/error.rs
ark0f: > to disambiguate it with core::result::Result

We can define it as 
```
pub type Result<T, E = Error> = std::result::Result<T, E>
```

> `anyhow::Result`, etc

I see only `anyhow::Error` is used as variant of `TestError`


Thus, I guess there will be no ambiguity

breathx, line 25: simple doc comment, please

gtest/src/wasm_executor.rs
```diff
@@ -212,24 +239,32 @@ mod meta_tests {
...
 
-        program.meta_state_empty::<Vec<Wallet>>();
+        let result: Vec<Wallet> = program
+            .meta_state(&Some(Id {
+                decimal: 2,
+                hex: vec![2u8],
+            }))
+            .expect("Meta_state failed");
+
+        assert_eq!(result.encode(), vec![0]);
```
breathx, line 267:
```suggestion
        assert_eq!(result, vec![]);
```

gtest/src/error.rs
```diff
@@ -0,0 +1,97 @@
...
+
+/// List of general errors.
+#[derive(Debug, derive_more::Display)]
+pub enum TestError {
```
breathx, line 29: seems that there is no error case, when we call meta fn on actor, which has no .meta file under the hood

breathx (approved):
In general I like it
minor stuff in comments above, and one more think: I really agree with @ark0f about naming here, investigate possibilities to use just `Result`. `TestError` is ok for me

breathx:
Write release notes also
This small change is exactly breaking for ecosystem

gtest/src/wasm_executor.rs
shamilsan: Missed it, sorry 🙏

gtest/src/error.rs
shamilsan: Ah, agree, generic alias with default `E` allows using both `Result<T>` and `Result<T, E>`. Fixed.

shamilsan: Added.

gtest/src/wasm_executor.rs
shamilsan: Fixed.

gtest/src/error.rs
shamilsan: You are right, missed this check. Fixed and added a test in https://github.com/gear-tech/gear/pull/1106/commits/10d3f108430f4198ee5b3fc13d4865ccf495d5a5

ark0f:
> `TestError` is ok for me

I see no reason to have `Test` prefix here because the error is for the whole library (not specific for something)

gtest/src/error.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// Wrapper for `wasmtime` error (used [`anyhow::Error`] for that).
+    #[display(fmt = "{}", _0)]
+    WasmtimeError(AnyhowError),
+
+    /// Wrapper for `parity-scale-codec` error (see [`parity_scale_codec::Error`]).
+    #[display(fmt = "{}", _0)]
+    ScaleCodecError(CodecError),
+}
+
+impl From<ProcessorError> for TestError {
```
ark0f, line 80: Use `derive_more`, please

shamilsan: Good idea, thanks!

shamilsan: [Done](https://github.com/gear-tech/gear/pull/1106/commits/d267bb144caf7b344355b1e8bd85ea8f806176aa).

shamilsan:
> I see no reason to have `Test` prefix here because the error is for the whole library (not specific for something)

`TestError` is because the crate's name is `gtest`. So we can understand it as `GTestError` but this name looks ugly. I prefer to not use too common names but more specific ones.

PR #1109, opened 2022-06-22
remove `encoder` bin from `demo-chat`

ukint-vs:
Needed to update CI cache without `encoder.wasm`

@gear-tech/dev


examples/chat/Cargo.toml
```diff
@@ -7,7 +7,3 @@ license = "GPL-3.0"
 
 [dependencies]
 codec = { package = "parity-scale-codec", version = "3.1.2", default-features = false, features = ["derive"] }
-
-[[bin]]
-name = "encoder"
-path = "src/encode.rs"
```
gshep, line 13: Should this file also be removed? If I don't miss anything it is not used anywhere.

Sorry for the late comment )

PR #1112, opened 2022-06-23
Run CI on Windows

ark0f:


.github/workflows/CI.yaml
```diff
@@ -326,7 +326,47 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           path: artifact
+  
+  build-win:
+    needs: spec_version
+    runs-on: windows-latest
```
shamilsan, line 332: I guess `windows-latest` host will be too slow. Nightly build is about 50 mins: https://github.com/gear-tech/builds/runs/7014779874?check_suite_focus=true

ark0f: We build in debug mode so I expect it will be faster

ark0f:
> Do we need to add a cache stage?

Of course, we do. But we test build time on GH runner for now

shamilsan:
@sergeyfilyanin Please configure a Windows self-hosted server. Feel free to commit to this PR branch.

ukint-vs:
> @ukint-vs Please consider integrating `cachepot` to this CI too. If it is quite a long task we may implement it as a separate PR.

I can do it in this PR

ukint-vs (approved):

PR #1113, opened 2022-06-24
Sf win ci

sergeyfilyanin:
just test


ark0f:
> error LNK2019: unresolved external symbol gr_exit referenced in function

It seems this error occurs because linker requires every symbol to be resolved on Windows when resolution is lazy on UNIX (correct me if I'm wrong).

Create `.cargo/config.toml` in user's directory to fix errors:
```toml
[target.x86_64-pc-windows-msvc]
rustflags = ["-Clink-arg=/FORCE:UNRESOLVED"]
```

PR #1117, opened 2022-06-26
Change program upload logic in `gtest`

### LINKED ISSUE #1096: Change program upload logic in `gtest`

ark0f:
Resolves #1096.

Also refactor `wasm-proc`


gtest/src/manager.rs
```diff
@@ -119,14 +127,18 @@ impl Actor {
 
 #[derive(Debug)]
 pub(crate) enum Program {
-    Genuine(CoreProgram, BTreeMap<PageNumber, PageBuf>),
+    Genuine(CoreProgram, CodeId, BTreeMap<PageNumber, PageBuf>),
```
gshep, line 130: Maybe it's worth to make it struct variant?

gshep (approved):
Good job! :gear:

gtest/src/manager.rs
```diff
@@ -67,7 +67,7 @@ impl Actor {
             let mut prog = maybe_prog
                 .take()
                 .expect("actor storage contains only `Some` values by contract");
-            if let Program::Genuine(p, _) = &mut prog {
+            if let Program::Genuine(p, _, _) = &mut prog {
```
shamilsan, line 70: We can condense wildcards (https://rust-lang.github.io/rustfmt/?version=master&search=condense_wildcard_suffixes#condense_wildcard_suffixes)
```suggestion
            if let Program::Genuine(p, ..) = &mut prog {
```

utils/wasm-proc/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+    pub fn metadata(&mut self) -> Result<Vec<u8>, Box<dyn std::error::Error>> {
+        log::debug!(
+            "*** Processing metadata optimization: {}",
+            self.file.display()
+        );
+
+        let mut metadata_module = self.module.clone();
+        let metadata_file_name = self.metadata_file_name();
+
+        pwasm_utils::optimize(
```
shamilsan, line 101: We do the same in `gear-wasm-builder`: https://github.com/gear-tech/gear/blob/master/utils/wasm-builder/src/wasm_project.rs#L233

Better to share code too.

utils/wasm-proc/cli/Cargo.toml
```diff
@@ -0,0 +1,18 @@
+[package]
+name = "wasm-proc-cli"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
shamilsan, line 6:
```suggestion
```

gtest/src/manager.rs
```diff
@@ -146,6 +173,7 @@ pub(crate) struct ExtManager {
     // State
     pub(crate) actors: BTreeMap<ProgramId, (Actor, Balance)>,
     pub(crate) codes: BTreeMap<CodeId, Vec<u8>>,
```
breathx, line 175: this might be renamed or some comments left, that actual `codes` is exactly `opt_binaries`

gtest/src/program.rs
```diff
@@ -207,16 +210,95 @@ impl<'a> Program<'a> {
...
             .clean();
 
-        let program_id = id.clone().into().0;
+        let filename = path.file_name().and_then(OsStr::to_str).unwrap_or_default();
+        assert!(
+            !filename.ends_with(".meta.wasm"),
+            "Cannot load `.meta.wasm` file without `.opt.wasm` one. \
+            Use Program::from_opt_and_meta() instead"
+        );
+        let is_opt = filename.ends_with(".opt.wasm");
```
breathx, line 219: what if I pass path not to .wasm file?

gtest/src/wasm_executor.rs
```diff
@@ -40,7 +41,8 @@ impl WasmExecutor {
         let engine = Engine::new(&config).expect("Failed to create engine");
         let mut store = Store::<StoreData<Ext>>::new(&engine, store_data);
 
-        let module = Module::new(&engine, program.code().code()).expect("Failed to create module");
+        let code = meta_binary.unwrap_or_else(|| program.code().code());
```
breathx, line 44: program's code won't contains any other fns except init, handle, handle_reply, so any calls in gtest should be done over meta wasm only.
We can throw only meta_wasm (wout program itself) and panic if it's none

utils/wasm-proc/src/main.rs
```diff
@@ -170,18 +74,22 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            let code = optimizer.optimize()?;
+            let path = optimizer.optimized_file_name();
+            fs::write(path, code)?;
         }
+
         if !skip_meta {
-            optimize_meta(file, module.clone())?;
+            let code = optimizer.metadata()?;
+            let path = optimizer.metadata_file_name();
+            fs::write(path, code)?;
```
breathx, line 92: seems optimiser doesn't store changes inside of it.
so lines 89..92 won't contain changes applied in line 84, isn't it?

gtest/src/wasm_executor.rs
ark0f: Nice catch!

shamilsan (approved):

gtest/src/program.rs
ark0f: Do you want I just assert file extension?

breathx: yeah

PR #1125, opened 2022-06-30
gtest: Add capability to check actors balances

shamilsan:
- Added methods to `gtest::Program` and `gtest::System` for adding value and getting balance.
- Fixed value processing in `gtest`.
- Added a `piggy-bank` example to test the value movement.
- Minor refactoring.

Using example: https://github.com/gear-tech/gear/pull/1125/files#diff-ff8da23646ffba628a78b81a75f0ebdf5c1cb5489ad431aee84613408e0ea8a4R505

**Release Notes:** ...

@gear-tech/dev

### DIFF
--- a/examples/piggy-bank/src/lib.rs
+++ b/examples/piggy-bank/src/lib.rs
@@ -1,0 +1,19 @@
+#![no_std]
+
+use gstd::{debug, exec, msg};
+
+#[no_mangle]
+unsafe extern "C" fn handle() {
+    match &msg::load_bytes()[..] {
+        b"insert" => debug!(
+            "inserted: {}, total: {}",
+            msg::value(),
+            exec::value_available()
+        ),
+        b"smash" => {
+            debug!("smashing, total: {}", exec::value_available());
+            msg::send_bytes(msg::source(), b"send", exec::value_available()).unwrap();
+        }
+        _ => (),
+    }
+}
--- a/examples/piggy-bank/src/lib.rs
+++ b/examples/piggy-bank/src/lib.rs
@@ -4,16 +4,11 @@ use gstd::{debug, exec, msg};
 
 #[no_mangle]
 unsafe extern "C" fn handle() {
-    match &msg::load_bytes()[..] {
-        b"insert" => debug!(
-            "inserted: {}, total: {}",
-            msg::value(),
-            exec::value_available()
-        ),
-        b"smash" => {
-            debug!("smashing, total: {}", exec::value_available());
-            msg::send_bytes(msg::source(), b"send", exec::value_available()).unwrap();
-        }
-        _ => (),
+    let available_value = exec::value_available();
+    debug!("inserted: {}, total: {}", msg::value(), available_value);
+
+    if msg::load_bytes() == b"smash" {
+        debug!("smashing, total: {}", available_value);
+        msg::reply_bytes(b"send", available_value).unwrap();
     }
 }
--- a/gtest/src/manager.rs
+++ b/gtest/src/manager.rs
@@ -49,6 +49,7 @@ pub(crate) enum Actor {
     // Contract: program is always `Some`, option is used to take ownership
     Uninitialized(Option<MessageId>, Option<Program>),
     Dormant,
+    User,
 }
 
 impl Actor {
@@ -246,7 +247,7 @@ impl ExtManager {
 
         self.gas_limits.insert(dispatch.id(), dispatch.gas_limit());
 
-        if self.actors.contains_key(&dispatch.destination()) {
+        if !self.is_user(&dispatch.destination()) {
             self.dispatches.push_back(dispatch.into_stored());
         } else {
             let message = dispatch.into_parts().1.into_stored();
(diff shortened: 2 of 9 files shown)


examples/piggy-bank/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+#![no_std]
+
+use gstd::{debug, exec, msg};
+
+#[no_mangle]
+unsafe extern "C" fn handle() {
```
breathx, line 6: `let value_available = exec::value_available()` and reuse it to avoid excess sys calls usage - it's cheaper

examples/piggy-bank/src/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+    match &msg::load_bytes()[..] {
+        b"insert" => debug!(
+            "inserted: {}, total: {}",
+            msg::value(),
+            exec::value_available()
+        ),
+        b"smash" => {
+            debug!("smashing, total: {}", exec::value_available());
+            msg::send_bytes(msg::source(), b"send", exec::value_available()).unwrap();
```
breathx, line 15:
```suggestion
            msg::reply_bytes("send", exec::value_available()).unwrap();
```

gtest/src/manager.rs
```diff
@@ -304,6 +305,32 @@ impl ExtManager {
...
+    }
+
+    pub(crate) fn add_value_to(&mut self, id: &ProgramId, value: Balance) {
+        let (_, balance) = self.actors.entry(*id).or_insert((Actor::User, 0));
+        *balance = balance.saturating_add(value);
+    }
+
+    pub(crate) fn actor_balance(&self, id: &ProgramId) -> Option<Balance> {
+        self.actors.get(id).map(|(_, balance)| *balance)
+    }
```
breathx, line 319:
```suggestion
    pub(crate) fn balance_of(&self, id: &ProgramId) -> Balance {
        self.actors.get(id).map(|(_, &balance)| balance).unwrap_or_default()
    }
```

This seems more logic and canonical for me

gtest/src/manager.rs
```diff
@@ -304,6 +305,32 @@ impl ExtManager {
...
+
+    pub(crate) fn claim_value_from_mailbox(&mut self, id: &ProgramId) {
```
breathx, line 321: Don't you want to provide different api for claiming value from concrete messages in mailbox, not from the all in the same moment?

gtest/src/manager.rs
```diff
@@ -645,9 +674,7 @@ impl JournalHandler for ExtManager {
                 *balance -= value;
             };
 
-            if let Some((_, balance)) = self.actors.get_mut(&to) {
-                *balance += value;
-            };
+            self.add_value_to(to, value);
```
breathx, line 677: I would provide `move_value(from, to, amount)` to cover all value sending needs

gtest/src/program.rs
```diff
@@ -405,6 +405,14 @@ impl<'a> Program<'a> {
             .call_meta(&self.id, None, "meta_state")
     }
 
+    pub fn add_value(&mut self, value: Balance) {
```
breathx, line 408: or `mint` probably?

gtest/src/system.rs
```diff
@@ -111,4 +111,22 @@ impl System {
...
+    }
+
+    /// Return actor balance (value) if exists.
+    pub fn actor_balance<ID: Into<ProgramIdWrapper>>(&self, id: ID) -> Option<Balance> {
+        let actor_id = id.into().0;
+        self.0.borrow().actor_balance(&actor_id)
+    }
+
+    /// Claim the user's value from the mailbox.
+    pub fn claim_value_from_mailbox<ID: Into<ProgramIdWrapper>>(&self, id: ID) {
```
breathx, line 128: this method should also be provided for `Mailbox` API

breathx (changes requested):

gtest/src/manager.rs
shamilsan: `add_value/add_value_to` is about minting. Let's rename them. We can provide an additional function to send value but I don't see the use case for it.

gtest/src/program.rs
```diff
@@ -405,6 +405,14 @@ impl<'a> Program<'a> {
...
+        self.manager.borrow_mut().add_value_to(&self.id(), value)
+    }
+
+    pub fn balance(&self) -> Option<Balance> {
```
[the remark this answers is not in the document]
shamilsan: My idea was to distinguish between existent actors vs inexistent ones for detecting mistakes in tests at an early stage. But if you believe it's not the case, I'll make the function return `0` for not existing actors.

examples/piggy-bank/src/lib.rs
shamilsan: Reworked.

gtest/src/program.rs
shamilsan: I was thinking about the `mint` name too, so you've confirmed that it is a more appropriate name.

Renamed `add_value` -> `mint`, `add_value_to` -> `mint_to`.

examples/piggy-bank/src/lib.rs
shamilsan: Done.

gtest/src/program.rs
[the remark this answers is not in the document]
breathx: > My idea was to distinguish between existent actors vs inexistent ones for detecting mistakes in tests at an early stage. But if you believe it's not the case, I'll make the function return `0` for not existing actors.

There are no statement "non existing actor".
gtest contain "program_exists", and its enough for checking existence of programs. any other addresses are valid users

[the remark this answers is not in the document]
shamilsan: Ok, changed it.

gtest/src/manager.rs
shamilsan: Added `Mailbox::claim_value` function (https://github.com/gear-tech/gear/pull/1125/commits/7a91463fbb8dfa1c726d1ac88601641d948498d5).

gtest/src/system.rs
shamilsan: Added `Mailbox::claim_value` function (https://github.com/gear-tech/gear/pull/1125/commits/7a91463fbb8dfa1c726d1ac88601641d948498d5).

breathx (approved):

PR #1126, opened 2022-06-30
Build WASM projects in their own unified directory

ark0f:
I expect it will reduce build size and compilation time during iterative development (e.g. branch switches, demo changes and so on)


shamilsan (approved):
Great, let's wait for feedback 😄

PR #1133, opened 2022-07-01
lazy pages optimization

grishasobol:
in order to  optimise lazy pages work - move logic which calc keys for pages in signal handler.
Also completely remove lazy_pages set because it can be very costly to support it when memory size is big

### Perf
I have run gr_read benchmarks to test perf increase:
```
perf record --all-user --call-graph=dwarf --freq 997 ./target/release/gear-node benchmark pallet --chain=dev --steps=50 --repeat=100 --pallet=pallet_gear --execution=native --extrinsic="gr_read"
``` 
gear_core_processor::executor::prepare_memory  (where lazy pages init info is called) takes
```
master       :    24.9% from all samples
this branch  :    0.01% from all samples
```

time for gr_read:
```
master       :    42.93 µs
this branch  :    32.65 µs
```

time for gr_reply_commit:
```
master       :    68.2 µs
this branch  :    49.92 µs
```

### Testing
Run node with this patch over all test-net chain - all OK

### Important
This patch requires node binary update after runtime update.

### DIFF
--- a/common/src/lazy_pages.rs
+++ b/common/src/lazy_pages.rs
@@ -82,16 +82,14 @@ pub fn protect_pages_and_init_info(mem: &impl Memory, prog_id: ProgramId) -> Res
     let prog_prefix = crate::pages_prefix(prog_id.into_origin());
     gear_ri::set_program_prefix(prog_prefix);
 
-    let addr = if let Some(addr) = mem.get_buffer_host_addr() {
-        addr
+    if let Some(addr) = mem.get_buffer_host_addr() {
+        gear_ri::set_wasm_mem_begin_addr(addr).map_err(|e| {
+            log::error!("{} (it's better to stop node now)", e);
+            e
+        })?;
     } else {
         return Ok(());
-    };
-
-    gear_ri::set_wasm_mem_begin_addr(addr).map_err(|e| {
-        log::error!("{} (it's better to stop node now)", e);
-        e
-    })?;
+    }
 
     let size = (mem.size().0 + 1) * WasmPageNumber::size() as u32;
     gear_ri::set_wasm_mem_size(size)?;
--- a/common/src/lazy_pages.rs
+++ b/common/src/lazy_pages.rs
@@ -37,6 +37,8 @@ pub enum Error {
     ReleasedPageHasInitialData(PageNumber),
     #[display(fmt = "RUNTIME ERROR: wasm memory buffer is undefined")]
     WasmMemBufferIsUndefined,
+    #[display(fmt = "wasm memory buffer size is bigger then u32::MAX")]
+    WasmMemorySizeOverflow,
 }
 
 impl From<RIError> for Error {
@@ -91,7 +93,13 @@ pub fn protect_pages_and_init_info(mem: &impl Memory, prog_id: ProgramId) -> Res
         return Ok(());
     }
 
-    let size = (mem.size().0 + 1) * WasmPageNumber::size() as u32;
+    let size = mem
+        .size()
+        .0
+        .checked_add(1)
+        .ok_or(Error::WasmMemorySizeOverflow)?
+        .checked_mul(WasmPageNumber::size() as u32)
+        .ok_or(Error::WasmMemorySizeOverflow)?;
     gear_ri::set_wasm_mem_size(size)?;
 
     mprotect_lazy_pages(mem, true)
@@ -151,7 +159,12 @@ pub fn update_lazy_pages_and_protect_again(
 
     let new_mem_size = mem.size();
     if new_mem_size > old_mem_size {
-        let size = (new_mem_size.0 + 1) * WasmPageNumber::size() as u32;
+        let size = new_mem_size
+            .0
+            .checked_add(1)
+            .ok_or(Error::WasmMemorySizeOverflow)?
+            .checked_mul(WasmPageNumber::size() as u32)
+            .ok_or(Error::WasmMemorySizeOverflow)?;
         gear_ri::set_wasm_mem_size(size)?;
     }
 
--- a/common/src/lazy_pages.rs
+++ b/common/src/lazy_pages.rs
@@ -22,7 +22,7 @@ use crate::Origin;
 use core::fmt;
 use gear_core::{
     ids::ProgramId,
-    memory::{HostPointer, Memory, PageBuf, PageNumber},
+    memory::{HostPointer, Memory, PageBuf, PageNumber, WasmPageNumber},
 };
 use gear_runtime_interface::{gear_ri, RIError};
 use sp_std::{collections::btree_map::BTreeMap, vec::Vec};
@@ -46,11 +46,11 @@ impl From<RIError> for Error {
 }
 
 fn mprotect_lazy_pages(mem: &impl Memory, protect: bool) -> Result<(), Error> {
-    let wasm_mem_addr = match mem.get_buffer_host_addr() {
-        None => return Ok(()),
-        Some(addr) => addr,
-    };
-    gear_ri::mprotect_lazy_pages(wasm_mem_addr, protect)
+    if mem.get_buffer_host_addr().is_none() {
+        return Ok(());
+    }
+
+    gear_ri::mprotect_lazy_pages(protect)
         .map_err(Into::into)
         .map_err(|e| {
             log::error!("{} (it's better to stop node now)", e);
@@ -76,33 +76,25 @@ pub fn is_lazy_pages_enabled() -> bool {
 }
 
 /// Protect and save storage keys for pages which has no data
-pub fn protect_pages_and_init_info(
-    mem: &impl Memory,
-    lazy_pages: impl Iterator<Item = PageNumber>,
-    prog_id: ProgramId,
-) -> Result<(), Error> {
-    let prog_id_hash = prog_id.into_origin();
-    let mut lay_pages_peekable = lazy_pages.peekable();
-
+pub fn protect_pages_and_init_info(mem: &impl Memory, prog_id: ProgramId) -> Result<(), Error> {
     gear_ri::reset_lazy_pages_info();
 
-    let addr = match mem.get_buffer_host_addr() {
-        None => {
-            return if lay_pages_peekable.peek().is_none() {
-                // In this case wasm buffer cannot be undefined
-                Err(Error::WasmMemBufferIsUndefined)
-            } else {
-                Ok(())
-            };
-        }
-        Some(addr) => addr,
+    let prog_prefix = crate::pages_prefix(prog_id.into_origin());
+    gear_ri::set_program_prefix(prog_prefix);
+
+    let addr = if let Some(addr) = mem.get_buffer_host_addr() {
+        addr
+    } else {
+        return Ok(());
     };
+
     gear_ri::set_wasm_mem_begin_addr(addr).map_err(|e| {
         log::error!("{} (it's better to stop node now)", e);
         e
     })?;
 
-    crate::save_page_lazy_info(prog_id_hash, lay_pages_peekable);
+    let size = (mem.size().0 + 1) * WasmPageNumber::size() as u32;
+    gear_ri::set_wasm_mem_size(size)?;
 
     mprotect_lazy_pages(mem, true)
 }
@@ -131,10 +123,11 @@ pub fn remove_lazy_pages_prot(mem: &impl Memory) -> Result<(), Error> {
     mprotect_lazy_pages(mem, false)
 }
 
-/// Protect lazy-pages and set new wasm mem addr if it has been changed
-pub fn protect_lazy_pages_and_update_wasm_mem_addr(
+/// Protect lazy-pages and set new wasm mem addr and size, if they have been changed
+pub fn update_lazy_pages_and_protect_again(
     mem: &impl Memory,
     old_mem_addr: Option<HostPointer>,
+    old_mem_size: WasmPageNumber,
 ) -> Result<(), Error> {
     struct PointerDisplay(HostPointer);
 
@@ -157,15 +150,14 @@ pub fn protect_lazy_pages_and_update_wasm_mem_addr(
                 e
             })?;
     }
-    mprotect_lazy_pages(mem, true)
-}
 
-/// Returns list of current lazy pages numbers
-pub fn get_lazy_pages_numbers() -> Vec<PageNumber> {
-    gear_ri::get_lazy_pages_numbers()
-        .iter()
-        .map(|p| PageNumber(*p))
-        .collect()
+    let new_mem_size = mem.size();
+    if new_mem_size > old_mem_size {
+        let size = (new_mem_size.0 + 1) * WasmPageNumber::size() as u32;
+        gear_ri::set_wasm_mem_size(size)?;
+    }
+
+    mprotect_lazy_pages(mem, true)
 }
 
 /// Returns list of realeased pages numbers
--- a/lazy-pages/src/lib.rs
+++ b/lazy-pages/src/lib.rs
@@ -39,8 +39,8 @@
 //! in storage. But program may have some significant data for next execution - so we have a bug.
 //! To avoid this we restrict double releasing.
 //! You can also check another cases in test: memory_access_cases.
-//!
-//! TODO: remove all deprecated code before release (issue #1147)
+
+// TODO: remove all deprecated code before release (issue #1147)
 
 #![allow(useless_deprecated, deprecated)]
 
--- a/lazy-pages/src/lib.rs
+++ b/lazy-pages/src/lib.rs
@@ -20,66 +20,185 @@
 //! In runtime data for contract wasm memory pages can be loaded in lazy manner.
 //! All pages, which is supposed to be lazy, must be mprotected before contract execution.
 //! During execution data from storage is loaded for all pages, which has been accesed.
-//! See also `handle_sigsegv`.
+//! See also `sys::user_signal_handler` in the source code.
+//!
+//! Restrict any page handling in signal handler more then one time.
+//! If some page will be released twice it means, that this page has been added
+//! to lazy pages more then one time during current execution.
+//! This situation may cause problems with memory data update in storage.
+//! For example: one page has no data in storage, but allocated for current program.
+//! Let's make some action for it:
+//! 1) Change data in page: Default data  ->  Data1
+//! 2) Free page
+//! 3) Alloc page, data will Data2 (may be equal Data1).
+//! 4) After alloc we can set page as lazy, to identify wether page is changed after allocation.
+//! This means that we can skip page update in storage in case it wasnt changed after allocation.
+//! 5) Write some data in page but do not change it Data2 -> Data2.
+//! During this step signal handler writes Data2 as data for released page.
+//! 6) After execution we will have Data2 in page. And Data2 in released. So, nothing will be updated
+//! in storage. But program may have some significant data for next execution - so we have a bug.
+//! To avoid this we restrict double releasing.
+//! You can also check another cases in test: memory_access_cases.
+//!
+//! TODO: remove all deprecated code before release (issue #1147)
+
+#![allow(useless_deprecated, deprecated)]
 
 use gear_core::memory::{HostPointer, PageBuf, PageNumber, WasmPageNumber};
 use sp_std::vec::Vec;
-use std::{cell::RefCell, collections::BTreeMap, ops::Add};
+use std::{cell::RefCell, collections::BTreeMap};
 
 mod sys;
 
+#[derive(Debug, derive_more::Display, derive_more::From)]
+pub enum Error {
+    #[display(fmt = "WASM memory begin address is not set")]
+    WasmMemAddrIsNotSet,
+    #[display(
+        fmt = "Exception is from unknown memory (WASM {:#x} > native page {:x})",
+        wasm_mem_begin,
+        native_page
+    )]
+    SignalFromUnknownMemory {
+        wasm_mem_begin: usize,
+        native_page: usize,
+    },
+    #[display(
+        fmt = "Page data must contain {} bytes, actually has {}",
+        expected,
+        actual
+    )]
+    InvalidPageSize { expected: usize, actual: u32 },
+    #[display(fmt = "Exception is from unknown memory: addr = {:?}, {:?}", _0, _1)]
+    LazyPageNotExistForSignalAddr(*const (), PageNumber),
+    /// Found a signal from same page twice - see more in head comment.
+    #[display(fmt = "Page cannot be release page twice: {:?}", _0)]
+    DoubleRelease(PageNumber),
+    #[display(fmt = "Protection error: {}", _0)]
+    #[from]
+    MemoryProtection(region::Error),
+}
+
+#[derive(Default, PartialEq, Eq)]
+pub(crate) struct LazyPagesExecutionContext {
+    /// Pointer to the begin of wasm memory buffer
+    pub wasm_mem_addr: Option<HostPointer>,
+    /// Wasm memory buffer size, to identify whether signal is from wasm memory buffer.
+    pub wasm_mem_size: Option<usize>,
+    /// Current program prefix in storage
+    pub program_storage_prefix: Option<Vec<u8>>,
+    /// Page data, which has been in storage before current execution.
+    /// For each lazy page, which has been accessed.
+    pub released_lazy_pages: BTreeMap<PageNumber, Option<PageBuf>>,
+
+    #[deprecated]
+    /// Keys in storage for each lazy page.
+    pub lazy_pages_info: BTreeMap<PageNumber, Vec<u8>>,
+}
+
 thread_local! {
-    /// NOTE: here we suppose, that each contract is executed in separate thread.
-    /// Or may be in one thread but consequentially.
+    // NOTE: here we suppose, that each contract is executed in separate thread.
+    // Or may be in one thread but consequentially.
 
     /// Identify whether signal handler is set for current thread
     static LAZY_PAGES_ENABLED: RefCell<bool> = RefCell::new(false);
-    /// Pointer to the begin of wasm memory buffer
-    static WASM_MEM_BEGIN: RefCell<HostPointer> = RefCell::new(0);
-    /// Key in storage for each lazy page
-    static LAZY_PAGES_INFO: RefCell<BTreeMap<LazyPage, Vec<u8>>> = RefCell::new(BTreeMap::new());
-    /// Page data, which has been in storage before current execution.
-    /// For each lazy page, which has been accessed.
-    static RELEASED_LAZY_PAGES: RefCell<BTreeMap<LazyPage, Option<PageBuf>>> = RefCell::new(BTreeMap::new());
+    /// Lazy pages context for current execution
+    static LAZY_PAGES_CONTEXT: RefCell<LazyPagesExecutionContext> = RefCell::new(Default::default());
 }
 
-pub fn save_lazy_pages_info(pages: Vec<u32>, prefix: Vec<u8>) {
-    LAZY_PAGES_INFO.with(|lazy_pages_info| {
-        let mut lazy_pages_info = lazy_pages_info.borrow_mut();
-        pages.into_iter().for_each(|p| {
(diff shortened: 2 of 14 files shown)


lazy-pages/src/lib.rs
```diff
@@ -20,68 +20,142 @@
...
+            .extend(pages.into_iter().map(PageNumber::from));
+        ctx.program_storage_prefix.replace(prefix);
     });
 }
 
-/// Returns vec of not-accessed wasm lazy pages
-pub fn available_pages() -> Vec<LazyPage> {
-    LAZY_PAGES_INFO.with(|lazy_pages_info| lazy_pages_info.borrow().keys().copied().collect())
+/// Returns lazy pages numbers (not accessed yet)
+pub fn get_lazy_pages() -> Vec<PageNumber> {
```
ark0f, line 121: Why was it renamed?

lazy-pages/src/lib.rs
```diff
@@ -20,68 +20,142 @@
...
+//! 4) After alloc we can set page as lazy, to identify wether page is changed after allocation.
+//! This means that we can skip page update in storage in case it wasnt changed after allocation.
+//! 5) Write some data in page but do not change it Data2 -> Data2.
+//! During this step signal handler writes Data2 as data for released page.
+//! 6) After execution we will have Data2 in page. And Data2 in released. So, nothing will be updated
+//! in storage. But program may have some significant data for next execution - so we have a bug.
+//! To avoid this we restrict double releasing.
+//! You can also check another cases in test: memory_access_cases.
+
+#![allow(useless_deprecated, deprecated)]
```
ark0f, line 43: > useless_deprecated

Why?

lazy-pages/src/lib.rs
```diff
@@ -20,68 +20,142 @@
...
 pub fn is_enabled() -> bool {
     LAZY_PAGES_ENABLED.with(|x| *x.borrow())
 }
 
+#[deprecated]
+/// Save lazy page storage key in global context
+pub fn set_lazy_page_info(page: PageNumber, key: &[u8]) {
+    LAZY_PAGES_CONTEXT.with(|ctx| {
+        ctx.borrow_mut().lazy_pages_info.insert(page, key.to_vec());
+        ctx.borrow_mut().lazy_pages.insert(page);
```
ark0f, line 150: Is it better to `borrow_mut` once to avoid unnecessary checks?

lazy-pages/src/lib.rs
```diff
@@ -94,18 +168,17 @@ pub unsafe fn init() -> bool {
...
-        return false;
-    }
-
-    if !WASM_MEM_BEGIN.with(|x| *x.borrow() == 0) {
-        log::error!("Wasm mem begin must be 0 before initialization");
-        return false;
-    }
+    let ctx_is_default = LAZY_PAGES_CONTEXT.with(|ctx| {
+        let ctx = ctx.borrow();
+        ctx.lazy_pages.is_empty()
```
ark0f, line 173:
```suggestion
        ctx == LazyPagesExecutionContext::default()
```

lazy-pages/src/sys.rs
```diff
@@ -15,39 +35,21 @@ cfg_if! {
...
 pub struct ExceptionInfo {
     /// Address where fault is occurred
     pub fault_addr: *const (),
 }
 
+/// Returns key which `page` has in storage.
+/// `prefix` is current program prefix in storage.
+fn page_key_in_storage(prefix: &Vec<u8>, page: PageNumber) -> Vec<u8> {
+    let mut key = Vec::with_capacity(prefix.len() + std::mem::size_of::<u32>());
+    key.extend(prefix.clone());
```
ark0f, line 48:
```suggestion
    key.extend(prefix);
```

lazy-pages/src/sys.rs
```diff
@@ -15,39 +35,21 @@ cfg_if! {
...
+    key.extend(page.0.to_le_bytes().to_vec());
```
ark0f, line 49:
```suggestion
    key.extend(page.0.to_le_bytes());
```

lazy-pages/src/lib.rs
```diff
@@ -20,68 +20,142 @@
 //! In runtime data for contract wasm memory pages can be loaded in lazy manner.
 //! All pages, which is supposed to be lazy, must be mprotected before contract execution.
 //! During execution data from storage is loaded for all pages, which has been accesed.
-//! See also `handle_sigsegv`.
+//! See also `user_signal_handler`.
```
ark0f, line 23:
```suggestion
//! See also `sys::user_signal_handler` in the source code.
```

Because this fn is private

lazy-pages/src/lib.rs
```diff
@@ -20,68 +20,142 @@
...
+    DoubleRelease(PageNumber),
+    #[display(fmt = "Protection error: {}", _0)]
+    #[from]
+    MemoryProtection(region::Error),
+}
 
-    /// Identify whether signal handler is set for current thread
-    static LAZY_PAGES_ENABLED: RefCell<bool> = RefCell::new(false);
+#[derive(Default)]
+pub(crate) struct LazyPagesExecutionContext {
```
ark0f, line 84: Maybe prefix `LazyPages` is redundant?

ark0f (changes requested):

lazy-pages/src/lib.rs
grishasobol: We mark code as deprecated, but use it because of old runtimes support

lazy-pages/src/lib.rs
grishasobol: This will create new object of lazy pages context

lazy-pages/src/lib.rs
grishasobol: old name is counterintuitively

lazy-pages/src/sys.rs
grishasobol: Done

lazy-pages/src/sys.rs
grishasobol: Done

lazy-pages/src/lib.rs
grishasobol: Done

lazy-pages/src/lib.rs
ark0f: Can you explain in more detail?

lazy-pages/src/lib.rs
ark0f: This is what `deprecated` for but I'm talking about `useless_deprecated` which ocurrs when `#[deprecated]` has no effect

lazy-pages/src/lib.rs
ark0f: It will but only once during the first initialization

lazy-pages/src/lib.rs
breathx: > This is what `deprecated` for but I'm talking about `useless_deprecated` which ocurrs when `#[deprecated]` has no effect

Runtime interface macro calls all fns under the hood, so it fails for clippy as usage of deprecated fns.
In @ukint-vs I offered to use this allowance for whole runtime-interface crate: that we could mark something as deprecated and still use it.

lazy-pages/src/sys.rs
breathx: @grishasobol, conversations on the PR __<ins>must</ins>__ be resolved by fixing commit or closed by the author if you discussed it

lazy-pages/src/sys.rs
```diff
@@ -67,21 +69,19 @@ pub unsafe fn user_signal_handler(info: ExceptionInfo) -> Result<(), Error> {
...
-        return Err(Error::UnknownMemory {
+        return Err(Error::SignalFromUnknownMemory {
             wasm_mem_begin,
             native_page,
         });
     }
 
     // First gear page which must be unprotected
-    let gear_page = PageNumber::from(((native_page - wasm_mem_begin) / gear_ps) as u32);
+    let gear_page = PageNumber(((native_page - wasm_mem_begin) / gear_ps) as u32);
```
breathx, line 84: Use saturating operations here

lazy-pages/src/sys.rs
```diff
@@ -91,7 +91,7 @@ pub unsafe fn user_signal_handler(info: ExceptionInfo) -> Result<(), Error> {
         (gear_page, 1usize, wasm_mem_begin + gear_page.offset())
     };
 
-    let accessed_page = PageNumber::from(((mem as usize - wasm_mem_begin) / gear_ps) as u32);
+    let accessed_page = PageNumber(((mem as usize - wasm_mem_begin) / gear_ps) as u32);
```
breathx, line 94: here and in other places for all lazy-pages dependent stuff as above

lazy-pages/src/sys.rs
```diff
@@ -105,24 +105,35 @@ pub unsafe fn user_signal_handler(info: ExceptionInfo) -> Result<(), Error> {
...
+            if !ctx.lazy_pages.remove(&page) {
+                return Err(Error::LazyPageNotExistForSignalAddr(mem, page));
+            }
 
-            let hash_key_in_storage = pages_info.remove(&page).ok_or(LazyPageError::UnknownInfo(page))?;
             let ptr = (unprot_addr as *mut u8).add(idx as usize * gear_ps);
             let buffer_as_slice = std::slice::from_raw_parts_mut(ptr, gear_ps);
 
-            let res = sp_io::storage::read(&hash_key_in_storage, buffer_as_slice, 0);
+            // TODO: simplify before release. Currently we must support here all old runtimes.
```
breathx, line 120: all todos should be covered with issues

breathx: ref it here

lazy-pages/src/sys.rs
```diff
@@ -105,24 +105,35 @@ pub unsafe fn user_signal_handler(info: ExceptionInfo) -> Result<(), Error> {
...
-            let page = LazyPage::from(gear_page) + idx;
+            let page = gear_page + idx.into();
+
+            if !ctx.lazy_pages.remove(&page) {
+                return Err(Error::LazyPageNotExistForSignalAddr(mem, page));
+            }
 
-            let hash_key_in_storage = pages_info.remove(&page).ok_or(LazyPageError::UnknownInfo(page))?;
             let ptr = (unprot_addr as *mut u8).add(idx as usize * gear_ps);
             let buffer_as_slice = std::slice::from_raw_parts_mut(ptr, gear_ps);
```
breathx, line 118: do all safety prerequisites correct here? should we extra check it?

lazy-pages/src/sys.rs
```diff
@@ -134,7 +145,14 @@ pub unsafe fn user_signal_handler(info: ExceptionInfo) -> Result<(), Error> {
...
                 .expect("Cannot panic here, because we create slice with PageBuf size");
-            page.release(page_buf)?;
+
+            if ctx
+                .released_lazy_pages
+                .insert(page, Some(page_buf))
+                .is_some()
+            {
+                return Err(Error::DoubleRelease(page));
+            }
```
breathx, line 155: the check may be preformed before we do a lots of calculations: as minimum, we can check if released lazy pages contain data under the key before we allocate vec at line 146

lazy-pages/src/sys.rs
```diff
@@ -134,7 +145,14 @@ pub unsafe fn user_signal_handler(info: ExceptionInfo) -> Result<(), Error> {
 
             let page_buf = PageBuf::new_from_vec(buffer_as_slice.to_vec())
                 .expect("Cannot panic here, because we create slice with PageBuf size");
```
breathx, line 147: Can we remove this expect? We have to follow as safe as it could be style of code.

I want to make one more proposal: could we not to perform here conversion to vec and create page buf directly from slice? 
Such functionality should be added in core also:

Please, check the [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bbb03e4b981be74ab8200c1d40be332f)

Isn't it more optimal solution? @grishasobol @ark0f @ukint-vs

pallets/gear-debug/src/tests.rs
```diff
@@ -284,6 +284,7 @@ fn get_last_message_id() -> MessageId {
     }
 }
 
+#[cfg(feature = "lazy-pages")]
```
breathx, line 287: If I remember right, you said that no behaviour should be changed with the feature, and all tests must run out of this?

runtime-interface/src/lib.rs
```diff
@@ -134,7 +137,7 @@ fn mprotect_pages_slice(
 
     let mut count = 1;
     for &page in pages.iter().skip(1) {
-        if start + count == page {
+        if start + count.into() == page {
```
breathx, line 140: saturating also

breathx (changes requested):

lazy-pages/src/lib.rs
ark0f: > One more solution - wrap your stuff into some method and code become more clear

But you still need to not forget compare new fields

lazy-pages/src/sys.rs
ark0f: - Conversion from slice

From [`impl From<&[T]> for Box<[T]>`](https://doc.rust-lang.org/nightly/std/boxed/struct.Box.html#impl-From%3C%26%27_%20%5BT%5D%3E) docs:

> This conversion allocates on the heap and performs a copy of slice.

Inside it uses `RawVec::with_capacity` which 

> Creates a `RawVec` (on the system heap) with exactly the capacity and alignment requirements

- Conversion from `Vec`

From [`Vec::into_boxed_slice`] docs:

> Note that this will drop any excess capacity.

From [`impl From<Vec<T>> for Box<[T]>`](https://doc.rust-lang.org/nightly/std/boxed/struct.Box.html#impl-From%3CVec%3CT%2C%20A%3E%3E) (which calls [`Vec::into_boxed_slice`]) docs:

> If v has excess capacity, its items will be moved into a newly-allocated buffer with exactly the right capacity.

Documentation is misleading. Actually the second doc is the truth because [`Vec::into_boxed_slice`] uses [`Vec::shrink_to_fit`] inside.

- Conversion from slice to `Vec` (e.g. [`to_vec`])

It uses [`Vec::with_capacity`] inside which:

> This method is allowed to allocate for more elements than capacity

So, @breathx, it is more optimal solution for now but I should mention that these conversions have no guarantees. Even the lastest [`Vec::with_capacity`]'s quote exists only in beta and nightly (not in stable yet)

[`to_vec`]: https://doc.rust-lang.org/nightly/std/primitive.slice.html#method.to_vec
[`Vec::into_boxed_slice`]: https://doc.rust-lang.org/nightly/std/vec/struct.Vec.html#method.into_boxed_slice
[`Vec::shrink_to_fit`]: https://doc.rust-lang.org/nightly/std/vec/struct.Vec.html#method.shrink_to_fit
[`Vec::with_capacity`]: https://doc.rust-lang.org/nightly/std/vec/struct.Vec.html#method.with_capacity

lazy-pages/src/lib.rs
grishasobol: Is now deprecated - so, no reasons to discuss this

lazy-pages/src/lib.rs
grishasobol: Ok done as more safe

lazy-pages/src/sys.rs
grishasobol: @breathx  absolutely disagree

lazy-pages/src/lib.rs
grishasobol: No

lazy-pages/src/lib.rs
```diff
@@ -20,68 +20,142 @@
...
+        ctx.program_storage_prefix.replace(prefix);
     });
 }
 
-/// Returns vec of not-accessed wasm lazy pages
-pub fn available_pages() -> Vec<LazyPage> {
-    LAZY_PAGES_INFO.with(|lazy_pages_info| lazy_pages_info.borrow().keys().copied().collect())
+/// Returns lazy pages numbers (not accessed yet)
+pub fn get_lazy_pages() -> Vec<PageNumber> {
+    LAZY_PAGES_CONTEXT.with(|ctx| ctx.borrow().lazy_pages.iter().copied().collect())
```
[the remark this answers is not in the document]
grishasobol: This is deprecated now

lazy-pages/src/lib.rs
```diff
@@ -20,68 +20,142 @@
...
 }
 
 /// Set current wasm memory begin addr
-pub fn set_wasm_mem_begin_addr(wasm_mem_begin: HostPointer) {
-    WASM_MEM_BEGIN.with(|x| *x.borrow_mut() = wasm_mem_begin);
+pub fn set_wasm_mem_begin_addr(wasm_mem_addr: HostPointer) {
+    LAZY_PAGES_CONTEXT.with(|ctx| ctx.borrow_mut().wasm_mem_addr.replace(wasm_mem_addr));
```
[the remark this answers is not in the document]
grishasobol: Depends on the compiler optimisations. Don't actually now but seems like insert has more opportunity to be optimised , so change to it

lazy-pages/src/sys.rs
grishasobol: Just 5 strings above we checks that native page >= wasm mem begin

lazy-pages/src/sys.rs
grishasobol: The same - mathematically `mem` cannot be less then `wasm_mem_begin `

lazy-pages/src/sys.rs
grishasobol: The only one possibility to throw error is different sizes. There is some misleading with to_vec yes, but as I can see in source code it creates with given slice size.

pallets/gear-debug/src/tests.rs
grishasobol: Yep, remove this

runtime-interface/src/lib.rs
grishasobol: this is deprecated now

lazy-pages/src/sys.rs
grishasobol: Ok done

lazy-pages/src/sys.rs
grishasobol: I cannot see any reasons to do this, because this errors cannot appear in normal cases, only if we have bugs in node or runtime. So, in case they will appear we must fix them immediately

lazy-pages/src/sys.rs
grishasobol: It's not rust safe) but it is used inside signal handler (which is already very unsafe), which must works with raw memory)

ukint-vs:
the fact that the number of samples has decreased does not mean that the performance has grown, `perf` cannot sample signal handler. Therefore it is impossible to measure what is happening inside the handler.

ukint-vs:
Please provide a side-by-side benchmark (`gr_read` or `gr_reply_commit`) comparison of this branch with master.

runtime-interface/src/lib.rs
```diff
@@ -17,6 +17,8 @@
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
 //! Runtime interface for gear node
+//!
+//! TODO: remove all deprecated code before release (issue #1147)
```
ark0f, line 21: TODO shouldn't be in doc string

ukint-vs:
> > 
> 
> What do you mean `side by side` ?

this branch:  `+ r    58.08 µs`
master: `+ r     57.9 µs`

pallets/gear/src/tests.rs
```diff
@@ -856,36 +856,40 @@ fn lazy_pages() {
         // Dirty hack: lazy pages info is stored in thread local static variables,
         // so after contract execution lazy-pages information
         // remains correct and we can use it here.
-        let lazy_pages: BTreeSet<PageNumber> = gear_ri::gear_ri::get_lazy_pages_numbers()
-            .iter()
-            .map(|p| PageNumber(*p))
-            .collect();
+        // let lazy_pages: BTreeSet<PageNumber> = gear_ri::gear_ri::get_lazy_pages_numbers()
```
ark0f, line 859: Why is it commented out?

ark0f (changes requested):

lazy-pages/src/lib.rs
```diff
@@ -20,68 +20,187 @@
...
-    RELEASED_LAZY_PAGES.with(|x| x.borrow_mut().clear());
-    WASM_MEM_BEGIN.with(|x| *x.borrow_mut() = 0);
+pub fn reset_context() {
+    LAZY_PAGES_CONTEXT.with(|ctx| *ctx.borrow_mut() = Default::default());
 }
 
 /// Returns vec of lazy pages which has been accessed
-pub fn released_pages() -> Vec<LazyPage> {
-    RELEASED_LAZY_PAGES.with(|x| x.borrow().keys().copied().collect())
+pub fn get_released_pages() -> Vec<PageNumber> {
```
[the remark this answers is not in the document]
grishasobol: because with this prefix it's much more clear what this function do

pallets/gear/src/tests.rs
grishasobol: Yep forget to remove this

lazy-pages/src/lib.rs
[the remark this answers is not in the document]
ark0f: Getters are usually without `get_` in Rust

grishasobol:
> > > 
> > 
> > 
> > What do you mean `side by side` ?
> 
> this branch: `+ r 58.08 µs` master: `+ r 57.9 µs`

Done, see in head

lazy-pages/src/lib.rs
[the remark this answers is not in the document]
grishasobol: As I can see in substrate (very big rust project) there are a lot of examples where getters begin with `get_`

runtime-interface/src/lib.rs
grishasobol: Done

lazy-pages/src/lib.rs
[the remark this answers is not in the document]
grishasobol: Ok, but we also use `get_` prefix everywhere in our project, so, if you want to remove it, discuss this with all team - not here

grishasobol:
> > the fact that the number of samples has decreased does not mean that the performance has grown, `perf` cannot sample signal handler. Therefore it is impossible to measure what is happening inside the handler.
> 
> Will `--all-kernel` option help us?

I can see the difference only in syscalls, which begin to be profiled

ukint-vs:
> > > the fact that the number of samples has decreased does not mean that the performance has grown, `perf` cannot sample signal handler. Therefore it is impossible to measure what is happening inside the handler.
> > 
> > 
> > Will `--all-kernel` option help us?
> 
> I can see the difference only in syscalls, which begin to be profiled

Good to know. Thx @ark0f

lazy-pages/src/lib.rs
```diff
@@ -20,68 +20,187 @@
...
+#[deprecated]
+pub fn get_lazy_pages_numbers() -> Vec<PageNumber> {
+    LAZY_PAGES_CONTEXT.with(|ctx| ctx.borrow().lazy_pages_info.keys().copied().collect())
 }
 
-/// Returns vec of not-accessed wasm lazy pages
-pub fn available_pages() -> Vec<LazyPage> {
-    LAZY_PAGES_INFO.with(|lazy_pages_info| lazy_pages_info.borrow().keys().copied().collect())
+/// Returns current wasm mem buffer pointer, if it's set
+pub fn get_wasm_mem_addr() -> Option<HostPointer> {
```
breathx, line 115: I would also remove all `get_` from fns

lazy-pages/src/lib.rs
```diff
@@ -20,68 +20,187 @@
...
 
 /// Returns whether lazy pages env is enabled
 pub fn is_enabled() -> bool {
     LAZY_PAGES_ENABLED.with(|x| *x.borrow())
 }
 
+#[deprecated]
+/// Set storage `key` for `page` in global context
+pub fn set_lazy_page_info(page: PageNumber, key: &[u8]) {
+    LAZY_PAGES_CONTEXT.with(|ctx| {
```
breathx, line 156: in general, we can just leave here empty fn due to the fact its already deprecated

lazy-pages/src/sys.rs
breathx: Nvm, I meant requirements to avoid ub from [here](https://doc.rust-lang.org/std/slice/fn.from_raw_parts_mut.html)

lazy-pages/src/sys.rs
```diff
@@ -105,24 +105,31 @@ pub unsafe fn user_signal_handler(info: ExceptionInfo) -> Result<(), Error> {
...
-            let page = LazyPage::from(gear_page) + idx;
+            let page = gear_page + idx.into();
 
-            let hash_key_in_storage = pages_info.remove(&page).ok_or(LazyPageError::UnknownInfo(page))?;
             let ptr = (unprot_addr as *mut u8).add(idx as usize * gear_ps);
             let buffer_as_slice = std::slice::from_raw_parts_mut(ptr, gear_ps);
 
-            let res = sp_io::storage::read(&hash_key_in_storage, buffer_as_slice, 0);
+            // TODO: simplify before release (issue #1147). Currently we must support here all old runtimes.
+            // In new runtimes we have to clalc page key from program page key.
```
breathx, line 117:
```suggestion
            // In new runtimes we have to calc page key from program page key.
```

lazy-pages/src/sys.rs
breathx: Yeah, but even if this is, for example, impossible moving code 5 lines upper is not a big deal in order to optimise it, or just beautify the code

runtime-interface/src/deprecated.rs
```diff
@@ -19,9 +19,14 @@
 //! Deprecated backend for runtime interface.
 //! TODO: remove before release
```
breathx, line 20:
```suggestion
//! TODO: remove before release (issue #1147).
```

breathx (approved):

lazy-pages/src/sys.rs
grishasobol: Because this part of code uses `page_buf ` we cannot move it up. So we must write separate useless code before.

common/src/lazy_pages.rs
```diff
@@ -46,11 +46,11 @@ impl From<RIError> for Error {
...
-    };
-    gear_ri::mprotect_lazy_pages(wasm_mem_addr, protect)
+    if mem.get_buffer_host_addr().is_none() {
+        return Ok(());
+    }
+
+    gear_ri::mprotect_lazy_pages(protect)
         .map_err(Into::into)
         .map_err(|e| {
             log::error!("{} (it's better to stop node now)", e);
```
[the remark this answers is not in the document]
grishasobol: Make issue: https://github.com/gear-tech/gear/issues/1165

ark0f (approved):

runtime-interface/src/deprecated.rs
ark0f, line 20:
```suggestion

// TODO: remove before release
```

common/src/lazy_pages.rs
```diff
@@ -157,15 +150,14 @@ pub fn protect_lazy_pages_and_update_wasm_mem_addr(
...
 
-/// Returns list of current lazy pages numbers
-pub fn get_lazy_pages_numbers() -> Vec<PageNumber> {
-    gear_ri::get_lazy_pages_numbers()
-        .iter()
-        .map(|p| PageNumber(*p))
-        .collect()
+    let new_mem_size = mem.size();
+    if new_mem_size > old_mem_size {
+        let size = (new_mem_size.0 + 1) * WasmPageNumber::size() as u32;
```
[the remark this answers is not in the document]
grishasobol: Yes,  because wasm memory cannot be bigger than u32 address has.

lazy-pages/src/lib.rs
grishasobol: No we cannot because it's used in old runtimes

lazy-pages/src/lib.rs
grishasobol: Ok, you can remove all in separate PR

PR #1134, opened 2022-07-02
Introduce common configs for processor

### LINKED ISSUE #1111: Design common config type
### File Location(s)

`core-processor`
`pallets/gear`

### Proposal

Atm we pass to much arguments to `core_processor::process` function. We need to figure out all required types/params to be thrown as config: after that we will be able to give it only two params as config arguments: stable for whole block (e.g. block number, ED) and setting for concrete message execution

Stable for whole block settings may be constructed only once when `manager: JournalHandler` creates, and used by ref afterward in process_queue.

ark0f:
Resolves #1111.

### DIFF
--- a/core-processor/src/configs.rs
+++ b/core-processor/src/configs.rs
@@ -18,9 +18,12 @@
 
 //! Configurations.
 
+use crate::common::ExecutableActor;
 use alloc::collections::BTreeSet;
 use codec::{Decode, Encode};
-use gear_core::{costs::HostFnWeights, memory::WasmPageNumber};
+use gear_core::{
+    costs::HostFnWeights, ids::ProgramId, memory::WasmPageNumber, message::IncomingDispatch,
+};
 
 const MAX_WASM_PAGES: u32 = 512;
 const INIT_COST: u64 = 5000;
@@ -124,4 +127,39 @@ impl ExecutionSettings {
     pub fn load_page_cost(&self) -> u64 {
         self.allocations_config.load_page_cost
     }
+}
+
+/// Stable parameters for the whole block across processing runs
+#[derive(Clone)]
+pub struct BlockConfig {
+    /// Block info
+    pub block_info: BlockInfo,
+    /// Allocations config
+    pub allocations_config: AllocationsConfig,
+    /// Existential deposit
+    pub existential_deposit: u128,
+    /// Gas allowance
+    pub gas_allowance: u64,
+    /// Outgoing limit
+    pub outgoing_limit: u32,
+    /// Host function weights
+    pub host_fn_weights: HostFnWeights,
+    /// Forbidden functions
+    pub forbidden_funcs: BTreeSet<&'static str>,
+    /// Mailbox threshold
+    pub mailbox_threshold: u64,
+}
+
+/// Unstable parameters for message execution across processing runs
+#[derive(Clone)]
+pub struct MessageExecutionConfig {
+    /// Optional executable actor
+    pub executable_actor: Option<ExecutableActor>,
+    /// Incoming dispatch
+    pub dispatch: IncomingDispatch,
+    /// The external origin for a key
+    pub origin: ProgramId,
+    // TODO: Temporary here for non-executable case. Should be inside executable actor, renamed to Actor.
+    /// Destination program
+    pub program_id: ProgramId,
 }
--- a/core-processor/src/configs.rs
+++ b/core-processor/src/configs.rs
@@ -129,35 +129,35 @@ impl ExecutionSettings {
     }
 }
 
-/// Stable parameters for the whole block across processing runs
+/// Stable parameters for the whole block across processing runs.
 // TODO: remove Default derive in favor of `BlockConfig::test`
 #[derive(Clone, Default)]
 pub struct BlockConfig {
-    /// Block info
+    /// Block info.
     pub block_info: BlockInfo,
-    /// Allocations config
+    /// Allocations config.
     pub allocations_config: AllocationsConfig,
-    /// Existential deposit
+    /// Existential deposit.
     pub existential_deposit: u128,
-    /// Outgoing limit
+    /// Outgoing limit.
     pub outgoing_limit: u32,
-    /// Host function weights
+    /// Host function weights.
     pub host_fn_weights: HostFnWeights,
-    /// Forbidden functions
+    /// Forbidden functions.
     pub forbidden_funcs: BTreeSet<&'static str>,
-    /// Mailbox threshold
+    /// Mailbox threshold.
     pub mailbox_threshold: u64,
 }
 
-/// Unstable parameters for message execution across processing runs
+/// Unstable parameters for message execution across processing runs.
 #[derive(Clone)]
 pub struct MessageExecutionContext {
-    /// Executable actor
+    /// Executable actor.
     pub actor: Actor,
-    /// Incoming dispatch
+    /// Incoming dispatch.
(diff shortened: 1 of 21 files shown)


core-processor/src/configs.rs
```diff
@@ -18,15 +18,21 @@
...
+};
 
 const MAX_WASM_PAGES: u32 = 512;
 const INIT_COST: u64 = 5000;
 const ALLOC_COST: u64 = 10000;
 const MEM_GROW_COST: u64 = 10000;
 const LOAD_PAGE_COST: u64 = 3000;
+const EXISTENTIAL_DEPOSIT: u128 = 500;
+const OUTGOING_LIMIT: u32 = 1024;
+const MAILBOX_THRESHOLD: u64 = 3000;
```
breathx, line 35: We should avoid such consts here. These params should be a runtime constants or mutable params there

core-processor/src/configs.rs
```diff
@@ -125,3 +131,53 @@ impl ExecutionSettings {
...
+#[derive(Clone)]
+pub struct BlockConfig {
+    /// Block info
+    pub block_info: BlockInfo,
+    /// Allocations config
+    pub allocations_config: AllocationsConfig,
+    /// Existential deposit
+    pub existential_deposit: u128,
+    /// Gas allowance
+    pub gas_allowance: u64,
```
breathx, line 145: this is not stable param

core-processor/src/configs.rs
```diff
@@ -125,3 +131,53 @@ impl ExecutionSettings {
...
+pub struct MessageExecutionConfig {
+    /// Optional executable actor
+    pub executable_actor: Option<ExecutableActor>,
+    /// Incoming dispatch
+    pub dispatch: IncomingDispatch,
+    /// The external origin for a key
+    pub origin: ProgramId,
+    // TODO: Temporary here for non-executable case. Should be inside executable actor, renamed to Actor.
+    /// Destination program
+    pub program_id: ProgramId,
```
breathx, line 182: resolve this todo, please, append id into ExecutableActor
in general, we should pass here just `Actor`, checking if it executable afterward

core-processor/src/configs.rs
```diff
@@ -125,3 +131,53 @@ impl ExecutionSettings {
...
+            host_fn_weights: Default::default(),
+            forbidden_funcs: Default::default(),
+            mailbox_threshold: MAILBOX_THRESHOLD,
+        }
+    }
+}
+
+/// Unstable parameters for message execution across processing runs
+#[derive(Clone)]
+pub struct MessageExecutionConfig {
```
breathx, line 173: I'd also rename this to execution context/data
than we'll  have more logical:
`fn process(to_execute: ExecutionData, config: &BlockConfig)`

breathx (changes requested):
I really want to see some getter-trait for such params, instead of concrete structs, how do you feel about it?

ark0f:
> We don't always need to have concrete objects if we only need immutable params from them:

Looks like overengineering to me. Such trait doesn't change something in general

pallets/gear/src/lib.rs
```diff
@@ -742,35 +754,21 @@ pub mod pallet {
...
                     )
                 } else {
                     core_processor::process::<Ext, SandboxEnvironment<_>>(
-                        Some(actor),
-                        queued_dispatch.into_incoming(gas_limit),
-                        block_info,
-                        allocations_config,
-                        existential_deposit,
-                        ProgramId::from_origin(source),
-                        actor_id,
```
breathx, line 768: Inserting some Processor entity will also allow us to remove copy-base code, cause it already will include any generics inside it

breathx (changes requested):

core-processor/src/processor.rs
```diff
@@ -275,19 +258,20 @@ fn process_success(
...
+    BlockConfig {
+        block_info,
+        allocations_config,
+        existential_deposit,
+        gas_allowance,
+        outgoing_limit,
+        host_fn_weights,
+        forbidden_funcs,
+        mailbox_threshold,
+    }: BlockConfig,
```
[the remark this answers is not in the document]
ark0f: What's the diff?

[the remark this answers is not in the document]
breathx: > What's the diff?

There is no actual diff, but my sense of beauty says me, that when you see just 3 args in fn args, it become more readable

core-processor/src/executor.rs
```diff
@@ -203,9 +203,9 @@ fn get_pages_to_be_updated<A: ProcessorExt>(
 
 /// Execute wasm with dispatch and return dispatch result.
 pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environment<A>>(
-    actor: ExecutableActor,
+    data: ExecutableActorData,
     dispatch: IncomingDispatch,
-    context: ExecutionContext,
+    context: WasmExecutionContext,
     settings: ExecutionSettings,
     msg_ctx_settings: ContextSettings,
```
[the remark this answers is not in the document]
ark0f: No because `data` is optional

gear-test/src/proc.rs
```diff
@@ -100,24 +100,28 @@ where
...
+                balance: 0,
+                pages_data: Default::default(),
+            }),
+            destination_program: program_id,
+        },
+        dispatch: message.into(),
+        origin: Default::default(),
+    };
+
+    let journal = core_processor::process::<Ext, E>(&block_config, message_execution_context);
```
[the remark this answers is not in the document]
ark0f: Making 1-function wrapper looks redundant to me

pallets/gear/src/lib.rs
breathx:
```rust
let processor_1 = Processor::<A, B>::new();
let processor_2 = Processor::<C, D>::new();
```

or even hide it under the hood by features, but that's not the best solution

core-processor/src/executor.rs
```diff
@@ -203,9 +203,10 @@ fn get_pages_to_be_updated<A: ProcessorExt>(
 
 /// Execute wasm with dispatch and return dispatch result.
 pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environment<A>>(
```
breathx, line 205: One thing forgot to notice:
this:
https://github.com/gear-tech/gear/blob/973250632d8b4163c85c6c10dc264b0c43f044cd/core-processor/src/executor.rs#L278-L294
should also be simplified

core-processor/src/executor.rs
```diff
@@ -275,23 +275,25 @@ pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environ
...
         program_id,
-        Default::default(),
-        settings.host_fn_weights,
-        settings.forbidden_funcs,
-        settings.mailbox_threshold,
-    );
+        program_candidates_data: Default::default(),
+        host_fn_weights: settings.host_fn_weights,
+        forbidden_funcs: settings.forbidden_funcs,
+        mailbox_threshold: settings.mailbox_threshold,
```
breathx, line 292: seems that we can pass here just settings and access it under the hood

ark0f: We could but there are already too many nesting levels. You can see something like `self.inner.context.gas_counter` in `LazyPagesExt`

breathx: ok, so skip it for some time

PR #1135, opened 2022-07-03
Introduce scheduling mechanism

### LINKED ISSUE #646: Rent charging
### Problem to Solve

Updated representation of #24 .

We have to charge gas/value rent for activities like:
- [x] message keeps in wait-list (already done by @ekovalev)
- [ ] code upload
- [ ] program keeps in storage (I think should be payed on init for some future time. In case of reservation end, disable program changing it's status)
- [ ] Expand wait-list charging after #496 for charging depend of stored data size
- [ ] message keeps in mailbox (partially relates to #642)
_... this list can be updated with new features added_

### Possible Solution

Need to expand offchain worker functionality for all of this cases.

### LINKED ISSUE #969: Move off-chain-worker logic on-chain
### Problem to Solve

Remove ocw appearance to calculate fees without extrinsics.

### Possible Solution

Let the usage pallet like 5 or 10 % of block to go through fee queue and charge it (additional logic would be applied, e.g. #349) before queue processing.

### Notes

_No response_

breathx:
## Main goal
Resolves #969.
Part of #646.

### Notes to make review easier
- We had several iterator wrappers for being able to modify result of initial iterator's output: now it's replaced with a single struct with generics for modifications: `IteratorWrap`
- Scheduler trait introduced: we still follow our common primitives strategy. Here's storing mechanism for tasks with their block numbers (when they should be done) and block numbers which still contains some tasks.
- Introduced new iterable trait, which allows to iterate over second keys by given first key of double map
- Pallet scheduler introduced. This will be used as associated type Scheduler for gear's pallet.
- `manager` module separated into dir: by default ext manager fns + implementations of task handler (for scheduler) + implementation of journal handler (for processing)
- process queue semantic changed: it now decreases gas allowance, instead of returning weight + also gets ext manager as param t consume at the end
- process tasks added into on_idle gear hook
- scheduling contains only functionality for waking obsolete messages: other will be added in separate PRs in order to resolve #646 by parts.
- pallet usage deleted, off chain worker deleted - everything is now on scheduling mechanism, which used by runtime and guaranteed to be safety wout any considers 

### Release notes:
Now we have mechanism to do some planned actions on some specific block: for example, wake message at block N or pause program at block M. This will be used for rent charging and implementing time-locking primitives inside the programs. Atm it implements only waitlist logic.

### Usage
Imagine, that we charge 100 gas per block for holding in wl. And we insert message in block N.

#### How it was before:
Message with gas GAS inserted to wl at block N. We don't precalculate when it should be deleted from it. We run off chain worker for each 5 blocks to iterate over all messages (with upper limit, of course) and spending value from them to block producer. If no gas left, we remove message and send error reply.

#### How it works now:
Message with GAS wants to be inserted in wl at block N. We calculate, that GAS / 100 = SOME_AMOUNT of blocks. So the message should be deleted at block N + SOME_AMOUNT (in fact, we take some offset here, so message will be removed at N + SOME_AMOUNT - small_offset).
If message wakes at some block < N + SOME_AMOUNT, then we spend gas to block producer = 100 * holding_duration, otherwise, we remove message inside on_idle hook (before processing queue) in this block N + SOME_AMOUNT and spend for holding durations.

#### Pros
This way is more efficient due to less storage interactions, more objectivism and safety, due to absence some different storages or actors for charging action. And also we have understanding what will happen at some specific block and can suppose the whole lifecycle of the message. 
#### Cons
Charging rent isn't distributed between block producers, but this isn't critical, due to out future migration on staking pallet with epochs concepts.


common/src/scheduler/scope.rs
```diff
@@ -0,0 +1,163 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use crate::storage::{CountedByKey, DoubleMapStorage, EmptyCallback, KeyIterableByKeyMap};
+use core::marker::PhantomData;
+
+/// Represents tasks managing logic.
+pub trait TasksScope {
```
ekovalev, line 23: Terminology: in my opinion `Scope`is not the best term to refer to a set of items (like tasks in this case). I would use something like `Registry`, `Pool` or `Set` for that matter

breathx: Agree here. Will rename it to `TaskPool`

pallets/gear/src/lib.rs
```diff
@@ -855,12 +877,113 @@ pub mod pallet {
...
+                    GasAllowanceOf::<T>::decrease(T::DbWeight::get().writes(1));
+                    set.insert(bn);
+                    (set, false)
+                })
+                .unwrap_or_else(|| {
+                    GasAllowanceOf::<T>::decrease(T::DbWeight::get().reads(1));
+                    ([bn].into(), true)
+                });
+
+            // Where we were had to stop processing due to zeroed gas allowance.
```
gshep, line 903: misprint? `When we had to ...`

pallets/gear/src/lib.rs
```diff
@@ -418,7 +419,28 @@ pub mod pallet {
...
             GasAllowanceOf::<T>::put(remaining_weight);
 
-            Self::process_queue()
+            // Ext manager creation.
+            // It will be processing messages execution results following its `JournalHandler` trait implementation.
+            // It also will handle delayed tasks following `TasksHandler`.
+            let mut ext_manager = Default::default();
+
+            // Processing regular and delayed tasks.
+            Self::process_tasks(&mut ext_manager);
```
gshep, line 428: I would like to discuss the function in more detail. As I thought previously it's intended to bound the available weight to process such kind of tasks, for example for 5-10% of the entire block weight.

If I understand the function's logic correctly it will execute till all available gas allowance is spent in the case of too much tasks there.

Common pattern used for similar logic is to precalculate some value called (weight/gas) budget that should be 5-10% of the weight for our needs and pass the budget to the function. The procedure itslef has to check this budget before any calculation it performs. What would you say about this approach?

pallets/gear-scheduler/src/lib.rs
```diff
@@ -0,0 +1,253 @@
...
+
+    // Below goes callbacks, used for task scope algorithm.
+    //
+    // Note, that they are public like storage wrappers
+    // only to be able use as public trait's generics.
+
+    // ----
+
+    /// Callback function for success `add` and `delete` actions.
+    pub struct OnAffect<T: crate::Config>(PhantomData<T>);
```
ekovalev, line 167: `OnAffect` sounds a bit over the top to me. Isn't the conventional `OnChange` more appropriate here?

common/src/scheduler/mod.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// Storing costs per block.
+    type CostsPerBlock: SchedulingCostsPerBlock<BlockNumber = Self::BlockNumber>;
+
+    /// Block numbers, which have already passed,
+    /// but still contain tasks to deal with.
+    ///
+    /// Used for checking if scheduler is able to process
+    /// current block aimed tasks, or there are some
+    /// incomplete job from previous blocks.
+    type MissedBlocks: ValueStorage<Value = Self::MissedBlocksCollection>;
```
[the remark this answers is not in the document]
breathx: In general, this block wasn't skipped - they have been processed, but stopped at some tasks, because there is no gas allowance left. So we more "miss" some tasks from the block, neither "skip" them

ekovalev (approved):
The `Scheduler` is a great and long awaited feature - not so much in context of charging fee for using resources but in the first place for enabling delayed messages processing.

As far as using this mechanism for charging fees (in place of previously used offchain workers), I guess we will need to benchmark it on some real workload and see in which case validators earn more: using block time for monitoring network resources usage or processing messages from the queue. Having this information at hand will help to make more informed decision and opt for the best strategy.

But generally very well done!

pallets/gear/src/lib.rs
```diff
@@ -198,14 +198,22 @@ pub mod pallet {
...
             Key = MessageId,
             Balance = u64,
             Error = DispatchError,
         >;
 
-        type BlockLimiter: common::BlockLimiter<Balance = u64>;
+        /// Block limits.
+        type BlockLimiter: BlockLimiter<Balance = u64>;
+
+        /// Scheduler.
```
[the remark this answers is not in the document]
breathx: Lets leave them as it is, because more informative docs are in common crate, but here in-place its like placeholder. Anyway before publish on crates io we will walk through our docs with proofread

PR #1137, opened 2022-07-04
pin to nightly-2022-06-30

gshep:
@gear-tech/dev


breathx:
that doesn't work like this, cause if you install some specific nightly dependency, you need to call it in scripts
like `cargo +nightly-2022-06-30 anything_you_want`

breathx:
Known workaround for me is to install latest nightly, install extra version and replace first nightly source files with the second one's content

breathx (approved):

breathx:
Placing todo with reverting this stuff would be good, but nvm, we will just pick of that commit once issue fixed

PR #1139, opened 2022-07-04
Add some love to error/debug logs

NikVolf:


gshep (approved):

core/src/memory.rs
```diff
@@ -330,7 +330,7 @@ impl AllocationsContext {
 
         while found < pages {
             if candidate + pages > self.max_pages {
-                log::debug!(
+                log::error!(
```
breathx, line 333: Why this should be not `debug`?  It's like user-space stuff, which doesn't show that something in platform goes wrong

breathx (approved):

ukint-vs (approved):

PR #1140, opened 2022-07-04
Core memory module: linear alloc and saturating ops

breathx:


core/src/memory.rs
```diff
@@ -158,14 +158,14 @@ impl PageNumber {
 impl core::ops::Add<PageNumber> for PageNumber {
     type Output = Self;
     fn add(self, other: Self) -> Self {
-        Self(self.0 + other.0)
+        Self(self.0.saturating_add(other.0))
```
NikVolf, line 161: `impl Add` which does `saturating_add` is really counter-intuitive and should be avoided

ukint-vs (approved):

PR #1158, opened 2022-07-06
Expand api for `create_program` syscall

### LINKED ISSUE #1065: Expand api for `create_program` syscall
### Problem to Solve

Make create programs sys call return not only actor id, but the message id also.
This will allow us to create async fns with this functionality. Provide them.

Also need to provide encodable variants of these fns

Don't forget about `ProgramGenerator`, once #973 merged.

### Possible Solution

Make minor changes in backend, apply changes to gcore and expand gstd functionalities.

dustyNdumb:
Resolves #1065  .

- Add ability to return `init_message_id` from `create_program` syscall
- Add async variants of `create_program` syscall
-

@reviewer-or-team


gstd/src/msg/async.rs
```diff
@@ -159,3 +159,99 @@ pub fn send_bytes_and_wait_for_reply<T: AsRef<[u8]>>(
...
 }
+
+/// Create a program and wait for reply.
+///
+/// This function works similarly to
+/// [`create_program_wbytes_and_wait_for_reply`], with one difference - it takes
+/// the structure in, then encodes it and sends it in bytes. The program will be
+/// interrupted (waiting for a reply) if an `.await` has been called on the
+/// `CodecMessageFuture` object returned by the function.
+#[deprecated(note = "please use `gstd::proc::create_program` instead")]
```
breathx, line 170:
```suggestion
```
Why?
and in other places

gstd/src/msg/async.rs
```diff
@@ -159,3 +159,99 @@ pub fn send_bytes_and_wait_for_reply<T: AsRef<[u8]>>(
...
+pub fn create_program_and_wait_for_reply<E: Encode, D: Decode>(
```
breathx, line 171:
```suggestion
pub fn create_program_for_reply<E: Encode, D: Decode>(
```
We now follow naming wouth `and_wait`

dustyNdumb:
Found 2 things that are confusing me and look like bugs:
1. Different `gtest` and `rtest` acting: if we create 2 programs (in one handle) with different code or salt they will generate 2 different pairs (one pair for `gtest` and one for `rtest`) of `ActorId` which is unacceptable for yaml tests. If we generate 2 programs (in different handles) with same salt and code (like in `create_program` example) we will have same `ActorId` in each of 4 cases but `gtest` will fail (only him) because of the second inition of the program
2.  Looks like `CodecMessageFuture` is working with mistakes. I tried to use it to send `String` from program to program. When I decoded result there were an extra trash char at the beginning of the string

SabaunT:
temporary closed

PR #1162, opened 2022-07-07
Don't charge for memory pages second time during a block execution

### LINKED ISSUE #1088: Charge for page load optionally
### Problem to Solve

At the moment, while processing queue (see `pallets/gear/fn process_queue`) we charge gas for loading programs pages every time, but this may be optimised due to overlay existence

### Possible Solution

To optimise with overlay:
Store program ids in ext_manager, which already were executed with current ext_manager instance: this means that pages were already read and/or updated, so the next read from db (overlay) is free.

### Notes

_No response_

gshep:
Resolves #1088 

- don't execute functions without exports
- introduce additional gas value charged for the next time of loading memory pages during a block execution

@gear-tech/dev


core-processor/src/configs.rs
```diff
@@ -53,6 +54,8 @@ pub struct AllocationsConfig {
     pub mem_grow_cost: u64,
     /// Load page cost.
     pub load_page_cost: u64,
+    /// Second load page cost.
```
grishasobol, line 57: So, if the name is `second_` then we should explain in comment that it's not second only

core-processor/src/executor.rs
```diff
@@ -93,9 +71,36 @@ fn make_checks_and_charge_gas_for_pages<'a>(
             return Err(ExecutionErrorReason::InitialMemoryGasExceeded);
         }
 
-        static_pages
+        return Ok(static_pages);
```
grishasobol, line 74: Comment please

core-processor/src/executor.rs
```diff
@@ -253,11 +260,33 @@ pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environ
         }
     };
 
+    let is_initial = program.get_allocations().is_empty();
+
+    let exports = program.code().exports();
```
grishasobol, line 265: Please check what can we do in core backend sandbox with exports then

grishasobol (approved):

grishasobol:
FYI. As I can see in
https://github.com/gear-tech/substrate/blob/677fcc194cef7c670c03f7a2d596afedb6fd8c66/primitives/state-machine/src/ext.rs#L183-L207
if data is just read from storage, it's not copied to overlay. But it can be stored in DB hot caches, and also we won't have need to spend one page twice to relay chain. So, as conclusion, second+ loading should be cheaper

core-processor/src/executor.rs
gshep: Sorry didn't get what there should be commented.

grishasobol: In case of init we return here and skip some below gas charging - explain in comment why

breathx:
> FYI. As I can see in
> https://github.com/gear-tech/substrate/blob/677fcc194cef7c670c03f7a2d596afedb6fd8c66/primitives/state-machine/src/ext.rs#L183-L207
> if data is just read from storage, it's not copied to overlay. But it can be stored in DB hot caches, and also we won't have need to spend one page twice to relay chain. So, as conclusion, second+ loading should be cheaper

Seems that it happens in other places of code, cause our manual test with @gshep showed that multiple reads occurs as single read in benchmarks.

core-processor/src/executor.rs
```diff
@@ -250,11 +254,33 @@ pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environ
...
+    if !exports.contains(&dispatch.kind()) {
+        return Ok(DispatchResult {
```
breathx, line 261: I would provide something like `DispatchResult::default(dispatch, pid, allocations)`

core-processor/src/executor.rs
```diff
@@ -56,45 +57,46 @@ fn make_checks_and_charge_gas_for_pages<'a>(
     }
 
     let mem_size = if let Some(max_wasm_page) = allocations.iter().next_back() {
```
breathx, line 59: in fact, charging in executor without lazy pages goes already after page loading

pallets/gear/src/manager/mod.rs
```diff
@@ -146,14 +146,29 @@ where
...
     }
 
     /// NOTE: By calling this function we can't differ whether `None` returned, because
     /// program with `id` doesn't exist or it's terminated
-    pub fn get_actor(&self, id: ProgramId, with_pages: bool) -> Option<Actor> {
+    pub fn get_actor(
+        &mut self,
+        id: ProgramId,
+        with_pages: bool,
+        subsequent_execution: bool,
```
breathx, line 170: useless field here, I suppose

pallets/gear/src/tests.rs
```diff
@@ -76,6 +76,20 @@ fn unstoppable_block_execution_works() {
...
+        let GasInfo {
+            burned: subsequent_burned_gas,
+            ..
+        } = Gear::calculate_gas_info(
+            USER_1.into_origin(),
+            HandleKind::Handle(program_id),
+            EMPTY_PAYLOAD.to_vec(),
+            0,
+            true,
+            true,
```
breathx, line 92: I see no reason to add this field to rpc call. We have no guarantees when message will be executed, so to calculate it properly, need to add this value in min_limit by default and provide some filed like `might_be_returned` with amount of gas which might not be burned due to some conditions

pallets/gear/src/manager/mod.rs
```diff
@@ -177,12 +192,16 @@ where
             Default::default()
         };
 
+        let subsequent_execution = subsequent_execution || self.programs.contains(&id);
+        self.programs.insert(id);
```
breathx, line 196: being in self.programs is not guarantee that we were executing this program this block: if we check some user id and its a program, we also append the id in programs

breathx: one more thing - if we try to execute program and have no gas to load pages also means that further execution should try to do this too

breathx (changes requested):

core-processor/src/executor.rs
gshep: Agree. Moreover I move this change to the standalone PR siince there are some discussions around memory pages charging.

gshep: done - #1183

PR #1163, opened 2022-07-07
Revert previous WASM path in `gear-wasm-builder` + minor cleanup

shamilsan:
The output WASM binary path has been accidentally changed in #1126.

- Reverted the WASM binary path to keep compatibility with building without a build script.
- Minor cleanup.

@gear-tech/dev

### DIFF
--- a/examples/binaries/calc-hash/over-blocks/src/lib.rs
+++ b/examples/binaries/calc-hash/over-blocks/src/lib.rs
@@ -16,7 +16,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 #![no_std]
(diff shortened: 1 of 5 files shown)


examples/binaries/calc-hash/over-blocks/src/lib.rs
```diff
@@ -16,7 +16,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 #![no_std]
-#![feature(const_btree_new)]
+#![cfg_attr(not(feature = "std"), feature(const_btree_new))]
```
gshep, line 19: why `cfg_attr` if the previous line directly states it is `no_std` only?

gshep (approved):

PR #1164, opened 2022-07-07
Some gear-wasm-builder improvements

ark0f:
Copy WASM files to standard directory (`target/wasm32-unknown-unknown/{debug, release}`)


utils/wasm-builder/README.md
```diff
@@ -47,17 +47,17 @@ mod code {
 }
 
 #[test]
-fn debug_wasm() {
+fn assert_wasm() {
     assert_eq!(
-        std::fs::read("target/wasm32-unknown-unknown/debug/test_program.wasm").unwrap(),
```
breathx, line 52: Am I right that you want to always push binaries to release dir, even if it had built without release flag?

utils/wasm-builder/src/cargo_command.rs
```diff
@@ -26,7 +26,6 @@ pub struct CargoCommand {
     path: String,
     manifest_path: PathBuf,
     args: Vec<&'static str>,
-    profile: String,
```
breathx, line 29: what about to really use profile setting?

breathx:
I don't see any reason to force always build in release mode. If developer writes some functionality and regularly runs tests, there is no need to wait excess time

ark0f:
> I don't see any reason to force always build in release mode. If developer writes some functionality and regularly runs tests, there is no need to wait excess time

But you even approved that logic: https://github.com/gear-tech/gear/pull/687

utils/wasm-builder/README.md
shamilsan: Yes, we build WASMs in release mode but the binaries path should correspond to the profile and be intuitive for the user. Also, we had an idea to cut gstd's `debug` feature when building in release profile.

ark0f: > Also, we had an idea to cut gstd's debug feature when building in release profile.

I think we need to dicuss possibility and how to do it correctly

shamilsan (changes requested):
Need to make `cargo` build the WASM with a selected profile instead of hard coded `--release`.

I guess it is better to make the debug-feature-related functionality (removing `debug` feature from `gstd` in release build) as a separate PR later.

breathx (approved):

shamilsan (approved):
Great 👍

PR #1183, opened 2022-07-14
Don't execute empty functions

### LINKED ISSUE #1182: Optimize empty functions
### File Location(s)

_No response_

### Proposal

To optimise with empty functions:
If we try to init program, where init export doesn't exist, we can just skip try to load pages and create env instance. We can check if program contains given export by the code on which program depend (or to avoid reading code, paste gear exports in program instance also), for more details see https://github.com/gear-tech/gear/pull/1085.

Extracted from #1088

gshep:
Resolves #1182 

@gear-tech/dev

### DIFF
--- a/core-processor/src/executor.rs
+++ b/core-processor/src/executor.rs
@@ -250,11 +250,27 @@ pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environ
         }
     };
 
+    let is_initial = program.get_allocations().is_empty();
+
+    let exports = program.code().exports();
+    if !exports.contains(&dispatch.kind()) {
+        let allocations = match is_initial {
+            true => Some((0..static_pages.0).map(WasmPageNumber).collect()),
+            false => None,
+        };
+
+        return Ok(DispatchResult::success(
+            dispatch,
+            program_id,
+            gas_counter.into(),
+            allocations,
+        ));
+    }
+
     // Getting wasm pages allocations.
-    let (allocations, is_initial) = if program.get_allocations().is_empty() {
-        ((0..static_pages.0).map(WasmPageNumber).collect(), true)
-    } else {
-        (program.get_allocations().clone(), false)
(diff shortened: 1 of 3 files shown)


SabaunT (approved):

core-processor/src/executor.rs
```diff
@@ -250,11 +250,27 @@ pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environ
...
     };
 
+    let is_initial = program.get_allocations().is_empty();
+
+    let exports = program.code().exports();
+    if !exports.contains(&dispatch.kind()) {
+        let allocations = match is_initial {
+            true => Some((0..static_pages.0).map(WasmPageNumber).collect()),
+            false => None,
+        };
```
breathx, line 260: simplify with if clause?

core-processor/src/executor.rs
```diff
@@ -250,11 +250,27 @@ pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environ
...
+            allocations,
+        ));
+    }
+
     // Getting wasm pages allocations.
-    let (allocations, is_initial) = if program.get_allocations().is_empty() {
-        ((0..static_pages.0).map(WasmPageNumber).collect(), true)
-    } else {
-        (program.get_allocations().clone(), false)
+    let allocations = match is_initial {
```
breathx, line 271: same here

breathx (approved):
Some tests would be also great, that it costs nothing

core-processor/src/executor.rs
gshep: Frankly speaking I intentionally made it this way since it looks more concise.

breathx: > Frankly speaking I intentionally made it this way since it looks more concise.

How do you feel about [this](https://doc.rust-lang.org/std/primitive.bool.html#method.then) stuff?

PR #1189, opened 2022-07-15
add error in case page data is not found

grishasobol:


breathx:
I prefer to request review on success CI

grishasobol:
> I prefer to request review on success CI

Done

common/src/lib.rs
```diff
@@ -171,11 +171,24 @@ pub enum Program {
     Terminated,
 }
 
-#[derive(Clone, Copy, Debug)]
-pub enum ProgramError {
-    CodeHashNotFound,
+#[derive(Clone, Debug, derive_more::Display)]
+pub enum CommonError {
```
breathx, line 175:
```suggestion
pub enum ProgramError {
```

I see no reason for renaming, cause it's still about the programs and its data only

common/src/lib.rs
```diff
@@ -318,34 +331,48 @@ pub fn get_program(id: H256) -> Option<Program> {
...
-    let data = sp_io::storage::get(&key)?;
-    Some(PageBuf::new_from_vec(data))
+pub fn get_program_page_data(program_id: H256, page: PageNumber) -> Result<PageBuf, CommonError> {
+    let key = page_key(program_id, page);
+    let data =
+        sp_io::storage::get(&key).ok_or(CommonError::CannotFindDataForPage { program_id, page })?;
+    PageBuf::new_from_vec(data).map_err(Into::into)
 }
 
+/// Returns data for all `program` pages, which has data
```
breathx, line 341:
```suggestion
/// Returns data for all `program` pages, which contain data.
```

grishasobol: In this case `has` is more suitable. Because storage `contains` data and page has data in storage

common/src/lib.rs
grishasobol: So, we can rename errors to ProgramError almost everywhere? Because everywhere errors about program execution. I can rename it to `Error`, but `ProgramError` is not uninformative

breathx: Why the "Common" is more informative so?

common/src/lib.rs
grishasobol: in English after `which` we must use `has`

common/src/lib.rs
grishasobol: just has the same name with crate, I think is more comfortable to use outside crate

common/src/lib.rs
breathx: `which have` is correct form for plural nouns, but not such popular as `that have`
[src](https://www.quora.com/Which-is-grammatically-correct-which-have-or-which-has)

grishasobol: Ok, will make `that have` =)

PR #1190, opened 2022-07-15
Refactor backend

### LINKED ISSUE #842: Remove `Rc` and `RefCell` pointers from `ExtCarrier`
### File Location(s)

_No response_

### Proposal

An idea is to change it to `Option<E>` or just `E`, where `E` implements `Ext`. In this way we will have compile time guarantees over pointers usage safety.

ukint-vs:
The main goal of this refactoring is to make the backend logic as simple and lightweight as possible.
We don't need to create `Environment` struct since we can do all the preparation and execution in one place.
Consuming the `Ext` in errors inside the backend destroys the purpose of the backend, we have to give input data and receive output data and work with this at a higher level in `gear-core-processor`.

- [x] Don't consume `Ext` inside backend
- [x] `Environment::execute`
- [x] use `&mut E` inside execution
- [x] Remove `ExtCarrier`. Resolve #842  
- [x] trait `RuntimeCtx` 

@gear-tech/dev


core-backend/common/src/lib.rs
```diff
@@ -110,25 +109,20 @@ pub struct ExtInfo {
...
-    pub info: ExtInfo,
+    fn trap_explanation(&self) -> Option<TrapExplanation>;
 }
 
+pub type BackendReport<T> = (TerminationReason, T, Option<WasmPageNumber>);
+
 #[derive(Debug, derive_more::Display)]
 #[display(fmt = "{}", reason)]
 pub struct BackendError<T> {
-    pub gas_amount: GasAmount,
```
ark0f, line 131: Why the field was removed?

core-processor/src/executor.rs
```diff
@@ -297,59 +297,58 @@ pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environ
...
+                    return Err(ExecutionError {
+                        program_id,
+                        gas_amount: ext.into_gas_amount(),
+                        reason: ExecutionErrorReason::Backend(e.to_string()),
+                    });
+                }
+            }
+            (
+                termination,
+                ext.into_ext_info(&memory).unwrap(),
```
ark0f, line 335: Why unwrapping is here?

ark0f (changes requested):

core-backend/common/src/lib.rs
```diff
@@ -110,25 +109,20 @@ pub struct ExtInfo {
...
     fn last_error(&self) -> Option<&ExtError>;
-}
 
-pub struct BackendReport {
-    pub termination: TerminationReason,
-    pub info: ExtInfo,
+    fn trap_explanation(&self) -> Option<TrapExplanation>;
 }
 
+pub type BackendReport<T> = (TerminationReason, T, Option<WasmPageNumber>);
```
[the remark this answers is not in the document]
breathx: Agree. Using struct with probably some trait bound seems better.

core-backend/common/src/lib.rs
```diff
@@ -139,39 +133,22 @@ pub trait Environment<E: Ext + IntoExtInfo + 'static>: Sized {
...
-        ext: E,
+    /// 2) Creates wasm memory
+    /// 3) Runs `pre_execution_handler` to fill the memory before running instance.
+    /// 4) Instantiate external funcs for wasm module.
+    /// 5) Run instance setup starting at `entry_point` - wasm export function name.
+    fn execute<F, T>(
+        ext: &mut E,
         binary: &[u8],
         entries: BTreeSet<DispatchKind>,
         mem_size: WasmPageNumber,
```
breathx, line 145: do we still need this arg?
we can take it from last allocation.

same for entries - why do we need them here?

core-backend/sandbox/src/funcs.rs
```diff
@@ -147,26 +135,26 @@ where
...
+            let payload = ctx.read_sandbox_memory(payload_ptr, payload_len)?;
+            let value: u128 = ctx.read_sandbox_memory_as(value_ptr)?;
+            let error_len = ctx
+                .ext()
                 .send(HandlePacket::new(dest, payload, value))
                 .process_error()
                 .map_err(FuncError::Core)?
                 .error_len_on_success(|message_id| {
-                    wto(memory, message_id_ptr, message_id.as_ref())
+                    ctx.write_sandbox_output(message_id_ptr, message_id.as_ref())
```
breathx, line 148:
```suggestion
                    ctx.write(message_id_ptr, message_id.as_ref())
```

breathx: at least, remove only `_sandbox_`

core-backend/sandbox/src/funcs.rs
```diff
@@ -179,27 +167,26 @@ where
...
-        .map_err(|err| {
-            ctx.err = err;
-            HostError
-        })
+        };
+        f().map(|code| Value::I32(code as i32).into())
+            .map_err(|err| {
+                ctx.set_err(err);
+                HostError
+            })
```
breathx, line 189: can't this be wrapped into some fn to avoid repetitions?

core-backend/sandbox/src/runtime.rs
```diff
@@ -0,0 +1,106 @@
...
+    pub memory: &'a DefaultExecutorMemory,
+    pub memory_wrap: &'a mut MemoryWrap,
+    pub err: FuncError<E::Error>,
+}
+
+impl<'a, E> Runtime<'a, E>
+where
+    E: Ext + 'a,
+{
+    pub fn new(
```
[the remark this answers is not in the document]
breathx: Agree.
1. Make fields private
2. Remove `new`, cause it only copy-pastes args from fn to struct def

core-processor/src/executor.rs
```diff
@@ -297,59 +297,58 @@ pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environ
     };
 
     // Creating externalities.
-    let ext = A::new(context);
+    let mut ext = A::new(context);
 
-    let mut env = E::new(
-        ext,
+    // let mut ext_carrier = ExtCarrier::new(ext);
```
breathx, line 302: dead code

core-processor/src/executor.rs
breathx: Also agree here, no unwraps should be used.
May `into_ext_info` be `-> Info`, instead of `-> Result<Info, ..>` ?

gtest/src/wasm_executor.rs
```diff
@@ -19,42 +19,37 @@
...
-    defined_host_functions: DefinedHostFunctions<Runtime<Ext>, <Ext as ExtTrait>::Error>,
-}
+pub(crate) struct WasmExecutor;
 
 impl WasmExecutor {
-    /// Creates a WasmExecutor instance from a program.
-    /// Also uses provided memory pages for future execution
-    pub(crate) fn new(
+    /// Executes non-void function by provided name.
+    /// Panics if function is void
```
breathx, line 44:
```suggestion
    /// Panics if function is void.
```

breathx (approved):
Seems ok in general.
But there is need in merging funcs.rs files for both wasmi and sandbox backends, cause they are identical.
Some beauty traits would be good.

GJ

core-backend/sandbox/src/env.rs
```diff
@@ -95,29 +81,35 @@ impl<'a, E: Ext + IntoExtInfo + 'static> EnvBuilder<'a, E> {
...
+        pre_execution_handler: F,
+    ) -> Result<BackendReport<Self::Memory>, BackendError<Self::Error>>
+    where
+        F: FnOnce(&mut Self::Memory) -> Result<(), T>,
+        T: fmt::Display,
+    {
         let mut builder = EnvBuilder::<E> {
             env_def_builder: EnvironmentDefinitionBuilder::new(),
-            forbidden_funcs: ext.forbidden_funcs(),
+            forbidden_funcs: &ext.forbidden_funcs().clone(),
```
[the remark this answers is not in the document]
ukint-vs: without clone ext will be borrowed and can't be used later in runtime

core-backend/common/src/lib.rs
ukint-vs: We don't want to destroy ext inside backend.

core-backend/sandbox/src/runtime.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// Read designated chunk from the sandbox memory.
+    ///
+    /// Returns `Err` if one of the following conditions occurs:
+    ///
+    /// - requested buffer is not within the bounds of the sandbox memory.
+    pub fn read_sandbox_memory(&self, ptr: u32, len: u32) -> Result<Vec<u8>, MemoryError> {
+        let mut buf = vec![0u8; len as usize];
+        self.memory
+            .get(ptr, buf.as_mut_slice())
+            .map_err(|_| MemoryError::OutOfBounds)?;
```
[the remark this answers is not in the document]
ukint-vs: get/set returns error only if memory is out of bounds.

core-processor/src/executor.rs
```diff
@@ -287,90 +287,102 @@ pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environ
...
+    match ext.into_ext_info(&memory, stack_end_page.unwrap_or_default()) {
+        Ok(info) => {
+            // Parsing outcome.
+            let kind = match termination {
+                TerminationReason::Exit(value_dest) => DispatchResultKind::Exit(value_dest),
+                TerminationReason::Leave | TerminationReason::Success => {
+                    DispatchResultKind::Success
+                }
+                TerminationReason::Trap(explanation) => {
+                    log::debug!(
```
grishasobol, line 351: What the reason to increase code nesting so much here?
It will be much more readable like this:
```rust
let info = match ext.into_ext_info(...).ok_or_else(|| ... )?;
// here logic with `info`
```

gtest/src/wasm_executor.rs
```diff
@@ -19,42 +19,36 @@
...
 use crate::{manager::ExtManager, Result, TestError, MAILBOX_THRESHOLD};
 
 /// Binary meta-functions executor for testing purposes
-pub(crate) struct WasmExecutor {
-    instance: ModuleRef,
-    store: Runtime<Ext>,
-    memory: MemoryRef,
-    defined_host_functions: DefinedHostFunctions<Runtime<Ext>, <Ext as ExtTrait>::Error>,
-}
+pub(crate) struct WasmExecutor;
```
ark0f, line 40: Do we ever need this struct after all of these changes?

ark0f (approved):

gtest/src/wasm_executor.rs
ukint-vs: We need to pass struct like this https://github.com/gear-tech/gear/blob/e4ac4d069ae1b152f85d641572fe09788a6e2721/pallets/gear/src/lib.rs#L791

PR #1193, opened 2022-07-16
Remove static page numbers from collections

### LINKED ISSUE #1192: Don't retain static allocations in Program struct
### File Location(s)

_No response_

### Proposal

So as the title says: we are always able to obtain count of static memory pages there so don't keep page numbers of such pages in the collection.

gshep:
Resolves #1192 

@gear-tech/dev

### DIFF
--- a/core-backend/common/src/lib.rs
+++ b/core-backend/common/src/lib.rs
@@ -102,6 +102,7 @@ pub enum TrapExplanation {
 pub struct ExtInfo {
     pub gas_amount: GasAmount,
     pub allocations: BTreeSet<WasmPageNumber>,
+    pub initial_allocations: BTreeSet<WasmPageNumber>,
     pub pages_data: BTreeMap<PageNumber, PageBuf>,
     pub generated_dispatches: Vec<Dispatch>,
     pub awakening: Vec<MessageId>,
--- a/core-processor/src/executor.rs
+++ b/core-processor/src/executor.rs
@@ -34,7 +34,7 @@ use gear_core::{
     gas::{ChargeResult, GasAllowanceCounter, GasCounter, ValueCounter},
     ids::ProgramId,
     memory::{AllocationsContext, Memory, PageBuf, PageNumber, WasmPageNumber},
-    message::{ContextSettings, IncomingDispatch, MessageContext},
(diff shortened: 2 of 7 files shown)


core-backend/common/src/lib.rs
```diff
@@ -102,6 +102,7 @@ pub enum TrapExplanation {
 pub struct ExtInfo {
     pub gas_amount: GasAmount,
     pub allocations: BTreeSet<WasmPageNumber>,
```
breathx, line 104: can't we pre-store initial allocations and keep here only actual after execution?

core-processor/src/executor.rs
```diff
@@ -47,17 +47,27 @@ fn make_checks_and_charge_gas_for_pages<'a>(
...
+        if wasm_page >= static_pages && !allocations.contains(&wasm_page) {
             return Err(ExecutionErrorReason::PageIsNotAllocated(*page));
         }
     }
 
-    let mem_size = if let Some(max_wasm_page) = allocations.iter().next_back() {
+    let mem_size = if !initial_execution {
+        let max_wasm_page = if let Some(page) = allocations.iter().next_back() {
+            *page
+        } else if static_pages.0 != 0 {
```
breathx, line 63: I would rewrite it just for beautifying:
```suggestion
        } else if static_pages != WasmPageNumber::zero() {
```

breathx, line 63: we also may use here saturating math to don't add such checks for gt than 0 etc

breathx (approved):
lgtm 👍

core-processor/src/executor.rs
gshep: ok

gshep: unfortunately we have to check this way: if there are neither dynamic allocations nor static pages then no charging should be performed for memory growing.

core-backend/common/src/lib.rs
gshep: ok

core-processor/src/executor.rs
breathx: ahh, we return Ok from fn, nor the closure

grishasobol (approved):

PR #1195, opened 2022-07-18
gtest: Fix claiming zero values from mailbox

shamilsan:
There was a problem with processing messages with zero value while claiming value from the mailbox. This PR fixes this bug.

@gear-tech/dev

### DIFF
--- a/gtest/src/manager.rs
+++ b/gtest/src/manager.rs
@@ -387,13 +387,15 @@ impl ExtManager {
     pub(crate) fn claim_value_from_mailbox(&mut self, id: &ProgramId) {
         let messages = self.mailbox.remove(id);
         if let Some(messages) = messages {
-            messages.iter().for_each(|message| {
-                self.send_value(
-                    message.source(),
-                    Some(message.destination()),
-                    message.value(),
-                )
-            });
+            for message in messages {
+                if message.value() > 0 {
+                    self.send_value(
+                        message.source(),
+                        Some(message.destination()),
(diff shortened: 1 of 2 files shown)


gtest/src/manager.rs
```diff
@@ -387,13 +387,15 @@ impl ExtManager {
...
         let messages = self.mailbox.remove(id);
         if let Some(messages) = messages {
-            messages.iter().for_each(|message| {
-                self.send_value(
-                    message.source(),
-                    Some(message.destination()),
-                    message.value(),
-                )
-            });
+            for message in messages {
```
breathx, line 390: for me seems easier and better for future debug/development
add this check (skip if 0) here:
https://github.com/gear-tech/gear/blob/8c06ad384f0d8ca34ac30b6f4cd3af5c9c6d2dd0/gtest/src/manager.rs#L749-L769
or here
https://github.com/gear-tech/gear/blob/8c06ad384f0d8ca34ac30b6f4cd3af5c9c6d2dd0/gtest/src/manager.rs#L367-L378

e.g., we will have some additional functionality to do if value is zero, and this stuff will become hard to catch

breathx (changes requested):

gtest/src/manager.rs
shamilsan: Minting zero value seems weird but sending should not cause an error, agree.

shamilsan: Fixed.

Issue #1130, opened 2022-07-01
Handle all error cases in `GasTree` interface


breathx:
### Problem to Solve

In most cases we just skip the result of `GasTree` operations:
https://github.com/gear-tech/gear/blob/07ccee0003608f37af6b58235386bcb4e2cba1cf/pallets/gear/src/manager.rs#L463
But we should branch our logic for error cases.

### Possible Solution

// .. //

breathx:
I think many of them should be `unreachable` and guaranteed by runtime logic as it done for other similar interfaces:
https://github.com/gear-tech/gear/blob/07ccee0003608f37af6b58235386bcb4e2cba1cf/pallets/gear/src/manager.rs#L466-L467

Issue #899, opened 2022-05-04
Clearly cover pallets with docs


breathx:
### File Location(s)

`gear/pallets/*`

### Proposal

Need to provide actual explanation of types, unreachables and algorithms.
Clear the code and provide actual docs for common traits.

PR #1197, opened 2022-07-18
Internal pallet module

breathx:
Part of #1130, part of #646, part of #899

- Mailbox and waitlist now contains bn of insertion and planned processing as `Interval`
- Module with complex logic added into `internal.rs`, this simplifies a lot
- Clean ups all over the code
- Pretty and readable doc

This changes appeared during work on Mailbox proper rent charging, so I decided to create separate PR with them.
It also a bit overflowed with different changes, but it's hard to split them.

Note: commit history a bit messed up with other PR, but it's not shown in diff, so I left it as it is


gshep (approved):

SabaunT (approved):

pallets/gear/src/lib.rs
```diff
@@ -1030,49 +1035,28 @@ pub mod pallet {
                 T::DebugInfo::remap_id();
             }
 
+            let lazy_pages_enabled =
+                cfg!(feature = "lazy-pages") && lazy_pages::try_to_enable_lazy_pages();
+
             while QueueProcessingOf::<T>::allowed() {
                 if let Some(dispatch) = QueueOf::<T>::dequeue()
                     .unwrap_or_else(|e| unreachable!("Message queue corrupted! {:?}", e))
```
SabaunT, line 1043: I think we can reduce one nesting level by doing like this
```suggestion
                let dispatch = match QueueOf::<T>::dequeue()
                    .unwrap_or_else(|e| unreachable!("Message queue corrupted! {:?}", e)) 
                {
                    Some(dispatch) => dispatch,
                    None => break,
                };  
```

pallets/gear/src/lib.rs
```diff
@@ -1109,82 +1087,39 @@ pub mod pallet {
                                     log::debug!(
                                         "Can not instrument code '{:?}' for program '{:?}'",
                                         code_id,
-                                        program_id
+                                        dispatch.destination()
                                     );
                                     continue;
                                 }
                             } else {
```
SabaunT, line 1094: What about adding this comment?
```suggestion
                            } else {
                            // This branch is considered unreachable, because there can't be a program without code.
                            // Reaching this branch is a sign of a serious storage/extrinsics corruption. 
```

breathx: In general, this PR does not affect any code storing logic, its just fast-find beautification, so I prefer to delay any explanations for future PRs, because everything can change.

pallets/gear/src/lib.rs
breathx: Yeah, but should we?

It reduces amount of tabs, but for me decreases readability, which isn't good as well

pallets/gear/src/lib.rs
SabaunT: This invariant didn't change for a long time. And seems not to be in the future. Anyway, in 'process_queue' function we have different unclear logical branches. I mean it's not obvious whether they are reachable and in what situations. So better to write this and change, rather than forget about it.

PR #1198, opened 2022-07-18
Replace reply data tuple with struct

breathx:
`Message`s reply now contains special named struct instead of just tuple `(MessageId, ExitCode)`. This increases readability of events in scanners like polkadot js apps 

@gear-tech/dev


core/src/message/mod.rs
```diff
@@ -92,7 +92,7 @@ impl DispatchKind {
 
 /// Message packet.
 ///
-/// Provides common behaviour for any message's packet: accessing to payload, gas limit and value.
+/// Provides common behavior for any message's packet: accessing to payload, gas limit and value.
```
shamilsan, line 95: You don't like British English 😄

core-backend/sandbox/src/funcs.rs
```diff
@@ -360,15 +360,16 @@ where
...
-            .with_fallible(|ext| ext.reply_to().map_err(FuncError::Core))
+            .with_fallible(|ext| ext.reply_details().map_err(FuncError::Core))
             .map_err(|e| {
                 ctx.err = e;
                 HostError
             })?;
 
-        if let Some((_, exit_code)) = reply_tuple {
+        if let Some(details) = opt_details {
+            let exit_code = details.into_parts().1;
```
shamilsan, line 372: Why not just
```suggestion
            let exit_code = details.exit_code();
```

core-backend/sandbox/src/funcs.rs
```diff
@@ -595,15 +596,16 @@ where
...
-            .with_fallible(|ext| ext.reply_to().map_err(FuncError::Core))
+            .with_fallible(|ext| ext.reply_details().map_err(FuncError::Core))
             .map_err(|err| {
                 ctx.err = err;
                 HostError
             })?;
 
-        if let Some((message_id, _)) = maybe_message_id {
+        if let Some(details) = opt_details {
+            let message_id = details.into_parts().0;
```
shamilsan, line 608: Also, do we really need `into_parts` as parts are copy types?
```suggestion
            let message_id = details.reply_to();
```

shamilsan (approved):
I like using structs instead of tuples in such kind of circumstances 👍

core-backend/sandbox/src/funcs.rs
breathx: I don't see the reason to clone/copy even copy-types when there is no need to access details further. Move-semantic makes sense for me even in such small things.
Isn't this destruction cheaper then copy?

breathx: We can provide something like
`fn into_reply_to(self) -> MessageId` for comfortable usage

osipov-mit (approved):

PR #1211, opened 2022-07-20
Add `core_processor::prepare` method

### LINKED ISSUE #1088: Charge for page load optionally

gshep:
Resolves #1088 , #1182 

- introduce `prepare` method. It checks if provided actor is executable, if code contains required export and tries to pre-charge for memory pages
- remove `process_executable` function
- add test cases that execution of programs with missing exports is free
- adjust `calculate_gas_info`: add `GasInfo::may_be_returned` if a program happens to be executed repeatedly in a block.

Supersedes #1162 , #1183 .

@gear-tech/dev 

@osipov-mit please also check this.


core-processor/src/common.rs
```diff
@@ -357,15 +378,22 @@ pub struct Actor {
 pub struct ExecutableActorData {
     /// Program.
     pub program: Program,
-    /// Data which some program allocated pages may have.
-    pub pages_data: BTreeMap<PageNumber, PageBuf>,
+    /// Numbers of memory pages with some data.
```
breathx, line 381: This formulae is really strange for me.
Can we replace it with something like: "allocated pages, that have non-default data"?

core-processor/src/common.rs
```diff
@@ -357,15 +378,22 @@ pub struct Actor {
 pub struct ExecutableActorData {
     /// Program.
```
breathx, line 379:
```suggestion
    /// Program aggregated data.
```

core-processor/src/processor.rs
```diff
@@ -42,33 +49,197 @@ enum SuccessfulDispatchResultKind {
...
+
+impl PreparedMessageExecutionContext {
+    /// Returns reference to the GasCounter.
+    pub fn gas_counter(&self) -> &GasCounter {
+        &self.gas_counter
+    }
+}
+
+/// Defines result variants of the function `prepare`.
+#[allow(clippy::large_enum_variant)]
```
breathx, line 71: instead of allowing clippy, can we use box for Ok case?

core-processor/src/processor.rs
```diff
@@ -42,33 +49,197 @@ enum SuccessfulDispatchResultKind {
...
         executable_data,
     } = actor;
 
-    match check_is_executable(executable_data, &dispatch) {
-        Err(exit_code) => process_non_executable(dispatch, destination_program, exit_code),
-        Ok(data) => process_executable::<A, E>(
-            origin,
-            gas_allowance,
-            data,
+    let data = match check_is_executable(executable_data, &dispatch) {
```
breathx, line 107: can't we destruct data struct into fields here?
like Data {
   program,
   pages_with_data
}

etc

core-processor/src/processor.rs
```diff
@@ -42,33 +49,197 @@ enum SuccessfulDispatchResultKind {
...
+        &mut gas_counter,
+        &mut gas_allowance_counter,
+        program.get_allocations(),
+        program.static_pages(),
+        dispatch.context().is_none() && matches!(dispatch.kind(), DispatchKind::Init),
+        subsequent_execution,
+    ) {
+        Ok(size) => size,
+        Err(reason) => {
+            log::debug!("failed to charge for memory pages: {:?}", reason);
```
breathx, line 140:
```suggestion
            log::debug!("Failed to charge for memory pages: {:?}", reason);
```
I'd also add logs for success charge

core-processor/src/processor.rs
```diff
@@ -42,33 +49,197 @@ enum SuccessfulDispatchResultKind {
...
+        },
+        pages_with_data: data.pages_with_data,
+    }
+}
+
+/// Process program & dispatch for it and return journal for updates.
+pub fn process<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environment<A>>(
+    block_config: &BlockConfig,
+    execution_context: PreparedMessageExecutionContext,
+    memory_pages: BTreeMap<PageNumber, PageBuf>,
```
breathx, line 170: can't memory pages be thrown with prepared message context?

core-processor/src/processor.rs
```diff
@@ -42,33 +49,197 @@ enum SuccessfulDispatchResultKind {
     Success,
 }
 
-/// Process program & dispatch for it and return journal for updates.
-pub fn process<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environment<A>>(
+/// Checked parameters for message execution across processing runs.
+pub struct PreparedMessageExecutionContext {
```
breathx, line 53: I'd also keep here and in other places something like 
```rust
struct Actor {
    id: ProgramId,
    balance: u128,
    executable_data: Option<(Program, MemoryPages)>
}
```

pallets/gear/src/tests.rs
```diff
@@ -4010,6 +4014,158 @@ fn test_async_messages() {
     })
 }
 
+#[test]
+fn missing_functions_are_not_executed() {
+    // handle is copied from ProgramCodeKind::OutgoingWithValueInHandle
```
breathx, line 4019: Why do we need so complicated demo for such check?

breathx (approved):
LGTM at all, but seems there is still place for refactoring and simplifications of aggregating data for future 👍

breathx:
Please, consensus on your changes with @ukint-vs in #1190

core-processor/src/processor.rs
gshep: Yep, we are able to, but do we really want to have unnecessary heap allocation?

pallets/gear/src/tests.rs
gshep: I supposed it doesn't matter how complex test is. Particularly this looks so due to the intention to test absent `handle_reply` function but the original program from `ProgramCodeKind` doesn't meet the test requirements.

core-processor/src/processor.rs
gshep: No, they cannot since it is responsibility of `prepare`  caller to fetch memory pages from storage on success. So to pass memory data with `PreparedMessage...` there should be added getter/setters/etc and `execution_context` has to be mutable - looks ugly. As an alternative new type could be introduced composing from `memory_pages` + `PrepareMessage...` but again it creates unnecessary  complications. So I would go with the current solution.

core-processor/src/processor.rs
breathx: Discussed

breathx:
@gshep, please, never resolve comments, before here any changes pushed

core-processor/src/processor.rs
gshep: To tell the truth I don't see any profit to introduce such structure currently. There is exactly mentioned entity but it is used to distingiush between executable and not executable program.

core-processor/src/processor.rs
```diff
@@ -40,33 +48,201 @@ enum SuccessfulDispatchResultKind {
...
+        }
+        Err(reason) => {
+            log::debug!("Failed to charge for memory pages: {reason:?}");
+            return PrepareResult::Error(match reason {
+                ExecutionErrorReason::InitialMemoryBlockGasExceeded
+                | ExecutionErrorReason::GrowMemoryBlockGasExceeded
+                | ExecutionErrorReason::LoadMemoryBlockGasExceeded => {
+                    process_allowance_exceed(dispatch, program_id, gas_counter.burned())
+                }
+                _ => process_error(dispatch, program_id, gas_counter.burned(), reason),
```
SabaunT, line 150: is it possible case?

SabaunT (approved):

core-processor/src/processor.rs
gshep: Yep, it is possible. `charge_gas_for_pages` may return `LoadMemoryGasExceede`/`GrowMemoryGasExceeded` and `InitialMemoryGasExceeded`.

SabaunT: Oh, I got confused by pretty similar enum variants 😅

core-processor/src/executor.rs
```diff
@@ -335,7 +334,7 @@ pub fn execute_wasm<A: ProcessorExt + EnvExt + IntoExtInfo + 'static, E: Environ
     };
 
     // Page which is right after stack last page
-    log::trace!("Stack end page = {:?}", stack_end_page);
+    log::trace!("Stack end page = {stack_end_page:?}");
```
breathx, line 337: Saw it somewhere, but didn't find the doc which declares that such syntax works for debug print `:?`.
Does it work?

breathx (approved):

core-processor/src/executor.rs
gshep: Yes, it is even checked in compile time.

PR #1219, opened 2022-07-21
gtest: Provide block height and timestamp to `meta_state` function

shamilsan:
Sometimes it is needed to get the current block's height and timestamp while debugging `meta_state` function. Now when calling `exec::block_height` and `exec::block_timestamp` functions in `meta_state`, they provide actual data according to the `gtest` runtime.

@gear-tech/dev

### DIFF
--- a/examples/block-info/src/lib.rs
+++ b/examples/block-info/src/lib.rs
@@ -13,3 +13,9 @@ unsafe extern "C" fn handle() {
     let bh = exec::block_height();
     msg::reply_bytes(format!("{}_{}", payload, bh), 0).unwrap();
 }
+
+#[no_mangle]
+unsafe extern "C" fn meta_state() -> *mut [i32; 2] {
+    let timestamp = exec::block_timestamp();
+    gstd::util::to_leak_ptr(timestamp.to_le_bytes())
+}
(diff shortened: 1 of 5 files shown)


breathx:
That's bad to provide some functionality, that work only for tests.
@osipov-mit, please, take a look is it possible to somehow apply some specific values to syscalls on js side, while reading meta

shamilsan:
> That's bad to provide some functionality, that work only for tests. @osipov-mit, please, take a look is it possible to somehow apply some specific values to syscalls on js side, while reading meta

I've expected this comment from your side 😄 This functionality will work not only in tests but in frontend too (as it provides block info when executing `meta_state`). It allows making some pre-calculations using native contract functions. I agree that we can do that calculations on the frontend side but it can be more expensive and error-prone.

Anyway this is the instrument only, one may choose to use it or not.

osipov-mit:
> That's bad to provide some functionality, that work only for tests.
> @osipov-mit, please, take a look is it possible to somehow apply some specific values to syscalls on js side, while reading meta

Yes, it’s already possible to get a `block’s timestamp` in js api and there is no problem in adding functionality to get block height

shamilsan:
> Yes, it’s already possible to get a `block’s timestamp` in js api and there is no problem in adding functionality to get block height

https://github.com/gear-tech/gear-js/pull/872

breathx (approved):
👍

PR #1223, opened 2022-07-22
Don't process data for stack pages

### LINKED ISSUE #1204: Don't retain data for stack pages
### File Location(s)

core-processor/src/executor.rs#L157-L166

### Proposal

The stack pages aren't used after execution so there is no reason to process its data.

gshep:
Resolves #1204 

@gear-tech/dev

### DIFF
--- a/core-backend/sandbox/src/env.rs
+++ b/core-backend/sandbox/src/env.rs
@@ -232,21 +232,25 @@ where
             Ok(ReturnValue::Unit)
         };
 
+        // Page which is right after stack last page
+        let stack_end_page = self.get_stack_mem_end();
+        log::trace!("Stack end page = {stack_end_page:?}");
+
         let Runtime {
             ext,
             memory,
             err: trap,
         } = self.runtime;
 
-        log::debug!("execution res = {:?}", res);
+        log::debug!("SandboxEnvironment::execute res = {res:?}");
 
-        let (info, trap_explanation) =
-            ext.into_inner()
-                .into_ext_info(&memory)
-                .map_err(|(reason, gas_amount)| BackendError {
-                    reason: SandboxEnvironmentError::Memory(reason),
-                    gas_amount,
-                })?;
+        let (info, trap_explanation) = ext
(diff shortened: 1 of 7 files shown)


core-processor/src/ext.rs
```diff
@@ -233,6 +234,7 @@ impl IntoExtInfo for Ext {
         let mut pages_data = BTreeMap::new();
         for page in (0..static_pages.0)
             .map(WasmPageNumber)
+            .skip_while(|page| *page < stack_page_count)
```
grishasobol, line 237: May be it's better to use `stack_page_count.0..static_pages.0`. Check before that stack_page_count <= static_pages

grishasobol (approved):

core-processor/src/ext.rs
gshep: Is `static_page_count > static_pages` case possible?

grishasobol: > static_page_count > static_pages case possible?
Yes it's possible if contract has wrong information in __gear_end_stack export.

I don't mind

core-backend/sandbox/src/env.rs
```diff
@@ -232,21 +232,25 @@ where
...
-        let (info, trap_explanation) =
-            ext.into_inner()
-                .into_ext_info(&memory)
-                .map_err(|(reason, gas_amount)| BackendError {
-                    reason: SandboxEnvironmentError::Memory(reason),
-                    gas_amount,
-                })?;
+        let (info, trap_explanation) = ext
+            .into_inner()
+            .into_ext_info(&memory, stack_end_page.unwrap_or_default())
```
breathx, line 249: isn't absence of stack end a critical error for us?

breathx (approved):

core-backend/sandbox/src/env.rs
gshep: Nope, since codegenerator doesn't create `__stack_end` for every program. I guess if it small enough.

PR #1225, opened 2022-07-24
Refactor `GasNode` type

### LINKED ISSUE #1179: Refactor `GasNode` type
### Proposal

`UnspecifiedLocal` type of gas nodes in a new gas tree (after #1073) has no children, so they don't need any children refs. The same could be stated for `ReservedLocal` nodes.
The idea is to remove general `GasNode` type and make 2 specific for nodes with/without children

SabaunT:
Resolves #1179 .

- Removed children refs from `UnspecifiedLocal` 
- Moved `consumed` field only to specific node types
- Removed `GasNodeType` (not needed without `consumed` field)


common/src/gas_provider/node.rs
```diff
@@ -105,6 +144,38 @@ impl<ExternalId: Clone, Id: Clone + Copy, Balance: Zero + Clone + Copy>
...
+        self.spec_refs().saturating_add(self.unspec_refs())
+    }
+
+    pub fn spec_refs(&self) -> u32 {
+        match self.inner {
+            GasNodeType::External { refs, .. } | GasNodeType::SpecifiedLocal { refs, .. } => {
+                refs.spec_refs
+            }
+            _ => 0,
+        }
```
breathx, line 156:
```suggestion
        use GasNodeType::{External, SpecifiedLocal};
        
        match self.inner {
            External { refs, .. } | SpecifiedLocal { refs, .. } => {
                refs.spec_refs
            }
            _ => 0,
        }
```

May be this var is more readable? And in other places

common/src/gas_provider/node.rs
```diff
@@ -19,13 +19,31 @@
 use super::*;
 use codec::MaxEncodedLen;
 
-// TODO remove children refs for Unspecified (https://github.com/gear-tech/gear/issues/1179)
 #[derive(Clone, Decode, Debug, Encode, MaxEncodedLen, TypeInfo, PartialEq, Eq)]
 pub enum GasNodeType<ExternalId, Id, Balance> {
```
breathx, line 23: We have decided, that `consumed` is specific field to `External` and `SpecifiedLocal` only
But the general node can also have `.consumed()` method

breathx:
Also provide docs for all public structs, please

common/src/gas_provider/node.rs
SabaunT: done

common/src/gas_provider/node.rs
SabaunT: please take a look at last changes ([64afdb6](https://github.com/gear-tech/gear/pull/1225/commits/64afdb68d5f0e878e199d15f19e4965ec8270ebe) and tell me whether you think changing it the way you suggested is still necessary.

common/src/gas_provider/node.rs
```diff
@@ -19,50 +19,99 @@
...
+    /// A node, which is a part of the tree structure (can be a parent and/or a child). As well as `External` node,
+    /// it has an internal balance and can exist while being consumed (for more info read [`Tree::consume`] implementor docs).
+    /// However, it has a `parent` field pointing to the node from which that one was created.
+    SpecifiedLocal {
+        parent: Id,
+        value: Balance,
+        refs: ChildrenRefs,
+        consumed: bool,
+    },
+    /// Pretty same as `SpecifiedLocal`, but doesn't have internal balance, so relies on it `parent`. Also such
```
gshep, line 44: misprint? `... relies on its parent...`

gshep (approved):
Looks good to me! :+1:

breathx (approved):

common/src/gas_provider/node.rs
SabaunT: fixed

PR #1227, opened 2022-07-24
CD workflow connect to vault

kvs96:


.github/workflows/CD.yml
```diff
@@ -1,78 +1,84 @@
...
+            kv/telegram_bots/alerts chat_id | TG_CHAT ;
+            kv/telegram_bots/alerts token | TG_TOKEN
 
-      - name: Install Info
-        run: |
-            ls -l .
-            ls -l ../..
-            pwd
+      - name: Checkout
+        uses: actions/checkout@v3.0.2
```
shamilsan, line 54: I don't see the reason to choose a concrete tag. Better to use a branch with the latest release within the major version. See for example: https://github.com/actions/checkout/
```suggestion
        uses: actions/checkout@v3
```

.github/workflows/CD.yml
```diff
@@ -1,78 +1,84 @@
...
 
   deploy:
     runs-on: ubuntu-latest
-    needs: docker
+    needs: build
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
+      - name: Read secrets
+        uses: hashicorp/vault-action@v2.4.1
```
shamilsan, line 34: Is this vault backed up and available for somebody except you? It should be protected against the bus factor.

breathx:
Is it stale PR? @kvs96

shamilsan:
Reopen it if needed.

PR #1228, opened 2022-07-25
Fix CI

shamilsan:
@gear-tech/dev


shamilsan:
`wasmi-validation` v0.4.2 [has been yanked](https://crates.io/crates/wasmi-validation/0.4.2) so we don't need this fix anymore.

PR #1236, opened 2022-07-26
production profile w\ fat lto

ukint-vs:
`wasmi 0.14.0` API has been changed to match the `wasmtime` style. I think we can switch to this later.

These params speed up execution time of benchmarks by 20-30%.
After this PR, we need to publish our production profile builds and also run them on our nodes.

```
./target/production/gear-node benchmark pallet --chain=dev --steps=50 --repeat=40 --pallet=pallet_gear --execution=wasm --wasm-execution=compiled --heap-pages=4096 --extrinsic="gr_reply_commit"
100.9 µs
./target/release/gear-node benchmark pallet --chain=dev --steps=50 --repeat=40 --pallet=pallet_gear --execution=wasm --wasm-execution=compiled --heap-pages=4096 --extrinsic="gr_reply_commit"
163.9 µs
```

Create production profile:
```
[profile.production]
inherits = "release"

# Sacrifice compile speed for execution speed by using optimization flags:

# https://doc.rust-lang.org/rustc/linker-plugin-lto.html
lto = "fat"
# https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units
codegen-units = 1
```
See https://github.com/paritytech/wasmi/issues/339.

@gear-tech/dev

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,7 +44,7 @@ opt-level = "s"
 
 [profile.release.package.galloc]
 opt-level = "s"
- 
+
 [profile.release.package.gtest]
 opt-level = "s"
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,21 @@ members = [
 [profile.release]
 panic = "unwind"
 
+[profile.release.package.gcore]
+opt-level = "s"
+
+[profile.release.package.gstd]
+opt-level = "s"
+
+[profile.release.package.gear-test]
+opt-level = "s"
+
+[profile.release.package.galloc]
+opt-level = "s"
+ 
+[profile.release.package.gtest]
+opt-level = "s"
+
 [profile.production]
 inherits = "release"
 
@@ -42,12 +57,3 @@ inherits = "release"
 lto = "fat"
 # https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units
 codegen-units = 1
-
-[profile.release.package.gcore]
-opt-level = "s"
-
-[profile.release.package.gstd]
-opt-level = "s"
-
-[profile.release.package.gear-test]
-opt-level = "s"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,15 +1,3 @@
-[profile.release]
-panic = "unwind"
-
-[profile.release.package.gcore]
-opt-level = "s"
-
-[profile.release.package.gstd]
-opt-level = "s"
-
-[profile.release.package.gear-test]
-opt-level = "s"
-
 [workspace]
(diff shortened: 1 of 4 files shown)


grishasobol:
How much it will cost for compilation time ?

grishasobol:
Also, we will test "release" profile and then run "production" profile in the test net, wether it safe?

Also fat lto can significantly increase runtime wasm-code size, which is also not so good. Please check size changes.

ukint-vs:
> How much it will cost for compilation time ?

5x but it doesn't matter since u can build and test in release profile.

ukint-vs:
> Also, we will test "release" profile and then run "production" profile in the test net, wether it safe?
> 
> Also fat lto can significantly increase runtime wasm-code size, which is also not so good. Please check size changes.

Size is lower. Runtime wasm w/ lto fat 2.0MB vs 2.8MB master

Cargo.toml
```diff
@@ -41,3 +29,25 @@ members = [
...
+lto = "fat"
+# https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units
+codegen-units = 1
+
+[profile.release.package.gcore]
+opt-level = "s"
+
+[profile.release.package.gstd]
+opt-level = "s"
+
```
breathx, line 51: please, add following code here:
```toml
[profile.release.package.galloc]
opt-level = "s"
 
[profile.release.package.gtest]
opt-level = "s"
```

Cargo.toml
```diff
@@ -41,3 +29,25 @@ members = [
...
+# https://doc.rust-lang.org/rustc/linker-plugin-lto.html
+lto = "fat"
+# https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units
+codegen-units = 1
+
+[profile.release.package.gcore]
+opt-level = "s"
+
+[profile.release.package.gstd]
+opt-level = "s"
```
breathx, line 50: and paste this release features above newly generated profile OR inside profile toml section

breathx (approved):

Cargo.toml
```diff
@@ -41,3 +29,25 @@ members = [
     "utils/economic-checks",
     "utils/economic-checks/fuzz",
 ]
+
+[profile.release]
+panic = "unwind"
```
ark0f, line 34: Why the line is here?

ukint-vs: > guess this is used for catching panic messages from wasm?

yes

Cargo.toml
ukint-vs: production profile inherits release

PR #1241, opened 2022-07-27
feat(gear): introduce extrinsic `create_program`

### LINKED ISSUE #1205: Extrinsic 'submit_program_without_code'
### Problem to Solve

We have extrinsics 'submit_program' and 'submit_code'. The former submits code and initializes a program, the latter just uploads code so it could be used later by 'create_program' syscall. The suggestion is to add possibility to initialize a program with an existing on-chain code.

### Possible Solution

add an extrinsic

### Notes

_No response_

clearloop:
Resolves #1205.

- [x] introduce check_code
- [x] introduce check_gas_limit_and_value
- [x] introduce `create_program`
- [x] renaming
- [x] tests
- [x] benchmarks

**Release Notes**: 

Introduced extrinsic new `create_program` inside `pallet_gear`, which can create program from exist code, the specific usages of this extrinsic would be like:

1. user calls extrinsic `pallet_gear::upload_code`, and then get the `code_id` from onchain events or storage
2. user calls extrinsic `pallet_gear::create_program` with the `code_id` he got from the previous step
3. user can call extrinsic `pallet_gear::create_program` again with the same `code_id` but different `salt` to create a different program.

@gear-tech/gear

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -795,9 +795,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "3.2.16"
+version = "3.2.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a3dbbb6653e7c55cc8595ad3e1f7be8f32aba4eb7ff7f0fd1163d4f3d137c0a9"
+checksum = "44bbe24bbd31a185bc2c4f7c2abe80bea13a20d57ee4e55be70ac512bdc76417"
 dependencies = [
  "atty",
  "bitflags",
@@ -948,18 +948,18 @@ dependencies = [
 
 [[package]]
 name = "cranelift-bforest"
-version = "0.86.1"
+version = "0.85.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "529ffacce2249ac60edba2941672dfedf3d96558b415d0d8083cd007456e0f55"
+checksum = "749d0d6022c9038dccf480bdde2a38d435937335bf2bb0f14e815d94517cdce8"
 dependencies = [
  "cranelift-entity",
 ]
 
 [[package]]
 name = "cranelift-codegen"
-version = "0.86.1"
+version = "0.85.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "427d105f617efc8cb55f8d036a7fded2e227892d8780b4985e5551f8d27c4a92"
+checksum = "e94370cc7b37bf652ccd8bb8f09bd900997f7ccf97520edfc75554bb5c4abbea"
 dependencies = [
  "cranelift-bforest",
  "cranelift-codegen-meta",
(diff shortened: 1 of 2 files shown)


breathx:
Also rename `claim_value_from_mailbox` into just `claim_value`, please

pallets/gear/src/lib.rs
```diff
@@ -1374,72 +1474,23 @@ pub mod pallet {
...
-
-            let code_id = code_and_id.code_id();
+                value,
+            )?;
 
-            // By that call we follow the guarantee that we have in `Self::submit_code` -
+            // By that call we follow the guarantee that we have in `Self::upload_code` -
             // if there's code in storage, there's also metadata for it.
-            if let Ok(code_hash) = Self::set_code_with_metadata(code_and_id, origin) {
+            if let Ok(code_hash) =
```
breathx, line 1491: I think we need just check here manually if wasm_code exists to clarify code

breathx (approved):
👍

osipov-mit (approved):

pallets/gear/src/lib.rs
```diff
@@ -1379,72 +1487,27 @@ pub mod pallet {
...
-
-            ensure!(
-                code.code().len() as u32 <= schedule.limits.code_len,
-                Error::<T>::CodeTooLarge
-            );
+            Self::check_gas_limit_and_value(gas_limit, value)?;
 
-            let code_and_id = CodeAndId::new(code);
+            let code_and_id = Self::check_code(code)?;
+            if T::CodeStorage::exists(code_and_id.code_id()) {
```
SabaunT, line 1497: I think it's wrong to return an error if someone tries to upload the same program. He will anyway pay for that (because of message size). Moreover, it won't lead to writing again the code to the code storage by the logic of `set_code_with_metadata`.

SabaunT (changes requested):

pallets/gear/src/lib.rs
SabaunT: Well, I don't agree with the fact that we must revert transaction with existing code as a parameter. In this case, we are like making users follow some determined flow of program creation. In this case `create_program` turns out to be not a feature, but a "must call, if code exists". I guess better remove the existence check.

pallets/gear/src/lib.rs
SabaunT: https://github.com/gear-tech/gear/pull/1241#discussion_r933784881

pallets/gear/src/lib.rs
breathx: Initially my comment was about don't set code if it already exists, but do not fail extrinsic.

I see them like:
- Upload code:
   - If code does not exist, we set it and extrinsic OK
   - If code already exist, we don't set it again and extrinsic OK (we will not also charge money for code existence further)
- Upload program:
   - If code exists, we use it and extrinsic OK
   - If code doesn't exists, we set it and extrinsic OK
   - If program exists, then extrinsic ERR

Weight mechanism of frame support also allows us to charge for only happened case (if code wasn't set in upload_program and/or upload code, then funds for setting will be returned, but its for future)

breathx: I see no reason to restrict some extrinsic with it.

Atm imagine the situation,
Two users found some code and decided to upload program with it. They made sure that there is no such code in storage and they use the different salts. Sending it really close in time will make one of the extrinsic fail, that wasn't expected by user. That's strange

breathx: > `Bob` doesn't know why `PROGRAM` with `SALT_BOB` has not been created.

We should continue execution if code exists and should create program with it, charging less money from `Bob` due to absence of code saving

breathx: > > Weight mechanism of frame support also allows us to charge for only happened case (if code wasn't set in upload_program and/or upload code, then funds for setting will be returned, but its for future)
> 
> I believe we can just do this for the weight corrections inside extrinsic
> 
> ```rust
> fn upload_program(..) -> DispatchWithPostInfo {
>     if code_exists_but_created_new_program {
>         return Ok(Some(
>             <T as Config>::WeightInfo::upload_program(0, salt.len() as u32),
>         ).into());
>     }
> }
> ```

Yeah, I meant exactly this way. We just will need two benches of different storage states for the extrinsic

pallets/gear/src/lib.rs
```diff
@@ -1379,112 +1483,79 @@ pub mod pallet {
...
-                Self::deposit_event(Event::CodeChanged {
-                    id: code_hash,
-                    change: CodeChangeKind::Active { expiration: None },
-                });
+            if !T::CodeStorage::exists(code_and_id.code_id()) {
+                // By that call we follow the guarantee that we have in `Self::upload_code` -
+                // if there's code in storage, there's also metadata for it.
+                if let Ok(code_hash) =
+                    Self::set_code_with_metadata(code_and_id, who.clone().into_origin())
+                {
```
SabaunT, line 1503: if code doesn't exist, the failure here must fail the extrinsic

SabaunT (changes requested):

pallets/gear/src/lib.rs
```diff
@@ -1379,112 +1483,79 @@ pub mod pallet {
...
+                    // TODO: replace this temporary (`None`) value
+                    // for expiration block number with properly
+                    // calculated one (issues #646 and #969).
+                    Self::deposit_event(Event::CodeChanged {
+                        id: code_hash,
+                        change: CodeChangeKind::Active { expiration: None },
+                    });
+                }
```
SabaunT, line 1511: I even suggest removing the check `!T::CodeStorage::exists(code_and_id.code_id())` and just calling the `Self::set_code_with_metadata`. Due to logic in latter (returns `Ok(hash)` when there was no code in storage and it was set or returns `Err` when there was a code in storage), no need in that conditionally executed block.

pallets/gear/src/lib.rs
SabaunT: Discussed with @breathx. Need to propagate error here. So please remove ```if let Ok(code_hash)``` and introduce instead ```let code_hash = <call>()?;```.

SabaunT (approved):

PR #1243, opened 2022-07-28
Introduce support for multiple runtimes

### LINKED ISSUE #1116: Stabilize runtime for a standalone chain with PoS support
Run a testnet based on PoS consensus with staking support.
For that:
- create a dedicated runtime;
- enable staking pallet and calibrate it according to our [rough] understanding of the future token economics (inflation rate, validators' reward curve etc.).

ekovalev:
Resolves #1116.

In order to be able to spin up both standalone network as well as a parachain (in the future) we want to maintain different variants of the runtime (think of Polkadot vs. Kusama vs. Westend vs. Rococo etc.).
Different runtimes require different clients and chain specs to configure the genesis block.

This PR attempts to create such support for multiple runtimes and introduce a new runtime for a standalone network codenamed `Vara` network. This standalone chain will eventually have staking. For now it only replaces block authoring consensus with `Babe` (in place of `Aura`) and introduces some basic session management.

In a nutshell, since there isn't anymore "The Runtime" but rather a set of possible runtimes each of which exposes their own set of Runtime Apis, we must also modify the Substrate client part of the node to be able to adjust to a concrete runtime chosen at the time of compilation. That being said, we, copying the approach taken by Polkadot and other Substrate based chains with multiple runtimes (Acala, Moonbeam), introduce traits `AbstractClient` and `RuntimeApiCollection` a client's Api must implement, as well as an enum `Client` whose variants are concrete runtimes (at the moment `gear` which refers to the old test network and `vara` - the standalone chain; in the future we might want `gear` to refer to the parachain runtime; in the meantime, there should also be one for `rococo` relay chain) and would implement the `AbstractClient` and a bunch of other traits the default Substrate client implements. This, in turn, requires the `Service` and `Cli` being changed accordingly.

Last but not least, the client part of the node has been split into several crates (again, following Polkadot, Moonbeam etc.) for the sake of easier maintenance in the future.

A heads-up: because at current point in order to make a transition to PoS smooth we still need to keep the old block authering mechanism based on `Aura` consensus algorithm. This makes the whole client part look a bit messy with lots of conditional compilation macros because we need to have two essentially different runtime Apis (with `Aura` and `Babe`). Once we've tested a standalone chain based on `Babe`, the legacy `Aura` will go away completely so we'll be able to clean up the node client.

The `gear-runtime-test-cli` utility has also been adjusted to multi-runtime code.

The fuzzer is still lagging behind but will be updated soonest possible (either in this PR or one of the following ones).

**Release Notes**:
This PR introduces support for multiple runtimes which allows for having several chains with similar underlying mechanics but possibly different in some concrete details (think of Polkadot vs. Kusama vs. Westend vs. Rococo etc.). For instance, there can be a standalone chain with PoS consensus and staking, whereas another one can implement necessary machinery to be plugged to a relay chain and act as a parachain in the Polkadot ecosystem thus providing access to our computing engine for other network participants by means of XCM system.
These runtimes co-exist all together within the same source code tree so that pallets can be shared between them eliminating the need to maintain separate branches for different chains.


node/primitives/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+repository = "https://github.com/gear-tech/gear"
+rust-version = "1.56"
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
+
+[dependencies]
+# codec = { package = "parity-scale-codec", version = "3.0.0", default-features = false, features = ["derive"] }
+scale-info = { version = "2.1.1", default-features = false, features = ["derive"] }
+# frame-system = { version = "4.0.0-dev", default-features = false, pgit = "https://github.com/gear-tech/substrate.git", branch = "gear-stable" }
```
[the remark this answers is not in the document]
ekovalev: Most likely. Will need to tidy it all up

node/service/src/lib.rs
```diff
@@ -0,0 +1,621 @@
...
+        spec if spec.is_gear() => {
+            chain_ops!(config, gear_runtime, GearExecutorDispatch, Gear)
+        }
+        #[cfg(all(not(feature = "gear-native"), feature = "vara-native"))]
+        spec if spec.is_vara() => {
+            chain_ops!(config, vara_runtime, VaraExecutorDispatch, Vara)
+        }
+        _ => Err("invalid chain spec".into()),
+    }
+    // #[cfg(feature = "gear-native")]
```
breathx, line 132: dead code

pallets/aux/shift-session-manager/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+        });
+
+        pallet_session::GenesisConfig::<TestRuntime> { keys }
+            .assimilate_storage(&mut t)
+            .unwrap();
+        TestExternalities::new(t)
+    }
+
+    #[test]
+    fn shift_session_manager_works() {
```
breathx, line 248: I'm right that the main purpose of this pallet is to provide different validators for different sessions, and following some circular buffer of current set of validators?

breathx: And what will (and should) happen if we add new validator?

pallets/aux/shift-session-manager/src/lib.rs
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    impl pallet_session::Config for TestRuntime {
+        type Event = ();
+        type ValidatorId = <Self as frame_system::Config>::AccountId;
+        type ValidatorIdOf = ConvertInto;
+        type ShouldEndSession = pallet_session::PeriodicSessions<Period, Offset>;
+        type NextSessionRotation = pallet_session::PeriodicSessions<Period, Offset>;
+        type SessionManager = ();
+        type SessionHandler = TestSessionHandler;
```
breathx, line 192: isn't here () impl?

runtime/vara/src/constants.rs
```diff
@@ -0,0 +1,70 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+/// Currency related constants
+pub mod currency {
+    use gear_node_primitives::Balance;
+
+    /// The existential deposit.
+    pub const EXISTENTIAL_DEPOSIT: Balance = MILLICENTS;
+
+    // TODO: review quantities based on economic model
```
breathx, line 26: open babe-related issue, please

runtime/vara/src/lib.rs
```diff
@@ -0,0 +1,780 @@
...
+impl pallet_gear_debug::Config for Runtime {
+    type Event = Event;
+    type WeightInfo = pallet_gear_debug::weights::GearSupportWeight<Runtime>;
+    type CodeStorage = GearProgram;
+    type Messenger = GearMessenger;
+}
+
+impl pallet_gear_scheduler::Config for Runtime {
+    type BlockLimiter = GearGas;
+    type ReserveThreshold = ConstU32<1>;
```
breathx, line 366: Such things may be also be stored in primitives as new best practice:
```suggestion
    type ReserveThreshold = ConstU32<gear_primitives::ReserveThreshold>;
```

scripts/src/build.sh
```diff
@@ -36,7 +36,7 @@ gear_test_build() {
 }
 
 node_build() {
-  cargo build -p gear-node "$@"
+  cargo build -p gear-node $@
```
breathx, line 39: we should always use quotes, passing such args in shell

utils/gear-runtime-test-cli/Cargo.toml
```diff
@@ -15,24 +15,34 @@ serde_json = "1.0.82"
...
 pallet-gear = { path = "../../pallets/gear" }
 pallet-gear-gas = { path = "../../pallets/gas" }
 pallet-gear-debug = { path = "../../pallets/gear-debug" }
 gear-core-processor = { path = "../../core-processor" }
 junit-common = { path = "../junit-common" }
+
+[features]
+gear-native = ["gear-runtime", "sp-consensus-aura", "authoring-aura"]
+vara-native = ["vara-runtime", "sp-consensus-babe", "sp-finality-grandpa"]
+authoring-aura = []
```
breathx, line 48: do we need separate flag for aura, if gear uses aura and vara uses babe?

breathx (approved):
Really huge work have been done!

I'm a bit confused with aux pallet and why does it excluded from overall workspace.
I really hope to see some default constants lib or module inside primitives. (It may also contain our override for extrinsic fee, maybe by some feature)

Also there is a place for a lot of future issues and considering moments: so, please, open them.

Great! ⚙️

scripts/src/build.sh
ekovalev: Surprisingly, it didn't work for me with the quotes so I removed them. Not so well-versed with shell scripting, it's possible that the reason it didn't work for me doesn't have to do with quotes. Will check again

utils/gear-runtime-test-cli/Cargo.toml
ekovalev: It's helpful to highlight mutually exclusive features. That is if both `gear-native` and `vara-native` are enabled we use `aura` and avoid importing the same definitions from conflicting packages. In some cases there's no conflict though

ekovalev:
> Really huge work have been done!
> 
> I'm a bit confused with aux pallet and why does it excluded from overall workspace. I really hope to see some default constants lib or module inside primitives. (It may also contain our override for extrinsic fee, maybe by some feature)
> 
> Also there is a place for a lot of future issues and considering moments: so, please, open them.
> 
> Great! ⚙️

Aux is not a pallet but rather an indicator that whatever is inside is not a real pallet but some ad-hoc solution. In this case it only has this custom-made deterministic session manager which will be replaced by a proper session manager once we add the `pallet_staking` to the runtime at the next iteration (because it depends on it). It's excluded from the workspace because cargo fails to find its `Cargo.toml` (which doesn't exist).

node/cli/src/benchmarking.rs
```diff
@@ -0,0 +1,294 @@
...
+				$( $usage )*
+			},
+		}
+	}
+}
+
+/// Generates a `SignedPayload` for the Kusama, Westend and Rococo runtime.
+///
+/// Should only be used for benchmarking as it is not tested for regular usage.
+#[allow(unused_macros)]
```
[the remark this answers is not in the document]
ekovalev: indeed, there's no need for it anymore - removed

node/cli/src/command.rs
```diff
@@ -0,0 +1,318 @@
...
+};
+use frame_benchmarking_cli::{BenchmarkCmd, ExtrinsicFactory, SUBSTRATE_REFERENCE_HARDWARE};
+use gear_node_primitives::Block;
+use sc_cli::{ChainSpec, RuntimeVersion, SubstrateCli};
+use service::{chain_spec, IdentifyVariant};
+use sp_keyring::Sr25519Keyring;
+
+impl SubstrateCli for Cli {
+    fn impl_name() -> String {
+        "Gear Node".into()
```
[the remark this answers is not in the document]
ekovalev: I don't think it's necessary: the resulting binary (and the command name that runs a node) is called `gear-node`regardless of the client (runtime) variant anyway

node/primitives/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "gear-node-primitives"
+version = "1.0.0"
+authors = ["Gear Technologies"]
+description = "Gear Node primitives"
+edition = "2018"
+license = "GPL-3.0"
+homepage = "https://gear-tech.io"
+repository = "https://github.com/gear-tech/gear"
+rust-version = "1.56"
```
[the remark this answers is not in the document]
ekovalev: Some old code legacy. Removed

node/service/Cargo.toml
```diff
@@ -0,0 +1,116 @@
+[package]
+name = "gear-service"
+version = "0.1.0"
+authors = ["Gear Technologies"]
+description = "Gear Node"
+edition = "2018"
+license = "GPL-3.0"
+homepage = "https://gear-tech.io"
+repository = "https://github.com/gear-tech/gear"
+rust-version = "1.56"
```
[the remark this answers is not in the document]
ekovalev: Good call. Removed everywhere.

pallets/aux/shift-session-manager/src/lib.rs
ekovalev: This is a deterministic algorithm to select a certain fraction of pre-defined set of authorities as block producers for the next epoch. Essentially, this is the same old PoA. It's a temporary session management implementation used with `Babe` until a proper authorities selection based on staking is introduced. Then it will be possible to add validators.

In this simplistic use case we do not expect any equivocation or anything of that sort, so I think we should just make sure that exactly one authority is selected for each epoch.

runtime/vara/src/lib.rs
ekovalev: This constant can be runtime-dependent, that is differ from runtime to runtime. I would suggest to define it in the `constants` module instead.
In primitives we collect some very basic generic concepts used in definitions of higher-level apis. Here I'm not inventing anything new - just copied the way it's done in Substrate and other chains.

pallets/aux/shift-session-manager/src/lib.rs
ekovalev: No, the `SessionHandler` trait is not implemented for `()`, hence this `TestSessionHandler`

runtime/vara/src/constants.rs
ekovalev: Done

runtime/vara/src/lib.rs
breathx: Yeah, I understand, but atm our constants for different runtimes are same and I see no reason to copy-paste and define them every time. But the `default_constants` mod seems ok for me.
I'll left it up to you so.

ekovalev: Moved as much of the common stuff from both runtimes as possible in a separate crate: `gear-runtime-common`. This reduces "copy-pasting" to the minimum.

runtime/gear/src/lib.rs
```diff
@@ -223,13 +167,17 @@ impl frame_system::Config for Runtime {
     type SS58Prefix = SS58Prefix;
     /// The set code logic, just the default since we're not a parachain.
     type OnSetCode = ();
-    type MaxConsumers = frame_support::traits::ConstU32<16>;
+    type MaxConsumers = ConstU32<16>;
+}
+
+parameter_types! {
+    pub const MaxAuthorities: u32 = 32;
 }
```
breathx, line 175:
```suggestion
const MAX_CONSUMERS: u32 = 16;
{    
    // .. // 
    
    type MaxConsumers = ConstU32<MAX_CONSUMERS>;
}

const MAX_AUTHORITIES: u32 = 32;

// .. //
impl pallet_aura::Config for Runtime {
     type AuthorityId = AuraId;
     type DisabledValidators = ();
     type MaxAuthorities = ConstU32<MAX_AUTHORITIES>;
 }
```

scripts/src/test.sh
```diff
@@ -120,5 +130,5 @@ run_fuzzer() {
 
   # Run fuzzer
   RUST_LOG="essential,pallet_gear=debug,gear_core_processor::executor=debug,economic_checks=debug,gwasm=debug" \
-  cargo fuzz run --release --sanitizer=none $TARGET
+  cargo fuzz run --release $FEATURES --sanitizer=none $TARGET
```
breathx, line 133:
```suggestion
  cargo fuzz run --release "$FEATURES" --sanitizer=none "$TARGET"
```

breathx (approved):

runtime/gear/src/lib.rs
ekovalev: Not sure this is better. What problem would it solve?
I would rather put `MaxConsumers` into `parameter_types!` macro just for the sake of uniformity. I guess it's more about personal preference than actual optimization, is it not?

PR #1246, opened 2022-07-28
Remove options from `GasTree` interface

breathx:
- Removes optional Ok clauses to simplify interface: inexistent node query causes error::node_not_found now
- `gear_common` crate docs formatted (that's why PR has many changes)
- Code adapted, tests formatted
- Added wrapped interface to query exactly gas_limit without gasful node, which provides it.
- Fixed test with known errors for gas tree


common/src/gas_provider/property_tests/assertions.rs
```diff
@@ -110,32 +119,35 @@ fn assert_removed_nodes_form_path(
...
-// If it was removed after a new `consume` call, then all the tree must be empty. So no nodes can be removed
-// after root was removed in the `consume` call.
+//
+// That is done the following way: each time `consume` procedure is called we
+// check `root_node` for existence. If it was removed after a new `consume`
+// call, then all the tree must be empty. So no nodes can be removed after
+// root was removed in the `consume` call.
 pub(super) fn assert_root_removed_last(root_node: Key, remaining_nodes: RemainingNodes) {
-    if Gas::get_node(root_node).is_none() {
+    Gas::get_node(root_node).is_none().then(|| {
```
SabaunT, line 128: Sure it's needed? Does ```.then``` computed lazily?

SabaunT (approved):
👍

common/src/gas_provider/property_tests/assertions.rs
breathx: I can't stop myself from using it 😆

Yeah, sure. `then_some` computes instantly, `then` is lazy variant, is any other in Rust, which takes closure as argument

shamilsan: Shouldn't we use `Option::or_else` instead of `Option::is_none` + `bool::then`?

gshep (approved):

common/src/gas_provider/property_tests/assertions.rs
shamilsan: Anyway, I don't see the reason for this modification. `bool::then/then_some` should be used to convert `bool` to `Option`. Here there is no benefit when using it.

breathx: Reverted to `if` statement

PR #1252, opened 2022-07-29
add wat examples and specs to run them

grishasobol:
1) Add read access and r/w access wat examples.
2) Allow to skip `expected` and `messages` fields in fixtures.
3) A little refactoring in gear-test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,10 @@ all: gear examples
 .PHONY: all-release
 all-release: gear-release examples
 
+.PHONY: build-wat-examples
+build-wat-examples:
+	@ ./scripts/gear.sh build wat-examples
+
 .PHONY: gear
 gear:
 	@ ./scripts/gear.sh build gear
@@ -246,5 +250,3 @@ doc:
 fuzz:
 	@ ./scripts/gear.sh test fuzz
 
-build-wat-examples:
-	@ ./scripts/gear.sh build wat-examples
--- a/Makefile
+++ b/Makefile
@@ -245,3 +245,6 @@ doc:
 .PHONY: fuzz
 fuzz:
 	@ ./scripts/gear.sh test fuzz
+
+build-wat-examples:
+	@ ./scripts/gear.sh build wat-examples
--- a/examples/wat-examples/read_access.wat
+++ b/examples/wat-examples/read_access.wat
@@ -1,0 +1,36 @@
+(module
+    (import "env" "memory" (memory 0))
+    (import "env" "alloc" (func $alloc (param i32) (result i32)))
+    (export "init" (func $init))
+    (func $init
+        (local $i i32)
+
+        ;; alloc 0x100 pages, so mem pages are: 0..=0xff
+        (block
+            i32.const 0x100
+            call $alloc
+            i32.eqz
+            br_if 0
+            unreachable
+        )
+
+        ;; make read for all gear pages
+        (loop
+            local.get $i
+            i32.const 1
+            i32.add
+            local.set $i
+
+            local.get $i
+            i32.const 0x1000
+            i32.mul
+            i32.load
+            drop
+
+            local.get $i
+            i32.const 0xfff
+            i32.ne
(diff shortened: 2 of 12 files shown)


utils/gear-runtime-test-cli/src/command.rs
```diff
@@ -261,7 +261,8 @@ fn run_fixture(test: &'_ sample::Test, fixture: &sample::Fixture) -> ColoredStri
     let mut errors = vec![];
     let mut expected_log = vec![];
 
-    for message in &fixture.messages {
+    let empty = Vec::new();
+    for message in fixture.messages.as_ref().unwrap_or(&empty) {
```
ukint-vs, line 265:
```suggestion
    for message in fixture.messages.as_ref().unwrap_or(&Vec::new()) {
```

ukint-vs: Change this in other places

ukint-vs (approved):

utils/gear-runtime-test-cli/src/command.rs
grishasobol: This can not be done, because we cannot return ref to the temprorary object which cannot outlive its context

Makefile
```diff
@@ -245,3 +245,6 @@ doc:
 .PHONY: fuzz
 fuzz:
 	@ ./scripts/gear.sh test fuzz
+
+build-wat-examples:
```
breathx, line 249: please, move to "build" section to keep things together

gear-test/src/proc.rs
```diff
@@ -204,7 +204,8 @@ where
 
     let fixture = &test.fixtures[fixture_no];
 
-    for message in &fixture.messages {
+    let empty = Vec::new();
+    for message in fixture.messages.as_ref().unwrap_or(&empty) {
```
breathx, line 208:
```suggestion
    for message in &fixture.messages.unwrap_or_default() {
```

utils/gear-runtime-test-cli/src/command.rs
breathx: Same as meant above

utils/gear-runtime-test-cli/src/command.rs
```diff
@@ -372,7 +373,8 @@ fn run_fixture(test: &'_ sample::Test, fixture: &sample::Fixture) -> ColoredStri
         }
     }
 
-    for exp in &fixture.expected {
+    let empty = Vec::new();
+    for exp in fixture.expected.as_ref().unwrap_or(&empty) {
```
breathx, line 377: and here

breathx (approved):

utils/gear-runtime-test-cli/src/command.rs
grishasobol: I've already answered

gear-test/src/proc.rs
grishasobol:
```
cannot move out of `fixture.messages` which is behind a shared reference
move occurs because `fixture.messages` has type `std::option::Option<Vec<sample::Message>>`, which does not implement the `Copy` trait
```

Or if like this:
```rust
    for message in fixture.messages.as_ref().unwrap_or_default() {
```
then error is
```
the trait bound `&Vec<sample::Message>: Default` is not satisfied
the trait `Default` is implemented for `Vec<T>`
```

The only one which works, is my variant, or `clone`: 
```rust
    for message in &fixture.messages.clone().unwrap_or_default() {
```

examples/wat-examples/read_access.wat
```diff
@@ -0,0 +1,36 @@
+(module
```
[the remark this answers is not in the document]
grishasobol: hmmm, what the reason ? Maybe better simple `examples/wats`

Makefile
grishasobol: Done

PR #1261, opened 2022-07-30
Build runtime WASM and node binary using production profile

shamilsan:


ukint-vs (approved):

NikVolf:
There should be a test that ensures that released wasm blob has no imports starting with "env::ext_benchmarking_"

Issue #641, opened 2022-02-09
Gas reservation inside program


breathx:
### Problem to Solve

Need to provide an ability to reserve some gas from given to use it for future actions inside the program (even over multiple execution, but it's relies on user-space).

It may look like this:

```rust
use gcore::exec; // or any other module

#[no_mangle]
pub unsafe extern "C" fn handle() {
    let reservation_id = exec::reserve_gas(5_000);
}
```

Using this stored reservation_id program will be able to use that gas.

### Possible Solution

Need to expand `ValueTree` functionality to be able split off not only for messages, but for reservations also.

breathx:
Fixed in #1505

PR #1272, opened 2022-08-01
Don't use string repr of reason in tests

### LINKED ISSUE #647: System signals mechanism
### Problem to Solve

There might be cases when something happen on the upper layer (e.g. in the `node`) and contract should change it's state somehow.

Example cases:
- Program waits for reply (stores future in persistent pages), but reply throws away from wait-list, because can't pay rent for holding there. Program need to somehow clear the memory, wake some messages etc. (the most obvious use-case for me atm.)

- #349 may be done with that logic

### Possible Solution

Provide new entry point, which will parse "signals" **only from system, not other actors**.
I fount this concept pretty similar to OS sigs.

No syscalls available inside, except `gr_read` (and probably `gr_wake`) to get the input.

Provide `gstd` and `gcore` default Rust realization for it.

ark0f:
Resolves #647.

Issue cannot be fully resolved until 

- [ ] #641

is done because we need to have some gas to call entry


SabaunT (approved):

core/src/ids.rs
```diff
@@ -193,6 +193,9 @@ impl MessageId {
 declare_id!(ProgramId: "Program identifier");
 
 impl ProgramId {
+    /// System program ID
+    pub const SYSTEM: Self = Self([0; HASH_LENGTH]);
```
SabaunT, line 197: you see, zero address is widely used in tests. What about making system address from b'gear' bytes (8 times, so it's going to be 32 bytes)

ark0f: Done

shamilsan (approved):

pallets/gear/src/tests.rs
```diff
@@ -4554,6 +4555,59 @@ fn missing_handle_is_not_executed() {
...
+            .into_signal_dispatch(pid, message_id)
+            .into_stored();
+
+        assert_ok!(CurrencyOf::<Test>::reserve(&USER_1, 50_000_000_000));
+        assert_ok!(GasHandlerOf::<Test>::create(
+            USER_1,
+            trap_reply.id(),
+            50_000_000_000
+        ));
+        assert_ok!(QueueOf::<Test>::queue(trap_reply.clone()));
```
breathx, line 4600: manual queueing in tests is evil

breathx (changes requested):

core/src/message/signal.rs
```diff
@@ -0,0 +1,82 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use crate::{
+    ids::{MessageId, ProgramId},
+    message::{Dispatch, DispatchKind, ExitCode, Message, Payload, ReplyDetails},
+};
+use codec::{Decode, Encode};
+use scale_info::TypeInfo;
+
+/// Message for Reply entry point.
```
breathx, line 26: Incorrect docs

core/src/message/signal.rs
```diff
@@ -0,0 +1,82 @@
...
+    /// Message payload.
+    payload: Payload,
+    /// Reply exit code.
+    exit_code: ExitCode,
+}
+
+impl SignalMessage {
+    /// Creates a new [`SignalMessage`].
+    pub fn new(origin_msg_id: MessageId, payload: Payload, exit_code: ExitCode) -> Self {
+        let id = MessageId::generate_reply(origin_msg_id, exit_code);
```
breathx, line 41: I'd provide different MessageId builder for signals

core/src/message/signal.rs
```diff
@@ -0,0 +1,82 @@
...
+use scale_info::TypeInfo;
+
+/// Message for Reply entry point.
+/// [`ReplyMessage`] is unique because of storing [`MessageId`] from message on what it replies, and can be the only one per some message execution.
+#[derive(Clone, Default, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Decode, Encode, TypeInfo)]
+pub struct SignalMessage {
+    /// Message id.
+    id: MessageId,
+    /// Message payload.
+    payload: Payload,
```
breathx, line 33: I see no reason in payload here.

Signal will have just message id it associates with and exit code

examples/binaries/signal-entry/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    extern crate std;
+
+    use gtest::{Program, System};
+
+    #[test]
+    fn program_can_be_initialized() {
```
breathx, line 46: incorrect test name

gtest/src/program.rs
```diff
@@ -88,6 +88,7 @@ pub trait WasmProgram: Debug {
     fn init(&mut self, payload: Vec<u8>) -> Result<Option<Vec<u8>>, &'static str>;
     fn handle(&mut self, payload: Vec<u8>) -> Result<Option<Vec<u8>>, &'static str>;
     fn handle_reply(&mut self, payload: Vec<u8>) -> Result<Option<Vec<u8>>, &'static str>;
+    fn handle_signal(&mut self, payload: Vec<u8>) -> Result<Option<Vec<u8>>, &'static str>;
```
breathx, line 91: atm we will deprecate anything except `gr_reply_to` and `gr_exit_code` for `handle_signal` entry, so there should be `Result<(), &'static str` and backend requires changes in checking context of execution as it did for mentioned above syscalls

breathx (changes requested):

core/src/message/signal.rs
ark0f: How we supposed to send trap reason to user then?

breathx: > How we supposed to send trap reason to user then?

For me seems reasonable that "signal" identifies itself via just code (exit_code)

breathx: And reply_to identifies which message produced this signal

breathx (approved):
@shamilsan, please, review

examples/binaries/signal-entry/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+#[cfg(feature = "std")]
+mod code {
+    include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
+}
+
+#[cfg(feature = "std")]
+pub use code::WASM_BINARY_OPT as WASM_BINARY;
+
+#[no_mangle]
+unsafe extern "C" fn init() {}
```
shamilsan, line 32: Empty `init` can be omitted.

shamilsan (approved):
Brilliant 💎

examples/binaries/signal-entry/src/lib.rs
ark0f: `error: WASM module doesn't contain required export funtions (init/handle) after optimized `
And this is logical because there is no signal without actual erroneous logic

shamilsan: It is weird as we support WASMs without `init` function. But ok, need to add an issue about it.

Issue #1173, opened 2022-07-11
Maximal duration in waitlist


breathx:
### Problem to Solve

#1135's merge brings one problem: there may be no money in gas handler of unspecified node

### Possible Solution

In fact, steps to solve:
- [x] Add ability to lock some value in unspecified nodes
- [x] Cover this feature with tests

---

- [x] Add maximal duration in waitlist, core, syscalls and libs.
- [x] Lock money by meant above functionality from unspecified nodes once added into waitlist.

### Notes

This is key-feature to solve #349.

PR #1274, opened 2022-08-02
Gas locking ability

breathx:
First part of #1173

### DIFF
--- a/common/src/gas_provider/internal.rs
+++ b/common/src/gas_provider/internal.rs
@@ -640,20 +640,29 @@ where
         Self::create_from_with_value(key, new_key, amount, true)
     }
 
-    // TODO (breathx): doc this.
     fn lock(key: Self::Key, amount: Self::Balance) -> Result<(), Self::Error> {
+        // Taking node to lock into.
         let node = Self::get_node(key).ok_or_else(InternalError::node_not_found)?;
 
+        // Validating node type to be able to lock.
         if node.is_reserved_local() {
             return Err(InternalError::forbidden().into());
         }
 
+        // Validating that node is not consumed.
         if node.is_consumed() {
             return Err(InternalError::node_was_consumed().into());
         }
 
+        // Quick quit on queried zero lock.
+        if amount.is_zero() {
+            return Ok(());
+        }
+
+        // Taking value provider for this node.
         let (mut upstream_node, node_id) = Self::node_with_value(node)?;
 
+        // Mutating value of provider.
         let upstream_node_value = upstream_node
             .value_mut()
             .ok_or_else(InternalError::unexpected_node_type)?;
@@ -664,6 +673,8 @@ where
 
         *upstream_node_value = upstream_node_value.saturating_sub(amount);
 
+        // If provider is a parent, we save it to storage, otherwise mutating
+        // current node further, saving it afterward.
         let mut node = if let Some(upstream_id) = node_id {
             StorageMap::insert(upstream_id, upstream_node);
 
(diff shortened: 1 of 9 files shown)


common/src/gas_provider/internal.rs
```diff
@@ -629,6 +640,100 @@ where
...
+
+        Ok(())
+    }
+
+    fn get_lock(key: Self::Key) -> Result<Self::Balance, Self::Error> {
+        let node = Self::get_node(key).ok_or_else(InternalError::node_not_found)?;
+
+        let lock = node.lock().ok_or_else(InternalError::forbidden)?;
+
+        Ok(lock)
```
gshep, line 734: Looks like the returned value can be used directly:
```suggestion
        let node = Self::get_node(key).ok_or_else(InternalError::node_not_found)?;

        node.lock().ok_or_else(InternalError::forbidden)
```

breathx: Yeah, I've tried, but it looks not so good:
```rust
node
    .lock()
    .ok_or_else(|| InternalError::forbidden().into())
```
OR
```rust
node
    .lock()
    .ok_or_else(InternalError::forbidden)
    .map_err(Into::into)
```

common/src/gas_provider/mod.rs
```diff
@@ -176,6 +176,29 @@ pub trait Tree {
...
+    /// This can't create imbalance as no value is burned or created.
+    fn lock(key: Self::Key, amount: Self::Balance) -> Result<(), Self::Error>;
+
+    /// Unlocking some value from node's locked balance.
+    ///
+    /// If `key` does not identify any value or the `amount` exceeds what's
+    /// locked under that key, an error is returned.
+    ///
+    /// This can't create imbalance as no value is burned or created.
+    fn unlock(key: Self::Key, amount: Self::Balance) -> Result<(), Self::Error>;
```
SabaunT, line 193: could you clarify please conditions of unlocking value? You see, currently it's the only function that can block 1) value move to user and 2) tree consumption.

SabaunT (changes requested):
Pretty neat :+1: . Please, refer to minor comments and introduce some additional property tests:
1. **deleted** nodes have 0 value locked
2. `max_balance = spent + rest_value + caught + rest_locked` (look here https://github.com/gear-tech/gear/blob/502e0b25b92f590106085464929208e43bdd7577/common/src/gas_provider/property_tests/mod.rs#L487)
3. Node's value/lock change during node's lifetime applies the next rule: `node_balance_when_created = current_value + current_lock_including_unspec_children + spent`. Maybe it's not needed, just suggesting.

common/src/gas_provider/internal.rs
```diff
@@ -629,6 +640,133 @@ where
...
+        }
+
+        *upstream_node_value = upstream_node_value.saturating_sub(amount);
+
+        // If provider is a parent, we save it to storage, otherwise mutating
+        // current node further, saving it afterward.
+        let mut node = if let Some(upstream_id) = node_id {
+            StorageMap::insert(upstream_id, upstream_node);
+
+            Self::get_node(key).ok_or_else(InternalError::node_not_found)?
```
SabaunT, line 681: unreachable, right?

common/src/gas_provider/internal.rs
```diff
@@ -629,6 +640,133 @@ where
...
+        // Mutating locked value of queried node.
+        let node_lock = node
+            .lock_mut()
+            .ok_or_else(InternalError::unexpected_node_type)?;
+
+        if *node_lock < amount {
+            return Err(InternalError::insufficient_balance().into());
+        }
+
+        *node_lock = node_lock.saturating_sub(amount);
```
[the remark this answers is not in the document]
SabaunT: errors after the subtraction and adding are invariant errors, so I guess it's ok.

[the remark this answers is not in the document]
breathx: Agree, these errors are guaranteed to do not appear, if everything goes well (in cases of tree invalidation, there are much more problems then not atomic math)

common/src/gas_provider/internal.rs
breathx: Sure. Queried the same above. Clone not used just to avoid excess cloning.

common/src/gas_provider/mod.rs
```diff
@@ -176,6 +176,29 @@ pub trait Tree {
...
+
+    /// Get locked value associated with given id.
+    ///
+    /// Returns errors in cases of absence associated with given key node,
+    /// or if such functionality is forbidden for specific node type:
+    /// for example, for `GasNode::ReservedLocal`.
```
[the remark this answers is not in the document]
breathx: If function failed or not allowed, I see no reason to create an illusion into final api usage, that lock gone well.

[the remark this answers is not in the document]
SabaunT: ok, nice point

common/src/gas_provider/mod.rs
SabaunT: discussed

common/src/gas_provider/internal.rs
breathx: Expects and any execution-breakers should be called on higher layers

I can just comment here, that this is unreachable error

SabaunT (approved):

breathx:
#1286 for future updates, @SabaunT

PR #1275, opened 2022-08-02
remove pages compare after execution

### LINKED ISSUE #1123: check page has been changed in sig-handler
### Problem to Solve

Currently core processor checks each accessed during execution page : whether it has been changed or not. To do this it must completely compare each byte in page. This may be costly, for example if contract will access many pages during execution. 

See also in #1048  about this.

### Possible Solution

We can decrease amount of pages to check , if we identify pages which has been accessed by store instruction.
Currently pages, which has been accessed are marked in lazy-pages signal handler. Unfortunately we cannot identify in signal handler whether it is write or read access. But we can remove only read access for first signal handling for page, then if second signal from this page appear it means that this signal is from store instruction. So, after that we can mark pages which has been accessed by `store` and compare only them after execution.

This solution brings some overhead, so before do it we also should check, whether there is a lot of pages, which is accessed only by `load` instruction during execution and isn't accessed by `store` after that.

### Notes

_No response_

grishasobol:
Resolves #1123

1) In lazy pages mode we upload page data to the storage in case it has been `write` accessed. Old logic is to compare pages data before and after execution and upload data if it has been changed. So, now in lazy pages mode comparing is removed. In executor I leave logic with comparing in case we works without lazy-pages: gtest/gear-test.
Removing pages comparing brings some performance changes. I've measured how execution time for function `gear_core_processor::execute_wasm` is changed for wat examples:
```
read_access.wat T(master)/T(patch) = 1.63
read_write_access.wat T(master)/T(patch) = 1
```
As you can see `read_write` access has no performance changes - this is because we must handle signal twice: one for read and second for write. This lead to 3 times more mprotection calls. 

2) Add page storage granularity:
```rust
/// Pages data storage granularity (PSG) is a size and wasm addr alignemnt
/// of a memory interval, for which the following conditions must be met:
/// if some gear page has data in storage, then all gear
/// pages, that are in the same granularity interval, must contain
/// data in storage. For example:
/// ````
///   granularity interval no.0       interval no.1        interval no.2
///                |                    |                    |
///                {====|====|====|====}{====|====|====|====}{====|====|====|====}
///               /     |     \
///    gear-page 0    page 1   page 2 ...
/// ````
/// In this example each PSG page contains 4 gear-pages. So, if gear-page `2`
/// has data in storage, then gear-page `0`,`1`,`3` also has data in storage.
/// This constant is necessary for consensus between nodes with different
/// native page sizes. You can see an example of using in crate `gear-lazy-pages`.

/// [PAGE_STORAGE_GRANULARITY] (PSG) case - if page is write accessed
/// first time in program live, then this page has no data in storage yet.
/// This also means that all pages from the same PSG interval has no data in storage.
/// So, in this case we have to insert in `released_lazy_pages` all pages from the same
/// PSG interval, in order to upload their data to storage later in runtime.
/// We have to make separate logic for this case in order to support consensus
/// between nodes with different native page sizes. For example, if one node
/// has native page size 4kBit and other 16kBit, then (without PSG logic)
/// for first one gear page will be uploaded and for second 4 gear pages.
/// This can cause conflicts in data about pages that have data in storage.
/// So, to avoid this we upload all pages from PSG interval (which is 16kBit now),
/// and restrict to run node on machines, that have native page number bigger than PSG.
```

3) Fix and add test cases.

### Tested
1) Test that if one node change data for some `key` in storage to the same value, which was before. And other node does not do this in the same block finalisation, then this nodes generates equal blocks and consensus is not failed between them.
2) Test that this patch works on chain in old test-net and in new test-net

### IMPORTANT
In case of runtime upgrade to current patch - need to rebuild nodes in test net

### DIFF
--- a/common/src/lazy_pages.rs
+++ b/common/src/lazy_pages.rs
@@ -19,13 +19,16 @@
 //! Lazy pages support runtime functions
 
 use crate::Origin;
-use core::fmt;
+use core::{convert::TryFrom, fmt};
 use gear_core::{
     ids::ProgramId,
     memory::{HostPointer, Memory, PageBuf, PageNumber, WasmPageNumber},
 };
 use gear_runtime_interface::{gear_ri, RIError};
 use sp_std::{collections::btree_map::BTreeMap, vec::Vec};
+
+extern crate alloc;
+use alloc::string::ToString;
 
 #[derive(Debug, Clone, PartialEq, Eq, derive_more::Display)]
 pub enum Error {
@@ -86,46 +89,26 @@ pub fn is_lazy_pages_enabled() -> bool {
 
 /// Protect and save storage keys for pages which has no data
 pub fn protect_pages_and_init_info(mem: &impl Memory, prog_id: ProgramId) -> Result<(), Error> {
-    gear_ri::reset_lazy_pages_info();
+    let program_prefix = crate::pages_prefix(prog_id.into_origin());
+    let wasm_mem_addr = mem.get_buffer_host_addr();
+    let wasm_mem_size =
+        u32::try_from(mem.size().offset()).map_err(|_| Error::WasmMemorySizeOverflow)?;
 
-    let prog_prefix = crate::pages_prefix(prog_id.into_origin());
-    gear_ri::set_program_prefix(prog_prefix);
+    // Cannot panic unless OS allocates buffer in not aligned by native page addr, or
+    // something goes wrong with pages protection.
+    // TODO: currently set stack pages as None, should be resolved (issue #1253).
+    gear_ri::initilize_for_program(wasm_mem_addr, wasm_mem_size, None, program_prefix)
+        .map_err(|err| err.to_string())
+        .expect("Cannot initilize lazy pages for current program");
 
-    if let Some(addr) = mem.get_buffer_host_addr() {
-        // Cannot panic, unless OS allocates wasm mem buffer
-        // in not aligned by native page addr.
-        gear_ri::set_wasm_mem_begin_addr(addr).expect("Cannot set wasm mem addr");
-    } else {
-        return Ok(());
-    }
-
-    let size = mem
-        .size()
-        .0
-        .checked_add(1)
-        .ok_or(Error::WasmMemorySizeOverflow)?
-        .checked_mul(WasmPageNumber::size() as u32)
-        .ok_or(Error::WasmMemorySizeOverflow)?;
-    gear_ri::set_wasm_mem_size(size)?;
-
-    mprotect_lazy_pages(mem, true)
+    Ok(())
 }
 
 /// Lazy pages contract post execution actions
 pub fn post_execution_actions(
     mem: &impl Memory,
-    pages_data: &mut BTreeMap<PageNumber, PageBuf>,
+    _pages_data: &mut BTreeMap<PageNumber, PageBuf>,
 ) -> Result<(), Error> {
-    // Loads data for released lazy pages. Data which was before execution.
-    let released_pages = gear_ri::get_released_pages();
-    for page in released_pages {
-        let data = gear_ri::get_released_page_old_data(page)
-            .ok_or_else(|| Error::ReleasedPageHasNoData(page.into()))?;
-        if pages_data.insert(page.into(), data).is_some() {
-            return Err(Error::ReleasedPageHasInitialData(page.into()));
-        }
-    }
-
     // Removes protections from lazy pages
     mprotect_lazy_pages(mem, false)
 }
--- a/common/src/lazy_pages.rs
+++ b/common/src/lazy_pages.rs
@@ -19,6 +19,7 @@
 //! Lazy pages support runtime functions
 
 use crate::Origin;
+use alloc::string::ToString;
 use core::fmt;
 use gear_core::{
     ids::ProgramId,
@@ -78,9 +79,9 @@ pub fn protect_pages_and_init_info(mem: &impl Memory, prog_id: ProgramId) -> Res
     // Cannot panic unless OS allocates buffer in not aligned by native page addr, or
     // something goes wrong with pages protection.
     // TODO: currently set stack pages as None, should be resolved (issue #1253).
-    gear_ri::initilize_for_program(wasm_mem_addr, wasm_mem_size.0, None, program_prefix)
+    gear_ri::initialize_for_program(wasm_mem_addr, wasm_mem_size.0, None, program_prefix)
         .map_err(|err| err.to_string())
-        .expect("Cannot initilize lazy pages for current program");
+        .expect("Cannot initialize lazy pages for current program");
 
     Ok(())
 }
--- a/common/src/lazy_pages.rs
+++ b/common/src/lazy_pages.rs
@@ -22,10 +22,10 @@ use crate::Origin;
 use core::{convert::TryFrom, fmt};
 use gear_core::{
     ids::ProgramId,
-    memory::{HostPointer, Memory, PageBuf, PageNumber, WasmPageNumber},
+    memory::{HostPointer, Memory, PageNumber, WasmPageNumber},
 };
 use gear_runtime_interface::{gear_ri, RIError};
-use sp_std::{collections::btree_map::BTreeMap, vec::Vec};
+use sp_std::vec::Vec;
 
 extern crate alloc;
 use alloc::string::ToString;
@@ -105,10 +105,7 @@ pub fn protect_pages_and_init_info(mem: &impl Memory, prog_id: ProgramId) -> Res
 }
 
 /// Lazy pages contract post execution actions
-pub fn post_execution_actions(
-    mem: &impl Memory,
-    _pages_data: &mut BTreeMap<PageNumber, PageBuf>,
-) -> Result<(), Error> {
+pub fn post_execution_actions(mem: &impl Memory) -> Result<(), Error> {
     // Removes protections from lazy pages
     mprotect_lazy_pages(mem, false)
 }
--- a/common/src/lazy_pages.rs
+++ b/common/src/lazy_pages.rs
@@ -91,13 +91,12 @@ pub fn is_lazy_pages_enabled() -> bool {
 pub fn protect_pages_and_init_info(mem: &impl Memory, prog_id: ProgramId) -> Result<(), Error> {
     let program_prefix = crate::pages_prefix(prog_id.into_origin());
     let wasm_mem_addr = mem.get_buffer_host_addr();
-    let wasm_mem_size =
-        u32::try_from(mem.size().offset()).map_err(|_| Error::WasmMemorySizeOverflow)?;
+    let wasm_mem_size = mem.size();
 
     // Cannot panic unless OS allocates buffer in not aligned by native page addr, or
     // something goes wrong with pages protection.
     // TODO: currently set stack pages as None, should be resolved (issue #1253).
-    gear_ri::initilize_for_program(wasm_mem_addr, wasm_mem_size, None, program_prefix)
+    gear_ri::initilize_for_program(wasm_mem_addr, wasm_mem_size.0, None, program_prefix)
         .map_err(|err| err.to_string())
         .expect("Cannot initilize lazy pages for current program");
 
@@ -146,8 +145,7 @@ pub fn update_lazy_pages_and_protect_again(
 
     let new_mem_size = mem.size();
     if new_mem_size > old_mem_size {
-        let size = get_memory_size_in_bytes(new_mem_size)?;
-        gear_ri::set_wasm_mem_size(size)?;
+        gear_ri::set_wasm_mem_size(new_mem_size.0);
     }
 
     mprotect_lazy_pages(mem, true)
--- a/common/src/lazy_pages.rs
+++ b/common/src/lazy_pages.rs
@@ -19,7 +19,7 @@
 //! Lazy pages support runtime functions
 
 use crate::Origin;
-use core::{convert::TryFrom, fmt};
+use core::fmt;
 use gear_core::{
     ids::ProgramId,
     memory::{HostPointer, Memory, PageNumber, WasmPageNumber},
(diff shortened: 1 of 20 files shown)


ukint-vs (approved):

lazy-pages/src/sys/windows.rs
```diff
@@ -46,9 +46,12 @@ unsafe extern "system" fn exception_handler(exception_info: *mut EXCEPTION_POINT
     let addr = (*exception_record).ExceptionInformation[1];
     let info = ExceptionInfo {
         fault_addr: addr as *mut _,
+        is_write: None,
     };
 
-    super::user_signal_handler(info).expect("Memory exception handler");
+    super::user_signal_handler(info)
+        .map_err(|err| format!("Memory exception handler failed: {}", err))
```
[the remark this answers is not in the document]
grishasobol: ok made thru expect

lazy-pages/src/sys.rs
```diff
@@ -39,117 +53,284 @@ cfg_if! {
...
+/// First part is always the same, so we can copy it to buffer
+/// once and then use it for all pages.
+struct PagePrefix {
+    buffer: Vec<u8>,
+}
+
+impl PagePrefix {
+    /// New page prefix from program prefix
+    pub fn new_from_program_prefix(program_prefix: &[u8]) -> Self {
+        let mut prefix = Self {
```
ark0f, line 72:
```suggestion
        Self { buffer: [program_prefix, &u32::MAX.to_le_bytes()].concat() }
```

lazy-pages/src/sys.rs
```diff
@@ -39,117 +53,284 @@ cfg_if! {
...
+}
+
+impl CheckedWasmAddr {
+    pub fn new_from_native(
+        native_addr: usize,
+        wasm_mem_addr: usize,
+        stack_end: WasmAddr,
+        wasm_mem_size: u32,
+    ) -> Result<Self, Error> {
+        let calc_wasm_mem_end_addr = || {
```
ark0f, line 104: I think it's simpler to calculate address right here

lazy-pages/src/sys.rs
```diff
@@ -39,117 +53,284 @@ cfg_if! {
...
+        // `addr` can be only decreased and `wasm_mem_addr` is never changed.
+        self.wasm_mem_addr + self.addr as usize
+    }
+
+    pub fn align_down(&mut self, alignment: u32) {
+        self.addr = (self.addr / alignment) * alignment;
+    }
+
+    /// Checks that interval [`addr`, `addr` + `size`) is in wasm memory.
+    pub fn check_interval_with_size_is_ok(&self, size: u32) -> Result<(), Error> {
```
ark0f, line 159:
```suggestion
    pub fn check_interval(&self, size: u32) -> Result<(), Error> {
```

lazy-pages/src/sys.rs
```diff
@@ -39,117 +53,284 @@ cfg_if! {
...
-                log::trace!(
-                    "{:?} has no data in storage, so just save current page data to released pages",
-                    page
-                );
-            } else {
-                log::trace!("{:?} has data in storage, so set this data for page and save it in released pages", page);
+    let fist_gear_page = wasm_addr.as_page_number();
+
+    for idx in 0..unprot_size / lazy_page_size {
+        // Arith opertaions are safe here.
```
ark0f, line 268: `...because`?

lazy-pages/src/sys.rs
```diff
@@ -39,117 +53,284 @@ cfg_if! {
...
+
+        ctx.accessed_pages_addrs.insert(lazy_page_wasm_addr);
+
+        for gear_page in (begin..end).map(PageNumber) {
+            let page_buffer_ptr = (wasm_mem_addr as *mut u8).add(gear_page.offset());
+            let buffer_as_slice = std::slice::from_raw_parts_mut(page_buffer_ptr, gear_ps as usize);
+            let res = sp_io::storage::read(prefix.calc_key_for_page(gear_page), buffer_as_slice, 0);
+
+            log::trace!(
+                "{:?} has{} data in storage",
```
ark0f, line 292:
```suggestion
                "{:?} has data in storage: {}",
```
and then only `res.is_none()`

Just simpler

pallets/gear-debug/src/tests.rs
```diff
@@ -284,6 +285,18 @@ fn get_last_message_id() -> MessageId {
     }
 }
 
+#[allow(unused)]
```
ark0f, line 288:
```suggestion
#[cfg(feature = "lazy-pages")]
```

pallets/gear-debug/src/tests.rs
```diff
@@ -19,11 +19,12 @@
 use super::*;
 use crate::mock::*;
 use common::{self, Origin as _};
+#[allow(unused)]
```
ark0f, line 22:
```suggestion
#[cfg(feature = "lazy-pages")]
```

common/src/lazy_pages.rs
```diff
@@ -22,23 +22,20 @@ use crate::Origin;
...
 use gear_core::{
     ids::ProgramId,
-    memory::{HostPointer, Memory, PageBuf, PageNumber, WasmPageNumber},
+    memory::{HostPointer, Memory, PageNumber, WasmPageNumber},
 };
 use gear_runtime_interface::{gear_ri, RIError};
-use sp_std::{collections::btree_map::BTreeMap, vec::Vec};
+use sp_std::vec::Vec;
+
+extern crate alloc;
```
ark0f, line 30: Move it in lib.rs

lazy-pages/src/lib.rs
```diff
@@ -22,87 +22,133 @@
...
+mod deprecated;
 mod sys;
 
 #[derive(Debug, derive_more::Display, derive_more::From)]
 pub enum Error {
     #[display(fmt = "WASM memory begin address is not set")]
     WasmMemAddrIsNotSet,
+    #[display(fmt = "WASM memory size is not set")]
+    WasmMemSizeIsNotSet,
+    #[display(fmt = "Overflow of addr arith operation")]
```
ark0f, line 48:
```suggestion
    #[display(fmt = "Overflow of address arithmetic operation")]
```

lazy-pages/src/lib.rs
```diff
@@ -22,87 +22,133 @@
...
     )]
     SignalFromUnknownMemory {
-        wasm_mem_begin: usize,
-        native_page: usize,
+        addr: usize,
+        wasm_mem_addr: usize,
+        wasm_mem_end_addr: usize,
+    },
+    #[display(
+        fmt = "Signal addr {:#x} is from wasm program virt-stack memory [0, {:#x})",
```
ark0f, line 64:
```suggestion
        fmt = "Signal addr {:#x} is from WASM program virtual stack memory [0, {:#x})",
```

lazy-pages/src/lib.rs
```diff
@@ -22,87 +22,133 @@
...
+        wasm_addr,
+        stack_end
+    )]
+    SignalFromStackMemory {
+        wasm_addr: WasmAddr,
+        stack_end: WasmAddr,
+    },
+    #[display(
+        fmt = "Accessed pages are not lies in wasm memory: [{:#x}, {:#x}) not in [{:#x}, {:#x})",
```
ark0f, line 73:
```suggestion
        fmt = "Accessed pages are not lie in WASM memory: [{:#x}, {:#x}) not in [{:#x}, {:#x})",
```

lazy-pages/src/sys.rs
```diff
@@ -18,10 +18,24 @@
...
+
+use crate::{
+    deprecated::user_signal_handler_internal_v1, Error, LazyPagesExecutionContext,
+    LazyPagesVersion, WasmAddr, LAZY_PAGES_CONTEXT, LAZY_PAGES_VERSION,
+};
+
+use gear_core::memory::{PageNumber, PAGE_STORAGE_GRANULARITY};
+
+// These constants are used both in runtime and in lazy-pages backend,
+// so makes here additional checks. If somebody would change these values
```
ark0f, line 33:
```suggestion
// so we make here additional checks. If somebody would change these values
```

pallets/gear-debug/src/tests.rs
```diff
@@ -436,6 +449,13 @@ fn check_not_allocated_pages() {
         persistent_pages.insert(gear_page0, page0_data.to_vec());
         persistent_pages.insert(gear_page7, page7_data.to_vec());
 
+        // For all pages, which is write accessed, and has no data in storage yet,
+        // we must upload to storage all pages from [PAGE_STORAGE_GRANULARITY] interval.
+        [gear_page0, gear_page7]
+            .iter()
```
ark0f, line 455:
```suggestion
            .into_iter()
```
And remove `copied()`.
Such method can be used since 2021 edition

ark0f (changes requested):

lazy-pages/src/sys.rs
grishasobol: It's very dirty to explain each arith operation in code

ark0f: I mean can you write something to explain in general?

common/src/lazy_pages.rs
grishasobol: What exactly?

ark0f: The whole line

pallets/gear-debug/src/tests.rs
grishasobol: I dont actually understand the reason, but warnings appear then:
```
this method call resolves to `<&[T; N] as IntoIterator>::into_iter` (due to backwards compatibility), but will resolve to <[T; N] as IntoIterator>::into_iter in Rust 2021
`#[warn(array_into_iter)]` on by default
this changes meaning in Rust 2021
for more information, see <https://doc.rust-lang.org/nightly/edition-guide/rust-2021/IntoIterator-for-arrays.html>rustc(array_into_iter)
tests.rs(455, 14): use `.iter()` instead of `.into_iter()` to avoid ambiguity: `iter`
tests.rs(454, 9): or use `IntoIterator::into_iter(..)` instead of `.into_iter()` to explicitly iterate by value: `IntoIterator::into_iter(`, `)`
```

grishasobol: But I will remove copied and make |&page | {}

lazy-pages/src/sys.rs
```diff
@@ -18,10 +18,24 @@
...
+// in runtime, then he also should pay attation to support new values here:
```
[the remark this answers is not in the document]
grishasobol: Ok)) I will try to find solution

common/src/lazy_pages.rs
```diff
@@ -19,13 +19,16 @@
...
```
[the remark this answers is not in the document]
grishasobol: Removed

lazy-pages/src/lib.rs
```diff
@@ -111,25 +157,88 @@ pub fn get_lazy_pages_numbers() -> Vec<PageNumber> {
     LAZY_PAGES_CONTEXT.with(|ctx| ctx.borrow().lazy_pages_info.keys().copied().collect())
 }
 
-/// Returns current wasm mem buffer pointer, if it's set
-pub fn get_wasm_mem_addr() -> Option<HostPointer> {
+#[derive(Debug, derive_more::Display)]
+pub enum InitForProgramErr {
```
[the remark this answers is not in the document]
grishasobol: Done

lazy-pages/src/sys.rs
grishasobol: Done

lazy-pages/src/sys.rs
grishasobol: Done

lazy-pages/src/sys.rs
grishasobol: Done

lazy-pages/src/sys.rs
grishasobol: Done

pallets/gear-debug/src/tests.rs
grishasobol: Done

common/src/lazy_pages.rs
grishasobol: Have removed this line

pallets/gear-debug/src/tests.rs
grishasobol: Done

lazy-pages/src/sys.rs
grishasobol: Ok, make some new comments

lazy-pages/src/lib.rs
grishasobol: Done

lazy-pages/src/lib.rs
grishasobol: Done

lazy-pages/src/lib.rs
grishasobol: Done

lazy-pages/src/sys.rs
grishasobol: Done

lazy-pages/src/sys.rs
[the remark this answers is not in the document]
grishasobol: Done

pallets/gear-debug/src/tests.rs
grishasobol: Done

common/src/lazy_pages.rs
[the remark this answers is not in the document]
grishasobol: MOve 
```
extern crate alloc;
```
to lib.rs

lazy-pages/src/lib.rs
```diff
@@ -111,25 +157,88 @@ pub fn get_lazy_pages_numbers() -> Vec<PageNumber> {
     LAZY_PAGES_CONTEXT.with(|ctx| ctx.borrow().lazy_pages_info.keys().copied().collect())
 }
 
-/// Returns current wasm mem buffer pointer, if it's set
-pub fn get_wasm_mem_addr() -> Option<HostPointer> {
+#[derive(Debug, derive_more::Display)]
+pub enum InitForProgramError {
+    #[display(
+        fmt = "Wasm mem native addr {:#x} is not aligned to the native page size",
```
ark0f, line 163:
```suggestion
        fmt = "WASM memory native address {:#x} is not aligned to the native page size",
```
Same for variants below

lazy-pages/src/lib.rs
```diff
@@ -111,25 +157,88 @@ pub fn get_lazy_pages_numbers() -> Vec<PageNumber> {
     LAZY_PAGES_CONTEXT.with(|ctx| ctx.borrow().lazy_pages_info.keys().copied().collect())
 }
 
-/// Returns current wasm mem buffer pointer, if it's set
-pub fn get_wasm_mem_addr() -> Option<HostPointer> {
+#[derive(Debug, derive_more::Display)]
+pub enum InitForProgramError {
```
ark0f, line 161:
```suggestion
pub enum InitializeForProgramError {
```
For consistency with function name

ark0f (approved):
Great!

PR #1292, opened 2022-08-08
Charge correctly in exit_code context

### LINKED ISSUE #1199: Syscall benchmarking for reply info
### Problem

Atm syscalls `gr_reply_to` and `gr_exit_code` costs the same (`Costs::ReplyTo`), but that's incorrect, it should be separated.

ukint-vs:
Resolves #1199 .

@gear-tech/dev

### DIFF
--- a/core-processor/src/ext.rs
+++ b/core-processor/src/ext.rs
@@ -34,9 +34,7 @@ use gear_core::{
     gas::{ChargeResult, GasAllowanceCounter, GasAmount, GasCounter, ValueCounter},
     ids::{CodeId, MessageId, ProgramId},
     memory::{AllocationsContext, Memory, PageBuf, PageNumber, WasmPageNumber},
-    message::{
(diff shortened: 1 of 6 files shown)


core-processor/src/ext.rs
```diff
@@ -471,6 +473,15 @@ impl EnvExt for Ext {
         Ok(())
     }
 
+    fn exit_code(&mut self) -> Result<Option<ExitCode>, Self::Error> {
+        self.charge_gas_runtime(RuntimeCosts::ExitCode)?;
+        if let Some(details) = self.context.message_context.current().reply() {
+            Ok(Some(details.into_exit_code()))
+        } else {
+            Ok(None)
+        }
```
gshep, line 482: Can it be rewritten this way? The same is for `reply_to`.
```suggestion
        Ok(self.context.message_context.current().reply().map(|d| d.into_exit_code()))
```

gshep (approved):

PR #1300, opened 2022-08-09
Gas reservation

### LINKED ISSUE #641: Gas reservation inside program

ark0f:
Resolves #641.

To do:
- [x] Make program reserves gas using itself ID (not message ID)
- [ ] Make implicit unreservation
- [ ] #1272 
- [ ] Test a new gas node with program
- [ ] Do gas reservation not only in success case
- [ ] Update docs


ark0f:
The PR will be splitted in 2 different PR - gas node and user API

PR #1304, opened 2022-08-10
Introduce support for multiple runtimes (revamped)

### LINKED ISSUE #1116: Stabilize runtime for a standalone chain with PoS support

ekovalev:
Reopening of reverted #1243.

In addition to the original changes contains a workaround that allows to attribute the current staging testnet chain id to the `gear` runtime even though the existing chain spec lacks the respective prefix.

---

Resolves #1116.

In order to be able to spin up both standalone network as well as a parachain (in the future) we want to maintain different variants of the runtime (think of Polkadot vs. Kusama vs. Westend vs. Rococo etc.).
Different runtimes require different clients and chain specs to configure the genesis block.

This PR attempts to create such support for multiple runtimes and introduce a new runtime for a standalone network codenamed `Vara` network. This standalone chain will eventually have staking. For now it only replaces block authoring consensus with `Babe` (in place of `Aura`) and introduces some basic session management.

In a nutshell, since there isn't anymore "The Runtime" but rather a set of possible runtimes each of which exposes their own set of Runtime Apis, we must also modify the Substrate client part of the node to be able to adjust to a concrete runtime chosen at the time of compilation. That being said, we, copying the approach taken by Polkadot and other Substrate based chains with multiple runtimes (Acala, Moonbeam), introduce traits `AbstractClient` and `RuntimeApiCollection` a client's Api must implement, as well as an enum `Client` whose variants are concrete runtimes (at the moment `gear` which refers to the old test network and `vara` - the standalone chain; in the future we might want `gear` to refer to the parachain runtime; in the meantime, there should also be one for `rococo` relay chain) and would implement the `AbstractClient` and a bunch of other traits the default Substrate client implements. This, in turn, requires the `Service` and `Cli` being changed accordingly.

Last but not least, the client part of the node has been split into several crates (again, following Polkadot, Moonbeam etc.) for the sake of easier maintenance in the future.

A heads-up: because at current point in order to make a transition to PoS smooth we still need to keep the old block authering mechanism based on `Aura` consensus algorithm. This makes the whole client part look a bit messy with lots of conditional compilation macros because we need to have two essentially different runtime Apis (with `Aura` and `Babe`). Once we've tested a standalone chain based on `Babe`, the legacy `Aura` will go away completely so we'll be able to clean up the node client.

The `gear-runtime-test-cli` utility has also been adjusted to multi-runtime code.

The fuzzer is still lagging behind but will be updated soonest possible (either in this PR or one of the following ones).

**Release Notes**:
This PR introduces support for multiple runtimes which allows for having several chains with similar underlying mechanics but possibly different in some concrete details (think of Polkadot vs. Kusama vs. Westend vs. Rococo etc.). For instance, there can be a standalone chain with PoS consensus and staking, whereas another one can implement necessary machinery to be plugged to a relay chain and act as a parachain in the Polkadot ecosystem thus providing access to our computing engine for other network participants by means of XCM system.
These runtimes co-exist all together within the same source code tree so that pallets can be shared between them eliminating the need to maintain separate branches for different chains.


shamilsan (approved):
As I see it is almost a copy of the approved PR so let's approve it too 😄

ekovalev:
> As I see it is almost a copy of the approved PR so let's approve it too 😄
It is, indeed. The only difference is the last commit that can be checked separately.

PR #1305, opened 2022-08-10
Update send_message and send_reply weights

### LINKED ISSUE #1203: Link weight to args in extrinsics
### File Location(s)

`pallets/gear/src/lib.rs`
`pallets/gear/src/benchmarking/mod.rs`

### Weight (especially db read-writes) depend on arguments
`send_message`:
 - if destination is not program, no message queue or mailbox affected (and probably no gas tree created)
`send_reply` && `claim_value_from_mailbox`:
- if value is 0, balances storages are unaffected

... And some other cases

### Small note for renaming proposal:
`claim_value_from_mailbox` -> `claim_value` cause we have no other abilities to claim anything, so its intuitive

### Some notes on benchmarks improvements:
write proper `verify` blocks

ukint-vs:
Resolves #1203  .

- payload size has almost no effect on execution time
- value can be zero or above minimal balance
- destination can be a program or regular account

```rust
#[pallet::weight({
            let v = if value.is_zero() {
                0
            } else {
                1
            };
            let weight = T::DbWeight::get().reads(1);
            if GearProgramPallet::<T>::program_exists(*destination) {
                weight.saturating_add(<T as Config>::WeightInfo::send_message(v, 1))
            } else {
                weight.saturating_add(<T as Config>::WeightInfo::send_message(v, 0))
            }
        })]
        pub fn send_message(
```
@gear-tech/dev


grishasobol (approved):

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -422,31 +421,44 @@ benchmarks! {
     }
 
     send_message {
-        let p in 0 .. MAX_PAYLOAD_LEN;
+        let v in 0 .. 1; // value is zero or no
+        let p in 0 .. 1; // dest is program or regular account
         let caller = benchmarking::account("caller", 0, 0);
```
breathx, line 426: If we add here following line: `log::debug!("v: {v} p: {p}");`
then following commands
```shell
$ cargo build -p gear-node --release --features runtime-benchmarks
$ RUST_LOG=pallet_gear=debug ./target/debug/gear-node benchmark pallet --chain=dev \
--steps=1 --repeat=1 --pallet=pallet_gear --extrinsic="send_message" --execution=wasm \
--wasm-execution=compiled --heap-pages=4096
```
show that there is no case when v = 0 p = 0

breathx (changes requested):
I don't really like this style of managing weights:
If I were you, I would probably separate weight function on each case we were discussing (here are 3 params: destination, gas limit, value - all of their combinations), and provide some soft function to be used in pallet inside weight macro (that will query data and choose the correct benchmark function).

pallets/gear/src/lib.rs
```diff
@@ -1635,7 +1635,19 @@ pub mod pallet {
...
         /// - `DispatchMessageEnqueued(MessageInfo)` when dispatch message is placed in the queue.
-        #[pallet::weight(<T as Config>::WeightInfo::send_message(payload.len() as u32))]
+        #[pallet::weight({
+            let v = if value.is_zero() {
+                0
+            } else {
+                1
+            };
+            let weight = T::DbWeight::get().reads(1);
+            if GearProgramPallet::<T>::program_exists(*destination) {
```
NikVolf, line 1645: weight calculation formula should not include database reads imho

breathx:
Temporary closing as stale one

PR #1306, opened 2022-08-10
Rename auxiliary pallets directory

ekovalev:
Previously a sub-directory named `aux` has been created in the `pallets` to contain temporary (or auxiliary) pallets that won't be there in any of the prod runtimes. For instance, currently it contains a deterministic session manager to be plugged into the `pallet_babe`'s `Config` trait implementation in `vara` runtime.

The name `aux`, however, seems to be invalid on Windows.

This posed another question: does it, at all, make sense to maintain such distinction between temporary pallets and "permanent" ones? Perhaps nothing bad will happen if we just mix them all together in the `pallets` folder, and just remove the obsolete ones when the time comes.

Following the discussion below, the auxiliary pallet `shift-session-manager` has been moved up to render the pallets structure flat.


ark0f (approved):
Thank you very much!

shamilsan (approved):

shamilsan:
> This poses another question, though: does it, at all, make sense to maintain such distinction between temporary pallets and "permanent" ones? Perhaps nothing bad will happen if we just mix them all together in the `pallets` folder, and just remove the obsolete ones when the time comes. Appreciate any feedback.

I vote for one-level pallets structure. We have no so many pallets at the moment. Each pallet name reflects its purpose. If the pallet it not needed more it can be limited indeed.

Issue #922, opened 2022-05-10
Optimize wasm without using third party tools.


ukint-vs:
### Proposal

Need to find a way to optimize artefacts in `wasm-builder` without using `wasm-opt`.

clearloop:
we can strip custom sections like carg-contract does https://github.com/paritytech/cargo-contract/blob/62cf21f6811261b0c382acd4b84b4d903e14823b/src/cmd/build.rs#L487

ark0f:
What motivation?

shamilsan:
> What motivation?

To not force to install additional dependencies I guess.

breathx:
> > What motivation?
> 
> To not force to install additional dependencies I guess.

Exactly. Somehow wrap optimisation logic inside gear wasm builder to prevent forcing `binaryen` manual installation

breathx:
> we can strip custom sections ( keep `name_section` for debugging ) like carg-contract does https://github.com/paritytech/cargo-contract/blob/62cf21f6811261b0c382acd4b84b4d903e14823b/src/cmd/build.rs#L487

Good catch. Feel free to apply this or open issue on it

PR #1308, opened 2022-08-11
feat(wasm-builder): optimize wasm builder

clearloop:
Trying to add `strip_custom_sections` but found some of the logics in `wasm-builder` are pretry hard to read )), the logic in `wasm_project.rs` and `optimize.rs` are mixed, some of the parameters are only for printing logs but not related  with the main process the functions are provided.

Ref #922 

- [x] clean verbose declarations
- [x] leave `Optimizer` with pure binary operations, move the os operations to `wasm_project.rs`
- [x] ~~strip custom sections~~

@gear-tech/dev

### DIFF
--- a/.github/workflows/CD-vara.yml
+++ b/.github/workflows/CD-vara.yml
@@ -1,0 +1,75 @@
+name: Vara CD
+
+on:
+  workflow_dispatch:
+
+env:
+  CARGO_TERM_COLOR: always
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to DockerHub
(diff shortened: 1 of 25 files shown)


ukint-vs (approved):

utils/wasm-builder/src/optimize.rs
```diff
@@ -31,24 +55,15 @@ impl Optimizer {
...
+    /// Process optimization.
+    pub fn optimize(&self, ty: OptType) -> Result<Vec<u8>> {
+        let mut module = self.module.clone();
         pwasm_utils::optimize(
-            &mut binary_module,
-            vec!["handle", "handle_reply", "init", "__gear_stack_end"],
+            &mut module,
+            match ty {
+                OptType::Opt => OPTIMIZED_EXPORTS.into(),
+                OptType::Meta => META_EXPORTS.into(),
```
breathx, line 65: what if provide this constants as `OptType` const fn?

breathx (approved):

Issue #1317, opened 2022-08-15
wrong block in test-net #866245


grishasobol:
### Problem

Last week We have added new algorithm for lazy pages, which differs a lot from old algorithm. To support old runtimes lazy pages uses versioning now: by default is `version 1` and new runtime set `version 2`. But because of my fault this version is updated only one time for each thread (which executes `on_idle`). On our test net authority nodes execute `on_idle` in the same thread, so lazy pages version has not been changed after last runtime upgrade.

![telegram-cloud-photo-size-2-5463126391546559309-y](https://user-images.githubusercontent.com/11819936/184665025-d700cf1a-6ec0-461b-ad17-7935cf0d648f.jpg)

In block #876041 we have a panic, which cause nodes restart and then node has correct lazy pages version 2, so all block are correct. Exception is block #866245. All blocks between #866245 and #876041 has been finalised without contracts execution. So, we have only one wrong finalised block.

### Steps

-

### Possible Solution

There is only one contract, which  executes incorrectly and cause wrong block `#866245`. With `lazy pages version 1` logs are:
```
gear_core_processor::executor: Executing program 0x25eaf5bdc65f5608b5f397bc3aad23539bd55144341e1a5f259b942b5e783dbc
gear_core_processor::executor: Executing dispatch IncomingDispatch { kind: Init, message: IncomingMessage { id: 0xbc7c5ec967bd683ff7454e24f21ffb5005c7d6885b16d8326
gear_lazy_pages::sys: Interrupted, exception info = ExceptionInfo { fault_addr: 0x7f5d540fffe4, is_write: Some(true) }
gear_lazy_pages::deprecated: Interrupted, exception info = ExceptionInfo { fault_addr: 0x7f5d540fffe4, is_write: Some(true) }
gear_lazy_pages::deprecated: mem=0x7f5d540fffe4 accessed=PageNumber(255),WasmPageNumber(15) pages=255..256 page_native_addr=0x7f5d540ff000
gear_lazy_pages::deprecated: PageNumber(255) has no data in storage, so just save current page data to released pages
gear_lazy_pages::sys: Interrupted, exception info = ExceptionInfo { fault_addr: 0x7f5d54101aac, is_write: Some(false) }
gear_lazy_pages::deprecated: Interrupted, exception info = ExceptionInfo { fault_addr: 0x7f5d54101aac, is_write: Some(false) }
[...]

PR #1318, opened 2022-08-15
fix wrong block panic in node

grishasobol:
See issue #1317 

### Test
Run test net sync for mac and for linux, all goes OK.

### Other
Removes `lazy-pages` from list of crates to update runtime spec version, because lazy-pages crate code is restricted to be in runtime.

### DIFF
--- a/lazy-pages/src/lib.rs
+++ b/lazy-pages/src/lib.rs
@@ -353,8 +353,8 @@ pub enum InitError {
 unsafe fn init_internal(version: LazyPagesVersion) -> Result<(), InitError> {
     use InitError::*;
 
-    // Set version even if it has been already set, because runtime upgrade and then contracts
-    // execution can be done in the same thread.
+    // Set version even if it has been already set, because `on_idle` calls can be
+    // in the same thread, even after runtime upgrade, which can change version.
     LAZY_PAGES_VERSION.with(|v| *v.borrow_mut() = version);
 
     if LAZY_PAGES_ENABLED.with(|x| *x.borrow()) {
--- a/lazy-pages/src/lib.rs
+++ b/lazy-pages/src/lib.rs
@@ -254,6 +254,27 @@ pub fn get_released_pages() -> Vec<PageNumber> {
     LAZY_PAGES_CONTEXT.with(|ctx| ctx.borrow().released_pages.iter().copied().collect())
 }
 
+pub fn get_released_pages_patch() -> Vec<PageNumber> {
+    LAZY_PAGES_CONTEXT.with(|ctx| {
+        let ctx = ctx.borrow_mut();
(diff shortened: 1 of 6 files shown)


lazy-pages/src/lib.rs
```diff
@@ -254,6 +254,29 @@ pub fn get_released_pages() -> Vec<PageNumber> {
     LAZY_PAGES_CONTEXT.with(|ctx| ctx.borrow().released_pages.iter().copied().collect())
 }
 
+/// TODO: remove this after current test-net chain will be dropped (issue #1317).
+/// This patch solves the problem for block `#866245`, see more in issue.
+pub fn get_released_pages_patch() -> Vec<PageNumber> {
+    LAZY_PAGES_CONTEXT.with(|ctx| {
+        let ctx = ctx.borrow_mut();
+        if ctx.program_storage_prefix.as_ref().unwrap()
+            == &vec![
```
ark0f, line 263: Use `lazy_pages` crate:
```suggestion
            == &lazy_pages::pages_prefix(...)
```

lazy-pages/src/lib.rs
```diff
@@ -254,6 +254,29 @@ pub fn get_released_pages() -> Vec<PageNumber> {
     LAZY_PAGES_CONTEXT.with(|ctx| ctx.borrow().released_pages.iter().copied().collect())
 }
 
+/// TODO: remove this after current test-net chain will be dropped (issue #1317).
+/// This patch solves the problem for block `#866245`, see more in issue.
+pub fn get_released_pages_patch() -> Vec<PageNumber> {
```
ark0f, line 259:
```suggestion
pub fn get_released_pages_patched() -> Vec<PageNumber> {
```

lazy-pages/src/lib.rs
grishasobol: It has additional  4 bytes for page prefix in the end. I think better is to use clear prefix without additional, to avoid additional bytes in right vector.

ark0f: > It has additional 4 bytes for page prefix in the end.

It's `page_key`, not `pages_prefix`

grishasobol: Okey, but what the difference ?)

ark0f: Oops. It's `common` crate where code is:
```rust
key.extend(STORAGE_PROGRAM_PAGES_PREFIX);
key.extend(program_id.as_fixed_bytes());
key.extend(b"::");
```

grishasobol: Discussed tet a tet

ark0f (approved):

lazy-pages/src/lib.rs
grishasobol: done

PR #827, opened 2022-04-07
Fix `core-processor` misprint and provide compact `From`'s for core

breathx:
By the review of @gshep on already merged PR.

@gear-tech/dev

PR #1324, opened 2022-08-17
Handle code re-instrumentation failure in process_queue

SabaunT:
Partially resolves #827.

The idea of the current solution is described in the issue's page. However, due to lack of type-level guarantees, that `expect`ing re-instrumentation won't fail, we state another to be considered in the future task to make the solution infallible.

- Introduce test to protect instructions from being accidentally deleted in `Rules::instruction_cost` method implementation.

### DIFF
--- a/core-backend/wasmi/src/env.rs
+++ b/core-backend/wasmi/src/env.rs
@@ -372,10 +372,7 @@ where
         };
 
         let defined_host_functions = builder.defined_host_functions.clone();
-        let module = match wasmi::Module::from_buffer(binary) {
-            Ok(module) => module,
-            Err(e) => return Err(ModuleInstantiation(e)),
-        };
+        let module = wasmi::Module::from_buffer(binary).map_err(ModuleInstantiation)?;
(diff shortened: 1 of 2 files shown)


breathx:
First linked issue seems misprinted for the PR

SabaunT:
> First linked issue seems misprinted for the PR

Yeah, corrected that

common/src/code_storage.rs
```diff
@@ -28,16 +28,66 @@ pub enum Error {
```
breathx, line 28: frankly speaking, I disagree with such usage of our common storage traits, but let it be so atm

breathx (approved):
👍

breathx:
@SabaunT, seems that you forget to specify wasm-instrument crate source

common/src/code_storage.rs
SabaunT: We need to have more guarantees that if instrumented code exists in the storage, than the original one exists. This is reached by "default" trait's method implementation. The solution here is not final. Finalizing is on @gshep.

PR #1328, opened 2022-08-17
Deny calls with value for `vara-stage-1`

breathx:
Reverts [this commit](https://github.com/gear-tech/gear/commit/c40e76860fd0c2515c78f2adade3081c1a85e437) and expands our custom signed extension in order to use the unified mechanic of denying calling extrinsics, that cause balances, instead of somewhere invalidate extrinsics, somewhere return errors.

### DIFF
--- a/runtime/vara/src/lib.rs
+++ b/runtime/vara/src/lib.rs
@@ -121,9 +121,10 @@ pub fn native_version() -> NativeVersion {
 /// RELEASE: This is only relevant for the initial PoA run-in period and will be removed
 /// from the release runtime.
 #[derive(Default, Encode, Debug, Decode, Clone, Eq, PartialEq, TypeInfo)]
-pub struct DisableBalancesCall;
-impl SignedExtension for DisableBalancesCall {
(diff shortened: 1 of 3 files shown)


runtime/vara/src/lib.rs
```diff
@@ -118,7 +118,7 @@ pub fn native_version() -> NativeVersion {
 
 /// Disallow balances transfer
 ///
-/// RELEASE: This is only relevant for the initial PoA run-in period and may be removed
+/// RELEASE: This is only relevant for the initial PoA run-in period and will be removed
 /// from the release runtime.
 #[derive(Default, Encode, Debug, Decode, Clone, Eq, PartialEq, TypeInfo)]
 pub struct DisableBalancesCall;
```
ekovalev, line 124: Probably the struct itself should be renamed to reflect the new semantics, should it not?

breathx: Into something like `DisableValueTransfers`?

ekovalev (approved):

PR #1341, opened 2022-08-19
Introduce Reserved gas node

ark0f:
The first step to #641


common/src/gas_provider/node.rs
```diff
@@ -238,6 +272,15 @@ impl<ExternalId: Clone, Id: Clone + Copy, Balance: Zero + Clone + Copy>
         matches!(self, Self::ReservedLocal { .. })
     }
 
+    /// Returns wether the node is of `Reserve` type
+    pub(crate) fn is_reserved(&self) -> bool {
+        matches!(self, Self::Reserved { .. })
+    }
+
+    pub(crate) fn is_orphan(&self) -> bool {
```
grishasobol, line 280: I think better to add comment here about `is_orphan` then everywhere in code

pallets/gear-scheduler/src/tests.rs
```diff
@@ -73,7 +73,8 @@ fn populate_wl_from(
         .expect("Failed to insert task");
     WaitlistOf::<Test>::insert(dispatch, u64::MAX).expect("Failed to insert to waitlist");
     Balances::reserve(&src, DEFAULT_GAS as u128).expect("Cannot reserve gas");
-    GasHandlerOf::<Test>::create(src, mid, DEFAULT_GAS).expect("Failed to create gas handler");
+    GasHandlerOf::<Test>::create(src, mid.into(), DEFAULT_GAS)
```
grishasobol, line 76: What the reason to use `.into()` now everywhere ?

gshep: From my point of view this a necessity from now

gshep (approved):
looks good to me!

common/src/gas_provider/node.rs
SabaunT, line 280: I think it's not representative name, because `GasNode::External` can be considered as orphan as well

common/src/gas_provider/node.rs
```diff
@@ -39,6 +65,12 @@ pub enum GasNode<ExternalId: Clone, Id: Clone, Balance: Zero + Clone> {
     /// (not node's parent, not node's child).
     ReservedLocal { id: ExternalId, value: Balance },
 
+    /// A node used for gas reservation feature.
+    ///
+    /// Such node types are independent and aren't part of the tree structure
+    /// (not node's parent, not node's child).
+    Reserved { id: ExternalId, value: Balance },
```
SabaunT, line 72: I think it's time to re-think names for `ReservedLocal` and `Reserved` nodes in accordance in accordance to their purposes. At least one of them must be changed, because they all have similar: names, instantiation logic, place in the gas tree (forest).

common/src/gas_provider/internal.rs
```diff
@@ -768,6 +775,44 @@ where
         node.lock().ok_or_else(|| InternalError::forbidden().into())
     }
 
+    fn reserve(
```
SabaunT, line 778: the internal logic of the function looks much like in `create_from_with_value`. Better adjust the latter one and call it in the body of the `reserve`.

SabaunT (changes requested):
Looks good :+1: 
However, please apply changes stated in the comments.
Also I want to ask you to change comment here https://github.com/gear-tech/gear/blob/9c736ef0d63e6bd18f1441274c3925814ed135bb/common/src/gas_provider/property_tests/assertions.rs#L42 as it's not complete after your changes

pallets/gear-scheduler/src/tests.rs
ark0f: Because we introduced `GasNodeId`

common/src/gas_provider/node.rs
```diff
@@ -268,17 +268,19 @@ impl<ExternalId: Clone, Id: Clone + Copy, Balance: Zero + Clone + Copy>
...
+    /// Returns whether the node is of `Reserve` type
     pub(crate) fn is_reserved(&self) -> bool {
         matches!(self, Self::Reserved { .. })
     }
 
-    pub(crate) fn is_orphan(&self) -> bool {
-        self.is_reserved() | self.is_reserved_local()
+    /// Returns whether the node is independent and isn't part of the tree structure
+    /// (not node's parent, not node's child).
+    pub(crate) fn is_independent(&self) -> bool {
```
SabaunT, line 282: what do you think of `is_separate`? Sounds more relevant for me. Can keep independent.

SabaunT, line 282: what do you think of is_separate? Sounds more relevant for me. Can keep `is_independent`.

SabaunT (approved):

common/src/gas_provider/node.rs
ark0f: I'd prefer current name

core/src/ids.rs
```diff
@@ -206,3 +206,19 @@ impl ProgramId {
         hash(&argument).into()
     }
 }
+
+declare_id!(ReservationId: "Reservation identifier");
```
breathx, line 210: I see no reason to add such struct into core part of the code

breathx (changes requested):
Requesting changes to avoid merging to master.

The first two things I'm against of: addition of reservation id in core, usage of `Into::into` everywhere **manually**.

I'll publish the rest of review asap, once I recheck algorithm changes in details.

common/src/gas_provider/internal.rs
```diff
@@ -368,19 +374,25 @@ where
...
-        } else {
-            node.increase_spec_refs();
-
-            GasNode::SpecifiedLocal {
-                value: amount,
-                lock: Zero::zero(),
-                parent: node_id,
-                refs: Default::default(),
-                consumed: false,
+        let new_node = match kind {
```
breathx, line 377: I'd move this lines under the following ones: try to reduce value, create node afterward

It's minor optimisation if insufficient balance queried, no `Self::get_external` will be called.

common/src/gas_provider/node.rs
```diff
@@ -18,6 +18,32 @@
...
+        }
+    }
+
+    pub fn as_reservation_id(self) -> Option<ReservationId> {
+        match self {
+            GasNodeId::Message(_) => None,
+            GasNodeId::Reservation(reservation_id) => Some(reservation_id),
+        }
+    }
+}
```
breathx, line 46: I'd provide `TryInto` instead of this fns
It's also against rust naming concepts to use `fn as_ ... ` with `(self)` arg

common/src/gas_provider/node.rs
```diff
@@ -234,8 +268,19 @@ impl<ExternalId: Clone, Id: Clone + Copy, Balance: Zero + Clone + Copy>
...
+    }
+
+    /// Returns whether the node is of `Reserve` type
+    pub(crate) fn is_reserved(&self) -> bool {
+        matches!(self, Self::Reserved { .. })
+    }
+
+    /// Returns whether the node is independent and isn't part of the tree structure
+    /// (not node's parent, not node's child).
+    pub(crate) fn is_independent(&self) -> bool {
```
breathx, line 282: For me, word "detached" seems more suitable this case

common/src/gas_provider/node.rs
```diff
@@ -234,8 +268,19 @@ impl<ExternalId: Clone, Id: Clone + Copy, Balance: Zero + Clone + Copy>
     }
 
     /// Returns whether the node is of `ReservedLocal` type
```
breathx, line 270:
```suggestion
    /// Returns whether the node is of `Cut` type
```

core/src/ids.rs
breathx, line 210: As I said before, I don't mean `ReservationId` as a common and **core** part of  the code. As I don't mean `Handle` in `gcore`, for example

I'd operate with just `type Hash = [u8; 32];` in core part, defining some `struct ReservationId` only for contracts (user-space)

breathx: Then `GasNodeId` will be 
```
enum GasNodeId {
     MessageId(MessageId),
     ReservationId(Hash),
}
```

common/src/gas_provider/mod.rs
```diff
@@ -207,6 +207,15 @@ pub trait Tree {
     /// for example, for `GasNode::ReservedLocal`.
     fn get_lock(key: Self::Key) -> Result<Self::Balance, Self::Error>;
 
+    /// Reserve some value from underlying balance.
+    ///
+    /// Used in gas reservation feature.
+    fn reserve(
```
breathx, line 213: To avoid `Into::into` everywhere (and also avoid `impl Into<Self::Key>` in fns) I'd add one more associated type to GasHandler trait with ReservationId, and use in storage some generic enum for `GasNodeType<NodeId, ReservationId>`

common/src/gas_provider/node.rs
```diff
@@ -37,7 +63,13 @@ pub enum GasNode<ExternalId: Clone, Id: Clone, Balance: Zero + Clone> {
...
     /// Such node types are independent and aren't part of the tree structure
     /// (not node's parent, not node's child).
-    ReservedLocal { id: ExternalId, value: Balance },
+    Cut { id: ExternalId, value: Balance },
+
+    /// A node used for gas reservation feature.
+    ///
+    /// Such node types are independent and aren't part of the tree structure
+    /// (not node's parent, not node's child).
+    Reserved { id: ExternalId, value: Balance },
```
breathx, line 72: Probably, reserved should also contain "parent" key for future purposes

breathx (changes requested):
Atm Reserved node is like the same that Cut node, but in fact it should be able to be parent of other nodes etc, which won't be able for Cut node.

NOTE: ReservationId may be introduced on other levels, not only for user-space if needed

common/src/gas_provider/node.rs
ark0f: > I'd provide TryInto instead of this fns

`TryInto` is about `Result` which is inconvenient, so I'll rename methods

common/src/gas_provider/node.rs
ark0f: So why not leave it "as is" for now?

core/src/ids.rs
ark0f: > As I said before, I don't mean ReservationId as a common and core part of the code. As I don't mean Handle in gcore, for example

`ReservationId` will be used in `Ext` trait methods so I don't want to operate with just `Hash`

NikVolf:
@breathx & @ark0f maybe try to get it un-stalled ?

common/src/gas_provider/internal.rs
```diff
@@ -56,22 +56,55 @@ impl<Balance: BalanceTrait> CatchValueOutput<Balance> {
...
+pub(crate) enum NodeCreationKey<MapKey, MapReservationKey> {
+    Cut(MapKey),
+    SpecifiedLocal(MapKey),
+    Reserved(MapReservationKey),
+}
+
+impl<MapKey, MapReservationKey> NodeCreationKey<MapKey, MapReservationKey>
+where
+    MapKey: Copy,
+    MapReservationKey: Copy,
```
breathx, line 69:
```suggestion
```

core/src/ids.rs
```diff
@@ -218,6 +218,22 @@ impl ProgramId {
...
+        let argument = [msg_id.as_ref(), &idx.to_le_bytes(), b"reservation_id_salt"].concat();
+        hash(&argument).into()
+    }
+}
+
+impl From<ReservationId> for MessageId {
+    fn from(id: ReservationId) -> Self {
+        Self(id.0)
+    }
+}
```
breathx, line 235:
```suggestion
```

breathx (approved):

grishasobol (approved):

common/src/gas_provider/node.rs
```diff
@@ -19,6 +19,29 @@
 use super::*;
 use codec::MaxEncodedLen;
 
+/// ID of the [`GasNode`].
+#[derive(Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd, Encode, Decode, TypeInfo)]
+pub enum GasNodeId<T, U> {
```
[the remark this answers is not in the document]
ark0f: These impls are conflicting

common/src/gas_provider/node.rs
```diff
@@ -19,6 +19,29 @@
...
+    Node(T),
+    Reservation(U),
+}
+
+impl<T, U> GasNodeId<T, U> {
+    pub fn to_node_id(self) -> Option<T> {
```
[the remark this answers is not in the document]
ark0f: > Prefix | Cost | Ownership
-- | -- | --
to_ | Expensive | owned -> owned (Copy types)

Also example from guideline:

> [f64::to_radians()](https://doc.rust-lang.org/std/primitive.f64.html#method.to_radians) converts a floating point quantity from degrees to radians. The input is f64. Passing a reference &f64 is not warranted because f64 is cheap to copy. Calling the function into_radians would be misleading because the input is not consumed.

I assume `GadNodeId` to be copy type (despite generics)

common/src/gas_provider/internal.rs
```diff
@@ -56,22 +56,55 @@ impl<Balance: BalanceTrait> CatchValueOutput<Balance> {
     }
 }
 
+#[derive(Debug, Copy, Clone)]
+pub(crate) enum NodeCreationKey<MapKey, MapReservationKey> {
```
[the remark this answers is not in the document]
ark0f: It is used in `create_from_with_value` function to differ node we want to create

common/src/gas_provider/internal.rs
ark0f: Motivation is same as for `GasNodeId`

SabaunT (approved):

gshep (approved):
Good job!

PR #1345, opened 2022-08-21
Fuzzer: adjust minimum balance to stay above existential deposit

ekovalev:


utils/economic-checks/src/util.rs
```diff
@@ -202,7 +202,7 @@ pub(crate) fn new_test_ext(
                 initial_authorities
                     .iter()
                     .cloned()
-                    .map(|(acc, _, _)| (acc, 1000)),
+                    .map(|(acc, _, _)| (acc, EXISTENTIAL_DEPOSIT * 2)),
```
breathx, line 205: why x2?

breathx (approved):

utils/economic-checks/src/util.rs
ekovalev: Just in case. These are validators' balances, that is in reality should be substantial. For the purpose of this application we just want to ensure the genesis block is applied successfully

PR #1347, opened 2022-08-22
Exited program balance inheritance

breathx:
Changes:

- Added new program status: `Exited`, which contains inheritor id.
- Fixed critical bug (try to transfer reserved funds on exit also)
https://github.com/gear-tech/gear/blob/9c736ef0d63e6bd18f1441274c3925814ed135bb/pallets/gear/src/manager/journal.rs#L169-L178
- Added logic to move value to inheritor if new funds received/unreserved for exited program

(Case when inheritor is exited program caught by defining chain of findings of inheritor, which is safe from circular inherit)

UPD: value inheritance also added for Teminated program case (program creator owns the value)


pallets/gear/src/tests.rs
```diff
@@ -2531,13 +2531,74 @@ fn test_message_processing_for_non_existing_destination() {
...
+        assert_ok!(Gear::send_message(
+            Origin::signed(USER_1),
+            program_id,
+            payload.encode(),
+            1_000_000_000,
+            value
+        ));
+        let message_1 = utils::get_last_message_id();
+
+        let payload = Input::Exit(USER_2.into_origin().into());
```
gshep, line 2582: what if it is also a terminated program?

pallets/gear/src/lib.rs
```diff
@@ -962,11 +962,38 @@ pub mod pallet {
...
+        pub fn is_exited(program_id: ProgramId) -> bool {
+            common::get_program(program_id.into_origin())
+                .map(|p| p.is_exited())
+                .unwrap_or_default()
+        }
+
+        /// Returns true if id is a program and the program has exited status.
+        pub fn exit_argument_of(program_id: ProgramId) -> Option<ProgramId> {
+            common::get_program(program_id.into_origin())
+                .map(|p| {
```
SabaunT, line 989: seems it's case for `and_then` instead of calling `unwrap_or_default` with `map`?

pallets/gear/src/tests.rs
breathx: So it's just burn possibility. We shouldn't restrict the user in this
The main idea of the PR and core-audit issue that `exit` - is unique termination, which moves all funds to somebody else (so any further fund receipts and unlocks should also be transferred to him)

SabaunT: As far as I understand, program becomes terminated only in case of init failed (case of program paused is not considered). Is it possible to perform same test for the logic of `Terminated`, not `Exited` program?

SabaunT (approved):
Nice!
A special shout out for not forgetting the loop in `inherent_of` :clap:

pallets/gear/src/tests.rs
breathx: > As far as I understand, program becomes terminated only in case of init failed (case of program paused is not considered). Is it possible to perform same test for the logic of `Terminated`, not `Exited` program?

Atm yes, but as I said before, this status doesn't mean any further transfers of owned funds.
Also I have a lot of plans of changing terminated status logic and usage.

pallets/gear/src/manager/journal.rs
```diff
@@ -162,19 +162,26 @@ where
...
+        common::set_program_exited_status(id_exited, value_destination)
             .expect("`exit` can be called only from active program; qed");
 
-        let program_account = &<T::AccountId as Origin>::from_origin(id_exited.into_origin());
-        let balance = CurrencyOf::<T>::total_balance(program_account);
+        let program_account = &<T::AccountId as Origin>::from_origin(id_exited);
+        let balance = CurrencyOf::<T>::free_balance(program_account);
+
+        let destination = Pallet::<T>::inheritor_for(value_destination);
+        let destination = <T::AccountId as Origin>::from_origin(destination.into_origin());
```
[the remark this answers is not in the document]
breathx: I don't think this problem should be solved with current PR.
It will partially disappear once we totally remove H256 usage from code.

pallets/gear/src/lib.rs
breathx: If I'm not mistaken, `and_then` returns Option anyway, so we have to unwrap it further

SabaunT: I guess it doesn't make nested Options.

breathx: > I guess it doesn't make nested Options.

Yeah, it doesn't, but for me that's the same code:
```rust
// #1
get_program(program_id.into_origin())
    .map(|p| p.is_active())
    .unwrap_or_default()
// #2
get_program(program_id.into_origin())
    .and_then(|p| p.is_active())
    .unwrap_or_default()
```

SabaunT: oops, got confused of `and_then`. I thought much of it, the it really was

gshep (approved):
Great job! :+1:

SabaunT:
are you sure we need to separate `Terminated(ProgramId)` and `Exited(ProgramId)`? Any use-cases?

breathx:
> are you sure we need to separate `Terminated(ProgramId)` and `Exited(ProgramId)`? Any use-cases?

Atm there is no difference, as is no restrictions or overhead.
I want to keep it as it is for future updates which will change this logic.
It also allows us to better cover with tests and check actual program status in storage

PR #1348, opened 2022-08-22
Correct error replies lifecycle

breathx:
There is no need to store error replies in mailbox, so this PR removed this possibility.


shamilsan (approved):
Good job 👍

gshep (approved):

PR #1355, opened 2022-08-22
fix `gr_read`  panic when arguments are incorrect

grishasobol:


core-backend/sandbox/src/funcs.rs
```diff
@@ -310,10 +308,15 @@ where
 
         let mut f = || {
             let msg = ctx.ext.msg().to_vec();
+            if at > msg.len() || at + len > msg.len() {
```
breathx, line 311:
```suggestion
            if at + len > msg.len() {
```
First check is automatically included in second one, cause if `at > msg.len()` than with any `len` `at + len > msg.len()`

Same thing in `wasmi` backend

breathx (approved):
Great catch!

core-backend/sandbox/src/funcs.rs
grishasobol: Yes, but we must make then checked add or saturated. Ok will do this

ukint-vs (approved):

core-backend/sandbox/src/funcs.rs
breathx: > Yes, but we must make then checked add or saturated. Ok will do this

I prefer saturating math here

breathx: Btw, it must be saturated even with the previous condition statement

grishasobol: Done

PR #1357, opened 2022-08-23
Do run CI on vara stable branches

NikVolf:


breathx:
Should we pack caches for these branches?

NikVolf:
> Should we pack caches for these branches?

if it is partioned by branch (@ukint-vs ?) then why not? But it's not that important

shamilsan (approved):

PR #1361, opened 2022-08-23
Update CD.yml

sergeyfilyanin:


breathx:
I think you can freely push commits with `[skip-ci]` for CD changes, but they should be tested on fork

breathx (approved):

PR #1367, opened 2022-08-24
Make formatters smarter

breathx:
- Excluded possibility to fail formatting, except `write!` result
- Precision works with both sides at the same time
- Added alternate `:#` formatter

For details, see test:
https://github.com/gear-tech/gear/blob/5f47966c6624ae72e3102e3dddab11992da0490f/core/src/ids.rs#L219-L242

### DIFF
--- a/core/src/ids.rs
+++ b/core/src/ids.rs
@@ -238,7 +238,7 @@ fn formatting_test() {
     assert_eq!(format!("{id:.2?}"), "0xc15d..fdef");
     // `Debug`/`Display` with precision 4.
     assert_eq!(format!("{id:.4?}"), "0xc15d1549..b936fdef");
-    // `Debug`/`Display` with precision 15 (the same for any case >= 16).
+    // `Debug`/`Display` with precision 15.
     assert_eq!(
         format!("{id:.15?}"),
         "0xc15d1549fa3950c0aa61e14a3ba476..95b4bcc894b9fab09e7fe9b936fdef"
--- a/core/src/ids.rs
+++ b/core/src/ids.rs
@@ -99,15 +99,17 @@ macro_rules! declare_id {
         impl core::fmt::Display for $name {
             fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
                 let len = self.0.len();
-                let mut e1 = len / 2;
-                let mut s2 = len / 2;
+                let median = (len + 1) / 2;
+
+                let mut e1 = median;
+                let mut s2 = median;
 
                 if let Some(precision) = f.precision() {
-                    let precision = precision.min(len / 2);
-
-                    e1 = precision;
-                    s2 = len - precision;
-                };
+                    if precision < median {
+                        e1 = precision;
+                        s2 = len - precision;
+                    }
+                }
 
                 let p1 = hex::encode(&self.0[..e1]);
                 let p2 = hex::encode(&self.0[s2..]);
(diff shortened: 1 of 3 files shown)


core/src/ids.rs
```diff
@@ -98,17 +98,26 @@ macro_rules! declare_id {
 
         impl core::fmt::Display for $name {
             fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
-                let mut end = self.0.len();
+                let len = self.0.len();
+                let mut e1 = len / 2;
+                let mut s2 = len / 2;
```
shamilsan, line 103: It's the case but this division is not good for odd `len`. As an option:

```rust
let mut e1 = len / 2;
let mut s2 = len - e1;
```

core/src/ids.rs
```diff
@@ -98,17 +98,26 @@ macro_rules! declare_id {
...
 
                 if let Some(precision) = f.precision() {
```
shamilsan, line 105: I'd make these calculation only if precision is good.
```suggestion
                if matches!(f.precision(), Some(precision) if precision < len/2) {
```

core/src/ids.rs
```diff
@@ -98,17 +98,26 @@ macro_rules! declare_id {
...
 
-                    end = precision;
+                    e1 = precision;
+                    s2 = len - precision;
                 };
 
-                write!(f, "0x{}", hex::encode(&self.0[..end]))
+                let p1 = hex::encode(&self.0[..e1]);
+                let p2 = hex::encode(&self.0[s2..]);
+                let sep = e1.ne(&s2).then_some("..").unwrap_or_default();
```
shamilsan, line 114: I know you love `then_some` 😄 
But `e1.ne(&s2)` doesn't work for odd `len`.

```rust
let mut sep = "";
if matches!(f.precision(), Some(precision) if precision < len/2) {
    // ...
    sep = "..";
}

**Review by shamilsan — APPROVED** [review]
Just minor grumbles 🤷‍♂️

File: core/src/ids.rs (hunk 1 above)
Comment (breathx) [reply core/src/ids.rs:103]: why?
[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d3e7b4350ef3b42649f436adaac4b253)
`let s = len - e1` will skip one element 

But that's really not our case

File: core/src/ids.rs (hunk 2 above)
Comment (breathx) [reply core/src/ids.rs:105]: For me, this works worse than previous variant: it contains for 1 more comparison and 1 more expect that we should proof
```rust
if matches!(f.precision(), Some(precision) if precision < median) {
    e1 = f.precision().expect("Infallible. Checked above.");
    s2 = len - e1;
}
```

Check to be more than zero in meaningless

core/src/ids.rs
breathx: As mentioned above, it works due to intervals specification

core/src/ids.rs
shamilsan: Ok, we can add an internal comparison:

```suggestion
                if let Some(precision) = f.precision() {
                    if precision > median {
```

Zero `precision` looks weird for me.

breathx: > Ok, we can add an internal comparison:
> 
> Zero `precision` looks weird for me.

But it works for floats:
`assert_eq!(format!("{:.0?}", 5.23), "5")`
The same I see reasonable for ids.
Imagine case when you interested only in payload in logs, for example, and debugging messages sent:
you can go with `log::debug!("{message:.0?}")` to ignore ids

core/src/ids.rs
shamilsan: Your solution is the best!

core/src/ids.rs
shamilsan: Agree.

PR #1368, opened 2022-08-24
Implement airdrop in vara stage 1 for initial tokens deposit

ekovalev:
A "surrogate" balance transfer from the pre-funded account (a.k.a. `root_key`) to an arbitraty account in absence of full-blown transfers.
The extrinsic `pallet_airdrop::transfer` must be wrapped in a `sudo` call.


ukint-vs (approved):

shamilsan (commented):
Should we decrease Sudoer's cap in spec?

ekovalev:
> Should we decrease Sudoer's cap in spec?

Yes, that's what we have discussed. But we haven't agreed on the exact number, have we?
This can be done in a separate PR though.

shamilsan (approved):

PR #1374, opened 2022-08-26
[skip ci] add reset worflow for vara

sergeyfilyanin:


shamilsan (approved):
`ansible/vara-reset.yaml` playbook is in the `vara-stage-1` only.

PR #1376, opened 2022-08-27
fix signal handler for `wasmer`

grishasobol:
`wasmer` uses signals to identify wrong memory operation that try to access memory outside of memory bounds. To support it we call old sighandler (which has been set before we set our sig handler for lazy pages) if signal came from unknown address.

Without patch node will panic in the case contract would like to access wrong memory. 

### TODO
This patch do not fix problem for `windows` OS. #1377 
Add also small test which checks that node does not panic in the case. Currently add it inside `runtime-update` test. #1378

### DIFF
--- a/lazy-pages/src/sys/unix.rs
+++ b/lazy-pages/src/sys/unix.rs
@@ -53,13 +53,10 @@ extern "C" fn handle_sigsegv(sig: i32, info: *mut siginfo_t, ucontext: *mut c_vo
         };
 
         if let Err(err) = super::user_signal_handler(exc_info) {
-            let old_sig_handler_works = match err {
-                Error::SignalFromUnknownMemory {
-                    addr: _,
-                    wasm_mem_addr: _,
-                    wasm_mem_end_addr: _,
-                } => old_sig_handler(sig, info, ucontext),
-                _ => false,
+            let old_sig_handler_works = if let Error::SignalFromUnknownMemory { .. } = err {
+                old_sig_handler(sig, info, ucontext)
+            } else {
+                false
             };
             assert!(old_sig_handler_works, "Signal handler failed: {}", err);
         }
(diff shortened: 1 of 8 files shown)


lazy-pages/src/sys/unix.rs
```diff
@@ -43,14 +47,22 @@ extern "C" fn handle_sigsegv(_sig: i32, info: *mut siginfo_t, ucontext: *mut c_v
...
+        if let Err(err) = super::user_signal_handler(exc_info) {
+            let old_sig_handler_works = match err {
+                Error::SignalFromUnknownMemory {
+                    addr: _,
+                    wasm_mem_addr: _,
+                    wasm_mem_end_addr: _,
+                } => old_sig_handler(sig, info, ucontext),
+                _ => false,
+            };
+            assert!(old_sig_handler_works, "Signal handler failed: {}", err);
```
breathx, line 64: why do we need production assertions there?

breathx, line 64:
```suggestion
            let old_sig_handler_works = match err {
                Error::SignalFromUnknownMemory { .. } => old_sig_handler(sig, info, ucontext),
                _ => false,
            };
            assert!(old_sig_handler_works, "Signal handler failed: {}", err);
```

OR

```suggestion
            if let Error::SignalFromUnknownMemory { .. } = err {
                    old_sig_handler(sig, info, ucontext)
            }
```

scripts/test-utils/processingMessageWithRuntimeUpgrade.js
```diff
@@ -4,8 +4,8 @@ const { readFileSync } = require("fs");
 const assert = require("assert/strict");
```
breathx, line 4: could you provide separate test for code upload?
just reuse function from utils

breathx: two random tests together seems really strange

breathx (approved):

scripts/test-utils/processingMessageWithRuntimeUpgrade.js
grishasobol: It's fix solution. I can add TODO. We can add separate test resolving issue #1378 .
I don't wanna to solve test problems in this PR

lazy-pages/src/sys/unix.rs
grishasobol: Because if sig handler finish work with an error we have no ways to continue execution currently.

grishasobol: Ok done

breathx: > Because if sig handler finish work with an error we have no ways to continue execution currently.

So for unrecoverable errors I prefer use panic to have more concrete log

scripts/test-utils/processingMessageWithRuntimeUpgrade.js
breathx: > It's fast solution. I can add TODO. We can add separate test resolving issue #1378 .
> I don't wanna to solve test problems in this PR

Okay, no problem

lazy-pages/src/sys/unix.rs
grishasobol: Ok change to panic

Issue #349, opened 2021-10-19
Time-limited variants of locking primitives 


NikVolf:
Once waiting list can have a time-limit (or program might run out of balance, or user can never reply voluntary), conditions of what is being waited for might resolve as a timeout.

Issue #1331, opened 2022-08-18
Present `Scheduler` and `TaskHandler` in `gtest`


breathx:
### Problem to Solve

Need to allow time-dependent logic for in-memory testing

PR #1379, opened 2022-08-28
Waitlist holding duration control

### LINKED ISSUE #1173: Maximal duration in waitlist

breathx:
Resolves #1173.
Part of #349.

This PR adds functionality to limit holding in waitlist from contracts side:
`gr_wait` - wait as much as possible
`gr_wait_for` - wait exactly for duration
`gr_wait_no_more` - wait as much as possible, but not more than given duration

**Release notes:** see changes and related issues above.

### DIFF
--- a/gcore/src/exec.rs
+++ b/gcore/src/exec.rs
@@ -33,6 +33,8 @@ mod sys {
         pub fn gr_leave() -> !;
         pub fn gr_value_available(val: *mut u8);
         pub fn gr_wait() -> !;
+        pub fn gr_wait_no_more(duration: *const u8) -> !;
+        pub fn gr_wait_for(duration: *const u8) -> !;
         pub fn gr_wake(waker_id_ptr: *const u8);
     }
 }
@@ -174,6 +176,9 @@ pub fn value_available() -> u128 {
 /// function later. All gas that hasn't yet been spent is attributed to the
 /// message in the *waiting queue*.
 ///
+/// This call delays message execution for maximal amount of blocks
+/// that could be payed.
+///
 /// # Examples
 ///
 /// ```
@@ -186,6 +191,19 @@ pub fn value_available() -> u128 {
 /// ```
 pub fn wait() -> ! {
     unsafe { sys::gr_wait() }
+}
+
+/// Same as [`wait`], but delays handling for given specific amount of blocks.
+///
+/// NOTE: It panics, if given duration couldn't be totally payed.
+pub fn wait_for(duration: u32) -> ! {
+    unsafe { sys::gr_wait_for(duration.to_le_bytes().as_ptr()) }
+}
+
+/// Same as [`wait`], but delays handling for maximal amount of blocks
+/// that could be payed, that doesn't exceed given duration.
+pub fn wait_no_more(duration: u32) -> ! {
+    unsafe { sys::gr_wait_no_more(duration.to_le_bytes().as_ptr()) }
 }
 
 /// Resume previously paused message handling.
--- a/gcore/src/exec.rs
+++ b/gcore/src/exec.rs
@@ -82,7 +82,7 @@ pub fn block_timestamp() -> u64 {
 }
 
 /// Terminate the execution of a program. The program and all corresponding data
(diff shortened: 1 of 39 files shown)


pallets/gear/src/internal.rs
```diff
@@ -317,28 +333,32 @@ where
...
+    ) {
+        // `HoldBound` cost builder.
+        let hold_builder = HoldBound::<T>::by(CostsPerBlockOf::<T>::waitlist());
+
+        // Maximal hold bound for the message.
+        let maximal_hold = hold_builder.clone().maximum_for_message(dispatch.id());
+
+        // Figuring out correct hold bound.
+        let hold = if let Some(duration) = duration {
+            hold_builder.duration(duration).min(maximal_hold)
```
gshep, line 349: It should be `duration` for `wait_for` call, shouldn't it?

gcore/src/exec.rs
```diff
@@ -188,6 +193,19 @@ pub fn wait() -> ! {
...
 
+/// Same as [`wait`], but delays handling for given specific amount of blocks.
+///
+/// NOTE: It panics, if given duration couldn't be totally payed.
+pub fn wait_for(duration: u32) -> ! {
+    unsafe { sys::gr_wait_for(duration.to_le_bytes().as_ptr()) }
+}
+
+/// Same as [`wait`], but delays handling for maximal amount of blocks
+/// that could be payed, that doesn't exceed given duration.
```
gshep, line 204: I would ask you to provide some examples to highlight the difference between `wait_for` and `wait_no_more` calls.

gshep (approved):
Good job! :+1:

pallets/gear/src/internal.rs
breathx: Not actually.
`Wait` result of execution contains `Option<BlockNumber>`. If it equals `None`, than we wait as much as possible (`maximum_for_message`), otherwise we wait minimum from as much as possible and maximal queried duration. In general, this is more about `wait_no_more`, but suitable for `wait_for` also, because it will panic in case of not enough funds. So I decided to keep this "check" for `wait_for` also just to make sure that there is enough funds one more time. (It can be replaced with some enum with appropriate to wait types variants, but it's just code repetition for me).

gcore/src/exec.rs
breathx: Frankly speaking, it's hard to do some interesting and useful for ecosystem demo before #1380 (and probably #1331) solved.
But there is already the core test in pallet gear, that shows how it works.

pallets/gear/src/tests.rs
```diff
@@ -2480,6 +2480,342 @@ fn wake_messages_after_program_inited() {
...
+        let wait_success = get_last_message_id();
+
+        run_to_next_block(None);
+
+        assert_eq!(get_waitlist_expiration(wait_success), expiration(duration));
+
+        // Command::WaitFor case.
+        let duration = 5;
+        let payload = Command::WaitFor(duration).encode();
+        let wl_gas = duration_gas(duration) + reserve_gas + 100_000_000;
```
SabaunT, line 2552: what is the origin of that magical `100_000`?

SabaunT (approved):
:+1:

pallets/gear/src/tests.rs
breathx: > what is the origin of that magical `100_000`?

Do you mean this `100_000_000`? It's just a big number of gas given to program.
Frankly speaking, that's copy-paste artefact from `Command::WaitNoMore` case, but that allows better see that despite the fact program has much more money, it will wait exactly appropriate to scenario time.

PR #1383, opened 2022-08-29
Separate client upload tests

### LINKED ISSUE #1378: make util to run tests in real node
### Problem to Solve

Currently we have only one (two) tests which run smart contracts using real node:
https://github.com/gear-tech/gear/blob/a3ff02efad762656743c450935d2ea6a875b81f7/.github/workflows/CI.yaml#L265-L266

To increase tests coverage and quality we need to run more contracts on real node.

### Possible Solution

1) use https://github.com/gear-tech/gear-program
2) use polka-dot api as it used now. But then we need more comfortable and wide scripts to run many contracts there

breathx:
Resolves #1378.

`./scripts/gear.sh test uploads "any" "amount" "of" "demos"` to check that their upload doesn't stop the node, independently from program init status

### DIFF
--- a/scripts/src/test.sh
+++ b/scripts/src/test.sh
@@ -23,6 +23,7 @@ test_usage() {
     pallet         run pallet-gear tests
     runtime-upgrade run runtime-upgrade test for queue processing
     client-weights run client weight test for infinite loop demo execution
+    uploads        run client upload code test
     fuzz           run fuzzer with a fuzz target
 
 EOF
@@ -91,7 +92,6 @@ runtime_upgrade_test() {
   TEST_SCRIPT_PATH="$ROOT_DIR/scripts/test-utils"
   RUNTIME_PATH="$ROOT_DIR/scripts/test-utils/gear_runtime.compact.compressed.wasm"
   DEMO_PING_PATH="$ROOT_DIR/target/wasm32-unknown-unknown/release/demo_ping.opt.wasm"
-  DEMO_WRONG_LOAD_PATH="$ROOT_DIR/target/wat-examples/wrong_load.wasm"
 
   # Run node
   RUST_LOG="pallet_gear=debug,runtime::gear=debug" $ROOT_DIR/target/release/gear-node \
@@ -101,7 +101,7 @@ runtime_upgrade_test() {
   cd "$TEST_SCRIPT_PATH"
 
   # Run test
-  npm run upgrade "$RUNTIME_PATH" "$DEMO_PING_PATH" "$DEMO_WRONG_LOAD_PATH"
+  npm run upgrade "$RUNTIME_PATH" "$DEMO_PING_PATH"
 
   # Killing node process added in js script
 }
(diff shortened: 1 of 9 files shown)


osipov-mit (approved):

scripts/test-utils/harmlessUploadDemos.js
```diff
@@ -0,0 +1,53 @@
...
+  main(args)
+    .then(() => {
+      exitCode = 0;
+    })
+    .catch((error) => {
+      console.error(`JS_TEST: ${error}`);
+      exitCode = 1;
+    })
+    .finally(() => {
+      exec('kill -9 $(pgrep -a gear-node)', (err, stdout, stderr) => {
```
grishasobol, line 39: Do we really need to kill node after each js test execution ? May be better make one kill in CI after all tests are passed

scripts/test-utils/harmlessUploadDemos.js
```diff
@@ -0,0 +1,53 @@
...
+    // Check that it is root
+    assert.ok((await api.query.sudo.key()).eq(account.addressRaw));
+
+    const isInitialized = checkInit(api);
+
+    for (const pathToDemo of demoPaths) {
+        console.log(`Uploading demo: ${pathToDemo}`);
+
+        const [_programId, messageId] = await uploadProgram(api, account, pathToDemo);
+        await isInitialized(messageId);
```
grishasobol, line 22: `wrong load ` cannot be initialised. We just check that node does not panic after we try to upload this kind of test.

scripts/src/test.sh
```diff
@@ -125,6 +125,26 @@ client_weights_test() {
...
+# $1 - ROOT DIR
+uploads_test() {
+  ROOT_DIR="$1"
+  shift
+
+  TEST_SCRIPT_PATH="$ROOT_DIR/scripts/test-utils"
+
+  # Run node
+  RUST_LOG="gear=debug,gwasm=debug" $ROOT_DIR/target/release/gear-node \
+  --dev --tmp --unsafe-ws-external --unsafe-rpc-external --rpc-methods Unsafe --rpc-cors all & sleep 7
```
grishasobol, line 137: `sleep 7` is old timer. We can use `sleep 3` now

grishasobol (changes requested):

scripts/test-utils/harmlessUploadDemos.js
breathx: In that case we should take care of node state: for example, that's not fully correct to test anything else after runtime upgrade test
Also, than we should take care of killing node, while running js tests locally via `gear.sh`

There is no overhead, cause we use pre-compiled node binary

scripts/src/test.sh
breathx: For me wait 4 seconds more to make sure, that any machine successfully ran node is ok, isn't it?

grishasobol: Yep) just FYI

scripts/test-utils/harmlessUploadDemos.js
grishasobol: We can make one function in test.sh which will run node, then run all js scripts, then kill it. Node state is not so bad I think, because it adds some variability

breathx: What do you think about adding flag --kill-node to scripts?
like npm run upload --kill-node ..other args

and we can leave this tests as it is atm, but kill node in CI just once?

scripts/src/test.sh
breathx: Done

scripts/test-utils/harmlessUploadDemos.js
breathx: Done

grishasobol (approved):

PR #1397, opened 2022-08-31
Fix script sometimes skips spec check

ark0f:
`check-spec.sh` skips Vara version check if Gear have an inactual version

### DIFF
--- a/scripts/check-spec.sh
+++ b/scripts/check-spec.sh
@@ -32,7 +32,7 @@ check_spec() {
         printf "\n    Spec version is correct.\n"
     else
         printf "\n    Spec version should $version.\n"
-        check_spec_result=1
+        EXIT_CODE=1
     fi
 }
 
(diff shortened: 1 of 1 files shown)


scripts/check-spec.sh
```diff
@@ -32,7 +32,7 @@ check_spec() {
         printf "\n    Spec version is correct.\n"
     else
         printf "\n    Spec version should $version.\n"
-        exit 1
+        check_spec_result=1
```
shamilsan, line 35: Better to use the upper case for variable in the shell.
```suggestion
        CHECK_SPEC_RESULT=1
```

ark0f: What's the difference?

scripts/check-spec.sh
```diff
@@ -52,8 +52,14 @@ for package in $(git diff --name-only origin/master | grep ".rs$" | cut -d "/" -
...
+
 header "Checking spec for Gear runtime"
 check_spec "$SPEC_ON_MASTER" "$ACTUAL_SPEC_GEAR" "$CHANGES"
 
 header "Checking spec for Vara runtime"
 check_spec "$SPEC_ON_MASTER" "$ACTUAL_SPEC_VARA" "$CHANGES"
+
+if [[ $check_spec_result -ne 0 ]]; then
+  exit 1
+fi
```
breathx, line 65: Rename CHECK_SPEC_RESULT into EXIT_CODE and won't approach below be easier?
```suggestion
exit $EXIT_CODE
```

breathx (changes requested):

breathx (approved):

PR #1401, opened 2022-08-31
ci(audit): add audit check

### LINKED ISSUE #1400: Do `audit-check` on PRs
### Problem to Solve

https://crates.io/crates/cargo-audit

### Possible Solution

not sure if it works fine (hope not panicking all the time) with substrate since the dependences of it is a mess

### Notes

_No response_

clearloop:
Resolves #1400 

- add `audit-check`
-
-

@gear-tech/dev

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,20 @@
+name: Audit
+
+on:
+  push:
+  pull_request:
+    paths:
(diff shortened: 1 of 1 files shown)


breathx:
Seems that this stuff will force us to update substrate everyday, isn't it?

I'd run this check only if some label appeared

PR #1409, opened 2022-09-02
Disable spec_version job

ukint-vs:
Needed for resetting spec_version.

@gear-tech/dev


breathx (changes requested):
Why do we need this?
Resetting spec version was mentioned in script already

PR #1419, opened 2022-09-04
Implement lazy pages comparison optimization for macOS

ark0f:
Also new:
* Add ability to write user defined signal handler for lazy pages
* Read-write flag test


lazy-pages/src/sys/unix.rs
```diff
@@ -29,24 +30,41 @@ thread_local! {
     static OLD_SIG_HANDLER: RefCell<SigHandler> = RefCell::new(SigHandler::SigDfl);
 }
 
-extern "C" fn handle_sigsegv(sig: i32, info: *mut siginfo_t, ucontext: *mut c_void) {
-    unsafe {
-        #[cfg(all(target_os = "linux", target_arch = "x86_64"))]
-        let is_write = {
-            let ucontext = ucontext as *const nix::libc::ucontext_t;
+cfg_if! {
```
breathx, line 33: Why can't we just use cfg attributes?
target os = linux and = macOS are incompatible on compile, so it should be fine

lazy-pages/src/sys/unix.rs
```diff
@@ -29,24 +30,41 @@ thread_local! {
...
+    if #[cfg(all(target_os = "linux", target_arch = "x86_64"))] {
+        unsafe fn ucontext_get_write(ucontext: *mut nix::libc::ucontext_t) -> Option<bool> {
             let error_reg = nix::libc::REG_ERR as usize;
             let error_code = (*ucontext).uc_mcontext.gregs[error_reg];
             // Use second bit from err reg. See https://git.io/JEQn3
             Some(error_code & 0b10 == 0b10)
-        };
+        }
+    } else if #[cfg(all(target_os = "macos", target_arch = "aarch64"))] {
```
breathx, line 41: and what if we use macos x86?

grishasobol (changes requested):
We dont have CI checks for macos and especially for aarch64. So, you need to test everything by hands.

grishasobol:
> We dont have CI checks for macos and especially for aarch64. So, you need to test everything by hands.

I think it will be good idea to make some test for that cases. Which checks that this bits works as we expect

lazy-pages/src/sys/unix.rs
ark0f: They are not because `mcontext_t` have different definitions on these platforms

Also we have `#[cfg(all(target_os = "linux", target_arch = "x86_64"))]` and `#[cfg(all(target_os = "macos", target_arch = "aarch64"))]` and their negation will look like 
```
#[cfg(not(any(
    all(target_os = "linux", target_arch = "x86_64"), 
    all(target_os = "macos", target_arch = "aarch64")
)))]
```

Which is quite ugly

lazy-pages/src/sys/unix.rs
ark0f: `None` is returned then. Also I don't have x86 macOS machine to implement the logic for it

breathx:
> > We dont have CI checks for macos and especially for aarch64. So, you need to test everything by hands.
> 
> I think it will be good idea to make some test for that cases. Which checks that this bits works as we expect

There is no GitHub aarch64 runners, and it's pretty hard to find such personal server :(

lazy-pages/src/sys/unix.rs
breathx: > `None` is returned then. Also I don't have x86 macOS machine to implement the logic for it

Do you mean that this is still opened issue, right?

ark0f: Yes

lazy-pages/src/sys/unix.rs
breathx: Why do we need their negation?
We should write placeholder for macOS x86 and dont support any other platforms so (it will cause panic in compile-time)

ark0f: Ok, good suggestion. Will do

.github/workflows/CI.yaml
```diff
@@ -459,6 +459,47 @@ jobs:
             -p gear-runtime-interface
             --features=lazy-pages
 
+  build-macos-x86:
```
grishasobol, line 462: I think we shouldn't make separate run for macos-x86. We can make nightly run like for windows. It's not critically important for us to test each commit for macos.

ark0f: > We can make nightly run like for windows

It's every master push actually. Same will be for this job but I'm testing right now

breathx (approved):

lazy-pages/src/sys.rs
```diff
@@ -166,6 +173,14 @@ impl CheckedWasmAddr {
     }
 }
 
+pub struct DefaultUserSignalHandler;
+
+impl UserSignalHandler for DefaultUserSignalHandler {
```
grishasobol, line 178: What the reason to use struct instead of function?

grishasobol: And what the reason to add it as template param?

ark0f: > What the reason to use struct instead of function?

Because you cannot just pass function pointer in generic

> And what the reason to add it as template param?

Because I don't want to spend additional resources for something like TLS

grishasobol: Can you please explain more widely cannot currently see any additional resources usages

ark0f: > currently

That's why I use generics. You ask me:

> And what the reason to add it as template param?

So I thought you mean it should be done in an another way. For example, thread local storage, which gives some overhead

grishasobol: Currently we have only one `DefaultSignalHandler` ? What the reason to make it as generic param ? Even if we will have two signal handlers in future they can have different args or returns

ark0f: > What the reason to make it as generic param ?

Tests

grishasobol (approved):

lazy-pages/src/sys.rs
grishasobol: Ok see now

PR #1433, opened 2022-09-05
Benchmarking script. Runtime-specific weights.

### LINKED ISSUE #1372: Keep benchmarked weights in runtime(s)
### Problem to Solve

Several pallets (timestamp, utility, etc) have weights for their extrinsics/operations. At the moment out of the box pallet weights are used in runtimes though we should use benchmarked weights for our own hardware.

### Possible Solution

Create benchmark.sh script to produce required weights (check https://github.com/gear-tech/gear/pull/1369/files#diff-64a10db0496f3cf1db9180a6839c71cba22926b9b79f8dd6a51942f3eb7741e2, or https://github.com/paritytech/cumulus/blob/master/scripts/benchmarks.sh). Adjust runtime to use this very benchmarked values.

### Notes

_No response_

ukint-vs:
Resolves #1372  .

Usage of `run_all_benchmarks.sh`

```bash
./scripts/benchmarking/run_all_benchmarks.sh -c gear-dev
./scripts/benchmarking/run_all_benchmarks.sh -c vara-dev -f -v
```

@gear-tech/dev


.maintain/frame-weight-template.hbs
```diff
@@ -0,0 +1,104 @@
...
+/// Weight functions needed for {{pallet}}.
+pub trait WeightInfo {
+    {{#each benchmarks as |benchmark|}}
+    fn {{benchmark.name~}}
+    (
+        {{~#each benchmark.components as |c| ~}}
+        {{c.name}}: u32, {{/each~}}
+    ) -> Weight;
+    {{/each}}
+}
```
gshep, line 45: I suppose this trait shouldn't be generated since the pallet's one is used.

.maintain/frame-weight-template.hbs
```diff
@@ -0,0 +1,104 @@
...
+            {{/if}}
+            {{#each benchmark.component_writes as |cw|}}
+            .saturating_add(T::DbWeight::get().writes(({{cw.slope}} as Weight).saturating_mul({{cw.name}} as Weight)))
+            {{/each}}
+    }
+    {{/each}}
+}
+
+// For backwards compatibility and tests
+impl WeightInfo for () {
```
gshep, line 78: This is also unnecessary because of the implementation in `pallets/gear/weights.rs`

gshep (approved):
Great!

I would like to ask you to also provide weights for `pallet_babe`, `pallet_grandpa`, `pallet_session`, `pallet_utility`, `pallet_gear_debug` and `pallet_gear_program`.

.maintain/frame-weight-template.hbs
ukint-vs: Consensus weights usually comes from their own weights. All other pallets can't be benchmarked as they don't have an export to do so. I'm investigating it.

.maintain/frame-weight-template.hbs
ukint-vs: Common practice to leave it.

.maintain/frame-weight-template.hbs
gshep: Maybe we need to add them to
```rust
#[cfg(feature = "runtime-benchmarks")]
mod benches {
    define_benchmarks!(
        [frame_benchmarking, BaselineBench::<Runtime>]
        [frame_system, SystemBench::<Runtime>]
// ...
```

sectino?

.maintain/frame-weight-template.hbs
ukint-vs: This template is needed for pallet weights also.

.maintain/frame-weight-template.hbs
ukint-vs: btw pallet-session-benchmarking needs pallet-staking. I think we can add this later.

PR #1457, opened 2022-09-08
feat(gstd): introduce wait locks

### LINKED ISSUE #1380: `gstd` async sending limitations
### Problem to Solve

Once #1379 merged, it uses `gr_wait_no_more(100)` as default behaviour for waiting in async scenarios.
There should be possibility to specify this value and wait-type correctly (and resolve the issue that once we wake and wait again on not-finished future, we wait for 100 blocks again, but not for the difference).

### Possible Solution

```rust
send_for_reply(..) // some async send
    .expect("Failed to send message") // returns Future
    // default behaviour is wait_no_more(gstd::SOME_CONFIGURABLE_CONST
    .no_longer(300) // wait no more than 300 blocks
    // OR
    // .for(400) to wait exactly for 400 blocks
    .await // returns result 
```

### Notes

Gasfull and gasless futures should have different methods (need to clippy warn on usage of gasless sending with unlimited wait, or probably deny it)

### LINKED ISSUE #1479: Handle timeout of the wait locks
based on #1457 

Maybe I've also missed some corner cases and usage issues, to be continued after remarks fixed.

This approach doesn't fix the most important parts of initial issue:
1. If timeout of waiting reply reached, future should be marked ready and error returned.
2. Imagine the situation:
- While processing message A, we send message B and `wait_for(100)` for reply.
- In 90 blocks some other message and contract's logic branch lead to `exec::wake(A)`
- We start processing A and run into pending future - we go into other branch of `message_loop` and call `wait_for(100)` again!
- If nothing happens further, we have waited 190 blocks instead of exactly 100 (I'd suggest to store block number of first polling also, to properly calculate waiting bound of `wait_for` and `wait_no_more`)

_Originally posted by @breathx in https://github.com/gear-tech/gear/pull/1457#pullrequestreview-1102442896_

clearloop:
Resolves #1380.
Resolves #1479 

- [x] introduce global config for `gstd`
- [x] introduce `LocksMap` in `gstd::async_runtime`n
- [x] introduce `up_to` and `exactly` for `FutureMessages`
- [x] throw errors when messages are timeout
- [x] tests
  - [x] wait locks
    - [x] message sending with wait for
    - [x] message sending with wait no more
    - [x] wait after wait
  - [x] wait timeout
  - [x] multiple messages wait timeout (with future operations)
    - [x] select
    - [x] join

@gear-tech/dev


gstd/src/async_runtime/locks.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+impl Default for Lock {
+    fn default() -> Self {
+        unsafe { Lock::NoMore(crate::config::DEFAULT_WAIT_NO_MORE_DURATION) }
+    }
+}
+
+/// Wait trait for async sending messages.
+pub trait Wait: Sized {
```
breathx, line 28: There are several problems with this trait:
1. We should always import this: I don't find this beauty for our contacts scheme
2. Doc comments of methods are messed up: `no_more` is about as much as possible, `till` about concrete waiting duration
3. `till(block_number)` naming for me means that we will wait until `block_number` comes: I'd suggest rename it into `exactly(..)` if you want to keep single word (I usually prefer it too)

gstd/src/async_runtime/locks.rs
```diff
@@ -0,0 +1,66 @@
...
+    /// Delays handling for given specific amount of blocks.
+    fn no_more(self, duration: u32) -> Self {
+        async_runtime::locks().insert(crate::msg::id(), Lock::NoMore(duration));
```
breathx, line 31: Such realisation of functions is bugged:
```rust
let f = msg::send_bytes_for_reply(to, [], 0)
    .expect("send message failed")
    .no_more(10);

let _ = msg::send_bytes_for_reply(to, [], 0)
    .expect("send message failed")
    .no_more(5);
  
let _res = f.await; // this will wait no more 10, instead of proper 5 
```

breathx: To solve this, we should set lock type on polling of future.

breathx: Also here might be a problem of polling several futures: `FuturesUnordered/join!(..)/etc.` - need to find the general solution and provide tests (probably deny `StreamExt` for custom waiting)

gstd/src/async_runtime/signals.rs
```diff
@@ -90,7 +90,11 @@ impl WakeSignals {
             Some(WakeSignal {
                 payload: Some(payload),
                 ..
-            }) => ReplyPoll::Some(payload),
+            }) => {
+                // Remove lock after waking.
+                super::locks().remove(&crate::msg::id());
```
breathx, line 95: Seems that's not correct place for this call (as minimum, usage should be somewhere in place `wait/wake/poll` called, further - it produces same bug as I mentioned before)

breathx (changes requested):
Maybe I've also missed some corner cases and usage issues, to be continued after remarks fixed.

This approach doesn't fix the most important parts of initial issue:
1. If timeout of waiting reply reached, future should be marked ready and error returned.
2. Imagine the situation:
- While processing message A, we send message B and `wait_for(100)` for reply.
- In 90 blocks some other message and contract's logic branch lead to `exec::wake(A)`
- We start processing A and run into pending future - we go into other branch of `message_loop` and call `wait_for(100)` again!
- If nothing happens further, we have waited 190 blocks instead of exactly 100 (I'd suggest to store block number of first polling also, to properly calculate waiting bound of `wait_for` and `wait_no_more`)

breathx:
Probably makes sense separate this issue into several PRs:
- [ ] Introduce `Config`.
- [ ] Solve the issue with timeout error.
- [ ] Solve the issue with extra polling (waking) of message, that leads to excess waiting.
- [ ] Add customisable waiting bound for different messages.

gstd/src/msg/async.rs
```diff
@@ -73,6 +73,21 @@ impl<D: Decode> Future for CodecMessageFuture<D> {
     }
 }
 
+impl<D: Decode> CodecMessageFuture<D> {
+    /// Delays handling for given specific amount of blocks.
+    pub fn no_more(self, duration: u32) -> Self {
```
shamilsan, line 78: Need to rename all `no_more` to `up_to`.

shamilsan:
Stale PR? @clearloop

gstd/src/async_runtime/locks.rs
```diff
@@ -0,0 +1,146 @@
...
+        let map = self.0.entry(message_id).or_insert_with(Default::default);
+        if map.is_empty() {
+            // If there is no `waiting_reply_to` id specfied, use
+            // the message id as the key of the message lock.
+            // (the key should to `waiting_replay_to` in general )
+            //
+            // # TODO
+            //
+            // refactor this implementation when we got better solution.
+            map.insert(message_id, Default::default());
```
breathx, line 112: for me seems that we should guarantee that map is never empty if this method called

breathx: cause this default is just for specific await, not for overall execution

gstd/src/async_runtime/locks.rs
```diff
@@ -0,0 +1,146 @@
...
+        locks.insert(waiting_reply_to, lock);
+    }
+
+    /// Remove lock of message.
+    pub fn remove(&mut self, message_id: MessageId, waiting_reply_to: MessageId) {
+        let locks = self.0.entry(message_id).or_insert_with(Default::default);
+        locks.remove(&waiting_reply_to);
+    }
+
+    /// Check if message is timeout.
```
breathx, line 137:
```suggestion
    /// Check if message is timed out.
```

gstd/src/async_runtime/locks.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn is_timeout(
+        &mut self,
+        message_id: MessageId,
+        waiting_reply_to: MessageId,
+    ) -> Option<(u32, u32)> {
+        let locks = self.0.entry(message_id).or_insert_with(Default::default);
+        locks.get(&waiting_reply_to).and_then(|l| l.timeout())
+    }
```
breathx, line 145:
```suggestion
    pub fn is_timeout(
        &self,
        message_id: MessageId,
        waiting_reply_to: MessageId,
    ) -> Option<(u32, u32)> {
        self.0.entry(message_id).and_then(|locks| locks.get(&waiting_reply_to).and_then(|l| l.timeout())).flatten()
    }
```

breathx: Ungrounded request for mutable borrowing here for me

gstd/src/msg/async.rs
```diff
@@ -48,6 +55,9 @@ where
                 return Poll::Ready(Err(ContractError::ExitCode(exit_code)));
             }
 
+            // Remove lock after waking.
```
breathx, line 58: Shouldn't this be also executed if logic in lines 54-56 (things about about returning error reply) triggered?

gstd/src/msg/async.rs
```diff
@@ -38,6 +38,13 @@ fn poll<F, R>(waiting_reply_to: MessageId, cx: &mut Context<'_>, f: F) -> Poll<R
 where
     F: Fn(Vec<u8>) -> Result<R>,
 {
+    let msg_id = crate::msg::id();
+
+    // check if message is timeout
+    if let Some((expected, now)) = async_runtime::locks().is_timeout(msg_id, waiting_reply_to) {
```
breathx, line 44: shouldn't we remove lock in timeout case also?

breathx (changes requested):

examples/binaries/wait-timeout/src/code.rs
```diff
@@ -0,0 +1,107 @@
...
+                .up_to(Some(5))
+                .expect("Invalid wait duration.");
+
+            if let Err(e) = msg::send_bytes_for_reply(to, b"", 0)
+                .expect("send message failed")
+                .up_to(Some(10))
+                .expect("Invalid wait duration.")
+                .await
+            {
+                if e.timed_out() {
```
breathx, line 31: otherwise panic mb?

gstd/src/async_runtime/locks.rs
```diff
@@ -0,0 +1,189 @@
...
+                );
+            }
+
+            match self.ty {
+                LockType::WaitFor(_) => exec::wait_for(blocks),
+                LockType::WaitUpTo(_) => exec::wait_up_to(blocks),
+            }
+        } else {
+            unreachable!(
+                "Checked in `crate::msg::async::poll`, will trigger the tiemout error automatically."
```
breathx, line 84:
```suggestion
                "Checked in `crate::msg::async::poll`, will trigger the timeout error automatically."
```

gstd/src/async_runtime/locks.rs
```diff
@@ -0,0 +1,146 @@
...
+            ty: LockType::WaitUpTo(b),
+        }
+    }
+
+    /// Call wait functions by the lock type.
+    pub fn wait(&self) {
+        if let Some(blocks) = self.bound().checked_sub(exec::block_height()) {
+            match self.ty {
+                LockType::WaitFor(_) => exec::wait_for(blocks),
+                LockType::WaitUpTo(_) => exec::wait_up_to(blocks),
```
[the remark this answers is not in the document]
breathx: so what about to ignore this thing in case of wait_up_to?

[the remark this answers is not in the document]
breathx: or open considering issue about this stuff please

gstd/src/config.rs
```diff
@@ -0,0 +1,75 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//! Declares modules, attributes, public re-exports.
+//! Gear libs are `#![no_std]`, which makes them lightweight.
+
+//! This module is for configuring `gstd` inside gear programs.
+use crate::errors::{ContractError, Result};
+
+/// `gstd` configuration
+pub struct Config {
```
breathx, line 26: what about also allow specify default: wait up to / wait for
like I just once write set_default_up_to(500)
and each time I call just await it triggers **up_to** 500

breathx (approved):
Awesome!

Issue #1444, opened 2022-09-06
Lazy pages signal handler overlaps with another ones during simultaneous tests run


ark0f:
### Problem

In file `lazy-pages/src/sys/unix.rs`:
```rust
let error_reg = nix::libc::REG_ERR as usize;
let error_code = (*ucontext).uc_mcontext.gregs[error_reg];
```
`error_code` is sometimes zero during during simultaneous tests run when it's expected to be `0b10`

### Steps

Run `cargo test --package gear-lazy-pages --lib tests::read_write_flag_works` (in #1419)

### Possible Solution

- Run tests in 1 thread
- Run tests with signal handler separately

### Notes

| Machine| Kernel | CPU | VM | Works? | Notes |
| --- | --- | --- | --- | --- | --- |
| EPYC CI | 5.15.0-25-generic | AMD EPYC 7502P (64) @ 2.500GHz | ✔️ | ❌ | https://github.com/gear-tech/gear/runs/8208657278?check_suite_focus=true#step:22:754. |
| EPYC CI | 5.15.0-25-generic | AMD EPYC 7502P (64) @ 2.500GHz | ✔️ | ✔️ | `docker run -it ubuntu` |
| EPYC CI | 5.15.0-25-generic | AMD EPYC 7502P (64) @ 2.500GHz | ✔️ | ❌ | `docker run -it ubuntu` with simultaneously tests |
| EPYC CI | 5.15.0-25-generic | AMD EPYC 7502P (64) @ 2.500GHz | ❌ | ✔️ | Linux environment |
| Xeon CI | 5.4.0-122-generic | Intel Xeon E5-2695 v3 (28) @ 3.300GHz | ✔️ | ✔️ | `docker run -it ubuntu` |
| Xeon CI | 5.4.0-122-generic | Intel Xeon E5-2695 v3 (28) @ 3.300GHz | ❌ | ✔️ | Linux environment |
| @ark0f's machine | 4.4.0-19041-Microsoft | AMD Ryzen 5 3600 (12) @ 3.600GHz | ✔️ | ❌ | WSL1 (possibly bad implementation) |
| @ark0f's machine | 5.10.16.3-microsoft-standard-WSL2 | AMD Ryzen 5 3600 (12) @ 3.600GHz | ✔️ | ✔️ | WSL2 |
| @ark0f's machine | 5.18.17 | AMD Ryzen 5 3600 (12) @ 3.600GHz | ❌ | ✔️ | Fedora |
[...]

PR #1461, opened 2022-09-08
fix problem with wasmer fault when lazy pages set sigaction

### LINKED ISSUE #1421: Node drop execution with sigfault
### Problem

When we tries to upload specific contracts to node on M1 macos, then we receive node execution drop with message
```
illegal hardware instruction
```

### Steps

1. env
```
main sc_cli::runner: ⛓  Native runtime: gear-node-1680 (gear-node-1.tx1.au1)
main sc_sysinfo: 💻 Operating system: macos
main sc_sysinfo: 💻 CPU architecture: aarch64
```
2. run node
```
./target/release/gear-node --dev  --execution=wasm
```
3. upload smart contract [lol.wasm.zip](https://github.com/gear-tech/gear/files/9485482/lol.wasm.zip)
4. receive an error during `on_idle`
```
illegal hardware instruction
```

### Possible Solution

May be there is some bugs  in `wasmer` for M1 chip - `wasmer` compile WASM code into machine code with some number of  illegal instructions.

grishasobol:
Resolves #1421

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -248,7 +248,9 @@ jobs:
         run: ./scripts/gear.sh test client-weights
 
       - name: "Test: Upload programs"
-        run: ./scripts/gear.sh test uploads "$(pwd)/target/wat-examples/wrong_load.wasm"
+        run: |
+          ./scripts/gear.sh test uploads "$(pwd)/target/wat-examples/wrong_load.wasm"
+          ./scripts/gear.sh test uploads "$(pwd)/target/wat-examples/inf_recursion.wasm"
 
       - name: "Test: Runtime benchmarks work"
         run: |
--- a/lazy-pages/src/sys/unix.rs
+++ b/lazy-pages/src/sys/unix.rs
@@ -27,11 +27,17 @@ use nix::{
     libc::{c_void, siginfo_t},
     sys::{signal, signal::SigHandler},
 };
-use std::{cell::RefCell, io};
+use once_cell::sync::OnceCell;
+use std::io;
 
-thread_local! {
-    static OLD_SIG_HANDLER: RefCell<SigHandler> = RefCell::new(SigHandler::SigDfl);
-}
+/// Signal handler which has been set before lazy pages initialization.
+/// Currently use to support wasmer signal handler.
+/// Wasmer protects memory around wasm memory and for stack limits.
+/// It makes it only in `store` initialization when executor is created,
+/// see https://github.com/gear-tech/substrate/blob/gear-stable/client/executor/common/src/sandbox/wasmer_backend.rs
+/// and https://github.com/wasmerio/wasmer/blob/e6857d116134bdc9ab6a1dabc3544cf8e6aee22b/lib/vm/src/trap/traphandlers.rs#L548
+/// So, if we receive signal from unknown memory we should try to use old (wasmer) signal handler.
+static mut OLD_SIG_HANDLER: OnceCell<SigHandler> = OnceCell::new();
 
 cfg_if! {
     if #[cfg(all(target_os = "linux", target_arch = "x86_64"))] {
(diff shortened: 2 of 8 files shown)


lazy-pages/src/sys/unix.rs
```diff
@@ -67,9 +68,21 @@ extern "C" fn handle_sigsegv(sig: i32, info: *mut siginfo_t, ucontext: *mut c_vo
 
 pub unsafe fn setup_signal_handler() -> io::Result<()> {
     let handler = signal::SigHandler::SigAction(handle_sigsegv);
+
+    // Set additional SA_ONSTACK and SA_NODEFER to avoid problems with wasmer executor.
+    // See comment from shorturl.at/KMO68 :
+    // ```
+    //  SA_ONSTACK allows us to handle signals on an alternate stack,
```
ark0f, line 75: It won't work without [sigaltstack(2)](https://man7.org/linux/man-pages/man2/sigaltstack.2.html). After the change it will also fix #1444

grishasobol: > It won't work without sigaltstack(2)
Are you sure about that ? As you can see `runtime-spec` and `pallet` test works well. They don't use `wasmer` as executor. Also I just check it for `wasmi` host executor - all goes right as well. I think it works on a common stack without sigaltstack 

> After the change it will also fix #1444
Really ? Cannot understand why))

ark0f: > test works well

Hm, maybe because:

> Rust installs an alternate stack with sigaltstack, so we rely on that.

> > After the change it will also fix https://github.com/gear-tech/gear/issues/1444
>
> Really ? Cannot understand why))

If we will install a new stack for each thread - signal won't overlap anymore

grishasobol: Ok, I see now

.github/workflows/CI.yaml
```diff
@@ -248,7 +248,9 @@ jobs:
         run: ./scripts/gear.sh test client-weights
 
       - name: "Test: Upload programs"
-        run: ./scripts/gear.sh test uploads "$(pwd)/target/wat-examples/wrong_load.wasm"
+        run: |
+          ./scripts/gear.sh test uploads "$(pwd)/target/wat-examples/wrong_load.wasm"
+          ./scripts/gear.sh test uploads "$(pwd)/target/wat-examples/inf_recursion.wasm"
```
breathx, line 253:
```suggestion
          ./scripts/gear.sh test uploads \
          "$(pwd)/target/wat-examples/wrong_load.wasm" \
          "$(pwd)/target/wat-examples/inf_recursion.wasm"
```

grishasobol: https://github.com/gear-tech/gear/issues/1462

breathx (approved):

lazy-pages/src/sys/unix.rs
```diff
@@ -120,21 +138,28 @@ where
     };
 
     let old_sigaction = signal::sigaction(signal, &sig_action).map_err(io::Error::from)?;
-    OLD_SIG_HANDLER.with(|sh| *sh.borrow_mut() = old_sigaction.handler());
+    let handler = old_sigaction.handler();
+    let _ = OLD_SIG_HANDLER
+        .set(handler)
```
ark0f, line 143: Won't we face problem old handler is not saved in multithreading mode?

> Returns Ok(()) if the cell was empty and Err(value) if it was full.

ark0f (approved):

lazy-pages/src/sys/unix.rs
ark0f: Signal handler is always only one for the whole process - so nvm

grishasobol: All is ok

PR #1472, opened 2022-09-13
Drop custom session management for all runtimes

ekovalev:
We don't need custom session management anymore - in any case, until proper elections have been implemented, authorities for each next session are selected by the `SessionManager`, and all our custom manager has been doing recently (since the selected fraction has been set to 100% in `vara`) is carrying forward authorities from the previous session to a new one. This is essentially what the "dummy" session manager from Substrate (that is, the unit type `()`) does.

Caveat: in `vara` network this all is done differently anyway, since there is staking pallet in the runtime, so more sophisticated session management can be constructed. This PR is mostly to keep `master` as more in-sync with `vara-stage-1` branch as possible


breathx (approved):
Yeah, seems good. But the only moment need to test: if I'm not mistaken, removing pallets from runtime requires not only spec_version bump, but impl(?) version also, to be successfully upgraded from the older one.

gshep (approved):
Great!

ekovalev:
> Yeah, seems good. But the only moment need to test: if I'm not mistaken, removing pallets from runtime requires not only spec_version bump, but impl(?) version also, to be successfully upgraded from the older one.

Could be - need to double check. Good point.

PR #1486, opened 2022-09-14
Run JS tests in Windows CI

### LINKED ISSUE #1422: Windows CI testing
### Problem to Solve

We have some platform-dependent logic, which should be checked not only for compiling (as it does atm).

### Possible Solution

Introduce at least node-client js tests (from main CI) for windows workflow

ark0f:
Resolves #1422.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -387,10 +387,10 @@ jobs:
 
       - name: "Install: cargo-hack"
         run: |
-          $tmp = New-TemporaryFile | Rename-Item -NewName { $_ -replace 'tmp$', 'zip' } -PassThru
+          $tmp = New-TemporaryFile | Rename-Item -NewName { $_ -replace 'tmp$', 'tar.gz' } -PassThru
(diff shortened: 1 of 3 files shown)


ark0f:
Successful run: https://github.com/gear-tech/gear/actions/runs/3060538761/jobs/4939178020

.github/workflows/CI.yaml
```diff
@@ -351,11 +354,18 @@ jobs:
...
+        with:
+          path-type: inherit
+          install: procps
+
       - name: "Install: rustup 1.24.3"
         run: |
           Invoke-WebRequest "https://static.rust-lang.org/rustup/archive/1.24.3/x86_64-pc-windows-msvc/rustup-init.exe" -OutFile "rustup-init.exe"
           .\rustup-init.exe -y
           rustup --version
+        shell: powershell
```
shamilsan, line 368: Why do we set `powershell` explicitly if it is enabled by default?

ark0f: It's not actually: https://github.com/gear-tech/gear/pull/1486/files#diff-b268bb7dc66e8638921e223098a11eabb92b424875ec72a19d6145dc1efbe3f2R341

shamilsan (approved):

PR #1488, opened 2022-09-14
Vara runtime improvements

ekovalev:
Vara runtime improvements:

- replacement of a dummy session manager by a real one. This doesn't change the behaviour of the runtime - all authorities are still being carried forward from previous session to the next one. It will start making difference at later stages;
- replacement of the `GenesisElectionProvider` associated type in `pallet_staking::Config` with something that would select all "electable" targets rather than throwing an error as before. This is done in order for the `try_trigger_new_era` to succeed triggering the first era during genesis block processing and not emitting annoying error events visible in UI every time election is called at the start of a session. Subsequently elections are not tried (the `force_era` is set to`Forcing::ForceNone` in `staking` genesis config which indefinitely avoids a new era), so the `ElectionProvider` type is still the one that throws an error - it is never tried anyway. In production it will be replaced with proper offchain Phragmen algorithm.

Caveat: a network restart from genesis is needed to enable `Forcing::ForceNone` variant. Alternatively, this could be done via a storage migration without network restart but I don't see a point at current phase.


ukint-vs (approved):

NikVolf:
fix ci bugs @ekovalev  ?

PR #1492, opened 2022-09-14
make big signal alt stack to be able run lazy pages in debug mode

grishasobol:
Rust on linux allocates only 0x2000 bytes for signal alternative stack (equal to libc::SIGSTKSZ). This is not enough for debug build, because in debug code use too much of stack. So, we create our own alternative signal stack if existed stack is not big enough for lazy pages signal handler (suppose that signal stack size for macos M1 0x20000 is big enough).

### DIFF
--- a/lazy-pages/src/lib.rs
+++ b/lazy-pages/src/lib.rs
@@ -252,7 +252,7 @@ pub enum InitError {
     CanNotSetUpSignalHandler(String),
 }
 
-/// Initialize lazy pages:
+/// Initialize lazy-pages once for process:
 /// 1) checks whether lazy pages is supported in current environment
 /// 2) set signals handler
 ///
--- a/lazy-pages/src/lib.rs
+++ b/lazy-pages/src/lib.rs
@@ -282,6 +282,8 @@ unsafe fn init_internal<H: UserSignalHandler>(version: LazyPagesVersion) -> Resu
         return Err(NativePageSizeIsNotSuitable(ps));
     }
 
+    sys::init_for_thread();
+
     if let Err(err) = sys::setup_signal_handler::<H>() {
         return Err(CanNotSetUpSignalHandler(err.to_string()));
     }
--- a/lazy-pages/src/lib.rs
+++ b/lazy-pages/src/lib.rs
@@ -16,7 +16,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
-//! Lazy pages support.
+//! Lazy-pages support.
 //! In runtime data for contract wasm memory pages can be loaded in lazy manner.
 //! All pages, which is supposed to be lazy, must be mprotected before contract execution.
 //! During execution data from storage is loaded for all pages, which has been accessed.
@@ -108,12 +108,12 @@ pub(crate) struct LazyPagesExecutionContext {
     pub wasm_mem_size: Option<u32>,
     /// Current program prefix in storage
     pub program_storage_prefix: Option<Vec<u8>>,
-    /// Wasm addresses of lazy pages, that have been read or write accessed at least once.
+    /// Wasm addresses of lazy-pages, that have been read or write accessed at least once.
     /// Lazy page here is page, which has `size = max(native_page_size, gear_page_size)`.
     pub accessed_pages_addrs: BTreeSet<WasmAddr>,
     /// End of stack wasm address. Default is `0`, which means,
     /// that wasm data has no stack region. It's not necessary to specify
-    /// this value, `lazy pages` uses it to identify memory, for which we
+    /// this value, `lazy-pages` uses it to identify memory, for which we
     /// can skip processing and this memory won't be protected. So, pages
     /// which lies before this value will never get into `released_pages`,
     /// which means that they will never be updated in storage.
@@ -126,10 +126,10 @@ thread_local! {
     // NOTE: here we suppose, that each contract is executed in separate thread.
     // Or may be in one thread but consequentially.
 
-    /// Lazy pages impl version. Different runtimes may require different impl of lazy pages functionality.
+    /// Lazy-pages impl version. Different runtimes may require different impl of lazy-pages functionality.
     /// NOTE: be dangerous when use it and pay attention process and thread initialization.
     static LAZY_PAGES_VERSION: RefCell<LazyPagesVersion> = RefCell::new(LazyPagesVersion::Version1);
-    /// Lazy pages context for current execution.
+    /// Lazy-pages context for current execution.
     static LAZY_PAGES_CONTEXT: RefCell<LazyPagesExecutionContext> = RefCell::new(Default::default());
 }
 
@@ -183,7 +183,7 @@ pub fn initialize_for_program(
 
         ctx.program_storage_prefix = Some(program_prefix);
 
-        log::trace!("Initialize lazy pages for current program: {:?}", ctx);
+        log::trace!("Initialize lazy-pages for current program: {:?}", ctx);
 
         Ok(())
     })
@@ -239,21 +239,21 @@ pub fn set_wasm_mem_size(size: WasmPageNumber) -> Result<(), WasmMemSizeError> {
     Ok(())
 }
 
-/// Returns vec of lazy pages which has been accessed
+/// Returns vec of lazy-pages which has been accessed
 pub fn get_released_pages() -> Vec<PageNumber> {
     LAZY_PAGES_CONTEXT.with(|ctx| ctx.borrow().released_pages.iter().copied().collect())
 }
 
 #[derive(Debug, Clone, derive_more::Display)]
 pub enum InitError {
-    #[display(fmt = "Native page size {:#x} is not suitable for lazy pages", _0)]
+    #[display(fmt = "Native page size {:#x} is not suitable for lazy-pages", _0)]
     NativePageSizeIsNotSuitable(usize),
     #[display(fmt = "Can not set signal handler: {}", _0)]
     CanNotSetUpSignalHandler(String),
 }
 
 /// Initialize lazy-pages once for process:
-/// 1) checks whether lazy pages is supported in current environment
+/// 1) checks whether lazy-pages is supported in current environment
 /// 2) set signals handler
 ///
 /// # Safety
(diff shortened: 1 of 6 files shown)


lazy-pages/src/sys/unix.rs
```diff
@@ -109,7 +109,102 @@ where
     }
 }
 
-pub unsafe fn setup_signal_handler<H>() -> io::Result<()>
+use errno::Errno;
+
+#[derive(Debug, Clone, Copy, derive_more::Display)]
+enum ThreadInitError {
+    #[display(fmt = "Cannot get information about old signal stack: {}", _0)]
+    OldStack(Errno),
```
ark0f, line 117:
```suggestion
    OldSigStack(Errno),
```

lazy-pages/src/sys/unix.rs
```diff
@@ -109,7 +109,102 @@ where
...
+    #[display(fmt = "Cannot mmap space for signal stack: {}", _0)]
+    Mmap(Errno),
+    #[display(fmt = "Cannot set new signal stack: {}", _0)]
+    SigAltStack(Errno),
+}
+
+fn init_for_thread_internal() -> Result<(), ThreadInitError> {
+    use core::{mem, ptr};
```
ark0f, line 125: Use `std` instead

lazy-pages/src/sys/unix.rs
```diff
@@ -109,7 +109,102 @@ where
...
+
+    // Should be enough for lazy-pages signal handler.
+    // Equal to libc::SIGSTKSZ on macos M1.
```
ark0f, line 128:
```suggestion
    // Equals to libc::SIGSTKSZ on macOS M1.
```

lazy-pages/src/sys/unix.rs
```diff
@@ -109,7 +109,102 @@ where
...
+                        );
+                    }
+                }
+            }
+        }
+    }
+
+    unsafe fn init_sigstack() -> Result<StackInfo, ThreadInitError> {
+        // Check whether old signal stack exist and suitable for lazy-pages signal handler.
+        let mut old_stack = mem::zeroed();
```
ark0f, line 154:
```suggestion
        let mut old_stack = MaybeUninit::uninit();
```
Recommended to be used instead

lazy-pages/src/sys/unix.rs
```diff
@@ -109,7 +109,102 @@ where
...
+        let res = libc::sigaltstack(ptr::null(), &mut old_stack);
+        if res != 0 {
+            return Err(ThreadInitError::OldStack(errno::errno()));
+        }
+        if old_stack.ss_flags & libc::SS_DISABLE == 0 && old_stack.ss_size >= SIGNAL_STACK_SIZE {
+            return Ok(StackInfo::UseOldStack);
+        }
+
+        // Alloc memory for new signal stack.
+        let ptr = libc::mmap(
```
ark0f, line 164: Add comment why you use mmap, if you motivation is taken here: https://stackoverflow.com/a/31820184/7844786

ark0f, line 164: You already have `region` library in deps which do literally the same

lazy-pages/src/sys/unix.rs
ark0f, line 125: I think all statements of function can be moved to the file itself

lazy-pages/src/lib.rs
```diff
@@ -236,70 +239,66 @@ pub fn set_wasm_mem_size(size: WasmPageNumber) -> Result<(), WasmMemSizeError> {
...
+                || (ps > PageNumber::size() && ps % PageNumber::size() != 0)
+                || (ps < PageNumber::size() && PageNumber::size() % ps != 0)
+            {
+                return Err(NativePageSizeIsNotSuitable(ps));
+            }
 
-    if let Err(err) = sys::setup_signal_handler::<H>() {
-        return Err(CanNotSetUpSignalHandler(err.to_string()));
-    }
+            if let Err(err) = sys::setup_signal_handler::<H>() {
```
ark0f, line 275:
```suggestion
            sys::setup_signal_handler::<H>().map_err(CanNotSetUpSignalHandler)?;
```

lazy-pages/src/sys/unix.rs
```diff
@@ -109,7 +109,102 @@ where
...
+
+    thread_local! {
+        static TLS: Result<StackInfo, ThreadInitError> = unsafe { init_sigstack() };
+    }
+
+    TLS.with(|tls| tls.as_ref().map(|_| ()).map_err(|err| *err))
+}
+
+pub(crate) unsafe fn init_for_thread() -> Result<(), String> {
+    init_for_thread_internal().map_err(|err| err.to_string())
```
ark0f, line 204: We have no need to convert error into string

ark0f (changes requested):

lazy-pages/src/sys.rs
```diff
@@ -30,19 +30,19 @@ use gear_core::memory::{PageNumber, PAGE_STORAGE_GRANULARITY};
...
-// 2) must support old runtimes: need to make lazy pages version with old constants values.
+// 2) must support old runtimes: need to make lazy-pages version with old constants values.
 static_assertions::const_assert_eq!(PageNumber::size(), 0x1000);
 static_assertions::const_assert_eq!(PAGE_STORAGE_GRANULARITY, 0x4000);
 
 cfg_if! {
     if #[cfg(windows)] {
         mod windows;
-        pub use windows::*;
+        pub(crate) use windows::*;
```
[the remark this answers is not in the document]
grishasobol: No it's not redundant - it's pub for crate

lazy-pages/src/sys/unix.rs
```diff
@@ -109,7 +109,102 @@ where
...
+    }
+
+    thread_local! {
+        static TLS: Result<StackInfo, ThreadInitError> = unsafe { init_sigstack() };
+    }
+
+    TLS.with(|tls| tls.as_ref().map(|_| ()).map_err(|err| *err))
+}
+
+pub(crate) unsafe fn init_for_thread() -> Result<(), String> {
```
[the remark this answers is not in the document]
grishasobol: NO

lazy-pages/src/sys/unix.rs
grishasobol: Why?

lazy-pages/src/sys/unix.rs
grishasobol: I don't wanna combinate using libc and region in this function

lazy-pages/src/sys/unix.rs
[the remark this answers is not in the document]
grishasobol: What the reason?

lazy-pages/src/sys/unix.rs
grishasobol: Sorry, can you please explain more?

lazy-pages/src/sys/unix.rs
grishasobol: Ok, but I cannot use Errno outside. Of course I can use io::Error, but what the difference between using io::Error and String in our case ?

lazy-pages/src/sys/unix.rs
grishasobol: I use here mmap because it's more clear what it does in this place. I don't wanna use malloc or region::alloc because they can have side effects which I must investigate before using. Using mmap here is the most clear variant

lazy-pages/src/sys/unix.rs
grishasobol: Will done

lazy-pages/src/sys/unix.rs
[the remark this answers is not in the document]
grishasobol: Cannot use it because of this: ``` the trait `Clone` is not implemented for `std::io::Error` ```

lazy-pages/src/sys/unix.rs
grishasobol: io::Error cannot represent full information about error. We have not only Errno as error, we have Error enum:
```rust
enum ThreadInitError {
    #[display(fmt = "Cannot get information about old signal stack: {}", _0)]
    OldSigStack(Errno),
    #[display(fmt = "Cannot mmap space for signal stack: {}", _0)]
    Mmap(Errno),
    #[display(fmt = "Cannot set new signal stack: {}", _0)]
    SigAltStack(Errno),
}
```
How can I transform it to io:Error without information lost?

lazy-pages/src/sys/unix.rs
grishasobol: Done

lazy-pages/src/sys/unix.rs
grishasobol: Done

lazy-pages/src/sys/unix.rs
grishasobol: Done

lazy-pages/src/sys/unix.rs
grishasobol: Done, check please

lazy-pages/src/sys/unix.rs
grishasobol: Make comment

lazy-pages/src/sys/unix.rs
grishasobol: Done

lazy-pages/src/lib.rs
grishasobol: Done

lazy-pages/src/sys/unix.rs
ark0f: You can return `Result<(), ThreadInitError` then

lazy-pages/src/sys/unix.rs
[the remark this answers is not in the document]
ark0f: Ok. Let's leave `errno` crate then

lazy-pages/src/sys/unix.rs
grishasobol: No, because it's only for unix errors

ark0f: Does this even matter in out case?

grishasobol: We can make more suitable errors when the same support  will be done for windows (if it will be done). Currently I think it's not a good way to move unix special errors into lib.rs

ark0f: We don't even need to move, just change function signature

grishasobol: No we need to make the same function for windows

ark0f: Why?

grishasobol: Make a look at `windows.rs`

ark0f: It doesn't need to be the same in case of error. The only requirement is to be printable:
```
if let Err(err) = unsafe { sys::init_for_thread() } {
    log::debug!("Cannot initialize lazy-pages for thread: {}", err);
    ...
}
```

grishasobol: Hmm may be,  ok will try it

grishasobol: Hmm, no. In this case we should make this error as public, which is not currently suitable. And I think it's very strange solution

ark0f (approved):

PR #1510, opened 2022-09-17
Remove restrictions on custom section

NikVolf:
While they are not desirable, there is nothing wrong if somebody wants to add wasm with custom section(s)

### DIFF
--- a/core/src/code.rs
+++ b/core/src/code.rs
@@ -85,8 +85,6 @@ pub enum CodeError {
     Encode,
     /// We restrict start sections in smart contracts.
     StartSectionExists,
-    /// We restrict custom sections in smart contracts.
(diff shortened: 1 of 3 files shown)


ukint-vs (approved):

core/src/code.rs
```diff
@@ -124,11 +124,6 @@ impl Code {
             return Err(CodeError::StartSectionExists);
         }
 
-        if module.custom_sections().count() != 0 {
-            log::debug!("Found custom sections in contract code, which is not allowed");
-            return Err(CodeError::CustomSectionsExist);
```
gshep, line 129: The error variant should also be removed since it doesn't get used anywhere else?

gshep (approved):

PR #1520, opened 2022-09-19
Fix CI

ark0f:


scripts/src/init.sh
```diff
@@ -56,7 +56,7 @@ cargo_init() {
   if [ -z $CI ]; then
     cargo install cargo-hack
     cargo install cargo-nextest
-  elif [ "$OSTYPE" == "linux-gnu" ]; then
+  elif [ "$OSTYPE" = "linux-gnu" ]; then
```
shamilsan, line 59: Please check it twice

.github/workflows/CI.yaml
```diff
@@ -335,7 +335,7 @@ jobs:
           path: artifact
 
   build-win:
-    if: github.ref == 'refs/heads/master'
+    #if: github.ref == 'refs/heads/master'
```
shamilsan, line 338: Do we really want to run Win CI in PR?

shamilsan: Ah, got it, it is a check.

scripts/src/init.sh
ark0f: We are because after Substrate update `protoc` is required to be installed

ark0f: I'm trying and everything is successful even in PR CI check, but not in master. There is an another try

ark0f:
Successful runs:
| OS | URL |
| --- | --- |
| macOS | https://github.com/gear-tech/gear/actions/runs/3084673468/jobs/4987100014 |
| Windows | https://github.com/gear-tech/gear/actions/runs/3084673468/jobs/4987100285 |

shamilsan (approved):

Issue #1528, opened 2022-09-21
Review all syscalls and prepare for release


shamilsan:
### File Location(s)

gstd/src/msg/async.rs
etc.
etc.

### Proposal

- [x] Change `gr_send`/`gr_reply` (et al.) syscalls to accept one additional parameter (delay).
- [x] Remove deprecated `gstd::send_and_wait_for_reply` and `gstd::send_bytes_and_wait_for_reply`.
- [x] Rename `gr_wait_no_more` to `gr_wait_up_to`.
- [x] #1208
- [ ] Part of #1065

PR #1530, opened 2022-09-21
Remove deprecated functions from `gstd` and rename syscall

shamilsan:
Part of #1528.

- Removed deprecated `gstd::send_and_wait_for_reply` and `gstd::send_bytes_and_wait_for_reply`.
- Renamed `wait_no_more` (I know it is an allusion to "Faith No More" but ...) to `wait_up_to`.

@gear-tech/dev

### DIFF
--- a/core/src/costs.rs
+++ b/core/src/costs.rs
@@ -112,8 +112,8 @@ pub struct HostFnWeights {
     /// Weight of calling `gr_wait_for`.
     pub gr_wait_for: u64,
 
-    /// Weight of calling `gr_wait_no_more`.
-    pub gr_wait_no_more: u64,
+    /// Weight of calling `gr_wait_up_to`.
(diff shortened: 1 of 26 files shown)


core/src/costs.rs
```diff
@@ -207,7 +207,7 @@ pub enum RuntimeCosts {
     Wait,
     /// Weight of calling `gr_wait_for`.
     WaitFor,
-    /// Weight of calling `gr_wait_no_more`.
+    /// Weight of calling `gr_wait_up_to`.
     WaitNoMore,
```
NikVolf, line 211: should be renamed as well?

shamilsan: Yep, missed it. Done.

gshep (approved):
👍

PR #1532, opened 2022-09-21
Process queue as unsigned extrinsic at the end of a block

### LINKED ISSUE #1531: Recover from panic in message queue processing
- [ ] Implement queue processing as an extrinsic/inherent so that a panic would not stall block production;
- [ ] Implement tests to simulate such situation and prove the recovery works.

ekovalev:
Partially resolves #1531.

Message queue processing can (in theory) panic owning to a smart contract code calling the environment (via syscall) in case such a call throws exception. If queue processing is done inside the `on_idle` hook, a panic in it will halt the network (no blocks will be produced from that moment on).
To mitigate this risk, we run queue processing inside a special extrinsic in `pallet-gear` - somewhat similar to an `inherent`, but we forcefully place it at the end of a block. Much like an `inherent`, this extrinsic is only included in a block by the block producer if it has been verified for being valid.

To achieve this effect of processing the queue within the same block, we have to provide a custom implementation of the `base-authorship` block authoring logic.

A simple guard that disables queue processing in all blocks following the one in which a panic occurred while processing messages has been put in place: there have been introduced two `StorageValue`'s in `pallet-gear` that control the "forcing strategy" for queue processing (whether it should be forced, not forced, forced once or never forced) and the status of the queue processing itself. In the beginning of every block the status is set to `Scheduled`; when the extrinsic that wraps queue processing completes, the status is set to `Completed`. In case of a panic all state changes are rolled back thereby keeping the queue processing status as `Scheduled`. Such case is interpreted as a panic signal and queue processing is set to `SkippedOrFailed`. After that the queue processing will not be triggered until the forcing strategy is set to `ForceOnce` or `ForceAlways` via runtime upgrade. In the former case the queue will run at least once, and if no more panic occurs, the flow will go back to normal.

**Caveat.** This guard relies on sequential state changes in a single thread. For parallel processing more sophisticated mechanism will be needed.

TODO:
- more tests are needed, specifically test of our customized block authoring machinery to ensure the network throughput will not suffer from this change.


pallets/gear-debug/src/mock.rs
```diff
@@ -213,9 +213,9 @@ pub fn run_to_block(n: u64, remaining_weight: Option<u64>) {
         Gear::on_initialize(System::block_number());
         let remaining_weight =
             remaining_weight.unwrap_or(pallet_gear::BlockGasLimitOf::<Test>::get());
-        Gear::on_idle(
-            System::block_number(),
-            Weight::from_ref_time(remaining_weight),
-        );
+        pallet_gear::GasAllowanceOf::<Test>::put(remaining_weight);
```
breathx, line 216: where will we update remaining weight in real node?

pallets/gear/src/lib.rs
```diff
@@ -1903,6 +1961,72 @@ pub mod pallet {
 
             Ok(())
         }
+
+        /// Process message queue
+        #[pallet::weight((
+            Weight::from_ref_time(BlockGasLimitOf::<T>::get()),
+            DispatchClass::Mandatory
+        ))]
+        pub fn run_process_queue(origin: OriginFor<T>) -> DispatchResultWithPostInfo {
```
breathx, line 1970: I'd rename this into probably just `run`?
Due to the fact that on_idle have not only been processing queue, but scheduled tests also

runtime-interface/src/lib.rs
```diff
@@ -133,7 +133,7 @@ fn mprotect_mem_interval_except_pages(
 /// Note: name is expanded as gear_ri
 #[runtime_interface]
 pub trait GearRI {
-    /// Init lazy pages for `on_idle`.
+    /// Init lazy pages.
```
breathx, line 136:
```suggestion
    /// Init lazy-pages.
```

runtime/common/src/lib.rs
```diff
@@ -21,24 +21,51 @@
...
+/// We assume that ~10% of the block weight is consumed by `on_initialize` handlers.
+/// This is used to limit the maximal weight of a single extrinsic.
+const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(10);
+/// We allow `Normal` extrinsics to fill up the block up to 25%.
+/// Such extrinsics only account for user messages. The rest of the block time can be used
+/// for message queue processing and/or `Operatoinal` extrinsics.
+// TODO: consider making the normal extrinsics share adjustable
 const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(25);
+/// We allocate one third of the average block time for compute.
+const MAXIMUM_BLOCK_WEIGHT: Weight = WEIGHT_PER_SECOND.saturating_div(3);
```
breathx, line 45: This won't be correct for vara runtime btw, so I'd suggest to remove this from common

pallets/gear/src/lib.rs
```diff
@@ -1903,6 +1961,72 @@ pub mod pallet {
 
             Ok(())
         }
+
+        /// Process message queue
+        #[pallet::weight((
+            Weight::from_ref_time(BlockGasLimitOf::<T>::get()),
```
breathx, line 1967: can't we ran into some misbehaviour if our calculations of block gas limit of exceed current real amount due to our rounding on division?

And why not 0 at all?
Nobody will pay fee for it, so we can set 0 at the beginning and provide real amount in DispatchResult always

pallets/gear/src/lib.rs
```diff
@@ -1903,6 +1961,72 @@ pub mod pallet {
 
             Ok(())
         }
+
```
breathx, line 1964: And as I mentioned before, you never paste here actual gas allowance (it also may be greater than block gas limit constant): should we calculate it in each extrinsic?

pallets/gear-debug/src/mock.rs
ekovalev: It's set to the block gas limit in the `pallet_gas`'s `on_initialize()` hook - same as before

pallets/gear/src/lib.rs
ekovalev: Makes sense

pallets/gear/src/lib.rs
ekovalev: Ok, I guess we can set the weigh of this extrinsic to 0. After all, it has a `DispatchClass::Mandatory` dispatch class and shouldn't interfere with the normal extrinsics whose total weight is determined by the `NORMAL_DISPATCH_RATIO`. The actual weight also doesn't really matter because this extrinsic is guaranteed to run last in a block, and always be included in a block (regardless of it's actual weight, as a mandatory dispatch).

pallets/gear/src/lib.rs
ekovalev: Finally realized what you meant (having fixed failing tests). Yes, certainly - doing it this way now.

ukint-vs (approved):

node/authorship/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+
+// Modified implementation of the basic block-authorship logic from
+// https://github.com/paritytech/substrate/tree/master/client/basic-authorship.
+// The block proposer explicitly pushes the `pallet_gear::run`
+// extrinsic at the end of each block.
+
+mod authorship;
+
+// #[cfg(test)]
+// mod tests;
```
breathx, line 27:
```suggestion
#[cfg(test)]
mod tests;
```

Also waiting for test you mentioned in dm

breathx (approved):
Awesome!

breathx (approved):

PR #1542, opened 2022-09-22
fix panic on big allocations in wasm runtime

grishasobol:
1) Add restriction for max allocation in gear runtime, currently check it only in gear_backend_sandbox.
2) Add tracing of sys-calls. To avoid performance decrease  disable them using log feature: max_level_debug


ukint-vs (approved):

core-backend/sandbox/src/runtime.rs
```diff
@@ -47,6 +47,9 @@ where
     }
 
     fn read_memory(&self, ptr: u32, len: u32) -> Result<Vec<u8>, MemoryError> {
+        if len as usize > RUNTIME_MAX_ALLOC_SIZE {
+            return Err(MemoryError::OutOfBounds);
+        }
```
gshep, line 52: Memory doesn't check this case itself?

gshep (approved):

PR #1544, opened 2022-09-22
move fix for sandbox backend to vara

grishasobol:


core-backend/sandbox/Cargo.toml
```diff
@@ -12,7 +12,8 @@ gear-backend-common = { path = "../common" }
 
 parity-wasm = { version = "0.45.0", default-features = false }
 sp-sandbox = { version = "0.10.0-dev", git = "https://github.com/gear-tech/substrate.git", branch = "gear-stable", default-features = false, features = ["host-sandbox"] }
-log = { version = "0.4.17", default-features = false }
+# Use max_level_debug feature to remove tracing in sys-calls by default.
+log = { version = "0.4.17", default-features = false, features = ["max_level_debug"] }
```
breathx, line 16: Despite the fact, that this won't affect log output, it still contains overhead of single comparison:
https://github.com/rust-lang/log/blob/d0a45aa2929365c840b82360370b223d0c22364e/src/macros.rs#L35

I'd suggest to add feature to sandbox like "deep-log"

And use tracing logging inside the module under `#[cfg(feature = "deep-log")]` macro to cut it in compile time

core-backend/sandbox/src/runtime.rs
```diff
@@ -47,6 +47,9 @@ where
     }
 
     fn read_memory(&self, ptr: u32, len: u32) -> Result<Vec<u8>, MemoryError> {
+        if len as usize > RUNTIME_MAX_ALLOC_SIZE {
```
breathx, line 50: should we provide same check for write output?
and propagate some other error instead of out of bounds to identify it in future easier

core-backend/sandbox/Cargo.toml
grishasobol:
```rust
 if lvl <= $crate::STATIC_MAX_LEVEL
```
is comparison between const and const, which will be removed by any compiler

core-backend/sandbox/src/runtime.rs
grishasobol: This problem must be checked in place where buffer is allocated, not in place where it uses. So, currently I cannot find any place for which this problem can occurs where `write_output` is used.

grishasobol: Talking about different error - is a good question. May be we really should make new error. But currently this error can occurs only in case where wasm memory bound error will also occurs, so I suppose leave one error for both cases is not criminal now. I will make issue: we must investigate all runtime and check that over size buffers cannot be created anywhere and also this question will be there

PR #1550, opened 2022-09-23
Define new design for loader

SabaunT:
Define new design for loader to be ready for multiple tasks and for usage of `gclient`


utils/node-loader/src/batch_pool.rs
```diff
@@ -0,0 +1,127 @@
...
+use context::{TaskContextUpdate, TasksContext};
+use generators::{BatchGenerator, BatchGeneratorImpl};
+use report::{BatchRunReport, TaskReporter};
+
+mod context;
+pub(crate) mod generators;
+mod report;
+mod task;
+
+pub(crate) struct BatchPool<Rng: crate::LoaderRng> {
```
breathx, line 16:
```suggestion
pub(crate) struct BatchPool<Rng: LoaderRng> {
```

breathx (approved):

PR #1560, opened 2022-09-26
feat(program): integrate gear-program into the main binary and clean up the tests

### LINKED ISSUE #1553: Clean crate gear-program after the merge
### Problem to Solve

We can remove a lot of implementation of [gear-program](https://github.com/gear-tech/gear-program) after the merge

### API generation

- remove `program/scripts/generate-api.sh`
- introduce `program/build.rs`
  - run if 
    - `pallet-gear` changed
    -  `gear-rpc` changed
  - get [metadata](https://github.com/paritytech/substrate/blob/397da2753081b3cbe4cbb0893f17b4188f0d5997/client/rpc/src/state/state_full.rs#L299) from runtime
  - do the [codegen](https://github.com/paritytech/subxt/blob/75e383dfcf9e11db20436ccebdb14e211135a8f1/cli/src/main.rs#L296)
  - call `rustfmt` as library to do the code format job

### Tests

- no need the spec-version tests anymore
- use `examples/binaries` to do the tests directly
  - remove `program/res/*.wasm`

### Notes

_No response_

clearloop:
Resolves #1553.

- [x] Fix the tests
  - [x] remove res/*.wasm
- [x] Genreate api with build script
- [x] Make program part of gear-node bin with features
  - [x] use `clap` instead of `structopt`
- [x] only build gear-program binary with --features bin

@gear-tech/dev

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1056,38 +1056,11 @@ dependencies = [
 
 [[package]]
 name = "cranelift-bforest"
-version = "0.87.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9f91425bea5a5ac6d76b788477064944a7e21f0e240fd93f6f368a774a3efdd1"
-dependencies = [
- "cranelift-entity 0.87.1",
-]
-
-[[package]]
-name = "cranelift-bforest"
 version = "0.88.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "44409ccf2d0f663920cab563d2b79fcd6b2e9a2bcc6e929fef76c8f82ad6c17a"
 dependencies = [
- "cranelift-entity 0.88.1",
-]
-
-[[package]]
-name = "cranelift-codegen"
-version = "0.87.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8b83b4bbf7bc96db77b7b5b5e41fafc4001536e9f0cbfd702ed7d4d8f848dc06"
-dependencies = [
- "cranelift-bforest 0.87.1",
- "cranelift-codegen-meta 0.87.1",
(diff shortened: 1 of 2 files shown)


program/tests/cmd/claim.rs
```diff
@@ -1,23 +1,33 @@
...
-use crate::common::{self, Result, ALICE_SS58_ADDRESS as ADDRESS};
+use crate::common::{self, ALICE_SS58_ADDRESS as ADDRESS};
 use gear_program::api::Api;
 
 const REWARD_PER_BLOCK: u128 = 3000;
 
 #[tokio::test]
-async fn test_command_claim_works() -> Result<()> {
-    let node = common::create_messager().await?;
+async fn test_command_claim_works() {
```
breathx, line 8: Why not to keep `?` syntax?

program/tests/cmd/meta.rs
```diff
@@ -37,8 +37,8 @@ Metadata {
 
 #[tokio::test]
 async fn test_display_metadata_works() {
```
breathx, line 39: Same about return type for all of the tests

program/tests/gear.rs
```diff
@@ -35,3 +15,9 @@ async fn api_timeout() {
         ))
     ));
 }
+
+#[test]
+fn paths() {
+    assert!(PathBuf::from(env::bin("gear")).exists());
+    assert!(PathBuf::from(env::bin("gear-node")).exists());
```
breathx, line 22: Seems that we are planning to rename gear-node to gear, so better to rename current gear to gprogram, for example

breathx (approved):

breathx:
We also had an idea to don't store metadata in code, producing it in some target dir, including into gear-program by path

PR #1561, opened 2022-09-26
Introduce delayed messaging and waking

### LINKED ISSUE #1293: Add ability to send scheduled messages
### Problem to Solve

Sometimes we need to send the message after some delay instead of doing it immediately. It might be useful for time scheduled routines, e.g. in DAO, auctions, lotteries.

### Possible Solution

Introduce new functions in `gstd::msg` (**all names, parameters, and return types are to be discussed!**):

```rust
/// `msg` module

/// Add the message to the shedule.
fn schedule(dest, payload, delay, value) -> Result<MessageId, ...>;
// fn schedule_with_gas(...)
// async fn schedule_for_reply(...)

/// Deny early sending.
fn deny_triggering(msg_id) -> bool;

/// Cancel sending scheduled message.
fn cancel_scheduled(msg_id) -> Result<(), ...>;

/// Send scheduled message immediately.
fn trigger(msg_id) -> Result<(), ...>;
// fn trigger_for_reply(...)
```

Also need to add correspondent functionality to Gear core.

### Notes

Using example (DAO):

```rust
let msg_id = msg::schedule(
    exec::program_id(),
    DaoAction::ProcessProposal(self.proposal_id)),
    exec::block_timestamp() + self.voting_period,
    0,
)
.expect("Error while deferring message");
msg::deny_triggering(msg_id);

let reply: DaoEvent = 
    exec::trigger_as(msg_id).expect("Error while shoving scheduled message");

exec::cancel_scheduled(msg_id).expect("Error while cancelling scheduled message");

let reply: DaoEvent = msg::schedule_for_reply_as(
    exec::program_id(),
    DaoAction::ProcessProposal(self.proposal_id),
    exec::block_timestamp() + self.voting_period,
    0,
)
.expect("Error while scheduling message")
.await
.expect("Error while processing result");
```

breathx:
Resolves #1293 and part of #1528.

- New field, defining delay before sending and waking message added.

This PR means the initial step of producing such behavior.
It needs more documentation, sending cancellation and soft macro for the future.

**Release notes**: _now contracts can send scheduled messages and wakes, so any time-dependent logic becomes much easier and variable_.


common/src/gas_provider/mod.rs
```diff
@@ -229,6 +229,9 @@ pub trait Tree {
         amount: Self::Balance,
     ) -> Result<(), Self::Error>;
 
+    /// Return bool, defining does node exists.
```
shamilsan, line 232:
```suggestion
    /// Return bool, defining does node exist.
```

pallets/gear/src/manager/journal.rs
```diff
@@ -272,21 +286,37 @@ where
...
+                    .unwrap_or_else(|e| unreachable!("Message queue corrupted! {:?}", e));
+
+                return;
+            }
+        } else if WaitlistOf::<T>::contains(&program_id, &awakening_id) {
+            let expected_bn =
+                SystemPallet::<T>::block_number().saturating_add(delay.unique_saturated_into());
+            let task = ScheduledTask::WakeMessage(program_id, awakening_id);
+
+            if !TaskPoolOf::<T>::contains(&expected_bn, &task) {
```
SabaunT, line 307: please state in the comment that the check is needed for the cross-block protection against double wake

pallets/gear/src/internal.rs
```diff
@@ -507,6 +507,92 @@ where
...
+                    let gas_limit = GasHandlerOf::<T>::get_limit(origin_msg)
+                        .unwrap_or_else(|e| unreachable!("GasTree corrupted! {:?}", e));
+
+                    // If available gas is greater then threshold,
+                    // than threshold can be used.
+                    (gas_limit >= threshold).then_some(threshold)
+                })
+                .unwrap_or_default();
+
+            if !dispatch.is_error_reply() && gas_limit >= threshold {
```
SabaunT, line 548: what if `gas_limit` is zero (in case `origin_msg` has less gas than `threshold`)? Seems no cut happens, but task for the message wake is added. Is it correct?

SabaunT: Also, is error reply supposed to be added to the task pool?

pallets/gear/src/internal.rs
```diff
@@ -602,6 +688,70 @@ where
...
+        };
+
+        // Taking data for funds manipulations.
+        let from = <T::AccountId as Origin>::from_origin(message.source().into_origin());
+        let to = <T::AccountId as Origin>::from_origin(message.destination().into_origin());
+        let value = message.value().unique_saturated_into();
+
+        // If gas limit can cover threshold, message will be added to mailbox,
+        // task created and funds reserved.
+        let expiration = if GasHandlerOf::<T>::exists(message.id()) {
```
SabaunT, line 714: why don't just call `get_limit`?

pallets/gear/src/internal.rs
```diff
@@ -602,6 +688,70 @@ where
...
+            // Querying gas limit. Fails in cases of `GasTree` invalidations.
+            let gas_limit = GasHandlerOf::<T>::get_limit(message.id())
```
SabaunT, line 716: It will panic in the case I described upper with gas limit being less than mailbox threshold. That's because no detached node will be created, so the query will panic

pallets/gear/src/internal.rs
```diff
@@ -602,6 +688,70 @@ where
...
 
+    /// Sends user message, once delay reached.
+    pub(crate) fn send_user_message_after_delay(message: StoredMessage) {
+        // Converting payload into string.
+        //
+        // Note: for users, trap replies always contain
+        // string explanation of the error.
+        let message = match message.exit_code() {
+            Some(0) | None => message,
+            _ => message
```
SabaunT, line 699: are error replies supposed to be sent with delay logic?

pallets/gear/src/internal.rs
SabaunT: Does message delay logic require paying for being delayed, i.e. being in the `TaskPool`?

SabaunT: discussed

pallets/gear/src/internal.rs
SabaunT: Discussed. State logic in comment.

pallets/gear/src/internal.rs
SabaunT: discussed.

SabaunT (approved):

PR #1575, opened 2022-09-27
Fix weights deserialization

ark0f:


utils/regression-analysis/src/main.rs
```diff
@@ -110,11 +110,11 @@ struct GithubActionBenchmark {
 
 #[derive(Deserialize)]
 #[serde(transparent)]
-struct WeightBenchmark(Vec<Weight>);
+struct WeightBenchmark(Vec<u64>);
```
breathx, line 113: How it will deserialise u64 if Weight type used for all benches?

breathx (approved):

utils/regression-analysis/src/main.rs
ark0f: It will use JSON file generated using our `regression-analysis-weight-template.hbs`

PR #1592, opened 2022-09-30
Add batch variants for all gear extrinsics in `gclient`

breathx:


gclient/src/api/calls.rs
```diff
@@ -109,27 +146,78 @@ impl GearApi {
...
+        let mut values = BTreeMap::new();
+
+        for message_id in message_ids {
+            values.insert(
+                message_id,
+                self.get_from_mailbox(self.0.account_id().clone(), message_id)
+                    .await?
+                    .map(|(message, _interval)| message.value()),
+            );
+        }
```
shamilsan, line 194: Consider avoiding excessive `clone` and `collect`.
```suggestion
        let mut message_ids = args.into_iter();

        let mut values = BTreeMap::new();

        for message_id in message_ids.by_ref() {
            values.insert(
                message_id,
                self.get_from_mailbox(self.0.account_id().clone(), message_id)
                    .await?
                    .map(|(message, _interval)| message.value()),
            );
        }

        let calls: Vec<_> = message_ids
            .map(|message_id| {
                RuntimeCall::Gear(GearCall::claim_value {
                    message_id: message_id.into(),
                })
            })
            .collect();
```

gclient/src/api/calls.rs
```diff
@@ -222,20 +347,82 @@ impl GearApi {
...
 
         Err(Error::EventNotFound)
     }
 
+    /// `pallet_gear::send_reply` with bytes in payload.
+    pub async fn send_reply_bytes_batch(
+        &self,
+        args: impl IntoIterator<Item = (MessageId, impl AsRef<[u8]>, u64, u128)>,
+    ) -> Result<(Vec<Result<(MessageId, u128)>>, H256)> {
+        let mut reply_to_ids = vec![];
```
shamilsan, line 366: I guess you can avoid this additional allocation too by using the `args` iterator.

gclient/src/api/calls.rs
```diff
@@ -109,27 +146,78 @@ impl GearApi {
...
+                })
+            })
+            .collect();
+
+        let mut values = BTreeMap::new();
+
+        for message_id in message_ids {
+            values.insert(
+                message_id,
+                self.get_from_mailbox(self.0.account_id().clone(), message_id)
```
shamilsan, line 190: Unrelated question but... Why does `get_from_mailbox` consume `AccountId32` instead of borrowing it?

gclient/src/api/calls.rs
```diff
@@ -109,27 +146,78 @@ impl GearApi {
...
+        let ex = self.0.tx().utility().force_batch(calls)?;
+        let tx = self.0.process(ex).await?;
+
+        let mut res = Vec::with_capacity(amount);
+
+        for event in tx.wait_for_success().await?.iter() {
+            match event?.event {
+                Event::Gear(GearEvent::UserMessageRead { id, .. }) => res.push(Ok(values
+                    .remove(&id.into())
+                    .flatten()
```
shamilsan, line 207: We can use `[]` to avoid `flatten` if we are sure the element exists. But it doesn't remove an entry. 
```suggestion
                Event::Gear(GearEvent::UserMessageRead { id, .. }) => res.push(Ok(values[&id.into()]
```

SabaunT (approved):
All the necessary comments are given by @shamilsan.

There is some copy-paste in batch calls, but that's not important right now. We can consider cleaning it in the future.

SabaunT:
@breathx please first merge #1599 into this branch before merging this to master

gclient/src/api/calls.rs
breathx: I prefer remove any possibilities to panic

gclient/src/api/calls.rs
```diff
@@ -222,20 +343,80 @@ impl GearApi {
...
+            values.insert(
+                message_id,
+                self.get_from_mailbox(self.0.account_id(), message_id)
+                    .await?
+                    .map(|(message, _interval)| message.value()),
+            );
+        }
+
+        let calls: Vec<_> = args
+            .into_iter()
```
shamilsan, line 375: `args` is already an iterator.
```suggestion
```

PR #1596, opened 2022-09-30
beautify runtime logs

grishasobol:
before:
```sh
 runtime::gear: ⚙️Initialization of block #1
 runtime::gear: ⚙️Queue and tasks processing of block #1 with weight='<wasm:stripped>'
 runtime::gear: ⚙️Weight '<wasm:stripped>' burned in block #1
 runtime::gear: ⚙️Finalization of block #1
```

after:
```sh
runtime::gear: ⚙️ Initialization of block #1
runtime::gear: ⚙️ Queue and tasks processing of block #1 with Weight(ref_time: 327333086333)
runtime::gear: ⚙️ Weight(ref_time: 25000000) burned in block #1
runtime::gear: ⚙️ Finalization of block #1
```

### DIFF
--- a/pallets/gear/src/lib.rs
+++ b/pallets/gear/src/lib.rs
@@ -422,14 +422,14 @@ pub mod pallet {
 
         /// Initialization
         fn on_initialize(bn: BlockNumberFor<T>) -> Weight {
-            log::debug!(target: "runtime::gear", "⚙️ Initialization of block #{:?}", bn);
+            log::debug!(target: "runtime::gear", "⚙️  Initialization of block #{:?}", bn);
 
             Weight::zero()
         }
 
         /// Finalization
         fn on_finalize(bn: BlockNumberFor<T>) {
-            log::debug!(target: "runtime::gear", "⚙️ Finalization of block #{:?}", bn);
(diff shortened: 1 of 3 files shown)


pallets/gear/src/lib.rs
```diff
@@ -422,14 +422,14 @@ pub mod pallet {
 
         /// Initialization
         fn on_initialize(bn: BlockNumberFor<T>) -> Weight {
-            log::debug!(target: "runtime::gear", "⚙️ Initialization of block #{:?}", bn);
+            log::debug!(target: "runtime::gear", "⚙️  Initialization of block #{:?}", bn);
```
shamilsan, line 425: Why do we need to write two whitespaces to show one whitespace? Looks weird.

grishasobol: Actually in my terminal without second space it looks like this:
image

With second space:
image

shamilsan: Agree, the gear (irony!) emoji seems to have some problem with kerning in the terminal.

The workaround in Substrate is the same: https://github.com/paritytech/substrate/blob/master/client/informant/src/display.rs#L118

shamilsan (approved):

PR #1598, opened 2022-10-01
Update metadata after introducing `LimitedVec`

SabaunT:
No update occurred after merging #1587 , which caused some problems in `gear-node-loader`

### DIFF
--- a/program/src/cmd/info.rs
+++ b/program/src/cmd/info.rs
@@ -107,7 +107,7 @@ impl fmt::Debug for Mail {
             )
             .field(
                 "payload",
-                &["0x", &hex::encode(&self.message.payload)].concat(),
+                &["0x", &hex::encode(&self.message.payload.0)].concat(),
             )
             .field("value", &self.message.value)
             .field("reply", &self.message.reply.as_ref().map(DebugReplyDetails))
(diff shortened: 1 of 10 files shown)


program/src/cmd/info.rs
```diff
@@ -107,7 +107,7 @@ impl fmt::Debug for Mail {
             )
             .field(
                 "payload",
-                &["0x", &hex::encode(&self.message.payload)].concat(),
+                &["0x", &hex::encode(&self.message.payload.0)].concat(),
```
shamilsan, line 110: Maybe it is worth implementing `Deref`, `AsRef`, and similar traits for `LimitedVec` to avoid such changes in the future.

SabaunT: As you can see in changes, most of code places use `Vec<u8>`, not the `&[u8]`. That's the only place, where trait is required.

SabaunT: We could say that changes in `program/tests/*` can be simplified with these traits, but it requires changing the `messager::REPLY_REPLY.encode()` to `&messager`, which shows that implementing suggested traits doesn't make significant changes.
We can consider doing that in the future with bringing changes to types used in `gear-program`

shamilsan (approved):

PR #1608, opened 2022-10-03
make macros to print sys-calls

grishasobol:
Using feature = max_level_debug for log dep cause that trace logs disables in all runtime. So, made a macros which enables tracing in sys-calls when self made feature is set.

### DIFF
--- a/core-backend/sandbox/src/funcs.rs
+++ b/core-backend/sandbox/src/funcs.rs
@@ -148,13 +148,24 @@ fn args_to_str(args: &[Value]) -> String {
     res
 }
 
+/// We use this macros to avoid perf decrease because of log level comparing.
+/// By default `sys-trace` feature is disabled, so this macros does nothing.
+/// To see sys-calls tracing enable this feature and rebuild node.
+macro_rules! sys_trace {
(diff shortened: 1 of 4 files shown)


SabaunT (approved):

core-backend/sandbox/src/funcs.rs
```diff
@@ -148,13 +148,24 @@ fn args_to_str(args: &[Value]) -> String {
...
+}
+
 impl<E> FuncsHandler<E>
 where
     E: Ext + IntoExtInfo + 'static,
     E::Error: AsTerminationReason + IntoExtError,
 {
     pub fn send(ctx: &mut Runtime<E>, args: &[Value]) -> SyscallOutput {
-        log::trace!(target: "syscall::gear", "send, args = {}", args_to_str(args));
+        sys_trace!(target: "syscall::gear", "send, args = {}", args_to_str(args));
```
breathx, line 168: Can't we just:
```suggestion
        #[cfg(feature = "sys-trace)]
        log::trace!(target: "syscall::gear", "send, args = {}", args_to_str(args));
```

grishasobol: No we cannot

PR #1614, opened 2022-10-03
Fix integer overflow in syscalls return value

### LINKED ISSUE #1208: Refactor syscalls output
### File Location(s)

core-backend/sandbox/funcs.rs

### Proposal

For example `return_i64()` can throw an error if `gas_available: u64` can't fit into `i64` but memory bytes are the same.
We need to pass pointer to u64 type in program space and write bytes to this address.

shamilsan:
Resolves #1208.

Changed syscalls:
- [x] `gr_block_height`
- [x] `gr_block_timestamp`
- [ ] `gr_gas_available`
- [ ] ? `gr_size`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1684,6 +1684,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "demo-overflow"
+version = "0.1.0"
+dependencies = [
+ "gear-wasm-builder",
+ "gstd",
+ "gtest",
+]
+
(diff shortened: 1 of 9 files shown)


breathx:
We've discussed more comfortable solution, so I also mentioned that I'm atm implementing it and will introduce it asap, so this approach seems useless

shamilsan:
Ok, so I'm closing it.

PR #1625, opened 2022-10-05
Use dtolnay Rust installer on Windows

ark0f:


ark0f:
Rust updates now on Windows instance successfully

PR #1628, opened 2022-10-06
Implement RPC getter of program pages

breathx:


breathx:
Will be unused since new concept of metadata

PR #1631, opened 2022-10-07
Charge for WASM instantiation

### LINKED ISSUE #1568: Charge for WASM module instantiation
Need to charge the user for creating the WASM module from the code binary.

ark0f:
Resolves #1568


core-backend/common/src/lib.rs
```diff
@@ -194,12 +194,12 @@ pub trait Environment<E: Ext + IntoExtInfo + 'static>: Sized {
...
     /// 3) Runs `pre_execution_handler` to fill the memory before running instance.
     /// 4) Instantiate external funcs for wasm module.
-    /// 5) Run instance setup starting at `entry_point` - wasm export function name.
+    fn new(ext: E, binary: &[u8], mem_size: WasmPageNumber) -> Result<Self, Self::Error>;
+
+    /// Run instance setup starting at `entry_point` - wasm export function name.
     fn execute<F, T>(
-        ext: E,
-        binary: &[u8],
+        self,
```
breathx, line 201:
```suggestion
        &self,
```

Can't we use self by ref for future purposes?

core-backend/sandbox/src/env.rs
```diff
@@ -173,19 +165,39 @@ where
...
             err: FuncError::Terminated(TerminationReason::Success),
         };
 
-        let mut instance = match Instance::new(binary, &env_builder, &mut runtime) {
+        let instance = match Instance::new(binary, &env_builder, &mut runtime) {
             Ok(inst) => inst,
             Err(e) => return Err((runtime.ext.gas_amount(), ModuleInstantiation(e)).into()),
         };
 
+        Ok(Self { instance, runtime })
```
breathx, line 181:
```suggestion
        match Instance::new(binary, &env_builder, &mut runtime) {
            Ok(instance) => Ok(Self { instance, runtime }),
            Err(e) => Err((runtime.ext.gas_amount(), ModuleInstantiation(e)).into()),
        }
```

Probably `map` and `map_err` will also be possible to use despite the compiler strange stuff

core-processor/src/processor.rs
```diff
@@ -127,25 +127,35 @@ pub fn prepare(
...
+            &block_config.allocations_config,
+            &mut gas_counter,
+            &mut gas_allowance_counter,
+            program.get_allocations(),
+            program.static_pages(),
+            dispatch.context().is_none() && matches!(dispatch.kind(), DispatchKind::Init),
+            subsequent_execution,
+        )?;
+        executor::charge_gas_for_instantiation(
+            block_config.module_instantiation,
```
breathx, line 141: how can you charge some "constant" to block value, if instantiation cost is proportional to code size?

there should be param of code len and cost per byte, than `charge_gas_for_instantiation` calculates it automatically

core-processor/src/configs.rs
```diff
@@ -112,6 +114,8 @@ pub struct BlockConfig {
     pub mailbox_threshold: u64,
     /// Cost for single block waitlist holding.
     pub waitlist_cost: u64,
+    /// Cost for single WASM module instantiation.
+    pub module_instantiation: u64,
```
breathx, line 118:
```suggestion
    pub module_instantiation_per_byte: u64,
```

pallets/gear/src/weights.rs
```diff
@@ -40,6 +40,7 @@ pub trait WeightInfo {
     fn grow_cost() -> Weight;
```
breathx, line 40: No need to rebench everything.

runtime/gear/src/weights/pallet_gear.rs
```diff
@@ -19,7 +19,7 @@
 //! Autogenerated weights for pallet_gear
```
breathx, line 19: Same here

runtime/vara/src/weights/pallet_gear.rs
```diff
@@ -19,7 +19,7 @@
 //! Autogenerated weights for pallet_gear
 //!
```
breathx, line 20: and here

breathx (changes requested):

breathx:
Also code for benchmark should contain different instructions: not the same one repeated

core-backend/common/src/lib.rs
ark0f: Such changes require a lot of refactoring so it must be separate PR

core-processor/src/configs.rs
```diff
@@ -104,6 +108,8 @@ pub struct BlockConfig {
     pub outgoing_limit: u32,
     /// Host function weights.
     pub host_fn_weights: HostFnWeights,
+    /// Static host function weights.
+    pub static_host_fn_weights: StaticHostFnWeights,
```
grishasobol, line 112: Why static? Why host fn?  Rename it everywhere. Our runtime use sandbox and don't care whether it use host or embedded executor (idiomatically). Actually I don't know what else do you wanna to put inside `static_host_fn_weights`, but can propose you: `executor_backend_weights` or `sandbox_weights`.

grishasobol (changes requested):

ukint-vs (approved):

core-processor/src/configs.rs
ukint-vs: Agree

grishasobol (approved):

core-backend/common/src/lib.rs
```diff
@@ -198,12 +198,12 @@ pub trait Environment<E: Ext + IntoExtInfo<E::Error> + 'static>: Sized {
...
         entries: BTreeSet<DispatchKind>,
```
breathx, line 206: Shouldn't we pass entries in `new` fn and store inside? We don't need to event instantiate when required entry doesn't exist.

core-backend/sandbox/src/env.rs
```diff
@@ -207,7 +217,12 @@ where
...
         };
 
-        let Runtime { err: trap, .. } = runtime;
+        let Runtime {
+            err: trap,
+            ext,
+            memory_wrap,
+            ..
+        } = runtime;
         drop(instance);
```
breathx, line 226: why do you drop instance manually?

core-processor/src/executor.rs
```diff
@@ -125,6 +125,27 @@ fn check_memory<'a>(
     Ok(())
 }
 
+pub(crate) fn charge_gas_for_instantiation(
+    weight: u64,
```
breathx, line 129: I'd rename this into something like `gas_per_byte`

core-processor/src/ext.rs
```diff
@@ -78,6 +78,29 @@ pub struct ProcessorContext {
     pub reserve_for: u32,
 }
 
+impl Default for ProcessorContext {
```
breathx, line 81: If it used just once in tests, I'd remove this impl

core/src/memory.rs
```diff
@@ -318,6 +318,7 @@ pub trait Memory {
 }
 
 /// Pages allocations context for the running program.
+#[derive(Debug, Default)]
```
breathx, line 321: I'd remove here default also

core/src/message/context.rs
```diff
@@ -120,7 +120,7 @@ pub struct ContextStore {
 }
 
 /// Context of currently processing incoming message.
-#[derive(Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Decode, Encode, TypeInfo)]
+#[derive(Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd, Decode, Encode, TypeInfo)]
```
breathx, line 123: And here default is meaningless

pallets/gear/src/lib.rs
```diff
@@ -963,7 +966,7 @@ pub mod pallet {
                     GasHandlerOf::<T>::get_origin_key(msg_id)
                         .map_err(|_| b"Internal error: unable to get origin key".to_vec())
                 };
-
+                {}
```
breathx, line 969:
```suggestion
```

pallets/gear/src/lib.rs
```diff
@@ -73,6 +69,12 @@ use sp_std::{
     prelude::*,
 };
 
+#[cfg(feature = "std")]
+type ExecutionEnvironment = gear_backend_wasmi::WasmiEnvironment<Ext>;
+
+#[cfg(not(feature = "std"))]
+type ExecutionEnvironment = gear_backend_sandbox::SandboxEnvironment<Ext>;
```
breathx, line 76: Don't you want to define Ext in the same manner dependent on `lazy-pages` feature to keep everything variable close?

pallets/gear/src/manager/mod.rs
```diff
@@ -79,6 +80,17 @@ pub enum HandleKind {
     Reply(MessageId, ExitCode),
 }
 
+impl fmt::Debug for HandleKind {
```
breathx, line 83: Why?

pallets/gear/src/schedule.rs
```diff
@@ -471,6 +473,18 @@ macro_rules! cost_byte_batched {
     };
 }
 
+impl<T: Config> Default for Schedule<T> {
```
breathx, line 476: Why?

pallets/gear/src/tests.rs
```diff
@@ -6198,7 +6208,7 @@ mod utils {
                             end $while
                         )
                         (func $init
-                            i32.const 100
+                            i32.const 1024
```
breathx, line 6211: why?

breathx (approved):

core-backend/sandbox/src/env.rs
breathx: I know, it's copy-pasted, but probably it's not actual now

PR #1635, opened 2022-10-09
Fix feature-dependent node build

breathx:
I found, that sometimes `gclient` users run into `IncompatibleMetadata` error, and during research found, that subxt autogenerated api from runtime crate and running node is different: on real node appears `pallet-gear-debug` (so runtime call hashes changes as well as an api of `gear-program`), that shouldn't occur in production builds:
- useless pallet excluded from all node builds (by hiding runtime spec testing command under the feature)
- all errors related to include options of features found
- now here is ability to choose runtime for runtime spec testing (vara / gear)
- metadata must be always valid


grishasobol (approved):

ekovalev (approved):

.github/workflows/CI.yaml
```diff
@@ -223,7 +223,7 @@ jobs:
           cargo test -p pallet-gear-payment --features=lazy-pages --release --locked
 
       - name: "Test: Gear workspace"
-        run: ./scripts/gear.sh test gear --exclude gclient --exclude gprogram --release --locked
+        run: ./scripts/gear.sh test gear --exclude gclient --exclude gprogram --features pallet-gear-debug/lazy-pages --release --locked
```
grishasobol, line 226: I've made special pallets tests run with lazy pages feature, see above. Testing also without lazy-pages is a good possibility to found some unexpected problems. Also I cannot understand why only for `pallet-gear-debug` then?

program/build.rs
```diff
@@ -18,10 +16,8 @@ const GENERATED_TITLE: &str = r#"
 
 /// Generate api
 fn codegen(raw_derives: Vec<String>) -> String {
-    let encoded = gear_runtime::Runtime::metadata().encode();
+    let metadata = gear_runtime::Runtime::metadata();
```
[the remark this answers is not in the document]
breathx: Yeah, it returns `RuntimeMetadataPrefixed` now. Probably, since last substrate update?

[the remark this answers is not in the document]
breathx: It couldn't fit with any other type due to rust type safety

.github/workflows/CI.yaml
breathx: Sure, but: cargo is really lazy thing with it's own issues

1. Building workspace tests (by default, it contains lazy pages due to appearance in node with it by default)
2. pallet-gear-debug didn't fit any crate without features, which is disabled by default in workspace
3. pallet-gear-debug built wout lazy-pages and it thinks that pallet-gear goes same, but it built with lazy-pages => test fails, so it's easier to enable just lazy-pages feature

test passes if we call directly pallet-gear-debug with and without lazy-pages via `cargo test -p pallet-gear-debug`. It's only about workspace troubles

breathx: Solved?

PR #1638, opened 2022-10-10
Rename `gear-node` binary to `gear`

### LINKED ISSUE #1624: Rename `gear-node` binary to `gear`
- After #1560
- [ ] Rename `gear-node` binary to `gear`
- [ ] Rename `gear-node` crate to `gear-cli`
- [ ] Keep the default base dir to be `gear-node` (for backward compatibility)
- [ ] Check all scripts and configs in the main Gear repo. 
- [ ] Inspect and add correspondent issues to other repos (e.g. wiki etc) to fix it anywhere.

shamilsan:
- Resolves #1624
- Removed `gear-node` crate.
- Changed `gear-cli` crate to produce a binary.
- Replaced `gear-node` binary with `gear` and `gear-node` crate with `gear-cli`.
- Changed default `--base-path` to `gear-node` instead of executable name (`gear[.exe]`) for backward compatibility.

@gear-tech/dev


NikVolf:
tagging @sergeyfilyanin so he can update his stuff

shamilsan:
Also, @osipov-mit please pay attention to making necessary updates to `gear-js`.

gshep (approved):
Looks good! :gear:

ukint-vs (approved):

Makefile
```diff
@@ -274,4 +274,4 @@ fuzz-vara:
 
 .PHONY: kill
 kill:
-	@ pgrep -f "gear-node" | xargs kill -9
+	@ pgrep -f "release/gear" | xargs kill -9
```
breathx, line 277:
```suggestion
	@ pgrep -f "gear" | xargs kill -9
```

That's executable binary name (process name)

cli/Cargo.toml
```diff
@@ -77,5 +82,8 @@ runtime-benchmarks = [
 	"frame-benchmarking-cli/runtime-benchmarks",
 ]
 debug-mode = ["service/debug-mode"]
-try-runtime = ["service/try-runtime", "try-runtime-cli"]
-program = [ "gear-program" ]
+try-runtime = [
+	"service/try-runtime",
```
breathx, line 86:
```suggestion
	"service?/try-runtime",
```

cli/src/command.rs
```diff
@@ -137,6 +138,13 @@ pub fn run() -> sc_cli::Result<()> {
         .execution
         .get_or_insert(ExecutionStrategy::Wasm);
 
+    // Set default base directory to `gear-node`.
+    cli.run
+        .base
+        .shared_params
+        .base_path
+        .get_or_insert_with(|| BasePath::from_project("", "", "gear-node").path().into());
```
breathx, line 146: What's this path?

scripts/calc-gas-spent.sh
```diff
@@ -2,7 +2,7 @@
 #
 # Prerequisites:
 #
-#     RUST_LOG=gwasm=debug,pallet_gear=debug cargo run -p gear-node -- --dev --tmp -l0
+#     RUST_LOG=gwasm=debug,pallet_gear=debug cargo r -rp gear-cli -- --dev --tmp -l0
```
breathx, line 5:
```suggestion
#     RUST_LOG=gwasm=debug,pallet_gear=debug cargo run -p gear-cli --release -- --dev --tmp -l0
```

I prefer more classic solutions

scripts/gear.sh
```diff
@@ -57,7 +57,7 @@ gear_usage() {
     docker         docker functionality
     format         format gear parts via rustfmt
     init           initializes and updates packages and toolchains
-    run            run gear-node processing
+    run            run gear processing
```
breathx, line 60:
```suggestion
    run            run gear node
```

scripts/src/test.sh
```diff
@@ -110,10 +110,10 @@ client_tests() {
 
   if [ "$2" = "--run-node" ]; then
     # Run node
-    RUST_LOG="pallet_gear=debug,runtime::gear=debug" $ROOT_DIR/target/release/gear-node \
+    RUST_LOG="pallet_gear=debug,runtime::gear=debug" $ROOT_DIR/target/release/gear \
       --dev --tmp --unsafe-ws-external --unsafe-rpc-external --rpc-methods Unsafe --rpc-cors all & sleep 3
 
-    cargo test -p gclient || pgrep -f "gear-node" | xargs kill -9 && pgrep -f "gear-node" | xargs kill -9
+    cargo test -p gclient || pgrep -f "release/gear" | xargs kill -9 && pgrep -f "release/gear" | xargs kill -9
```
breathx, line 116: same about "release/gear"

scripts/test-utils/processingMessageWithRuntimeUpgrade.js
```diff
@@ -79,7 +79,7 @@ main(pathToRuntimeCode, pathToDemoPing)
     exitCode = 1;
   })
   .finally(() => {
-    exec('pgrep -f "gear-node" | xargs kill -9', (err, stdout, stderr) => {
+    exec('pgrep -f "release/gear" | xargs kill -9', (err, stdout, stderr) => {
```
breathx, line 82: and here

breathx (approved):

breathx:
This also requires wiki and articles updates @EugenWay

breathx:
I also do not like solution where we have no node crate at all

cli/src/command.rs
shamilsan: - Changed default `--base-path` to `gear-node` instead of executable name (`gear[.exe]`) for backward compatibility.

scripts/calc-gas-spent.sh
shamilsan: Ok.

Makefile
shamilsan: As it is a regular expression, just `gear` also kills `gear.sh` which yields an error.

shamilsan:
> I also do not like solution where we have no node crate at all

Why? `gear-node` was just a wrapper around `gear-cli`. It is quite useless.

shamilsan:
> agree with this since the `node` crate has already been a part of the custom of substrate projects )))
> 
> the `cli`, `service`, etc. parts can be parts of the `node` crate but kicking out the `node` crate just feeling weird

We could keep `gear-node` crate but:
1) It is not quite intuitive when you call `gear program` from the binary produced by `gear-node` crate. `gear-cli` looks more intuitive here.
2) I don't like the idea to support a pretty empty crate that does nothing. Note that we will be obligated to publish it to crates.io, take care of its release cycle, etc. In my opinion, fewer crates are better.

shamilsan:
> yeah for me I'm okay with this solution

Ideally, it would be great to have the `gear` root crate 😄 But this crate name is [already taken](https://crates.io/crates/gear).

Makefile
shamilsan: Replaced it with `pkill -f 'gear |gear$' -9`, seems to work now.

PR #1643, opened 2022-10-11
Support vara in gprogram

### LINKED ISSUE #1562: Support vara in gear-program cli
### Problem to Solve

now gear-program cli using gear-runtime to build the apis, introduce feature vara to support vara as well

### Possible Solution

genreate apis from both gear-rutnime and vara-runtime, choose runtime on compile or making it a flag

### Notes

_No response_

clearloop:
Resolves #1562 

- [x] add features gear and vara in `gprogram`
- [x] update features of `node`
- [x] update README of `gprogram`

@gear-tech/dev

### DIFF
--- a/runtime/gear/src/lib.rs
+++ b/runtime/gear/src/lib.rs
@@ -98,7 +98,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     impl_name: create_runtime_str!("gear"),
     apis: RUNTIME_API_VERSIONS,
     authoring_version: 1,
-    spec_version: 460,
+    spec_version: 470,
     impl_version: 1,
     transaction_version: 1,
     state_version: 1,
--- a/runtime/vara/src/lib.rs
+++ b/runtime/vara/src/lib.rs
@@ -99,7 +99,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     // The version of the runtime specification. A full node will not attempt to use its native
     //   runtime in substitute for the on-chain Wasm runtime unless all of `spec_name`,
     //   `spec_version`, and `authoring_version` are the same between Wasm and native.
-    spec_version: 460,
+    spec_version: 470,
     impl_version: 1,
     apis: RUNTIME_API_VERSIONS,
     transaction_version: 1,


breathx:
Could you, please, apply solution we were talking about to avoid holding autogenerated metadata files tracked in git: store metadata in build.rs in hidden with `.` files, added in `.gitignore`, and add them into `gprogram` crate with `include!()` macro under features?

breathx:
I'll set `in-progress` until mentioned above comment get solved

program/build.rs
```diff
@@ -61,12 +47,48 @@ fn update_api() {
...
+/// Update runtime api
+fn update_api() {
+    let path = PathBuf::from(env::var("OUT_DIR").expect("OUT_DIR not found")).join("metadata.rs");
+
+    println!("{path:?}");
+
+    #[cfg(any(
+        all(feature = "gear", not(feature = "vara")),
+        all(feature = "gear", feature = "vara")
+    ))]
```
breathx, line 67: Don't you want to just paste at build.rs line:
```
#[cfg(all(feature = "gear", feature = "vara"))]
compile_error!("Support for different runtimes at the same time is not supported yet");
```

program/Cargo.toml
```diff
@@ -52,14 +52,17 @@ cfg-if = "1"
...
 frame-metadata = "15.0.0"
 parity-scale-codec = "3.1.5"
 subxt-codegen = "0.22.0"
 syn = "1"
 
 [features]
-default = [ ]
+default = [ "gear" ]
+gear = [ "gear-runtime" ]
+vara = [ "vara-runtime" ]
```
breathx, line 65: Here and for `client` should be duplicated to runtime features-set:
`lazy-pages`, `debug-data`, etc, that throws into runtime

Also disable default features for runtimes (std stuff)

breathx: This may be the reason why metadata can't be build directly from runtime: recheck, please

breathx (changes requested):
LGTM, only feature-dependent stuff seems have issues

program/build.rs
breathx: Ohhh, sure

program/Cargo.toml
breathx: > > This may be the reason why metadata can't be build directly from runtime: recheck, please
> 
> what you mean about this? we can build the metadata from runtime directly since the `check` CI works https://github.com/gear-tech/gear/actions/runs/3262279860/jobs/5359139117

Didn't get the point here

breathx: > > Also disable default features for runtimes (std stuff)
> 
> can not build with `default-features = false` https://github.com/gear-tech/gear/actions/runs/3265494330/jobs/5367802193, I think it's free to choose `std` since the cli is based on std itself

Yeah, ok. We found that `substrate_wasm_builder` cuts `std` from runtime **always**

breathx: > yeah, `std`, `runtime-benchmarks` are unnecessary since the runtime apis will not change with these features, but I'm not familiar with the `try-runtime` feature

If I'm not mistaken `runtime-benchmarks` adds some extrinsics, so api should change. Same for `try-runtime`? I see the reason just to paste it there for future purposes. Nobody got hurt from these extra unused features

breathx (approved):

PR #1665, opened 2022-10-14
Remove redundant feature

ekovalev:
- Remove superfluous "cli" feature from `gear-cli` crate which is always on anyway - building the node without it doesn't make much sense.

- Fix failing rtest on vara by making sure the value passed to and from a program would surpass the existential deposit.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -244,7 +244,7 @@ jobs:
 
       - name: "Test: runtime spec tests no lazy pages"
         run: |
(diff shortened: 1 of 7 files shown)


shamilsan (approved):
Let's remove it from here too?

https://github.com/gear-tech/gear/blob/master/.github/workflows/CI.yaml#L247

ekovalev:
> Let's remove it from here too?
> 
> https://github.com/gear-tech/gear/blob/master/.github/workflows/CI.yaml#L247

Good catch

ukint-vs (approved):

breathx (commented):
👍

PR #1671, opened 2022-10-15
Add `create_program` async version

### LINKED ISSUE #1065: Expand api for `create_program` syscall

breathx:
Resolves #1065.


breathx:
I'm not sure should I provide some demo tests just to call fn

ukint-vs (approved):

shamilsan (approved):
Flawlessly 👍

PR #1675, opened 2022-10-16
wasm gen new features

### LINKED ISSUE #1565: WASM-GEN: force wasm-smith to generate bigger wasm code
### Problem to Solve

`wasm-smith` generates code with random sizes, but I've not seen code bigger then 20 kBytes before. We have restriction in  runtime: code can be less then 512 kBytes. So, we must find the problem in `wasm-smith` and force it to generate big codes.

### Possible Solution

-

### Notes

_No response_

grishasobol:
Resolves #1565.

1) Increase code average size.
2) Syscalls refactoring and tuning: make syscalls generation more suitable for gear node.
3) Add handle generation.
4) Add print info in test generation: information string is printed using `gr_debug` in the beginning of `init`.

### DIFF
--- a/utils/node-loader/src/batch_pool/generators.rs
+++ b/utils/node-loader/src/batch_pool/generators.rs
@@ -29,7 +29,7 @@ pub fn generate_gear_program<Rng: LoaderRng>(seed: Seed) -> Vec<u8> {
 
     let mut u = Unstructured::new(&buf);
 
-    let mut config = gear_wasm_gen::GearConfig::default();
+    let mut config = gear_wasm_gen::GearConfig::new_normal();
     config.print_test_info = Some(format!("Gear program seed = '{}'", seed));
 
     gear_wasm_gen::gen_gear_program_code(&mut u, config)
--- a/utils/node-loader/src/batch_pool/generators.rs
+++ b/utils/node-loader/src/batch_pool/generators.rs
@@ -29,7 +29,10 @@ pub fn generate_gear_program<Rng: LoaderRng>(seed: Seed) -> Vec<u8> {
 
     let mut u = Unstructured::new(&buf);
 
-    gear_wasm_gen::gen_gear_program_code(&mut u, gear_wasm_gen::GearConfig::default())
+    let mut config = gear_wasm_gen::GearConfig::default();
+    config.print_test_info = Some(format!("Gear program seed = '{}'", seed));
+
+    gear_wasm_gen::gen_gear_program_code(&mut u, config)
 }
 
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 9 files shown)


utils/node-loader/src/batch_pool/generators.rs
```diff
@@ -29,7 +29,10 @@ pub fn generate_gear_program<Rng: LoaderRng>(seed: Seed) -> Vec<u8> {
 
     let mut u = Unstructured::new(&buf);
 
-    gear_wasm_gen::gen_gear_program_code(&mut u, gear_wasm_gen::GearConfig::default())
+    let mut config = gear_wasm_gen::GearConfig::new_normal();
+    config.print_test_info = Some(format!("Gear program seed = '{}'", seed));
```
breathx, line 33:
```suggestion
    config.print_test_info = Some(format!("Gear program seed = '{seed}'"));
```

utils/wasm-gen/src/syscalls.rs
```diff
@@ -73,501 +132,173 @@ impl Default for SyscallsConfig {
...
-    // gr_wait_for(duration: u32) -> !;
     res.insert(
         "gr_wait_for",
-        SysCallInfo {
-            params: [I32].to_vec(),
-            results: [].to_vec(),
-            param_rules: [ptr_rule()].to_vec(),
-            frequency,
-        },
+        SysCallInfo::new(config, [Delay], [], frequency),
```
breathx, line 204: That's not delay, that's duration

utils/wasm-gen/src/syscalls.rs
```diff
@@ -73,501 +132,173 @@ impl Default for SyscallsConfig {
...
+    res.insert("gr_wait", SysCallInfo::new(config, [], [], frequency));
     res.insert(
         "gr_wait_up_to",
-        SysCallInfo {
-            params: [I32].to_vec(),
-            results: [].to_vec(),
-            param_rules: [ptr_rule()].to_vec(),
-            frequency,
-        },
+        SysCallInfo::new(config, [Size], [], frequency),
```
breathx, line 200: same here: not size

breathx (approved):

utils/wasm-gen/src/syscalls.rs
grishasobol: Is there difference between `delay` and `duration` in syscalls ?
```
       pub fn gr_wait_for(duration: u32) -> !;

       pub fn gr_wake(message_id_ptr: *const [u8; 32], delay: u32) -> SyscallError;
```

breathx: `gr_wait_for` and `gr_wake` is totally different things.

Delay for sending messages and `gr_wake` means amount of blocks system should wait before do requested stuff.

Duration allowed for `gr_wait`-like syscalls and means how many blocks message should be in waitlist

utils/node-loader/src/batch_pool/generators.rs
grishasobol: Done

utils/wasm-gen/src/syscalls.rs
grishasobol: Done

utils/wasm-gen/src/syscalls.rs
grishasobol: Done

PR #1689, opened 2022-10-19
Introduce gear-wasm-instrument

### LINKED ISSUE #1607: Method `inject2` in wasm-instrument
### Problem to Solve

The method should inject two globals - `gas_counter` & `gas_allowance` and an internal function (export) - `gas`. The function charges gas for execution and calls imported function(s) - `out_of_gas`/`out_of_allowance` - if there is no enough gas/allowance to reduce from. The original binary instrumented with calls to the internal function `gas`.

### Possible Solution

-

### Notes

_No response_

gshep:
Resolves #1607 

Blocked by https://github.com/gear-tech/wasm-instrument/pull/2

@gear-tech/dev


utils/wasm-instrument/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+
+    mbuilder.push_export(
+        builder::export()
+            .field(GLOBAL_NAME_ALLOWANCE)
+            .internal()
+            .global(allowance_index)
+            .build(),
+    );
+
+    let mut elements = vec![
```
grishasobol, line 145: Create TODO and isssue to optimize this code

utils/wasm-instrument/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+
+    // determine cost for successful execution
+    let cost_check_blocks = match elements.iter().take(7).try_fold(0u64, |cost, instruction| {
+        rules.instruction_cost(instruction).map(|c| cost + c as u64)
+    }) {
+        Some(c) => 2 * c,
+        None => return Err(mbuilder.build()),
+    };
+
+    let cost_set_blocks = match elements
```
grishasobol, line 215: Why do we need to calculate `cost_check_blocks` and `cost_set_blocks` separately if we use them like this `cost_check_blocks + cost_set_blocks`

utils/wasm-instrument/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+            .body()
+            .with_instructions(elements::Instructions::new(elements))
+            .build()
+            .build(),
+    );
+
+    // back to plain module
+    let module = mbuilder.build();
+
+    gas_metering::post_injection_handler(module, rules, gas_charge_index, out_of_gas_index, 2)
```
grishasobol, line 256: Add please comment about call indexes resolving here, because it's not simple to understand why call indexes stay correct after new imports push.

utils/wasm-instrument/src/tests.rs
```diff
@@ -0,0 +1,597 @@
...
+    wasmparser::validate(&binary).unwrap();
+}
+
+#[test]
+fn grow_no_gas_no_track() {
+    let module = parse_wat(
+        r"(module
+        (func (result i32)
+            global.get 0
+            memory.grow)
```
grishasobol, line 121: Why so many tests with memory.grow ? this instruction is restricted in gear.

grishasobol (changes requested):

utils/wasm-instrument/src/lib.rs
gshep: Since instructions inside "if" block are skipped: they executed only on out of gas/allowance

utils/wasm-instrument/src/tests.rs
gshep: These are the original tests from wasm-instrument

utils/wasm-instrument/src/lib.rs
gshep: ok

utils/wasm-instrument/src/lib.rs
grishasobol: What about this?
```rust
    let cost = elements
        .iter()
        .take(7)
        .chain(elements.iter().skip(10).take(7))
        .try_fold(0u64, |cost, instruction| {
            rules.instruction_cost(instruction).map(|c| cost + c as u64)
        });
```

grishasobol: Also can you make checked add: `|c| cost + c as u64`. And make check after that the result `cost` for gas_charge function is less then u32::MAX

utils/wasm-instrument/src/tests.rs
grishasobol: Ok

utils/wasm-instrument/src/lib.rs
gshep: good remark! I didn't so 'cause my eyes are full of soap =)

utils/wasm-instrument/src/lib.rs
gshep: Yep, good remark too. However there are comments in the original wasm-instrument explaining that:
- https://github.com/gear-tech/wasm-instrument/blob/gear-stable/src/gas_metering/mod.rs#L179
- https://github.com/gear-tech/wasm-instrument/blob/gear-stable/src/gas_metering/mod.rs#L187
- https://github.com/gear-tech/wasm-instrument/blob/gear-stable/src/gas_metering/mod.rs#L130

So I guess there is no reason to duplicate the information there.

grishasobol: Ok

grishasobol (approved):

PR #1698, opened 2022-10-20
Reject incorrect binaries

### LINKED ISSUE #1694: Malformed binaries should be ignored
### Problem

At the moment the extrinsics accept incorrect wasm-binaries so processor fails later on module instantiation. For better user experience and possible side-effects deny such binaries on `Code` level.

### Steps

Provide corresponding test.

### Possible Solution

_No response_

### Notes

_No response_

### Relevant Log Output

<p>

```
<Paste your log here>
```

</p>
</details>

gshep:
Resolves #1694 

@gear-tech/dev

### DIFF
--- a/pallets/gear/src/tests.rs
+++ b/pallets/gear/src/tests.rs
@@ -5758,6 +5758,34 @@ fn test_mad_big_prog_instrumentation() {
     })
 }
 
+#[test]
+fn reject_incorrect_binary() {
+    let wat = r#"
+    (module
+        (import "env" "memory" (memory 1))
+        (export "handle" (func $handle))
+        (func $handle
+            i32.const 5
+        )
+    )"#;
+
+    init_logger();
+    new_test_ext().execute_with(|| {
+        assert_noop!(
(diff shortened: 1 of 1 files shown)


ukint-vs (approved):

core/src/code.rs
```diff
@@ -185,6 +187,8 @@ impl Code {
         module: Option<Module>,
         instrument_with_const_rules: bool,
     ) -> Result<Self, CodeError> {
+        wasmparser::validate(&original_code).map_err(|_| CodeError::Decode)?;
```
grishasobol, line 190: As I can see in the comment:
```rust
    /// Create the code without checks.
```

grishasobol (changes requested):

core/src/code.rs
gshep: Agree however as you can see it contains several checks anyway. So I suppose that original code should be valid wasm-binary.

grishasobol (approved):

PR #1709, opened 2022-10-22
fix alloc overflow when 0 pages allocated 

grishasobol:
Add test for the case.

### DIFF
--- a/gclient/tests/upload.rs
+++ b/gclient/tests/upload.rs
@@ -81,11 +81,9 @@ async fn harmless_upload() -> Result<()> {
     Ok(())
 }
 
-#[test]
-fn alloc_zero_pages() {
-    #[tokio::main()]
-    async fn start_tokio() -> Result<()> {
-        let wat_code = r#"
+#[tokio::test]
+async fn alloc_zero_pages() -> Result<()> {
+    let wat_code = r#"
         (module
             (import "env" "memory" (memory 0))
             (import "env" "alloc" (func $alloc (param i32) (result i32)))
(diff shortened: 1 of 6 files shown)


ukint-vs (approved):

gclient/tests/upload.rs
```diff
@@ -61,16 +50,65 @@ async fn harmless_upload() -> Result<()> {
...
+    // By default, login as Alice, than re-login as Bob.
+    let api = GearApi::dev().await?.with("//Bob")?;
+
+    upload_programs_and_check(&api, codes).await?;
+
+    Ok(())
+}
+
+#[test]
+fn alloc_zero_pages() {
```
breathx, line 85: Why so complicated?

breathx (commented):
Where is the fix itself?

gclient/tests/upload.rs
grishasobol: Can you offer simpler solution? I suppose there is some ways to make tests with timeout better.

SabaunT, line 85: I guess it could be run in an async context.

SabaunT (approved):

gclient/tests/upload.rs
SabaunT: either test can be re-written or `block_stopped` checking function

breathx: with tokio::test, exactly

grishasobol: Done

PR #1715, opened 2022-10-25
refactor(gstd): impl traits for Futures with declarative macro

### LINKED ISSUE #1710: Implemenet `Future` trait for `gstd::msg::async::*Future` with declarative macros
### File Location(s)

https://github.com/gear-tech/gear/blob/master/gstd/src/msg/async.rs

### Proposal

Same logic for 4 structs now, since we are not going to use trait for this, writing some macros for them will enhance our developing experience ))

```
macro_rules! impl_future {}
macro_rules! impl_codec_future {}
```

clearloop:
Resolves #1710 

https://github.com/gear-tech/gear/pull/1457 introduced two more methods which will make the implementation of the futures more verbose, so here maybe we can try wrapping the duplicated implementations in macros

- [x] introduce macro `impl_futures`
- [x] impl traits for `*Futures` with macros

@gear-tech/dev

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2855,9 +2855,9 @@ dependencies = [
  "gear-backend-common",
  "gear-core",
  "gear-core-errors",
+ "gear-wasm-instrument",
  "log",
  "parity-scale-codec",
- "parity-wasm 0.45.0",
  "sp-sandbox",
 ]
 
@@ -2869,9 +2869,9 @@ dependencies = [
  "gear-backend-common",
  "gear-core",
  "gear-core-errors",
+ "gear-wasm-instrument",
  "log",
  "parity-scale-codec",
- "parity-wasm 0.45.0",
  "wasmi 0.14.0",
 ]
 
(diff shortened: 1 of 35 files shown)


gstd/src/msg/async.rs
```diff
@@ -33,6 +34,23 @@ use core::{
 };
 use futures::future::FusedFuture;
 
+fn poll<F, R>(waiting_reply_to: MessageId, cx: &mut Context<'_>, f: F) -> Poll<Result<R>>
+where
+    F: Fn(Vec<u8>) -> Result<R>,
+{
+    match signals().poll(waiting_reply_to, cx) {
+        ReplyPoll::None => panic!("Somebody created CodecMessageFuture with the MessageId that never ended in static replies!"),
```
shamilsan, line 42: Need to choose more neutral future name.
```suggestion
        ReplyPoll::None => panic!("Somebody created a future with the MessageId that never ended in static replies!"),
```

shamilsan (approved):

PR #1725, opened 2022-10-26
Sycalls benchmarks refactoring and problems resolving

grishasobol:
1) Move syscalls benchmarks logic to separate file. 
2) For each syscall benchmark slightly increase complexity  (using for example unaligned offsets).
3) Add gr_debug_per_kb bench and charging gas for debug payload len.
4) Add `salt len` param in gr_create_program_per_kb bench and charging gas for salt len.
5) Add `free` benchamrking and gas charging for `free`.
6) Fix syscall signature in all benchmarks.
7) Fix syscall call params (or pre-handling), so that sysalls will be executed in _happy_ case (not for all syscalls).
8) Measure new weights and change them in corresponding files.


ukint-vs (approved):

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -643,1295 +472,411 @@ benchmarks! {
...
-            context,
-            memory_pages,
-        } = prepare::<T>(instance.caller.into_origin(), HandleKind::Handle(ProgramId::from_origin(instance.addr)), vec![], 0u32.into())?;
-    }: {
-        core_processor::process::<
-            Externalities,
-            ExecutionEnvironment,
-        >(&block_config, context, memory_pages);
+        let mut res = None;
+        let exec = getter_bench::<T>("gr_value", r)?;
```
breathx, line 564:
```suggestion
        let exec = getter_bench::<T>(_, r)?;
```
this might work

breathx (approved):
Great!
Wait merge until discussed rebench

pallets/gear/src/benchmarking/mod.rs
grishasobol: Unfortunately it does not work like this. Let's do something in future refactoring.
```
  error: in expressions, `_` can only be used on the left-hand side of an assignment
     --> /Users/gsobol/src/gear/pallets/gear/src/benchmarking/mod.rs:597:45
      |
  597 |         let exec = number_getter_bench::<T>(_, ValueType::I32, r)?;
      |                                             ^ `_` not allowed here

```

PR #1750, opened 2022-11-05
Rework block authorship

ekovalev:
Solves the issue of incorrect encoding of the custom unsigned extrinsic after a runtime upgrade (due to changes in `parity-scale-codec` indexing of installed pallets in case new pallets are added to the `construct_runtime!` macro).

The fix is inspired by how `inherent` extrinsics manage to keep the correct encoding by fetching the extrinsics in question via a dedicated `runtime-api` call thereby ensuring the correct runtime version.
A new `runtime-api` call has been added to do exactly this: to return the correct `RuntimeCall` variant corresponding to the `pallet_gear`'s `run` extrinsic.
Naturally, the custom block authoring logic has been modified (in fact, simplified) to make use this runtime-api call.

This guarantees the custom block authoring mechanism won't be broken by runtime upgrades.

However, since the client part has changed (block authorship), new node binaries will have to be deployed for the current `vara-stage-1` network.


node/authorship/src/authorship.rs
```diff
@@ -470,7 +468,7 @@ where
             }
 
             trace!("[{:?}] Pushing to the block.", pending_tx_hash);
-            match sc_block_builder::BlockBuilder::push(&mut block_builder, pending_tx_data) {
+            match BlockBuilderExt::push(&mut block_builder, pending_tx_data) {
```
breathx, line 471:
```suggestion
            block_builder.push(pending_tx_data) {
```

pallets/gear/rpc/runtime-api/src/lib.rs
```diff
@@ -20,11 +20,15 @@
...
+use sp_runtime::traits::Block as BlockT;
 use sp_std::vec::Vec;
 
 sp_api::decl_runtime_apis! {
     pub trait GearApi {
         #[allow(clippy::too_many_arguments)]
         fn calculate_gas_info(source: H256, kind: HandleKind, payload: Vec<u8>, value: u128, allow_other_panics: bool, initial_gas: Option<u64>,) -> Result<GasInfo, Vec<u8>>;
+
+        /// Generate inherent-like extrinsic the runs message queue processing.
+        fn run_queue_extrinsic() -> <Block as BlockT>::Extrinsic;
```
breathx, line 32:
```suggestion
        fn gear_run_extrinsic() -> <Block as BlockT>::Extrinsic;
```

node/authorship/src/block_builder.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// Estimate the size of the block in the current state.
+    pub fn estimate_block_size(&self, include_proof: bool) -> usize {
+        self.block_builder.estimate_block_size(include_proof)
+    }
+
+    pub fn create_terminal_extrinsic(&mut self) -> Result<Block::Extrinsic, Error> {
+        let block_id = BlockId::Hash(self.parent_hash);
+        self.api
+            .execute_in_transaction(move |api| {
+                TransactionOutcome::Rollback(api.run_queue_extrinsic_with_context(
```
breathx, line 82: can't find from where `_with_context` method appeared?

breathx (approved):
Seems its' required to update the runtime binary for runtime upgrade test

breathx:
Add some dummy pallet for the test runtime would be also good to check that everything goes well for our case also

breathx:
> Add some dummy pallet for the test runtime would be also good to check that everything goes well for our case also

Or just change gear pallet number in the runtime

node/authorship/src/block_builder.rs
ekovalev: It's some kind of Substrate magic. Didn't find it either but yet all the runtime api methods called from transactional environment have it.

breathx: Seems like extending for rpc methods to have bn stamp as an argument, okay

ukint-vs (approved):

PR #1754, opened 2022-11-06
Add runtime-api call to support block production

ekovalev:
Add a new runtime-api call that would return a properly encoded custom unsigned extrinsic that runs the message queue. 

This is a necessary prerequisite for a transition to proper block production free of the previously observed `RuntimeCall` encoding issues before the node binaries can be updated.


ukint-vs (approved):

shamilsan (approved):

ukint-vs:
Completed via force push.

PR #1783, opened 2022-11-07
Fix events handling

SabaunT:
Off-chain applications are unaware of forks, so they need a better block headers handling. That's because getting events of some block by, say, block header (as a key) can be a fallible action due to state of the block with that block hash being discarded. Solution here can be getting only finalized blocks and reading data from them.

That's the purpose of that fix - subscribe on finalized blocks only,

@gear-tech/dev

### DIFF
--- a/utils/node-loader/src/utils.rs
+++ b/utils/node-loader/src/utils.rs
@@ -7,6 +7,7 @@ use gclient::WSAddress;
 use rand::{Rng, RngCore, SeedableRng};
 use reqwest::Client;
 use std::{
+    collections::HashMap,
     fs::File,
     io::Write,
     iter,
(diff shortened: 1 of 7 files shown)


breathx (approved):

utils/node-loader/src/utils.rs
```diff
@@ -130,11 +131,10 @@ pub async fn with_timeout<T>(fut: impl Future<Output = T>) -> Result<T> {
 
 pub async fn stop_node(monitor_url: String) -> Result<()> {
     let client = Client::new();
-    client
-        .post(monitor_url)
-        .body("gear-node-loader=stop-gear")
-        .send()
-        .await?;
+    let mut params = HashMap::new();
+    params.insert("__script_name", "stop");
```
shamilsan, line 135: In the future. You can create a `HashMap` from an array of tuples.
```rust
let params: HashMap<_,_> = [("__script_name", "stop")].into();
```

PR #1792, opened 2022-11-08
Add a new runtime-api call (vara-stage-2)

ekovalev:
Add a new runtime-api call that would return a properly encoded custom unsigned extrinsic that runs the message queue.

This is a necessary prerequisite for a transition to proper block production free of the previously observed RuntimeCall encoding issues before the node binaries can be updated.

**N.B.:** it seems a better idea to rebase this branch on top of the `vara-stage-1` that would already have this change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3193,6 +3193,8 @@ dependencies = [
  "log",
  "memory-db",
  "pallet-babe",
+ "pallet-gear",
+ "pallet-gear-rpc-runtime-api",
  "pallet-timestamp",
  "parity-scale-codec",
  "parity-util-mem",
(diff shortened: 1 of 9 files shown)


shamilsan:
Isn't it better to rebase `vara-stage-2` onto the `vara-stage-1` branch? Or they are too different?

ekovalev:
> Isn't it better to rebase `vara-stage-2` onto the `vara-stage-1` branch? Or they are too different?

Good question. Ideally that's how it should be.

shamilsan (approved):

ukint-vs (approved):

NikVolf:
Vara-stage-2 should be rebased instead

PR #1798, opened 2022-11-08
feat(gprogram): export rpc calls from Api

### LINKED ISSUE #1780: Add rpc-calls support for `gclient` from `GearApi`
### Problem to Solve

.

### Possible Solution

.

### Notes

_No response_

clearloop:
Resolves #1780 

- move rpc module from signer to api
-

@gear-tehc/dev

### DIFF
--- a/pallets/gear/src/manager/journal.rs
+++ b/pallets/gear/src/manager/journal.rs
@@ -471,13 +471,16 @@ where
             unreachable!("Threshold for reservation invalidated")
         }
 
-        let total_amount = amount.saturating_add(hold.lock());
+        // TODO: uncomment line below (issue #1830).
+        // let total_amount = amount.saturating_add(hold.lock());
+        let total_amount = amount;
 
         GasHandlerOf::<T>::reserve(message_id, reservation_id, total_amount)
             .unwrap_or_else(|e| unreachable!("GasTree corrupted: {:?}", e));
 
-        GasHandlerOf::<T>::lock(reservation_id, hold.lock())
-            .unwrap_or_else(|e| unreachable!("GasTree corrupted! {:?}", e));
+        // TODO: uncomment lines below (issue #1830).
+        // GasHandlerOf::<T>::lock(reservation_id, hold.lock())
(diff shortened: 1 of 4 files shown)


SabaunT:
Can we see a couple of tests in 'gclient' of rpc calls?

program/src/api/rpc.rs
```diff
@@ -35,7 +30,7 @@ impl Signer {
             .request(
                 "gear_calculateInitCreateGas",
                 rpc_params![
-                    self.source(),
+                    H256(Default::default()),
```
breathx, line 33: why source removed?

breathx: this thing may affect calculate gas logic

gclient/src/api/mod.rs
```diff
@@ -90,3 +91,17 @@ impl GearApi {
             .map_err(Into::into)
     }
 }
+
+impl Deref for GearApi {
```
breathx, line 95: Do we really need all methods of Signer available for GearApi?

breathx (changes requested):
Agree with @SabaunT, some tests are required just for demonstration
Otherwise lgtm!

program/src/api/rpc.rs
```diff
@@ -0,0 +1,126 @@
...
+        at: Option<H256>,
+    ) -> Result<GasInfo> {
+        self.rpc()
+            .request(
+                "gear_calculateHandleGas",
+                rpc_params![
+                    origin,
+                    H256(destination.into()),
+                    hex::encode(payload),
+                    u64::try_from(value).unwrap_or(u64::MAX),
```
breathx, line 90: do we really can't throw here u128? use [u8; 16] instead for rpc calls also

breathx (approved):
I'd make source argument optional and paste self id in case of none

breathx:
Add and similar bindings for `gclient`, please!

program/src/api/rpc.rs
breathx: we never used u64 for value
If I'm not mistaken, rpc params macro wasn't able to use it for some reason

PR #1784, opened 2022-11-07
Introduce backend integration tests

SabaunT:
Is intended solve this https://github.com/gear-tech/gear/issues/1496#issuecomment-1290685887.

This PR shows how to implement #1802. Sys-calls integrity tests can be run with executors from both wasmi and sanbox. For wasmi simply run from the workspace root `cargo test -p pallet-gear --features runtime-benchmarks`. For sandbox first build the node with `runtime-benchmarks` feature being enabled and then run `./targetproduction/gear benchmark pallet --chain dev --execution=wasm --wasm-execution=compiled --pallet=pallet_gear --extrinsic="check_syscalls_integrity" --steps=2 --repeat=1`. To enable logs in the last one add `-lgear_backend_sandbox=trace,gear_core_processor=trace,pallet_gear=trace,pallet_babe=trace`.

PR #1803, opened 2022-11-09
Use the input buffer as outgoing payload in several syscalls

### LINKED ISSUE #1669: Allow reply_push to use input buffer
This will allow zero-copy message relaying in multiple scenarios

gshep:
Resolves #1669 

- `Ext::send_push_input` & `Ext::reply_push_input`
- adjust `gcore` & `gstd` accordingly
- implement syscalls in wasmi & sandbox backends
- add benchmarks for the new functions
- test case with relaying parts of an incoming message to different destinations

**Release notes**: This PR introduces a new ability to relay incoming messages. The slice of the incoming message buffer is used as a payload for the outgoing message. The new `gstd` API supports common range syntax (`..`, `0..`, `..=n`, etc) in Rust so one is able to relay the first half of the message the following way:
```rust
let len = gstd::msg::size() / 2 as usize;
gstd::msg::send_input(user_or_program_id, 0, ..=len).expect("`send_input' failed");
```

@gear-tech/dev


NikVolf:
And more thoughtful examples where you send different part of the incoming payloads to different destination would be great

NikVolf (changes requested):
Please review my comments

core-backend/sandbox/src/env.rs
```diff
@@ -161,6 +161,12 @@ where
             Funcs::reservation_reply_commit,
         );
         builder.add_func("gr_reply_push", Funcs::reply_push);
+        builder.add_func("gr_rereply", Funcs::rereply);
```
NikVolf, line 164: i think naming should be made more consistent

for example,`gr_reply_push` should have an analog of `gr_reply_push_input` and so on

gshep: Yep, agree. What is about 'gr_relay_{send, push}`?

NikVolf: Relay has too opionated meaning, it's not just about relay

core/src/message/context.rs
```diff
@@ -313,6 +313,43 @@ impl MessageContext {
...
+            excluded_end,
+        } = range;
+
+        data.try_extend_from_slice(&self.current.payload()[offset..excluded_end])
+            .map_err(|_| Error::MaxMessageSizeExceed)?;
+
+        Ok(())
+    }
+
+    /// Check range for `send_push_input`/`reply_push_input` methods.
```
NikVolf, line 337: Can you explain in doc what is this check about?

So far I don't see this functions checks anything but rather returns some different range

NikVolf (approved):
other than comment above

core/src/message/context.rs
gshep: ok

Cargo.lock
```diff
@@ -1679,6 +1679,16 @@ dependencies = [
  "scale-info",
 ]
 
+[[package]]
```
breathx, line 1682: To be merged after #1784

PR #1812, opened 2022-11-10
fix gclient zero_alloc test timeout

grishasobol:
set so that timeout will be for the part of messages processing.


gclient/tests/upload.rs
```diff
@@ -27,7 +27,11 @@ const PATHS: [&str; 2] = [
     "../target/wat-examples/inf_recursion.wasm",
 ];
 
-async fn upload_programs_and_check(api: &GearApi, codes: Vec<Vec<u8>>) -> Result<()> {
+async fn upload_programs_and_check(
+    api: &GearApi,
+    codes: Vec<Vec<u8>>,
+    timeout: Option<Duration>,
```
breathx, line 33: I'd add milliseconds and construct Duration type inside

breathx (approved):

gclient/tests/upload.rs
grishasobol: Actually we can do this when we will heave need =) currently cannot see any points

PR #1814, opened 2022-11-10
tune load and store weights

### LINKED ISSUE #1728: BENCHMARKING: improve load/store instructions cost
### Problem

https://github.com/gear-tech/gear/blob/4dd6c63ef788fb7343cd6a20445a8268fb9614e0/pallets/gear/src/benchmarking/mod.rs#L1960-L1974

Looking at benchmarks for loads we can see, that we investigate only loads, that access two first wasm pages. Two wasm pages can be fit inside L1 and L2 caches of processor, so each load will be without L2 cache miss, which means that his execution time is not much bigger than simple add. In wasm example above, I have made access to different part of 512-pages wasm memory, so L2 cache misses are appear almost for each load operation, which causes that in fact load operation takes much more time then it's weight tell us.
Another problem is that in benchmark all loads are without offset. But in wasmer loads with offsets generates additional checks in code (check that base addr + offset calcs without overflow), which leads to additional branches in compiled code and also increase execution time.

See also: https://github.com/gear-tech/gear/issues/1382

grishasobol:
Resolves #1728.

Make random access overall 512 wasm pages memory. Which makes load and stores much more expensive, because of L1 and L2 caches misses (see also #1817 ). To reduce `lazy mmap` factor, make all memory pages access before.

Add also `lazy-pages` read/write access benchmarks and weights for future using.

### DIFF
--- a/pallets/gear/src/benchmarking/code.rs
+++ b/pallets/gear/src/benchmarking/code.rs
@@ -385,6 +385,8 @@ where
 
 /// Mechanisms to generate a function body that can be used inside a `ModuleDefinition`.
 pub mod body {
+    use gear_core::memory::WasmPageNumber;
+
     use super::*;
 
     /// When generating contract code by repeating a wasm sequence, it's sometimes necessary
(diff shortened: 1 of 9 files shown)


ukint-vs (approved):

pallets/gear/src/benchmarking/code.rs
```diff
@@ -452,6 +454,36 @@ pub mod body {
...
+        head
+    }
+
+    pub fn read_access_all_pages_instrs(
+        mem_size: u32,
+        mut head: Vec<Instruction>,
+    ) -> Vec<Instruction> {
+        for page in (0..mem_size)
+            .map(WasmPageNumber)
+            .flat_map(|p| p.to_gear_pages_iter())
```
breathx, line 478: why do we need access to all gear pages, if only wasm pages make sense for the bench, isn't it?

breathx (approved):

pallets/gear/src/benchmarking/code.rs
grishasobol: No, each native page must be accessed at least

NikVolf (changes requested):
What is this, 12ms for single memory access? Where do you get memory with this latency?

breathx:
> What is this, 12ms for single memory access? Where do you get memory with this latency?

That's price for batch of 100 calls. Single memory access equals 12ms / 100 = 120 us (microseconds)

grishasobol:
> > What is this, 12ms for single memory access? Where do you get memory with this latency?
> 
> That's price for batch of 100 calls. Single memory access equals 12ms / 100 = 120 us (microseconds)

Now we have 1000 instructions in one batch.

NikVolf (approved):

PR #1833, opened 2022-11-13
Fix event depositing from reservation

breathx:


ark0f (approved):

ark0f:
LGTM!

shamilsan (approved):
As it should be 🔥

PR #1840, opened 2022-11-15
Rename `gear-node` base dir to `gear` instead of changing base path

### LINKED ISSUE #1776: Deal with broken path to db after binary renaming
### Problem to Solve

It couldn't be purged with default command

### Possible Solution

Also throw `--tmp` flag for tests running in scripts

### Notes

_No response_

shamilsan:
Resolves #1776

- Rename the `gear-node` base directory to `gear` if the first exists and the latter doesn't.
- Default base directory now depends on the executable name as it was before the executable binary renaming.

@gear-tech/dev

### DIFF
--- a/node/cli/src/command.rs
+++ b/node/cli/src/command.rs
@@ -126,6 +126,12 @@ macro_rules! unwrap_client {
 pub fn run() -> sc_cli::Result<()> {
     let mut cli = Cli::from_args();
 
+    let old_base = BasePath::from_project("", "", "gear-node");
+    let new_base = BasePath::from_project("", "", &Cli::executable_name());
+    if old_base.path().exists() && !new_base.path().exists() {
+        _ = std::fs::rename(old_base.path(), new_base.path());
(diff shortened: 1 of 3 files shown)


ukint-vs (approved):

node/cli/src/command.rs
```diff
@@ -126,6 +126,12 @@ macro_rules! unwrap_client {
 pub fn run() -> sc_cli::Result<()> {
     let mut cli = Cli::from_args();
 
+    let old_base = BasePath::from_project("", "", "gear-node");
+    let new_base = BasePath::from_project("", "", &Cli::executable_name());
+    if old_base.path().exists() && !new_base.path().exists() {
+        _ = std::fs::rename(old_base.path(), new_base.path());
```
ukint-vs, line 132: I think it's better to delete it later

shamilsan: Yes, we will.

PR #1843, opened 2022-11-16
Integral tests for token economy and democracy in stage-2

ekovalev:
At current stage this tool includes some tests to check basic invariants about economic model, namely the total inflation over a period of time, validators and nominators rewards etc.
As a separate effort, there is a test that checks some democracy functionality, however it's value is questionable since it mainly repeats (or, combines, for that matter) existing Substrate unit tests from the `democracy` pallet.


breathx:
Seems rebase didn't help

utils/integral-tests/src/runtime.rs
```diff
@@ -0,0 +1,965 @@
...
+pub mod constants {
+    /// Currency related constants
+    pub mod currency {
+        use runtime_primitives::Balance;
+
+        /// The existential deposit.
+        pub const EXISTENTIAL_DEPOSIT: Balance = MILLICENTS / 10; // 1_000
+
+        // TODO: review quantities based on economic model (issue #1277)
+        pub const UNITS: Balance = 1_000_000_000; // 10^(-9) precision
```
breathx, line 78: Was it expected to use ^9 here?

utils/integral-tests/src/runtime.rs
```diff
@@ -0,0 +1,965 @@
...
+    pub const QueueLengthStep: u128 = 10;
+    pub const OperationalFeeMultiplier: u8 = 5;
+}
+
+impl pallet_transaction_payment::Config for Runtime {
+    type RuntimeEvent = RuntimeEvent;
+    type OnChargeTransaction = CurrencyAdapter<Balances, ()>;
+    type OperationalFeeMultiplier = OperationalFeeMultiplier;
+    type WeightToFee = IdentityFee<Balance>;
+    type LengthToFee = IdentityFee<Balance>;
```
breathx, line 305: We no longer use identity fee, does it matter somehow?

breathx (approved):
How do you feel about merge this pr not in vara, but in master branch? and reuse constants from vara runtime

utils/integral-tests/src/runtime.rs
ekovalev: No reason for having smaller precision, just dealing with smaller numbers. I suggest to keep it this way for now to avoid risk of tests failing in case we align it with 12 decimals as we have in Vara.

utils/integral-tests/src/runtime.rs
ekovalev: It doesn't because no transaction fees apply in current tests anyway. Anyway, modified it in accordance with Vara (and Gear) runtimes.

PR #1860, opened 2022-11-18
feat(cli): override log level on flag validator

### LINKED ISSUE #1794: Feature which disables debug and trace logging in node-cli
### Problem to Solve

In order to reduce risk of node performance slow down and therefore validators slashing, we should make additional feature which will disable all logging, except (may be) info and error level . And also we should publish binaries for validators compiled with this feature and add in doc building command with this feature.

clearloop:
Resolves #1794 

the `Ord` implementation of `log::Level` follows the `usize` repr https://docs.rs/log/0.4.17/log/enum.Level.html

```
#[repr(usize)]
pub enum Level {
    Error,
    Warn,
    Info,
    Debug,
    Trace,
}
```

so `max_log_level` means, supports `Error`, `Warn` and `Info`, if we want to only support `Error` and `Info` we have to hack `log_tracer` or `sc_tracing` unfortunately since the logic we have to modify are private in the two modules

##  Refs

- https://github.com/paritytech/substrate/blob/0dc5358e43187a1f54afb2455d42eddb0965c16a/client/tracing/src/logging/mod.rs#L165
- https://github.com/tokio-rs/tracing/blob/60c60bef62972e447414a748a95b31ff9027165b/tracing-log/src/log_tracer.rs#L248

@gear-tech/dev

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -75,9 +75,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.19"
+version = "0.7.20"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b4f55bd91a0978cbfd91c457a164bab8b4001c833b7f323132c0a4e1922dd44e"
+checksum = "cc936419f96fa211c1b9166887b38e5e40b19958e5b895be7c1f93adec7071ac"
 dependencies = [
  "memchr",
 ]
@@ -126,9 +126,9 @@ dependencies = [
 
 [[package]]
 name = "array-bytes"
-version = "4.1.0"
+version = "4.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6a913633b0c922e6b745072795f50d90ebea78ba31a57e2ac8c2fc7b50950949"
+checksum = "f52f63c5c1316a16a4b35eaac8b76a98248961a533f061684cb2a7cb0eafb6c6"
 
 [[package]]
 name = "arrayref"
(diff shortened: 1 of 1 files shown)


SabaunT (approved):
I suppose after running there were only `INFO`,  `WARN`, `ERROR` logs, right?

node/cli/src/cli.rs
```diff
@@ -30,6 +31,17 @@ pub struct RunCmd {
...
 }
 
+impl RunCmd {
+    /// Set `Wasm` as default execution strategy.
+    pub fn with_wasm_execution(&mut self) {
+        self.base
+            .import_params
+            .execution_strategies
+            .execution
+            .get_or_insert(ExecutionStrategy::Wasm);
```
breathx, line 41: get_or_insert won't work if Native already set, isn't it?

runtime/gear/src/lib.rs
```diff
@@ -97,7 +97,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
     impl_name: create_runtime_str!("gear"),
     apis: RUNTIME_API_VERSIONS,
     authoring_version: 1,
-    spec_version: 100,
+    spec_version: 110,
```
breathx, line 100: dump, please

breathx (changes requested):

node/cli/src/cli.rs
grishasobol: Cannot understand the reason of creating this method

grishasobol (approved):

breathx:
Btw, wasn't `--validator` always set for `--dev` node? What about assert debugging modes also?

breathx (approved):

PR #1854, opened 2022-11-16
Charge for reservation holding

ark0f:
Resolves #1830

PR #1861, opened 2022-11-18
Extend property tests of gas provider

ark0f:
@gear-tech/dev


SabaunT (changes requested):

SabaunT (changes requested):
- [x] Please check that in the `consume` if called on the `Cut`, only `Cut` node is removed. So cascade removal doesn't remove any other node when `Cut` is removed
- [ ] `Cut` node doesn't point to the tree (with `ExternalId`)
- [x] If cascade removal (after `consume`) removed `External` node, `Reserved` node won't appear in removed nodes (and vice versa).
- [x] Check that `Reserved` node is created only with `reserve` call
- [x] When you create `Reserved` node, you actually create a root for a new tree (it's not a sub-tree). So check that tree with this `Reserved` root always has a balance less than (or equal) to the balance of the node, from which it was created (at the moment of the creation). That's quite complicated, can be skipped, but very desired to be implemented

ark0f:
> `Reserved` node seems doesn't need `lock` field, as you can't make locks directly on this node. Only on children nodes. Is it true?

It is needed with #1854

SabaunT (approved):

PR #1880, opened 2022-11-21
Increase loader ux

SabaunT:
Some changes requested by @sergeyfilyanin and born by experience of running the loader.


utils/node-loader/src/batch_pool/api/nonce.rs
```diff
@@ -53,7 +53,11 @@ fn hold_missed_nonces<'a>() -> Result<MissedNoncesGuard<'a>> {
 
 pub fn catch_missed_nonce<T>(batch_res: &GClientResult<T>, nonce: u32) -> Result<()> {
     if let Err(err) = batch_res {
-        if err.to_string().contains(utils::SUBXT_RPC_CALL_ERR_STR) {
+        if err
+            .to_string()
+            .to_lowercase()
+            .contains(utils::SUBXT_RPC_CALL_ERR_STR)
```
breathx, line 59: It's more safe to call const to lowercase also

breathx (approved):

utils/node-loader/src/batch_pool/api/nonce.rs
SabaunT: done

PR #1881, opened 2022-11-21
Vara stage 2: Rolling changes

NikVolf:


ekovalev (approved):

ukint-vs (approved):

runtime/vara/Cargo.toml
```diff
@@ -177,7 +184,6 @@ std = [
 	"sp-version/std",
 	"substrate-wasm-builder",
 	"pallet-airdrop/std",
-	"validator-set/std",
 ]
 runtime-benchmarks = [
 	"frame-benchmarking/runtime-benchmarks",
```
breathx, line 189: Don't we need to append here `runtime-benchmarks` feature for added pallets, bench weights for them and use in the runtime?

runtime/vara/src/lib.rs
```diff
@@ -959,6 +1147,9 @@ construct_runtime!(
         StakingRewards: pallet_gear_staking_rewards = 106,
         GearVoucher: pallet_gear_voucher = 107,
 
+        // TODO: Remove in stage 3
```
breathx, line 1150: Do we need to place some comment that previously validator_set pallet had index 98 and we shouldn't give the index to any other pallet in future?

breathx (approved):

PR #1891, opened 2022-11-23
Adjust payment tests

gshep:
@gear-tech/dev


breathx (approved):
Also recheck please that we never go with any consts inside our pallet payment implementation and always use proper WeightToFee conversion

PR #1896, opened 2022-11-23
Heuristics for gas calculation

MikitaS:


MikitaS:
@grishasobol @breathx PTAL

MikitaS:
@grishasobol @breathx PTAL

runtime/vara/src/tests.rs
```diff
@@ -0,0 +1,100 @@
...
+    check_weight_inbounds_interval!(instruction_weights, i64extendsi32, 1_200);
+    check_weight_inbounds_interval!(instruction_weights, i64extendui32, 650);
+    check_weight_inbounds_interval!(instruction_weights, i32wrapi64, 375);
+    check_weight_inbounds_interval!(instruction_weights, i64eq, 3_200);
+    check_weight_inbounds_interval!(instruction_weights, i64ne, 3_200);
+
+    check_weight_inbounds_interval!(instruction_weights, i64lts, 3_200);
+    check_weight_inbounds_interval!(instruction_weights, i64ltu, 3_200);
+    check_weight_inbounds_interval!(instruction_weights, i64gts, 3_200);
+    check_weight_inbounds_interval!(instruction_weights, i64gtu, 3_200);
```
grishasobol, line 77: Could we make one macros call for all of this instructions? They are all from one group. May be we can make macros with generic amount of arguments.

runtime/vara/src/tests.rs
```diff
@@ -0,0 +1,100 @@
...
+    check_weight_inbounds_interval!(instruction_weights, global_set, 2_900);
+    check_weight_inbounds_interval!(instruction_weights, memory_current, 14_200);
+
+    check_weight_inbounds_interval!(instruction_weights, i64clz, 6_100);
+    check_weight_inbounds_interval!(instruction_weights, i64ctz, 5_700);
+    check_weight_inbounds_interval!(instruction_weights, i64popcnt, 1_600);
+    check_weight_inbounds_interval!(instruction_weights, i64eqz, 3_400);
+    check_weight_inbounds_interval!(instruction_weights, i64extendsi32, 1_200);
+    check_weight_inbounds_interval!(instruction_weights, i64extendui32, 650);
+    check_weight_inbounds_interval!(instruction_weights, i32wrapi64, 375);
```
grishasobol, line 70: Also please make an issue, for instructions that have  weight less then 1000 ps, I think we should make additional investigation for that cases.

runtime/gear/src/tests.rs
```diff
@@ -0,0 +1,101 @@
...
+            println!("Weight is {weight_v} ps. Expected interval is [{interval_start}, {interval_end}] ps");
+            panic!();
+        }
+    };
+}
+
+#[test]
+fn heuristics_test() {
+    let instruction_weights = InstructionWeights::<crate::Runtime>::default();
+    println!("{instruction_weights:?}");
```
grishasobol, line 42: This println! should be removed.

runtime/vara/src/tests.rs
MikitaS: It have no sense in current context. We have only a few same instructions, generally all instructions are different.

runtime/vara/src/tests.rs
MikitaS: done
https://github.com/gear-tech/gear/issues/1916

runtime/gear/src/tests.rs
MikitaS: done

breathx (changes requested):
In fact weights for `gear` and `vara` runtimes for instructions and syscalls must match, so I'd move it into pallets/gear/tests unit tests

MikitaS:
> In fact weights for `gear` and `vara` runtimes for instructions and syscalls must match, so I'd move it into pallets/gear/tests unit tests

I'm agree with Grigory Sobol. This 2 heuristics is different and relative difference might be increased. So it should stay as it implemented now.

grishasobol (approved):

PR #1897, opened 2022-11-24
feat(pallet-gear): refactor the benchmark of `gr_random` and charge weights

### LINKED ISSUE #1892: BENCHMARK: add benchmark for gr_random_per_kb
### Problem to Solve

Also recheck other input of sequences from memory

### Possible Solution

.

### Notes

_No response_

clearloop:
Resolves #1892 

- add benchmarks for `gr_random_per_kb`

@gear-tech/dev

### DIFF
--- a/pallets/gear/src/tests.rs
+++ b/pallets/gear/src/tests.rs
@@ -4612,7 +4612,7 @@ fn program_messages_to_paused_program_skipped() {
             code,
             vec![],
             InputArgs {
-                destination: <[u8; 32]>::from(paused_program_id).into()
+                destination: <[u8; 32]>::from(paused_program_id)
             }
             .encode(),
             50_000_000_000u64,


ukint-vs (approved):
Looks fine. Btw we need to check `schedule.limits.subject_len` in pallet_gear/ext. If this is not related to this pr then I can take this.

grishasobol (changes requested):
Review will be soon

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -718,6 +718,17 @@ benchmarks! {
         verify_process(res.unwrap());
     }
 
+    gr_random_per_kb {
+        let n in 0 .. T::Schedule::get().limits.subject_len / 1024;
```
grishasobol, line 722: I cannot understand. If `subject_len` == 1024, then we iterate over 0 .. 1 ? How do you have iteration  `n` is `[0, 1024]` in `weights.rs` ?

pallets/gear/src/schedule.rs
```diff
@@ -660,6 +664,7 @@ impl<T: Config> Default for HostFnWeights<T> {
             gr_block_height: cost_batched!(gr_block_height),
             gr_block_timestamp: cost_batched!(gr_block_timestamp),
             gr_random: cost_batched!(gr_random),
+            gr_random_per_byte: cost_batched!(gr_random_per_kb),
```
grishasobol, line 667: You must use here `cost_byte_batched`

pallets/gear/src/schedule.rs
```diff
@@ -512,7 +515,7 @@ impl Default for Limits {
             // 4k function pointers (This is in count not bytes).
             table_size: 4096,
             br_table_size: 256,
-            subject_len: 32,
+            subject_len: 1024,
```
grishasobol, line 518: @ukint-vs  why we doesn't use it anywhere in `gr_random` syscall?

grishasobol: @ukint-vs may be we should use old small value `32` and add restriction in `gr_random` ? Then this benchmarks has no sence.

ukint-vs: > @ukint-vs why we doesn't use it anywhere in `gr_random` syscall?

I added a comment above.

ukint-vs:
> > we can leave only one benchmark for the maximum subject length of 32 and check this limit in ext.
> 
> do you mean
> 
> * remove `gr_random_per_kb`
> * refactor the benchmark of `gr_random` with fixed  `subject_len = 32`
> * check subject_len in `funcs.random`

Yes.

> why we limit the length of subject or not hmmm...

if the length of the subject affects the speed of execution, then it is illogical to leave the possibility of having a large length and complicate the support of random syscall.

core-processor/src/configs.rs
```diff
@@ -60,6 +62,7 @@ impl Default for AllocationsConfig {
             alloc_cost: ALLOC_COST,
             mem_grow_cost: MEM_GROW_COST,
             load_page_cost: LOAD_PAGE_COST,
+            random_subject_len: 32,
```
breathx, line 65: Why?

breathx: It should be the same as for payload

breathx: and ofc not in allocations config, even if exist

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -708,9 +708,10 @@ benchmarks! {
...
     gr_random {
-        let r in 0 .. API_BENCHMARK_BATCHES;
+        let n in 0 .. API_BENCHMARK_BATCHES;
+        let l in 0 .. T::Schedule::get().limits.subject_len;
         let mut res = None;
-        let exec = Benches::<T>::gr_random(r)?;
+        let exec = Benches::<T>::gr_random(n, l)?;
     }: {
         res.replace(run_process(exec));
     }
```
breathx, line 717: it should be separated into gr_random and gr_random_per_byte, isn't it?

breathx (changes requested):

breathx:
> if the length of the subject affects the speed of execution, then it is illogical to leave the possibility of having a large length and complicate the support of random syscall.

But if I want to pass it? I want to depend on my state, for example and avoid hash fns inside wasm at all?

ukint-vs:
> > if the length of the subject affects the speed of execution, then it is illogical to leave the possibility of having a large length and complicate the support of random syscall.
> 
> But if I want to pass it? I want to depend on my state, for example and avoid hash fns inside wasm at all?

No problem, we can increase the maximum size to 1024 bytes and implement per-byte bench.

core-backend/sandbox/src/funcs.rs
```diff
@@ -455,13 +460,13 @@ where
     pub fn random(ctx: &mut Runtime<E>, args: &[Value]) -> SyscallOutput {
         sys_trace!(target: "syscall::gear", "random, args = {}", args_to_str(args));
 
-        let (subject_ptr, len, bn_random_ptr): (_, u32, _) = args.iter().read_3()?;
+        let (subject_ptr, bn_random_ptr): (_, _) = args.iter().read_2()?;
 
         ctx.run(|ctx| {
-            let mut subject = RuntimeBuffer::try_new_default(len as usize)?;
-            ctx.read_memory_into_buf(subject_ptr, subject.get_mut())?;
+            let mut subject =
```
breathx, line 466: We don't need runtime buffer here for hash case

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -94,6 +94,7 @@ use sp_runtime::{
 use sp_std::prelude::*;
 
 const MAX_PAYLOAD_LEN: u32 = 16 * 64 * 1024;
+const MAX_PAYLOAD_LEN_PER_KB: u32 = MAX_PAYLOAD_LEN / 1024;
```
breathx, line 97:
```suggestion
const MAX_PAYLOAD_LEN_KB: u32 = MAX_PAYLOAD_LEN / 1024;
```

core-backend/sandbox/src/funcs.rs
```diff
@@ -455,13 +460,13 @@ where
...
+                RuntimeBuffer::try_from(ctx.read_memory_as::<Hash>(subject_ptr)?.to_vec())?;
```
breathx, line 467: I'd go with read_memory_decoded here

breathx (approved):

core-backend/sandbox/src/funcs.rs
grishasobol: Yes, checking constant size buffers is overwork.

pallets/gear/src/benchmarking/mod.rs
grishasobol: Or in_kb

grishasobol (approved):

PR #1907, opened 2022-11-24
fix for new rust clippy version

grishasobol:


node/authorship/src/authorship.rs
```diff
@@ -306,7 +306,7 @@ where
             None,
             Box::pin(async move {
                 // leave some time for evaluation and block finalization (33%)
-                let deadline = (self.now)() + max_duration - max_duration / 3;
+                let deadline = (self.now)() + (max_duration / 3) * 2;
```
shamilsan, line 309: Isn't it better to multiply first, then divide (not to lose the remainder after integer division)?

```suggestion
                let deadline = (self.now)() + max_duration * 2 / 3;
```

shamilsan (approved):
Your solution seems to be better than the other one: https://github.com/gear-tech/gear/pull/1899/commits/bba75c537c0eb334c00a03eda9a44c107aa01a79

node/authorship/src/authorship.rs
grishasobol: I think it's better to divide first in order to avoid `max_duration * 2` overflow

grishasobol: Also It's better for deadline to be smaller than to be bigger I think

breathx: Any reason to do this instead of just ignore clippy on the case?
Any under/overflows aren't possible here and nothing changes if we look at amount of instructions

grishasobol: I don't know. I think better to avoid clippy ignore, to have less complexity in code.

Issue #1778, opened 2022-11-06
Add gas and value consumption into transaction fee


breathx:
### Problem

We can send transaction successfully without having balance to by applied amount of gas

SabaunT:
Closing this @breathx ?

PR #1937, opened 2022-12-01
Validate TX with info about gas and message value

SabaunT:
Resolves #1778 .

The rationale little differs form what is stated in the PR. We do not withdraw gas and message value when `pre_dispatch` happens as we will have huge obstacles in refunding on transaction error. That's why we only check funds sufficiency in the transaction pool (on `validate` and `pre_dispatch`).

TODO:
- [x] Test tx validation
- [x] Make funds reservations on `pallet_gear` calls infallible (using `DefensiveResult::map_err`) 
- [x]  ̶P̶r̶o̶v̶i̶d̶e̶ ̶g̶a̶s̶ ̶a̶n̶d̶ ̶m̶e̶s̶s̶a̶g̶e̶ ̶v̶a̶l̶u̶e̶ ̶t̶o̶ ̶f̶e̶e̶s̶ ̶i̶n̶f̶o̶ ̶i̶n̶ ̶`̶q̶u̶e̶r̶y̶_̶f̶e̶e̶_̶d̶e̶t̶a̶i̶l̶s̶`̶ ̶a̶n̶d̶ ̶`̶q̶u̶e̶r̶y̶_̶i̶n̶f̶o̶`̶  (separated PR -> #1944 )

PR #1944, opened 2022-12-05
Add gas and value to tx fee

techraed:
Continuation of the #1778 .

The rationale for this is to tell to `gear` users more about their expenses when calling `send_message`, `upload_program`, `create_program`, `send_reply`. More precisely, add gas and value info to current fees, so users won't be mistaken by expecting the call to be executed successfully.    

@gear-tech/dev

### DIFF
--- a/core-backend/sandbox/src/funcs.rs
+++ b/core-backend/sandbox/src/funcs.rs
@@ -70,11 +70,7 @@ pub enum FuncError<E: Display> {
     SyscallErrorExpected,
     #[display(fmt = "Terminated: {_0:?}")]
     Terminated(TerminationReason),
-    #[display(
-        fmt = "Cannot take data by indexes {:?} from message with size {}",
-        _0,
-        _1
-    )]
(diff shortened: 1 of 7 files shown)


SabaunT (commented):
Need to decide whether changes in `query_fee_details` are necessary.
From one side it's hard to include gas and value info to the details data, because we can break compatibility with the front-end (`polkadot-js/apps`) in the future, i.e. we don't know whether parity have plans to use this call internally somewhere.
From the other side, it's an available rpc call, so proper information must be given. Also I didn't find any internal usage of the `query_fee_details` right now in `polkadot-js/apps`.

What do you think? @breathx @ekovalev @shamilsan @gshep

breathx (approved):

gshep (changes requested):
Looks good :+1:  but some changes are required.

As far as I understand there are no disagreements with this PR for improvements of user experience (@ekovalev ?). So I would like to see `TryExtract` & `MessageResources...` there and point the merge request to the main branch.

Then we could search for some consensus on #1937  independently.

SabaunT:
After some internal discussions we decided to close it

PR #1958, opened 2022-12-06
Remove `unsafe` from `extern` functions

shamilsan:
- Removed `unsafe` from exported `extern` functions as it breaks Rust idiom (keep unsafe blocks as small as possible).

@gear-tech/dev

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -48,8 +48,9 @@ Here is a minimal program for a classic ping-pong contract:
 use gstd::{ext, msg};
 
 #[no_mangle]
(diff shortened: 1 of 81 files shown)


gshep (approved):
Good job! :100: 

I would also notice that using pattern
```rust
let mut_var = unsafe { &mut VAR; }
// ...
*mut_var = func();
```
several more examples could be simplified.

shamilsan:
> I would also notice that using pattern 
> ...
> several more examples could be simplified.

Let's do it later if you don't mind 😄

PR #1960, opened 2022-12-07
feat(gtest): introduce delayed messages in gtest

### LINKED ISSUE #1781: Delayed messages support in `gtest`
### Problem to Solve

Just mock the logic inserting messages from some pool while skipping blocks

clearloop:
Resolves #1781 

The lifecycle of delayed messages

- system.send_delayed_dispatch
- add delayed_dispatch into queue
- after spend_blocks ( whatever how many blocks ), delayed dispatches got processed at last, return a vector of results.

@gear-tech/dev


gtest/src/manager.rs
```diff
@@ -400,6 +429,31 @@ impl ExtManager {
         }
     }
 
+    /// Insert delayed dispatch into task pool.
+    pub(crate) fn send_delayed_dispatch(&mut self, delay: u32, dispatch: DelayedDispatch) {
+        let expected_block = self.block_info.height + delay;
+        if let Some(queue) = self.scheduled_dispatches.get_mut(&expected_block) {
```
breathx, line 435: mut_or_default?

breathx: and push afterward

gtest/src/manager.rs
```diff
@@ -204,6 +204,34 @@ impl Program {
     }
 }
 
+/// Delayed dispatch with original message id and reservation info.
+#[derive(Debug)]
+pub(crate) struct DelayedDispatch {
+    pub message_id: MessageId,
+    pub dispatch: Dispatch,
+    pub reservation: Option<ReservationId>,
```
breathx, line 212: why do we need this reservation? what's the aim of it?
delayed dispatches on pallet have existing gas node once created

gtest/src/manager.rs
```diff
@@ -204,6 +204,34 @@ impl Program {
     }
 }
 
+/// Delayed dispatch with original message id and reservation info.
+#[derive(Debug)]
+pub(crate) struct DelayedDispatch {
+    pub message_id: MessageId,
+    pub dispatch: Dispatch,
```
breathx, line 211: dispatch contains message id

breathx: See no reason for intro such struct at all

gtest/src/mailbox.rs
```diff
@@ -357,4 +357,38 @@ mod tests {
...
+            source_user_id,
+            destination_user_id,
+            message_payload.encode().try_into().unwrap(),
+            Default::default(),
+            0,
+            None,
+        );
+
+        let blocks = 10;
+        system.0.borrow_mut().send_delayed_dispatch(
```
breathx, line 383: users never allowed to send delayed dispatches atm
it's only under the hood api for programs
and original idea of the task is to support this functionality in `manager`'s implementation of `JournalHandler`

gtest/src/mailbox.rs
```diff
@@ -357,4 +357,38 @@ mod tests {
...
+            blocks,
+            (message_id, Dispatch::new(DispatchKind::Handle, message)).into(),
+        );
+
+        let mailbox = system.get_mailbox(destination_user_id);
+        assert!(!mailbox.contains(&log));
+
+        system.spend_blocks(blocks);
```
breathx, line 391: @LouiseMedova how do you see delayed processing while spending blocks? maybe it should return some info enum?

like

- spend_blocks called for 10 blocks
- in 3 blocks occurs delayed message, got processed, so spend block interrupts with returning RunResult + current bn maybe?

gtest/src/manager.rs
```diff
@@ -219,6 +247,7 @@ pub(crate) struct ExtManager {
     pub(crate) opt_binaries: BTreeMap<CodeId, Vec<u8>>,
     pub(crate) meta_binaries: BTreeMap<CodeId, Vec<u8>>,
     pub(crate) dispatches: VecDeque<StoredDispatch>,
+    pub(crate) scheduled_dispatches: HashMap<u32, VecDeque<DelayedDispatch>>,
```
breathx, line 250: no guarantees in order here: I'd go with vector, inserting dispatches randomly, cause it depends on hashes and rocksdb/sp_io impl on-chain
Also need your opinion @LouiseMedova

breathx (changes requested):

gtest/src/mailbox.rs
LouiseMedova: I think it is not necessary to interrupt `spend_blocks`, and it's better to handle the `RunResult` when the  indicated number of blocks has passed

shamilsan:
@clearloop stale PR? Let's complete it.

gtest/src/manager.rs
```diff
@@ -223,6 +223,7 @@ pub(crate) struct ExtManager {
     pub(crate) wait_list: BTreeMap<(ProgramId, MessageId), StoredDispatch>,
     pub(crate) wait_init_list: BTreeMap<ProgramId, Vec<MessageId>>,
     pub(crate) gas_limits: BTreeMap<MessageId, Option<u64>>,
+    pub(crate) delayed_dispatches: Vec<Dispatch>,
```
breathx, line 226: Why do you need to process all of the delayed dispatches? It should be only if we already spent the expected for dispatch block number

breathx (changes requested):

gtest/src/system.rs
```diff
@@ -79,19 +85,34 @@ impl System {
         self.0.borrow_mut().run_dispatch(dispatch)
     }
 
-    pub fn spend_blocks(&self, amount: u32) {
+    pub fn send_delayed_dispatch(&self, delay: u32, dispatch: Dispatch) {
```
breathx, line 88: users and external api aren't allowed to send delayed messages. it's program-only functionality
```suggestion
```

gtest/src/system.rs
```diff
@@ -79,19 +85,34 @@ impl System {
...
+                manager.random_data = (random.to_vec(), manager.block_info.height + 1);
+            }
+        };
+
+        (manager.block_info.height..manager.block_info.height + amount)
+            .map(|bn| {
+                check_epoch(&mut manager);
+                manager.block_info.height += 1;
+                manager.block_info.timestamp += 1000;
+                manager.process_delayed_dispatches(bn + 1)
```
breathx, line 112: why `bn + 1`?

gtest/src/system.rs
```diff
@@ -79,19 +85,34 @@ impl System {
...
-                INITIAL_RANDOM_SEED + (manager.block_info.height / EPOCH_DURATION_IN_BLOCKS) as u64,
-            );
-            let mut random = [0u8; 32];
-            rng.fill_bytes(&mut random);
-
-            manager.random_data = (random.to_vec(), manager.block_info.height + 1);
-        }
-        manager.block_info.height += amount;
-        manager.block_info.timestamp += 1000 * amount as u64;
+        let check_epoch = |manager: &mut RefMut<ExtManager>| {
```
breathx, line 94: we can move it into separate fn

gtest/src/mailbox.rs
```diff
@@ -358,4 +358,44 @@ mod tests {
...
+            source_user_id,
+            destination_user_id,
+            message_payload.encode().try_into().unwrap(),
+            Default::default(),
+            0,
+            None,
+        );
+
+        let unexpected_blocks = 5;
+        let expected_blocks = 10;
```
breathx, line 384: misleading naming

breathx (changes requested):

breathx (changes requested):

gtest/src/program.rs
```diff
@@ -353,32 +353,40 @@ impl<'a> Program<'a> {
         )
     }
 
-    pub fn send<ID: Into<ProgramIdWrapper>, C: Codec>(&self, from: ID, payload: C) -> RunResult {
-        self.send_with_value(from, payload, 0)
+    pub fn send<ID: Into<ProgramIdWrapper>, C: Codec>(
+        &self,
+        from: ID,
+        payload: C,
+        delay: Option<u32>,
```
breathx, line 360: that's exactly programs interface. is we discussed before, users have no ability to send delayed messages. it's programs specific feature atm

breathx: Furthermore, delay also shouldn't be optional anywhere in code: in case of absence of delay (when you want to send something instantly) delay just equals 0

breathx (changes requested):

breathx:
@clearloop upd pr pls

ukint-vs (approved):

gtest/src/manager.rs
```diff
@@ -286,6 +287,41 @@ impl ExtManager {
...
+    /// Insert message into the delayed queue.
+    pub(crate) fn send_delayed_dispatch(&mut self, dispatch: Dispatch, delay: u32) {
+        self.delayed_dispatches
+            .entry(delay)
+            .or_default()
+            .push(dispatch)
+    }
+
+    /// Process all delayed dispatches.
+    pub(crate) fn process_delayed_dispatches(&mut self, delay: u32) -> Vec<RunResult> {
```
breathx, line 299:
```suggestion
    pub(crate) fn process_delayed_dispatches(&mut self, block_number: u32) -> Vec<RunResult> {
```

gtest/src/manager.rs
```diff
@@ -286,6 +287,41 @@ impl ExtManager {
         self.id_nonce
     }
 
+    /// Insert message into the delayed queue.
+    pub(crate) fn send_delayed_dispatch(&mut self, dispatch: Dispatch, delay: u32) {
```
breathx, line 291:
```suggestion
    pub(crate) fn send_delayed_dispatch(&mut self, dispatch: Dispatch, bn: u32) {
```

gtest/src/manager.rs
```diff
@@ -797,9 +833,14 @@ impl JournalHandler for ExtManager {
         &mut self,
         _message_id: MessageId,
         dispatch: Dispatch,
-        _delay: u32,
+        delay: u32,
         _reservation: Option<ReservationId>,
     ) {
+        if delay > 0 {
+            self.send_delayed_dispatch(dispatch, delay);
```
breathx, line 840:
```suggestion
            self.send_delayed_dispatch(dispatch, current_bn.saturating_add(delay)); // it should be bn at which message should be sent
```

breathx (changes requested):

breathx (approved):
To be finished

gtest/src/mailbox.rs
```diff
@@ -358,4 +358,49 @@ mod tests {
 
         assert_eq!(system.balance_of(receiver_id), 1000);
     }
+
+    #[test]
+    fn delayed_dispatches_works() {
```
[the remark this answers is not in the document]
breathx: cause gtest provides only on-chain-like behaviour as external api

PR #1967, opened 2022-12-08
Fix the panic after init message `prepare` failure

breathx:
Tests `non_existent_code_id_zero_gas` and `waited_with_zero_gas` are panicking now. Others are just for confidence.


pallets/gear/src/manager/journal.rs
```diff
@@ -140,10 +143,22 @@ where
                 // dequeued. The other case is async init.
                 wake_waiting_init_msgs(program_id);
 
-                self.clean_reservation_tasks(program_id);
+                // If we run into `InitFailure` after real execution (not
+                // prepare or precharge) processor methods, then we are
```
gshep, line 147:
```suggestion
                // prepare or precharge processor methods), then we are
```

pallets/gear/src/manager/mod.rs
```diff
@@ -268,11 +268,20 @@ where
...
-        let active_program =
-            common::get_active_program(program_id.into_origin()).unwrap_or_else(|e| {
-                unreachable!("`exit` can be called only from active program: {}", e)
-            });
+    fn clean_reservation_tasks(&mut self, program_id: ProgramId, maybe_inactive: bool) {
+        let maybe_active_program = common::get_active_program(program_id.into_origin());
+
+        if maybe_active_program.is_err() && maybe_inactive {
+            return;
+        };
```
gshep, line 276: unnecessary semicolon

gshep (approved):
Good catch! :gear:

pallets/gear/src/manager/journal.rs
breathx: Will be applied next pr just to avoid waiting one more ci

PR #1968, opened 2022-12-08
feat(core-backend): introduce `gr_gas_limit`

### LINKED ISSUE #1796: `gr_gas_limit`
### Problem to Solve

Introduce syscall to get message gas limit (don't mess with `gr_gas_avaiable`)

### Possible Solution

.

### Notes

_No response_

clearloop:
Resolves #1796 

- introduce gas_limit in ext
- register gas_limit as syscall
- introduce gas_limit in `gcore` and provide tests

@gear-tech/dev


gcore/src/exec.rs
```diff
@@ -220,6 +220,33 @@ pub fn gas_available() -> u64 {
     gas
 }
 
+/// Get the current amount of gas limit for execution.
```
breathx, line 223: Incorrect comment: it's about `gas_available`
`gr_gas_limit` should return (consider for gasless sendings) amount of gas message was initially sent with

breathx: I'd go with something like `enum InitialGas { GasFull (u64), Gasless (u64) }`

pallets/gear/src/benchmarking/tests/syscalls_integrity.rs
```diff
@@ -829,6 +830,20 @@ where
...
+    T::AccountId: Origin,
+{
+    run_tester::<T, _, _, T::AccountId>(|_, _| {
+        let mp = Kind::GasLimit(SEND_MESSAGE_GAS_LIMIT).encode().into();
+
+        (TestCall::send_message(mp), None::<DefaultPostCheck>)
+    })
+}
+
+const SEND_MESSAGE_GAS_LIMIT: u64 = 50_000_000_000;
```
breathx, line 845: it's better to move at the beginning of the file

breathx (changes requested):
This approach will work incorrect in wait-wake logic, but maybe it's okay:
@LouiseMedova which variant do you prefer below?
- return initial gas limit value while message was generated
- return initial gas limit once message got dequeued

gcore/src/exec.rs
LouiseMedova: I think the first variant is better

shamilsan:
@clearloop stale PR?

breathx (changes requested):
As we decided to go with option where we're returning amount of gas message initially was created with, such changes requires adoption of gas tree anyway, so for me current realisation seems not working

Issue #1971, opened 2022-12-11
Switch to `wasm-opt ` rust bindings


ukint-vs:
### Proposal

Switch from using wasm-opt from command line to rust bindings https://crates.io/crates/wasm-opt

PR #1972, opened 2022-12-12
Pin `substrate` to `v0.9.33` tag

ukint-vs:
This allows us to match our codebase to stable release tags of substrate.

- `wasm-opt v111` disabled cuz its breaks wasm artifacts. (Error: Unknown opcode 192) (#1971).
- `wasm-proc` -p arg reduced to last arg which can contain multiple values.

@gear-tech/dev

### DIFF
--- a/utils/wasm-proc/src/main.rs
+++ b/utils/wasm-proc/src/main.rs
@@ -91,10 +91,6 @@ enum Error {
 
 #[derive(Debug, clap::Parser)]
 struct Args {
-    /// Path to WASMs, accepts multiple files
-    #[clap(short, long, value_parser, multiple = true)]
-    path: Vec<String>,
-
     /// Don't generate `.meta.wasm` file with meta functions
     #[clap(long)]
     skip_meta: bool,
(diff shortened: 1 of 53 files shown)


gshep (approved):
Great! :100:

utils/wasm-proc/src/main.rs
```diff
@@ -173,14 +173,15 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-        log::info!(
-            "wasm-opt: {} {} Kb -> {} Kb",
-            res.dest_wasm.display(),
-            res.original_size,
-            res.optimized_size
-        );
+        // Issue (#1971)
+        // let res = gear_wasm_builder::optimize::optimize_wasm(file.clone(), "s", true)?;
+
+        // log::info!(
```
shamilsan, line 179: Doesn't `wasm-opt` work at all?

shamilsan (approved):
Amazing 🆒 
Personal thanks for migrating to `clap` v4.

utils/wasm-proc/src/main.rs
ukint-vs: doesn't work for me locally

gshep: what version did you try?

ukint-vs: > what version did you try?

wasm-opt version 111

shamilsan: v111 doesn't work for me too, but v110 does.

PR #1978, opened 2022-12-14
Update Cargo.lock

gshep:
@gear-tech/dev


breathx:
How can it be possible if we use --locked on ci?

PR #1981, opened 2022-12-14
Fix panic on wait + exit across executions

### LINKED ISSUE #1951: Signal on terminated program
What will happen in case when:
program has messages A and B
A creates reservation and goes into waitlist
B exits
Will system reservation disappear/remove on execution if waked (e.g. by schedule)/removed from waitlist - will signal be generated on already exited/terminated program?
I can trigger same logic with async init in handle_reply also

_Originally posted by @breathx in https://github.com/gear-tech/gear/pull/1854#discussion_r1036223472_

ark0f:
Resolves #1951


pallets/gear/src/manager/journal.rs
```diff
@@ -178,7 +178,7 @@ where
 
                 DispatchStatus::Failed
             }
-            CoreDispatchOutcome::NoExecution => {
+            NoExecution => {
```
breathx, line 181: lets go with full specification of enums always

pallets/gear/src/tests.rs
```diff
@@ -8008,6 +8067,121 @@ fn system_reservation_gas_allowance_rollbacks() {
...
+
+        run_to_next_block(None);
+
+        // nothing happened after
+        assert!(MailboxOf::<Test>::is_empty(&USER_1));
+    });
+}
+
+#[test]
+fn signal_on_terminated_program() {
```
breathx, line 8145: scheme of the test should be:
- msg A is going to init program
- in init msg A sends msg B and waits for reply
- user replies on message B with msg C
- msg C goes into handle_reply section
- in handle reply msg C should: wake msg A, make reservation and wait
- once msg A processed after wake it panics, triggering termination

breathx (approved):

PR #1983, opened 2022-12-15
feat(gcore): introduce `read_unchecked` from `read`

### LINKED ISSUE #1859: Provide unchecked ops at the backend
### Problem to Solve

E.g. each time you call gr_read you expect it to be correct and you don't want to write err len

### Possible Solution

Provide this wrappers dependent on error_len ptr value 

### Notes

_No response_

clearloop:
Resolves #1859 

- add `gr_read_unchecked` based on `gr_read`
- export `gr_read_unchecked` to `gcore`

@gear-tech/dev

### DIFF
--- a/core-backend/sandbox/src/funcs.rs
+++ b/core-backend/sandbox/src/funcs.rs
@@ -289,7 +289,6 @@ where
         let (at, len, buffer_ptr, err_len_ptr): (_, _, u32, _) = args.iter().read_4()?;
 
         ctx.run(|ctx| {
-            let write_err_len = ctx.register_write_as(err_len_ptr);
             let res = ctx.ext.read(at, len);
             let length = match res.into_ext_error(&mut ctx.err)? {
                 Ok(buf) => {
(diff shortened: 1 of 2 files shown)


core-backend/sandbox/src/funcs.rs
```diff
@@ -338,6 +338,14 @@ where
         })
     }
 
+    pub fn read_unchecked(ctx: &mut Runtime<E>, args: &[Value]) -> SyscallOutput {
```
breathx, line 341: we don't need separate syscall for this
use already existing one by playing around value of "resulting" ptr arg: if u32::MAX than mean as infallible

breathx (changes requested):

shamilsan:
I don't see `gr_read_unchecked` in this PR. Maybe it is worth changing PR's title and description?

SabaunT (approved):
lgtm, but would wait for @breathx review

breathx (changes requested):
It's okay with gcore, but such fn required in gstd also
Going further, where is following changes for backend to ignore errors if error_len ptr is u32::MAX?

breathx (changes requested):

breathx (changes requested):

breathx:
#2083 merged, to be rebased and refactored as discussed personally

PR #1989, opened 2022-12-17
Re-check gas charge/refunding on backend/processor levels

### LINKED ISSUE #1768: Recheck refunding mechanism
### File Location(s)

_No response_

### Proposal

Refund should never return gas into gas allowance counter (recheck for globals and inner core counters)

SabaunT:
Resolves #1768.

Found problems:
- `free` calls refund in case of error after freeing requested page. Fixed by checking inner `free` call result.
- allowance and gas counters are not zeroed in case of charge errors. Now if charging for the sys-call (`charge_runtime_gas`) fails, counters are zeroed.
- `read` calls doesn't charge for the sys-call in some cases, however must always charge. Fixed by unwrapping the resultf with default max value.
- no charge for sending message in `create_program` call. Fixed by introducing the corresponding charge.
- no payload size checks on sandbox env. Actually there was a check with `RuntimeBuffer` type, but that's larger than acceptable payload. Fixed by introducing payload size checks.

There's one found really serious issue that was fixed in parallel here: #1988. All wasmi calls are poisoned by the following problem: if error occurs during sys-call execution, global counters won't be updated. That is so for `call_fallible`, `call_infallible` and also all the fns where `call.update_globals()` is called directly (like `alloc`, `free`, `read`, different waits and etc.).


core-backend/sandbox/src/funcs.rs
```diff
@@ -800,6 +823,7 @@ where
         let (data_ptr, data_len): (_, u32) = args.iter().read_2()?;
 
         ctx.run(|ctx| {
+            // Todo shall we use Payload here?
```
SabaunT, line 826: @breathx @ark0f @gshep what do you think?

core-backend/sandbox/src/funcs.rs
```diff
@@ -235,11 +243,13 @@ where
...
         ctx.run(|ctx| {
-            let payload = ctx.read_memory(payload_ptr, len)?;
+            let payload = (Payload::max_len() >= (len as usize))
+                .then_some(ctx.read_memory(payload_ptr, len))
+                .ok_or(FuncError::PayloadSize(PayloadSizeError))?;
 
             let len = ctx
                 .ext
-                .send_push(handle, &payload)
+                .send_push(handle, &payload?)
```
breathx, line 252: why not double "?" before?

core-backend/sandbox/src/funcs.rs
breathx: In fact argument for debug should be similarly limited as anything else

core-backend/wasmi/src/funcs/mod.rs
```diff
@@ -56,7 +56,7 @@ pub enum FuncError<E: Display> {
     #[display(fmt = "{_0}")]
     Core(E),
     #[display(fmt = "Runtime Error")]
-    HostError,
+    WrongInstrumentation,
```
breathx, line 59: fmt message and error itself are incompatible now

core-backend/wasmi/src/funcs/mod.rs
```diff
@@ -1378,6 +1378,7 @@ where
             move |caller: Caller<'_, HostState<E>>, string_ptr: u32, len: u32| -> EmptyOutput {
                 let mut caller = CallerWrap::prepare(caller, forbidden)?;
 
+                // todo Shall we use Payload?
```
breathx, line 1381: same for sandbox, despite the fact that we use wasmi only for tests

core-processor/src/ext.rs
```diff
@@ -611,13 +612,23 @@ impl EnvExt for Ext {
...
+        let max_payload_size: u32 = MAX_PAYLOAD_SIZE
+            .try_into()
+            .expect("statically asserted to be in u32 bounds");
+        let (size, with_error) = self
             .size()?
             .try_into()
-            .map_err(|_| MessageError::IncomingPayloadTooBig)?;
-
+            .map(|v| (v, false))
+            // This must not happen. because max size of the payload
```
breathx, line 622: should we assert it so? :/

core-processor/src/ext.rs
```diff
@@ -800,7 +811,9 @@ impl EnvExt for Ext {
             packet.salt().len() as u32,
         ))?;
 
+        self.check_forbidden_call(packet.destination())?;
```
breathx, line 814: I'd rename the method: `check forbidden dest` or something else

utils/node-loader/src/utils.rs
```diff
@@ -16,9 +16,9 @@ use std::{
     time::{Duration, SystemTime, UNIX_EPOCH},
 };
 
-/// subxt's GenericError::Rpc::RequestError::RestartNeeded
+/// subxt's GenericError::Rpc::RequestError::RestartNeeded // TODO STRING IS MISSED IN A NEW SUBXT
```
breathx, line 19: ref todos?

breathx: and below also

breathx (approved):

core-backend/sandbox/src/funcs.rs
```diff
@@ -1063,8 +1094,14 @@ where
                 hash: code_id,
                 value,
             } = ctx.read_memory_as(cid_value_ptr)?;
-            let salt = ctx.read_memory(salt_ptr, salt_len)?.try_into()?;
-            let payload = ctx.read_memory(payload_ptr, payload_len)?.try_into()?;
+            let salt = (Payload::max_len() >= (salt_len as usize))
```
ark0f, line 1097: Consider introduce `Payload::try_from_vec`

ark0f (changes requested):

core-processor/src/ext.rs
SabaunT: I'd live without an assertion in the back-end until proper types and refactoring is introduced (discussed with you several weeks ago).

core-backend/wasmi/src/funcs/mod.rs
SabaunT: changed

core-backend/sandbox/src/funcs.rs
SabaunT: done

core-processor/src/ext.rs
```diff
@@ -331,7 +332,7 @@ impl Ext {
         Ok(())
     }
 
-    fn check_forbidden_call(&mut self, id: ProgramId) -> Result<(), ProcessorError> {
+    fn check_forbidden_dest(&mut self, id: ProgramId) -> Result<(), ProcessorError> {
```
ark0f, line 335:
```suggestion
    fn check_forbidden_destination(&mut self, id: ProgramId) -> Result<(), ProcessorError> {
```
Please

ark0f (approved):

core-processor/src/ext.rs
SabaunT: done

core-processor/src/ext.rs
SabaunT: done

Issue #1381, opened 2022-08-28
Stabilise pausing programs interface 


breathx:
### Problem

At this lines we drain all waitlist for program, without unlocking value and gas limit from them, also not removing task from task pool, that will lead to runtime panic once attached block for one of them will come:
https://github.com/gear-tech/gear/blob/654a0b8551fa398142c96bc4050aa2567e686b43/pallets/gear-program/src/pause.rs#L71-L82

This behaviour should be changed for correct one with truth wake of the messages.
I also think that ability to restore waitlist should be deleted due to economical and ways-to-abuse queue reasons.

### Steps

// .. // 

### Relevant Log Output

<details><summary>Click to expand/collapse</summary>
<p>

```
<Paste your log here>
```

</p>
</details>

breathx:
Protocol doesn't support program pauses or expirations; message rent is the only exception

PR #2017, opened 2022-12-23
Add ProgramStorage

### LINKED ISSUE #788: Move `common::Program` & corresponding entities to gear-program pallet
### File Location(s)

_No response_

### Proposal

This is a first step of #578

gshep:
Resolves #788 

- remove suspend\resume program API. It will be revamped in #1381 
- introduce ProgramStorage. Adjust source code to the new entity
- refactor PageBuf to workaround limit in polkadot.js/metadata

@gear-tech/dev


core/src/memory.rs
```diff
@@ -68,68 +63,83 @@ static_assertions::const_assert_eq!(WASM_PAGE_SIZE % GEAR_PAGE_SIZE, 0);
 static_assertions::const_assert_eq!(WASM_PAGE_SIZE % PAGE_STORAGE_GRANULARITY, 0);
 static_assertions::const_assert_eq!(PAGE_STORAGE_GRANULARITY % GEAR_PAGE_SIZE, 0);
 
+/// Alias for inner type of page buffer.
+pub type PageBufInner = LimitedVec<u8, (), GEAR_PAGE_SIZE>;
```
grishasobol, line 67: Why do we need to do this? Is less performance effective .

runtime-interface/src/lib.rs
```diff
@@ -39,10 +39,10 @@ pub use sp_std::{convert::TryFrom, result::Result, vec::Vec};
 pub trait GearRI {
     /// Init lazy-pages.
     /// Returns whether initialization was successful.
-    fn init_lazy_pages() -> bool {
+    fn init_lazy_pages(pages_final_prefix: [u8; 32]) -> bool {
```
grishasobol, line 42: You need to make separate function `init_lazy_pages` with #[version(2)]. And leave this one without changes. in `gear-lazy-pages` crate you need to support both versions: for old runtimes and for new runtimes

grishasobol (changes requested):

SabaunT (approved):
nice to see desired changes being implemented! :+1:

runtime-interface/src/lib.rs
gshep: Good catch, thanks! However we will drop everything after this PR, won't we? @breathx

common/src/program_storage.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    fn get_program(program_id: ProgramId) -> Option<Program> {
+        Self::ProgramMap::get(&program_id)
+    }
+
+    fn program_exists(program_id: ProgramId) -> bool {
+        Self::ProgramMap::contains_key(&program_id)
+    }
+
+    fn update_active_program<F, ReturnType>(
```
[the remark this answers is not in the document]
gshep: Yep, it is required since sometimes Program itself needs to be changed, for example a transition `Program::Active` -> `Program::Exited`

core/src/memory.rs
gshep: Please check comments on line 73 there.

What about performance: the vec is always preallocated before use so it shouldn't be impaired.

runtime-interface/src/lib.rs
grishasobol: And vara?

lazy-pages/src/lib.rs
```diff
@@ -124,19 +124,33 @@ pub enum InitializeForProgramError {
...
     wasm_mem_size: WasmPageNumber,
     stack_end_page: Option<WasmPageNumber>,
-    program_prefix: Vec<u8>,
+    program_id: [u8; 32],
 ) -> Result<(), InitializeForProgramError> {
     use InitializeForProgramError::*;
     LAZY_PAGES_CONTEXT.with(|ctx| {
         let mut ctx = ctx.borrow_mut();
+
+        let prefix = match ctx.program_storage_prefix.take() {
```
grishasobol, line 141: Please add a comment than here, that nobody can move this after
```rust
*ctx = LazyPagesExecutionContext::default();
```

core/src/memory.rs
```diff
@@ -68,68 +63,83 @@ static_assertions::const_assert_eq!(WASM_PAGE_SIZE % GEAR_PAGE_SIZE, 0);
...
-impl PageBuf {
-    /// Tries to transform vec<u8> into page buffer.
-    /// Makes it without any reallocations or memcpy: vector's buffer becomes PageBuf without any changes,
-    /// except vector's buffer capacity, which is removed.
-    pub fn new_from_vec(v: Vec<u8>) -> Result<Self, Error> {
-        Box::<[u8; GEAR_PAGE_SIZE]>::try_from(v.into_boxed_slice())
-            .map_err(|data| Error::InvalidPageDataSize(data.len() as u64))
-            .map(Self)
+impl From<PageBufInner> for PageBuf {
+    fn from(mut v: PageBufInner) -> Self {
```
grishasobol, line 125: I'm completely diagree with using trait `From` here. Yes, some times it's very approptiate, but not in this case. I cannot see all places where this method will be used, because of `.into()`

core/src/memory.rs
grishasobol: ok see now

runtime-interface/src/lib.rs
grishasobol: I think we will not drop vara, so you need to support.

program/src/metadata/mod.rs
```diff
@@ -26,7 +26,7 @@ use wasmtime::AsContextMut;
 /// Data used for the wasm exectuon.
 pub type StoreData = ext::Ext;
 
-const PAGE_SIZE: usize = 4096;
+pub const PAGE_SIZE: usize = 4_096;
```
breathx, line 29: can't we reimport from core such stuff?

breathx (changes requested):
Need to sync on deployment steps and how to deal with runtime interface

core/src/memory.rs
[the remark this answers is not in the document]
gshep: This is not necessary from my point of view. Since PageBuf is kept in Substrate's StorageMap after the PR it appears in metadata but polkadot.js doesn't support fixed size array longer 2048 items. That is the reason of such solution and it looks like there is no other ways.

Comments about the case are several lines below.

core/src/memory.rs
```diff
@@ -68,68 +63,83 @@ static_assertions::const_assert_eq!(WASM_PAGE_SIZE % GEAR_PAGE_SIZE, 0);
...
-pub struct PageBuf(Box<[u8; GEAR_PAGE_SIZE]>);
+#[derive(Clone, PartialEq, Eq, TypeInfo)]
+pub struct PageBuf(PageBufInner);
+
+// These traits are implemented intentionally by hand to achieve two goals:
+// - store PageBuf as fixed size array in a storage to eliminate extra bytes
+//      for length;
+// - work with PageBuf as with Vec. This is to workaround a limit in 2_048
+//      items for fixed length array in polkadot.js/metadata.
+//      Grep 'Only support for [Type' to get more details on that.
```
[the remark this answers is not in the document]
gshep: The phrase is quite correct (will adjust anyway):
```
$ grep -Hnr 'Only support for [[]Type' ./scripts/test-utils/node_modules/
./scripts/test-utils/node_modules/@polkadot/types-create/cjs/util/getTypeDef.js 159:    throw new Error(`${type}: Only support for [Type; <length>], where length <= 2048`);
./scripts/test-utils/node_modules/@polkadot/types-create/util/getTypeDef.js:136     throw new Error(`${type}: Only support for [Type; <length>], where length <= 2048`);
./scripts/test-utils/node_modules/@polkadot/types/bundle-polkadot-types.js:7931       throw new Error(`${type}: Only support for [Type; <length>], where length <= 2048`);
./scripts/test-utils/node_modules/@polkadot/types/bundle-polkadot-types.js:1506 :      throw new Error('Only support for [Type; <length>], where length <= 2048');
./scripts/test-utils/node_modules/@polkadot/types/cjs/metadata/PortableRegistry/PortableRegistry.js:837:    throw new Error('Only support for [Type; <length>], where length <= 2048');
./scripts/test-utils/node_modules/@polkadot/types/metadata/PortableRegistry/PortableRegistry.js:804:    throw new Error('Only support for [Type; <length>], where length <= 2048');
```

common/src/program_storage.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    fn remove_program_page_data(program_id: ProgramId, page_num: PageNumber) {
+        Self::MemoryPageMap::remove(program_id, page_num);
+    }
+
+    fn remove_program_pages(program_id: ProgramId) {
+        Self::MemoryPageMap::clear_prefix(program_id);
+    }
+
+    fn pages_final_prefix() -> [u8; 32];
```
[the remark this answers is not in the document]
gshep: like substrate's naming

common/src/storage/primitives/double_map.rs
```diff
@@ -74,6 +74,9 @@ pub trait DoubleMapStorage {
     /// Gets value stored under given keys, if present,
     /// and removes it from storage.
     fn take(key1: Self::Key1, key2: Self::Key2) -> Option<Self::Value>;
+
+    /// Remove items from the map matching a `first_key` prefix.
+    fn clear_prefix(first_key: Self::Key1);
```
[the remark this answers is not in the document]
gshep: there is already method with such name

lazy-pages/src/lib.rs
```diff
@@ -124,19 +124,33 @@ pub enum InitializeForProgramError {
...
+    #[display(fmt = "Prefix of storage with memory pages was not set")]
+    MemoryPagesPrefixNotSet,
 }
 
 pub fn initialize_for_program(
     wasm_mem_addr: Option<usize>,
     wasm_mem_size: WasmPageNumber,
     stack_end_page: Option<WasmPageNumber>,
-    program_prefix: Vec<u8>,
+    program_id: [u8; 32],
```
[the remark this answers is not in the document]
gshep: Sorry didn't get what you mean

common/src/storage/primitives/map.rs
```diff
@@ -66,6 +66,18 @@ pub trait MapStorage {
     fn take(key: Self::Key) -> Option<Self::Value>;
 }
 
+pub trait AppendMapStorage<Item, Key, Value>: MapStorage<Key = Key, Value = Value>
+where
+    Item: codec::Encode,
+    Key: codec::Encode,
+    Value: codec::EncodeAppend<Item = Item>,
+{
+    fn append<EncodeLikeKey, EncodeLikeItem>(key: EncodeLikeKey, item: EncodeLikeItem)
```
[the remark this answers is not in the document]
SabaunT: yeah

runtime-interface/src/lib.rs
grishasobol: What do you mean `not active network`? If you mean no lazy-pages is used yet on vara - then it's not the point. Old runtimes require that all import functions are supported by binary node.

core/src/memory.rs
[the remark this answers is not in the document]
breathx: oh, got it

lazy-pages/src/lib.rs
grishasobol: Also can be problem: in `init_for_program` we take `program_storage_prefix` and add `program_id`. If `init` is called for each message handling then this will not cause the problems, but if we will call `init` once per process queue cycle, then `program_id`s will be accumulated in `program_storage_prefix`, this is very dangerous bug, because it can lead to node binary panic.

grishasobol (changes requested):

core/src/memory.rs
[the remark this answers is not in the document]
breathx: I'm aware of performance decrease here

core/src/memory.rs
```diff
@@ -68,68 +63,83 @@ static_assertions::const_assert_eq!(WASM_PAGE_SIZE % GEAR_PAGE_SIZE, 0);
...
+        v.extend_with(0);
+        Self(v)
     }
+}
 
+impl PageBuf {
     /// Returns new page buffer with zeroed data.
     pub fn new_zeroed() -> PageBuf {
-        Self(Box::<[u8; GEAR_PAGE_SIZE]>::new([0u8; GEAR_PAGE_SIZE]))
+        Self(PageBufInner::filled_with(0))
```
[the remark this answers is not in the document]
breathx: > Agree however LimitedVec is a generic not necessarily containing numbers.

We can go with default trait bound

lazy-pages/src/lib.rs
grishasobol: Hmm, my previous comments disappear...

We have accumulating problem here. If we will make `init` once per process queue, then `program_storage_prefix ` will accumulate `program_id`'s for each handled message.

common/src/program_storage.rs
```diff
@@ -0,0 +1,158 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use super::*;
+use crate::storage::{AppendMapStorage, DoubleMapStorage, MapStorage};
+
+#[derive(Clone, Copy, Debug)]
+pub enum Error {
```
breathx, line 23: move to trait and pallet gear-program error enum

pallets/gear/src/lib.rs
```diff
@@ -1680,7 +1685,7 @@ pub mod pallet {
             <T as Config>::Scheduler::reset();
             <T as Config>::GasProvider::reset();
             <T as Config>::Messenger::reset();
```
breathx, line 1687: code storage reset

runtime-interface/src/lib.rs
```diff
@@ -51,7 +58,7 @@ pub trait GearRI {
         wasm_mem_addr: Option<HostPointer>,
         wasm_mem_size_in_pages: u32,
         stack_end_page: Option<u32>,
-        program_prefix: Vec<u8>,
+        program_id: Vec<u8>,
```
breathx, line 61: program id type

utils/wasm-proc/src/main.rs
```diff
@@ -21,7 +21,7 @@ use gear_wasm_builder::optimize::{OptType, Optimizer};
 use parity_wasm::elements::External;
 use std::{collections::HashSet, fs, path::PathBuf};
 
-const RT_ALLOWED_IMPORTS: [&str; 50] = [
+const RT_ALLOWED_IMPORTS: [&str; 51] = [
```
breathx, line 24: +1 after comment above

breathx (approved):

grishasobol (approved):

runtime-interface/src/lib.rs
gshep: unfortunately that requires `ProgramId` implementing `sp_runtime_interface::RIType`

PR #2051, opened 2022-12-30
Implement Gear stack end in `wasm-gen`

### LINKED ISSUE #2050: Add gear stack end to wasm gen
Add gear stask end to wasm gen to test all cases depend on this

1) ~1% of cases with invalid stack size not a multiple of the page size
2) ~1% of cases with invalid stask size that is biger than import memory
3) ~1% of cases stack size is not generated at all
4) all other cases should be valid

MikitaS:
Resolves #2050 

Add gear stack end to wasm-gen to test all cases depending on this

~1% of cases with invalid stack size not a multiple of the page size
~1% of cases with invalid stack size that is bigger than import memory
~1% of cases stack size is not generated at all
all other cases should be valid

### DIFF
--- a/utils/wasm-gen/src/lib.rs
+++ b/utils/wasm-gen/src/lib.rs
@@ -321,6 +321,10 @@ impl<'a> WasmGen<'a> {
         }
         mem_section_idx.map(|index| module.sections_mut().remove(index));
 
+        // TODO(MIKITA)
+        module = builder::from_module(module).global().value_type().i32().init_expr(Instruction::I32Const(10)).build().build();
+        module.export_section().unwrap().entries_mut().push()
+
         let mem_size = self.u.int_in_range(0..=self.config.max_mem_size).unwrap();
         let mem_size_upper_bound = if self.config.has_mem_upper_bound.get(self.u) {
             Some(if self.config.upper_bound_can_be_less_then.get(self.u) {
--- a/utils/wasm-gen/src/lib.rs
+++ b/utils/wasm-gen/src/lib.rs
@@ -391,11 +391,21 @@ impl<'a> WasmGen<'a> {
 
         let gear_stack_end_seed = self.get_gear_stack_end_seed(mem_size);
         if let GearStackEndExportSeed::GenerateValue(gear_stack_val) = gear_stack_end_seed {
+            let mut module = builder::from_module(module)
+                .global()
+                .value_type()
+                .i32()
+                .init_expr(Instruction::I32Const(gear_stack_val as i32))
+                .build()
+                .build();
+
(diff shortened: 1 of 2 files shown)


utils/wasm-gen/src/lib.rs
```diff
@@ -311,6 +316,45 @@ impl<'a> WasmGen<'a> {
         }
     }
 
+    // ~1% of cases with invalid stack size not a multiple of the page size
+    // ~1% of cases with invalid stask size that is biger than import memory
+    // ~1% of cases stack size is not generated at all
+    // all other cases should be valid
+    fn get_gear_stack_end_seed(&mut self, min_memory_size_pages: u32) -> GearStackEndExportSeed {
+        const MAX_SEED: u32 = 100;
```
grishasobol, line 324: I think it's better to use Ratio and also add this 1% in Config

utils/wasm-gen/src/test.rs
```diff
@@ -23,11 +23,14 @@ use rand::{rngs::SmallRng, RngCore, SeedableRng};
 #[allow(unused)]
 use indicatif::ProgressIterator;
 
+const MODULES_AMOUNT: usize = 100;
+const TEST_THRESHOLD: usize = 1000000;
```
grishasobol, line 27: UNSTRUCTURED_SIZE

MikitaS: Done

utils/wasm-gen/src/lib.rs
MikitaS: done

MikitaS:
@grishasobol @breathx please take another look

utils/wasm-gen/src/lib.rs
```diff
@@ -282,6 +286,11 @@ struct WasmGen<'a> {
     calls_indexes: Vec<FuncIdx>,
 }
 
+enum GearStackEndExportSeed {
```
grishasobol, line 289: You can use Option<u32> for such cases. But it's not a big problem

grishasobol (approved):

shamilsan (approved):

utils/wasm-gen/src/lib.rs
shamilsan: Agree, `Option<u32>` allows avoiding introducing a new type.

MikitaS: @grishasobol @shamilsan Sure it can be used. But I used created a new type for better readability, 'cause  GearStackEndExportSeed is more constructive then Option<u32>. It is also not effect to performance

PR #2054, opened 2023-01-02
Rework block authorship

ekovalev:
Resolves the issue with a flaky machinery of deciding wether the message queue should be processed in the end of each block depending on the outcome of previous runs.

The old mechanism relied on the inherent-like extrinsic signalling a successful completion of the message queue processing by setting some storage value to a specific status. If not set, this was deemed as a panic having occurred during one of the previous runs so that a new attempt wouldn't (by default) take place. This creates an attack vector of a malicious validator opting for deliberately not including this inherent into a block thereby blocking the queue processing in all the blocks thereafter, until the issue would have been investigated and fixed via a runtime upgrade which would render the network useless in the meantime.

The proposed solution changes this approach so that the inherent-like extrinsic is always included in a block; however, there is another storage value - `ExecuteInherent`, that is checked inside this extrinsic: if set to `false`, an error is returned thereby causing the block builder to drop this extrinsic. This value can only be altered by a `root origin`. This is useful to manually disable queue processing in case it panics and re-enable once the issue has been fixed.

**Issues**: since queue processing is run inside an unsigned extrinsic (a pseudo-inherent) it suffers from the same problems normal inherents do: a malicious validator may alter its behaviour, and there is no way to distinguish between a legitimate exclusion of this extrinsic due to a panic or an error and a deliberate dropping by validator.

For an external observer the following signals are available:
```
                                |      ExecuteInherent == true     |      ExecuteInherent == false
--------------------------------------------------------------------------------------------------------
                                |   Gear::BlockNumber::<T> keeps   |
                                |        growing every block       |
  Gear::run() gets into block   |                                  |                 N/ A
                                | Event::MessgesDispatched emited  |
                                |      (if total_handled > 0)      |
--------------------------------------------------------------------------------------------------------
               |      error     |   Gear::BlockNumber::<T> stalls  | 
  Gear::run()  | - - - - - - - -|     as chain blocks progress     |   Gear::BlockNumber::<T> stalls
    dropped    |      panic     |                                  |     as chain blocks progress
  from block   | - - - - - - - -|  Event::QueueProcessingReverted  |
               |   intentional  |      emited at every block       |
```

### DIFF
--- a/gclient/src/api/storage/block.rs
+++ b/gclient/src/api/storage/block.rs
@@ -17,7 +17,7 @@
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
 use super::{GearApi, Result};
-use crate::Error;
+use crate::{api::listener::EventListener, Error};
 use gp::api::{
     config::GearConfig,
     generated::api::{runtime_types::gear_runtime::RuntimeEvent, storage},
@@ -151,8 +151,29 @@ impl GearApi {
     }
 
     /// Check whether the message queue processing is stopped or not.
-    pub async fn queue_processing_stopped(&self) -> Result<bool> {
+    pub async fn queue_processing_enabled(&self) -> Result<bool> {
         let at = storage().gear().execute_inherent();
-        Ok(!self.0.storage().fetch_or_default(&at, None).await?)
+        Ok(self.0.storage().fetch_or_default(&at, None).await?)
+    }
+
+    /// Looks at two blocks from the stream and checks if the Gear block number
+    /// has grown from block to block or not.
+    pub async fn queue_processing_stalled(&self, listener: &mut EventListener) -> Result<bool> {
+        let at = storage().gear().block_number();
+
+        let current = listener.next_block_hash().await?;
+        let gear_current = self
+            .0
+            .storage()
+            .fetch_or_default(&at, Some(current))
+            .await?;
+
+        let mut next = current;
+        while next == current {
+            next = listener.next_block_hash().await?;
+        }
+        let gear_next = self.0.storage().fetch_or_default(&at, Some(next)).await?;
+
+        Ok(gear_next <= gear_current)
     }
 }
--- a/gclient/src/api/storage/block.rs
+++ b/gclient/src/api/storage/block.rs
@@ -20,10 +20,7 @@ use super::{GearApi, Result};
 use crate::Error;
 use gp::api::{
     config::GearConfig,
-    generated::api::{
-        runtime_types::{gear_runtime::RuntimeEvent, pallet_gear::ProcessStatus},
-        storage,
-    },
+    generated::api::{runtime_types::gear_runtime::RuntimeEvent, storage},
 };
 use subxt::{ext::sp_core::H256, rpc::ChainBlock};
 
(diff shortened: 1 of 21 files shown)


pallets/gear/src/lib.rs
```diff
@@ -463,19 +425,26 @@ pub mod pallet {
...
 
+    /// A flag indicating whether the message queue should be processed at the end of a block
+    ///
+    /// If not set, the inherent extrinsic that processes the queue will keep throwing an error
+    /// thereby making the block builder exclude it from the block.
     #[pallet::storage]
-    #[pallet::getter(fn queue_state)]
-    pub(crate) type QueueState<T> = StorageValue<_, ProcessStatus, ValueQuery>;
+    #[pallet::getter(fn execute_inherent)]
+    pub(crate) type ExecuteInherent<T> = StorageValue<_, bool, ValueQuery, ExecuteInherentOnEmpty>;
```
breathx, line 447:
```suggestion
    pub(crate) type ExecuteInherent<T> = StorageValue<_, bool, ValueQuery, ConstBool<true>>;
```

gclient/src/api/storage/block.rs
```diff
@@ -124,12 +121,12 @@ impl GearApi {
     }
 
     pub async fn queue_processing_stopped(&self) -> Result<bool> {
-        let at = storage().gear().queue_state();
+        let at = storage().gear().execute_inherent();
         self.0
             .storage()
             .fetch(&at, None)
             .await?
             .ok_or(Error::StorageNotFound)
```
breathx, line 129: Seems your ran into this error, cause default value set for a storage - so such key does not exist in database. here should be `unwrap_or(true)` instead of matching Option as returned value

breathx (approved):
Seems none binary replacements required, isn't it? Just runtime upgrade

breathx:
Some extra tests would be also great. For example, using `gclient`

ukint-vs (approved):

breathx (approved):
Still actual question for testing

breathx (changes requested):
Ohhh, suddenly found that seems the solution breaks logic with custom gear block number mutations

SabaunT:
> Ohhh, suddenly found that seems the solution breaks logic with custom gear block number mutations

what's the problem with that?

SabaunT (approved):

pallets/gear/src/lib.rs
```diff
@@ -483,6 +457,10 @@ pub mod pallet {
         }
     }
 
+    #[pallet::storage]
+    #[pallet::getter(fn last_gear_block_number)]
+    pub(crate) type LastGearBlockNumber<T: Config> = StorageValue<_, T::BlockNumber>;
```
breathx, line 462: some comments on it

pallets/gear/src/lib.rs
```diff
@@ -496,94 +474,38 @@ pub mod pallet {
...
-                }
-                Forcing::ForceOnce => {
-                    // Forcing queue processing in current block, subsequently not forcing anything
-                    ForceQueue::<T>::put(Forcing::default());
-                    QueueState::<T>::put(ProcessStatus::Scheduled);
-                    T::DbWeight::get().reads_writes(1, 3)
-                }
-                Forcing::NotForcing => T::DbWeight::get().reads_writes(1, 1),
-            }
+            T::DbWeight::get().writes(1)
```
breathx, line 485: add comment why not two please

breathx (approved):

Issue #1962, opened 2022-12-07
Add more features to `gclient`


shamilsan:
### Problem to Solve

`gclient` has some restrictions that bother using it for end-to-end tests. E.g., we don't have the possibility to get a sender account address or query program state.

### Possible Solution

- [x] Add public `account_id` or method to `GearApi`
- [x] Add `read_state` and `read_state_bytes` methods.
- [ ] Consider to replace `Borrow<AccountId32>` with `Into<AccountId32>` for convenience.
- [x] Consider to split [`get_from_mailbox`](https://docs.gear.rs/gclient/struct.GearApi.html#method.get_from_mailbox) into two methods: with `account_id` parameter and without it.
- [x] Consider adding an equivalent to the [gtest's `System::block_timestamp()`](https://docs.gear.rs/gtest/struct.System.html#method.block_timestamp) function.

shamilsan:
Also, we can split [`get_from_mailbox`](https://docs.gear.rs/gclient/struct.GearApi.html#method.get_from_mailbox) into two methods: with `account_id` parameter and without it.

fluiderson:
Consider to add an equivalent to the [gtest's `System::block_timestamp()`](https://docs.gear.rs/gtest/struct.System.html#method.block_timestamp) function.

PR #2056, opened 2023-01-04
gclient: Add functions for state reading

shamilsan:
- Part of #1962

- Added wrappers around three RPC calls: `gear_readState`, `gear_readStateUsingWasm`, `gear_readStateUsingWasm` 

@gear-tech/dev

### DIFF
--- a/gclient/src/api/rpc.rs
+++ b/gclient/src/api/rpc.rs
@@ -20,7 +20,11 @@
 use crate::{api::Result, GearApi};
 use gear_core::ids::{CodeId, MessageId, ProgramId};
 use gp::api::types::GasInfo;
-use subxt::ext::sp_core::H256;
+use parity_scale_codec::{Decode, Encode};
+use std::path::Path;
+use subxt::{ext::sp_core::H256, rpc::rpc_params};
+
+use crate::utils;
 
 impl GearApi {
     /// gear_calculateInitCreateGas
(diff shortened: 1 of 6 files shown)


gclient/src/api/rpc.rs
```diff
@@ -95,4 +99,134 @@ impl GearApi {
...
+    ) -> Result<Vec<u8>> {
+        self.0
+            .rpc()
+            .request("gear_readState", rpc_params![H256(program_id.into()), at])
+            .await
+            .map_err(Into::into)
+    }
+
+    /// Read the program's state as decoded data.
+    pub async fn read_state<D: Decode>(
```
breathx, line 117: How do you feel about separate these fns with and without non-optional "at" argument?

breathx (approved):
Some gclient tests would be also great. Probably minimal ones.

gclient/src/api/rpc.rs
shamilsan: I agree, but should we rework other functions too? 

For example, `read_state_at(..., at)` and `read_state(...)`?

breathx: > I agree, but should we rework other functions too?
> 
> For example, `read_state_at(..., at)` and `read_state(...)`?

yeah, this way

breathx: probably some smart wrapper that accepts closure would be good

shamilsan: I've not found a reason to use a closure there. I've just split functions into two options.

SabaunT (approved):

PR #2102, opened 2023-01-18
Update batch sender's balance

SabaunT:
Fuzz testing with `node-loader` ends with insufficient balance after some time when loading the node from `ALICE`. So we need to renew sender's balances during the loading.


ekovalev (approved):

NikVolf:
why is it stalled? is it that our node tests stall because now they have so much more money?

SabaunT:
> why is it stalled? is it that our node tests stall because now they have so much more money? 

Im fully into this task today. Preparing Pr that will unstall the test. Reason was not related to money.

SabaunT:
Solution is reworked, need a new review

ekovalev (approved):

gshep (approved):

PR #2143, opened 2023-01-26
Avoid annoying runtime rebuild

### LINKED ISSUE #2140: Crate "gear-runtime" is rebuilt every time
Even when no changes were made

shamilsan:
- Fixes #2140

- Also, examples are declared explicitly in root `Cargo.toml` to avoid build problems when switching between branches.

@gear-tech/dev


shamilsan:
@clearloop Please take a look. The problem is caused by using `vara-runtime` in the `gear-program` crate. As I see when we compile it with both `gear` and `vara` features, it doesn't use `vara-runtime` but imports it.

There may be a better solution but this one works for me.

node/cli/Cargo.toml
```diff
@@ -69,7 +69,7 @@ vara-native = [
 	"service/vara-native",
 	"gear-runtime-test-cli?/vara-native",
 	"vara-runtime",
-	"gear-program/vara"
+	"gear-program/gear"
```
gshep, line 72: sorry didn't get that change. `gear-program` declares its own `vara` feature and there it is used for `vara-native` that looks logical.

shamilsan:
@clearloop @gshep Please take a look to this alternative solution:

https://github.com/gear-tech/gear/pull/2143/commits/39b9546f9ef4e8aca1ff3b6244e8a5bcdeefc8fa

shamilsan:
> still requires rebuilding on my machine

It looks strange, I've tried to rebuild it from scratch, and it works for me.

Anyway, it doesn't pass CI tests for `gear-program`.

shamilsan:
@clearloop So, another option is to remove `gear-program` from the default build.

PR #2157, opened 2023-01-30
Fix bn and timestamp query in state fns

breathx:
By request of @LouiseMedova

### DIFF
--- a/examples/binaries/new-meta/state-v3/build.rs
+++ b/examples/binaries/new-meta/state-v3/build.rs
@@ -1,0 +1,21 @@
+// This file is part of Gear.
+
+// Copyright (C) 2021-2022 Gear Technologies Inc.
+// SPDX-License-Identifier: GPL-3.0-or-later WITH Classpath-exception-2.0
+
+// This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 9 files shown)


examples/binaries/new-meta/state-v3/build.rs
```diff
@@ -0,0 +1,21 @@
+// This file is part of Gear.
+
+// Copyright (C) 2021-2022 Gear Technologies Inc.
```
gshep, line 3: 2023

gshep: and for others too

examples/binaries/new-meta/state-v3/src/wasm.rs
```diff
@@ -0,0 +1,22 @@
+use demo_meta_io::Wallet;
```
gshep, line 1: do we need license there?

gshep (approved):

examples/binaries/new-meta/state-v3/build.rs
breathx: Let's change licenses in separate PR all over the project

PR #2165, opened 2023-01-31
Support reading program state from via gtest

### LINKED ISSUE #2096: Add support of the new meta to `gtest`
### Problem to Solve

`gtest` needs to support the new meta approach too.

### Possible Solution

Add corresponding methods.

### Notes

_No response_

DennisInSky:
- Resolves #2096


gtest/src/manager.rs
```diff
@@ -402,13 +393,58 @@ impl ExtManager {
 
     /// Call non-void meta function from actor stored in manager.
     /// Warning! This is a static call that doesn't change actors pages data.
-    pub(crate) fn call_meta(
+    pub(crate) fn read_state(&mut self, program_id: &ProgramId) -> Result<Vec<u8>> {
```
shamilsan, line 396: Isn't it better to decode result (or provide alternative functions with decoded output)?

gtest/src/program.rs
```diff
@@ -440,28 +440,24 @@ impl<'a> Program<'a> {
...
+    pub fn read_state(&self) -> Result<Vec<u8>> {
+        self.manager.borrow_mut().read_state(&self.id)
     }
 
-    pub fn meta_state_with_bytes(&self, payload: impl AsRef<[u8]>) -> Result<Vec<u8>> {
-        self.manager.borrow_mut().call_meta(
+    pub fn read_state_with_map(
+        &self,
+        mapping_wasm: Vec<u8>,
+        mapping_wasm_function_name: impl Into<String>,
```
shamilsan, line 450: IMO suboptimal. Why if we want to pass `&str`?

gtest/src/program.rs
```diff
@@ -440,28 +440,24 @@ impl<'a> Program<'a> {
...
-        D::decode(&mut self.meta_state_with_bytes(payload.encode())?.as_slice()).map_err(Into::into)
+    pub fn read_state(&self) -> Result<Vec<u8>> {
+        self.manager.borrow_mut().read_state(&self.id)
     }
 
-    pub fn meta_state_with_bytes(&self, payload: impl AsRef<[u8]>) -> Result<Vec<u8>> {
-        self.manager.borrow_mut().call_meta(
+    pub fn read_state_with_map(
+        &self,
+        mapping_wasm: Vec<u8>,
```
shamilsan, line 449: I'd provide the method where Wasm is referenced by its path as we do it in `gclient`.

shamilsan:
Also, I'd recommend using the same function names and similar signatures as we have in `gclient`:

https://docs.gear.rs/gclient/struct.GearApi.html#method.read_state

gtest/src/program.rs
shamilsan: As I see it makes sense to use `AsRef<str>` or similar only if we would implement the `WasmEntry` trait for it. Not crucial for the testing crate, though.

gtest/src/manager.rs
```diff
@@ -402,13 +393,58 @@ impl ExtManager {
...
+            Some(program.allocations().clone()),
+            Some(*program_id),
+            Default::default(),
+            u64::MAX,
+            self.block_info,
+        )
+        .map_err(TestError::ReadStateError)
+    }
+
+    pub(crate) fn read_state_with_map(
```
breathx, line 419: lets go with similar to pallet and rpc calls naming pls + wrappers for decoded variants
"mapping_" wasm/arg isn't intuitive enough for me

examples/binaries/new-meta/tests/read_mapped_state.rs
```diff
@@ -0,0 +1,48 @@
+use demo_new_meta::{Person, Wallet, META_EXPORTS_V1, META_EXPORTS_V2, META_WASM_V1, META_WASM_V2};
```
breathx, line 1: always declare consts outside the fn bodies pls

breathx (changes requested):

examples/binaries/new-meta/tests/common/mod.rs
```diff
@@ -0,0 +1,15 @@
+use demo_new_meta::MessageInitIn;
```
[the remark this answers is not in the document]
DennisInSky: > wont `examples/binaries/new-meta/tests/common.rs` work as mod decl?

No, it won't. Everything under the `tests` folder is treated by cargo as a separate test crate. If one wants ot share common functionality amongst integration tests, they have to use "old-style" mod declaration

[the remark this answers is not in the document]
DennisInSky: > I also see no reason to split everything in ./tests by files when total lines < 100

Will gather all stuff related to state reading under a single integration test crate

gtest/src/program.rs
DennisInSky: replaced by using &str. Found here https://www.philipdaniels.com/blog/2019/rust-api-design/#:~:text=Into%20has%20a%20similar%20effect,String%3E%20%7B%20...%20%7D that is more idioamatic for rust to use &str instead of AsRef<str>

shamilsan: AsRef<str> is more general than &str. You can find more use case details in the "Rust for Rustaceans" book.

image

shamilsan (approved):
Looks good to me 👍

examples/binaries/new-meta/tests/read_state.rs
```diff
@@ -0,0 +1,87 @@
...
+        .read_state_bytes_using_wasm(FIRST_WALLET_FUNC_NAME, META_WASM_V1.to_vec(), None)
+        .expect("Unable to read program state");
+
+    let expected_state = Wallet::test_sequence().first().encode();
+
+    assert_eq!(expected_state, actual_state);
+}
+
+#[test]
+fn read_state_bytes_with_parameterized_wasm_func_returns_transformed_state() {
```
breathx, line 38: `read_state_bytes_with_wasm_func_returns_transformed_state` already exists, lets force this test return decoded value

examples/binaries/new-meta/tests/read_state.rs
```diff
@@ -0,0 +1,87 @@
+use demo_new_meta::{
```
breathx, line 1: I'd also provide test for mock reading state. why not

gtest/src/program.rs
```diff
@@ -440,26 +440,43 @@ impl<'a> Program<'a> {
...
+    /// Reads and decodes the program’s transformed state. The transformed state
+    /// is a result of applying the `fn_name` function from the `wasm`
+    /// binary with the optional `argument`.
+    pub fn read_state_using_wasm<E: Encode, D: Decode>(
+        &self,
+        fn_name: &str,
+        wasm: Vec<u8>,
+        argument: Option<E>,
+    ) -> Result<D> {
+        let argument_bytes = argument.map(|arg| arg.encode());
```
breathx, line 477:
```suggestion
        let argument_bytes = argument.map(Encode::encode);
```

breathx (approved):
👍🏻

PR #2169, opened 2023-02-01
feat(gprogram): remove runtimes from build dependencies

### LINKED ISSUE #2162: Support gprogram in default build in workspace's root directory
### Problem to Solve

Original issue: #2140 
Previous PR: #2143

### Possible Solution

https://github.com/paritytech/substrate/pull/11624

do the similar stuffs in the `build.rs` of `gprogram`

### Notes

_No response_

clearloop:
Resolves #2162 

Based on https://github.com/gear-tech/gear/pull/2145, figured out a more proper solution to solve the rebuilding problem.

After reviewing the build logs, just found even we are at the first build, both of gear-runtime and vara-runtime are compiling for twice times, according to [build dependencies](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#build-dependencies):

> A package itself and its build script are built separately, so their dependencies need not coincide.

We can just build the runtimes as wasm projects only like what `gear-runtime/vara-runtime`'s build script does since we can get the metadata we want from runtimes' wasm directly, and since the building process is managed by `substrate-wasm-builder`, whatever we build runtimes' wasm from `node/cli` or `/gear-program`, the build cache is always shared!

@gear-tech/dev @shamilsan

### DIFF
--- a/core/src/message/incoming.rs
+++ b/core/src/message/incoming.rs
@@ -154,7 +154,7 @@ impl IncomingDispatch {
         )
     }
 
-    /// Decompose IncomingDispatch for it's components: DispatchKind, IncomingMessage and Option<ContextStore>.
+    /// Decompose IncomingDispatch for it's components: DispatchKind, IncomingMessage and `Option<ContextStore>`.
     pub fn into_parts(self) -> (DispatchKind, IncomingMessage, Option<ContextStore>) {
         self.into()
     }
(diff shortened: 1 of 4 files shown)


shamilsan (approved):
Looks good to me. Great job 👍

program/build.rs
```diff
@@ -82,9 +150,9 @@ fn update_api() {
     {
         write_api(
```
breathx, line 151: shouldn't above be similar cfg mapping, that contains
`all(feature = "gear", feature = "vara")`?

breathx (approved):

PR #2176, opened 2023-02-03
Fix backend errors handling for fallible syscalls

breathx:
## Error explanation
For example let's take fallible `gr_unreserve_gas` syscall.
https://github.com/gear-tech/gear/blob/9d96bec693b34db229598d0e638502228dc1a43d/core-backend/sandbox/src/funcs.rs#L856-L871
For case if we failed to read read reservation id, atm we are leaving backend with wasm trap, while it should try write ext error about incorrect `reservation_id_ptr` to `write_err_unreserved`, because syscall supposed to be fallible and dance around results.
For case when fails write to `err_<something>` ptr, we already cannot parse error in user space so we should leave with a trap.

## TODO list
- [x] Apply similar changes to wasmi backend
- [x] Provide tests and check that it fails on master

### DIFF
--- a/core-backend/sandbox/src/runtime.rs
+++ b/core-backend/sandbox/src/runtime.rs
@@ -48,68 +48,11 @@ pub(crate) struct Runtime<E: Ext> {
     pub memory_manager: MemoryAccessManager<E>,
 }
 
-/*
-
-WASM -> Backend
--- fallible case --
-ext.some_call() -> Result<T, ProcessorError>
-Result.into_ext_error(&mut state) {
-Ok(T) -> Ok(T)
-Err(ProcessorError::ExtError) -> Ok(Err(LENGTH))
-Err(Processor::GasAllowanceExceeded) -> Err()
-}
-
-
-
-*/
-
-/*
-
-type BodyRes = Result<T, E1>;
-type PostRes = Result<T, E2>;
-
-// - T to be casted into `ReturnValue`
-// - E1 after body closure writes into `Runtime { err, .. }`,
-// supposed to be
-
-ctx.run_any(
-    body: |&mut self| -> BodyRes,
-    post: |&mut self, BodyRes| -> PostRes,
-)
-where:
-    - E to be written into Runtime { err, .. }
-    -
-
-*/
-
 impl<E> Runtime<E>
 where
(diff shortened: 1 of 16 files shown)


ark0f (approved):

core-backend/sandbox/src/runtime.rs
```diff
@@ -91,6 +87,63 @@ where
...
+            *self.err_mut() = err;
+        });
+
+        if body_res.is_err() {
+            if let Ok(to_be_returned) = self.last_err() {
+                body_res = Ok(Err(to_be_returned.encoded_size() as u32));
+            }
+        }
+
+        let body_res = body_res;
```
grishasobol, line 121: Can you clarify this line please?

grishasobol (approved):

grishasobol (changes requested):
We make everywhere : 
```rust
ctx.run_fallible(
|ctx| {
    let write_obj = ctx.register_write_as(T)
},
|ctx| {
    // main logic
},
|ctx, write_obj, res| {
    ctx.write_as(write_obj, T::from(res))
},
```

We can make generic method `run_fallible` with T param

examples/binaries/backend-error/src/lib.rs
```diff
@@ -0,0 +1,57 @@
...
+            gsys::gr_send(
+                pid_value.as_ptr(),
+                u32::MAX as *const u8,
+                42,
+                0,
+                res.as_mut_ptr(),
+            )
+        };
+
+        assert!(res.length != 0)
```
ark0f, line 55: Also read error to see it's correct

gsys/src/lib.rs
```diff
@@ -170,6 +172,16 @@ impl From<Result<Handle, Length>> for LengthWithHandle {
     }
 }
 
+#[repr(C, packed)]
+#[derive(Default)]
+pub struct LengthBytes([u8; size_of::<Length>()]);
```
ark0f, line 177:
```suggestion
pub struct LengthBytes([u8; mem::size_of::<Length>()]);
```

ark0f (approved):

grishasobol (approved):

PR #2196, opened 2023-02-07
Get rid of using Borrow in gclient

### LINKED ISSUE #1962: Add more features to `gclient`

DennisInSky:
Resolves #1962 

@reviewer-or-team

### DIFF
--- a/gclient/src/api/storage/into_account_id.rs
+++ b/gclient/src/api/storage/into_account_id.rs
@@ -1,0 +1,42 @@
+// This file is part of Gear.
+
+// Copyright (C) 2022 Gear Technologies Inc.
+// SPDX-License-Identifier: GPL-3.0-or-later WITH Classpath-exception-2.0
+
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use gear_core::ids::ProgramId;
+use subxt::ext::sp_runtime::AccountId32;
+
+pub trait IntoAccountId32 {
+    fn into_account_id(self) -> AccountId32;
+}
+
+impl IntoAccountId32 for AccountId32 {
+    fn into_account_id(self) -> AccountId32 {
+        self
+    }
+}
+
+impl IntoAccountId32 for &AccountId32 {
+    fn into_account_id(self) -> AccountId32 {
+        self.clone()
+    }
+}
+
+impl IntoAccountId32 for ProgramId {
+    fn into_account_id(self) -> AccountId32 {
+        AccountId32::new(self.into_bytes())
+    }
+}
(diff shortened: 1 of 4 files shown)


gclient/src/api/storage/mod.rs
```diff
@@ -47,8 +47,8 @@ impl GearApi {
     /// mailbox.
     pub async fn get_from_account_mailbox(
         &self,
-        account_id: impl Borrow<AccountId32>,
-        message_id: impl Borrow<MessageId>,
+        account_id: &AccountId32,
```
breathx, line 50:
```suggestion
        account_id: impl Into<AccountId32>,
```

gclient/src/api/storage/mod.rs
```diff
@@ -47,8 +47,8 @@ impl GearApi {
...
+        message_id: MessageId,
     ) -> Result<Option<(StoredMessage, Interval<u32>)>> {
         let at = storage().gear_messenger().mailbox(
             account_id.borrow(),
```
breathx, line 54: line below may be replaced with `generated_ids::MessageId::from(message_id)`

breathx (changes requested):

gclient/src/api/storage/mod.rs
breathx: Also provide From/Into AccountID <-> u64, [u8; 32], ProgramId

shamilsan: So, we need to change all functions that have the `&AccountId32` argument.

gclient/src/api/storage/into_account_id.rs
```diff
@@ -0,0 +1,42 @@
+// This file is part of Gear.
```
breathx, line 1: rename mode to something like `id.rs`, because such long namings confuse me

gclient/src/api/storage/into_account_id.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+
+impl IntoAccountId32 for &AccountId32 {
+    fn into_account_id(self) -> AccountId32 {
+        self.clone()
+    }
+}
+
+impl IntoAccountId32 for ProgramId {
```
breathx, line 38: same for reference prog_id maybe?

gclient/src/api/storage/into_account_id.rs
```diff
@@ -0,0 +1,42 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use gear_core::ids::ProgramId;
+use subxt::ext::sp_runtime::AccountId32;
+
+pub trait IntoAccountId32 {
```
breathx, line 22: some doc comments here?

breathx (approved):

PR #2198, opened 2023-02-07
Revert check for `--locked` in CI

shamilsan:
Check whether `Cargo.lock` need to be updated et early stages: https://github.com/gear-tech/gear/actions/runs/4117599630/jobs/7109093163

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -106,7 +106,7 @@ jobs:
         run: ./scripts/gear.sh format doc --check
 
       - name: "Check clippy: Gear"
-        run: ./scripts/gear.sh clippy gear --all-targets --all-features
+        run: ./scripts/gear.sh clippy gear --all-targets --all-features --locked
 
       - name: "Check clippy: Examples"
         run: ./scripts/gear.sh clippy examples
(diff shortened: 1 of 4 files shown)


.github/workflows/CI.yaml
```diff
@@ -106,7 +106,7 @@ jobs:
         run: ./scripts/gear.sh format doc --check
 
       - name: "Check clippy: Gear"
-        run: ./scripts/gear.sh clippy gear --all-targets --all-features
+        run: ./scripts/gear.sh clippy gear --all-targets --all-features --locked
```
breathx, line 109: seems it doesn't work for cargo build, why should do here?

breathx (approved):

.github/workflows/CI.yaml
shamilsan: `cargo clippy` without `--locked` will update `Cargo.lock`, and then the next steps will pass successfully even with `--locked` flag.

shamilsan: This has been added for `check` stage breaks at the early stage too. But we can leave `--locked` in the `build` stage only indeed.

PR #2200, opened 2023-02-07
Introduce `gr_panic` sys-call

### LINKED ISSUE #1857: Improve diagnostic when `gstd` without 'debug'
### Problem to Solve

When an execution of a program built with `gstd/nodebug` hits `expect`/`assert`/`panic`/`unreachable`/etc we get in the logs just short message 'probably `unreachable` reached`. It doesn't point whether it is program logic fails and where it was or an error rose on executor side.

### Possible Solution

The idea is to introduce new syscall (or two):
```diff
+
+        pub fn gr_panic(file: u64, line: u32, column: u32);
+        pub fn gr_panic2(location: &core::panic::Location<'_>) -> !;
     }
 }
 // ...
+pub fn panic(loc: &core::panic::Location<'_>) -> ! {
+    unsafe {
+        sys::gr_panic(loc.file().as_ptr() as u64, loc.line(), loc.column());
+        sys::gr_panic2(loc)
+    }
+}
 ```
 
Implementation of `gr_panic` is just to put data to the log and interrupt the execution.  `gr_panic2` forces compiler to put `Location` instance somewhere and to provide a pointer to it. Panic handler could possibly look then like this:
 ```diff
 --- a/gstd/src/common/handlers.rs
+++ b/gstd/src/common/handlers.rs
@@ -48,8 +48,11 @@ pub fn oom(_: Layout) -> ! {
 #[cfg(not(debug_assertions))]
 #[cfg(target_arch = "wasm32")]
 #[panic_handler]
-pub fn panic(_: &PanicInfo) -> ! {
-    wasm32::unreachable();
+pub fn panic(panic_info: &PanicInfo) -> ! {
+    match panic_info.location() {
+        Some(loc) => gcore::msg::panic(loc),
+        _ => wasm32::unreachable(),
+    }
 }
 ```
 
 `demo-panicker` WAT looks with these changes applied the following way:
 ```wat
 (module
  (type (;0;) (func (param i32)))
  (type (;1;) (func))
  (type (;2;) (func (param i64 i32 i32)))
  (import "env" "gr_panic" (func (;0;) (type 2)))
  (import "env" "gr_panic2" (func (;1;) (type 0)))
  (import "env" "memory" (memory (;0;) 17))
  (func (;2;) (type 1)
    i32.const 1048620
    call 3
    unreachable)
  (func (;3;) (type 0) (param i32)
    (local i32 i32)
    local.get 0
    call 5
    unreachable)
  (func (;4;) (type 1)
    i32.const 1048652
    call 3
    unreachable)
  (func (;5;) (type 0) (param i32)
    local.get 0
    call 6
    unreachable)
  (func (;6;) (type 0) (param i32)
    local.get 0
    call 7
    unreachable)
  (func (;7;) (type 0) (param i32)
    local.get 0
    i64.load32_u
    local.get 0
    i32.load offset=8
    local.get 0
    i32.load offset=12
    call 0
    local.get 0
    call 1
    unreachable)
  (export "handle" (func 2))
  (export "init" (func 4))
  (data (;0;) (i32.const 1048576) "I just panic every timepanicker/src/lib.rs\00\00\17\00\10\00\13\00\00\00\08\00\00\00\05\00\00\00Panic in init!\00\00\17\00\10\00\13\00\00\00\0e\00\00\00\05"))
  ```
  
  `wasm-proc` could then process the binary, extract source file paths to the new info file (so called 'debug info') and squeeze `data` element a bit:
  ```diff
  @@ -3,10 +3,9 @@
   (type (;1;) (func))
   (type (;2;) (func (param i64 i32 i32)))
   (import "env" "gr_panic" (func (;0;) (type 2)))
-  (import "env" "gr_panic2" (func (;1;) (type 0)))
   (import "env" "memory" (memory (;0;) 17))
   (func (;2;) (type 1)
-    i32.const 1048620
+    i32.const 1048576
     call 3
     unreachable)
   (func (;3;) (type 0) (param i32)
@@ -15,7 +14,7 @@
     call 5
     unreachable)
   (func (;4;) (type 1)
-    i32.const 1048652
+    i32.const 1048588
     call 3
     unreachable)
   (func (;5;) (type 0) (param i32)
@@ -34,9 +33,7 @@
     local.get 0
     i32.load offset=12
     call 0
-    local.get 0
-    call 1
     unreachable)
   (export "handle" (func 2))
   (export "init" (func 4))
-  (data (;0;) (i32.const 1048576) "I just panic every timepanicker/src/lib.rs\00\00\17\00\10\00\13\00\00\00\08\00\00\00\05\00\00\00Panic in init!\00\00\17\00\10\00\13\00\00\00\0e\00\00\00\05"))
+  (data (;0;) (i32.const 1048576) "\17\00\10\00\13\00\00\00\08\00\00\00\17\00\10\00\13\00\00\00\0e\00\00\00"))
```

However for this to work it is required somehow to force compiler to put all `Location`s in the new standalone section and file paths to the different one.

### Notes

Disadvantage: too complicated

ark0f:
Resolves #1857

Also remove OOM handler because it's unreachable. Our `alloc()` implementation just stops execution in case of impossibility to allocate more memory

TODO:
- [x] Fix asserts in tests


core-backend/sandbox/src/funcs.rs
```diff
@@ -759,6 +762,26 @@ where
...
+    pub fn panic(ctx: &mut Runtime<E>, args: &[Value]) -> SyscallOutput {
+        sys_trace!(target: "syscall::gear", "panic, args = {}", args_to_str(args));
+
+        let (data_ptr, data_len): (_, u32) = args.iter().read_2()?;
+
+        ctx.run(|ctx| {
+            let read_data = ctx.register_read(data_ptr, data_len);
+            let data = ctx.read(read_data)?;
+
+            let s = String::from_utf8_lossy(&data).to_string();
```
gshep, line 774: TrimmedString should be used there, shouldn't it?

gstd/src/common/handlers.rs
```diff
@@ -26,55 +26,25 @@
...
-
-#[cfg(not(feature = "debug"))]
-#[cfg(not(debug_assertions))]
-#[cfg(target_arch = "wasm32")]
-#[panic_handler]
-pub fn panic(_: &PanicInfo) -> ! {
-    wasm32::unreachable();
-}
-
-#[cfg(any(feature = "debug", debug_assertions))]
```
gshep, line 55: Sorry cannot agree with that though it provides better diagnostics.
It will lead to our programs begin to swell in size.

gshep: relates to #1857

core-backend/sandbox/src/funcs.rs
```diff
@@ -710,6 +713,22 @@ where
...
 
+    /// Infallible `gr_panic` syscall.
+    pub fn panic(ctx: &mut Runtime<E>, args: &[Value]) -> SyscallOutput {
+        sys_trace!(target: "syscall::gear", "panic, args = {}", args_to_str(args));
+
+        let (data_ptr, data_len): (_, u32) = args.iter().read_2()?;
+
+        ctx.run(|ctx| {
+            let read_data = ctx.register_read(data_ptr, data_len);
+            let data = ctx.read(read_data)?;
```
breathx, line 724: are you sure we want to exit by trap instead of panic with empty message in case of failing read?
furthermore add check for 0-len to avoid going to memory (open issue)

breathx: use runtime buffer also?

breathx (approved):

PR #2214, opened 2023-02-09
Support spawning dev node via gclient means

DennisInSky:
@reviewer-or-team


gclient/src/node/mod.rs
```diff
@@ -16,4 +16,77 @@
...
+    ws_address: WSAddress,
+}
+
+impl Node {
+    /// Instantiates a node runnin in development mode (--dev) via spawning its
+    /// process using the specified `path` on a randomly picked port. Waits for
+    /// the node getting initialized before returning it to a caller.
+    /// Ideally, the node's binary should be downloaded by means of CI pipeline
+    /// from https://get.gear.rs./
+    pub fn from_path(path: impl AsRef<OsStr>) -> Result<Self> {
```
shamilsan, line 44: Why not `AsRef<Path>`?

shamilsan: Also, I like when methods that return `Result` have a `try_` prefix (similar to those in [`TryFrom` trait](https://doc.rust-lang.org/std/convert/trait.TryFrom.html)).

gclient/src/node/mod.rs
```diff
@@ -16,4 +16,77 @@
...
+    pub fn ws_address(&self) -> &WSAddress {
+        &self.ws_address
+    }
+
+    fn wait_while_initialized(&mut self) -> Result<String> {
+        self.wait_for_log_record("Imported #1 ")
+    }
+
+    fn wait_for_log_record(&mut self, log: &str) -> Result<String> {
+        let stderr = self.process.stderr.take();
```
shamilsan, line 76: This function is not reenterable, insn't it? Is it ok?

gclient/src/node/port.rs
```diff
@@ -0,0 +1,19 @@
+//! Picking random ports
+use rand::Rng;
+use std::{net::TcpListener, ops::Range};
+
+/// localhost addr
+const LOCALHOST: &str = "127.0.0.1";
+const PORT_RANGE: Range<u16> = 15000..25000;
+
+/// Pick a random port
+pub fn pick() -> u16 {
```
shamilsan, line 10: Can we use port 0 for dynamic ports instead of RNG? Or does it not work with the node?

shamilsan (commented):
Some tests would be highly appreciated, especially when we can invoke two instances in parallel.

gclient/src/node/mod.rs
DennisInSky: > Why not `AsRef<Path>`?

@shamilsan I have chosen the type based on the one used for the `Command::new` function. Do you think they chose it wrong and the `AsRef<Path>` should've been used instead? I am happy to amend, but please explain why

shamilsan: > @shamilsan I have chosen the type based on the one used for the `Command::new` function. Do you think they chose it wrong and the `AsRef<Path>` should've been used instead? I am happy to amend, but please explain why

Ah, I see. It is ok, as both `Path` and `PathBuf` implement `AsRef<OsStr>`. For me, `Path` is just a little bit more habitual.

DennisInSky: > Also, I like when methods that return `Result` have a `try_` prefix (similar to those in [`TryFrom` trait](https://doc.rust-lang.org/std/convert/trait.TryFrom.html)).

addressed

shamilsan:
> > Some tests would be highly appreciated, especially when we can invoke two instances in parallel.
> 
> @shamilsan Any ideas on how to add such? Using of this requires an existing binary for the node, but I do not think we want to make gclient dependent on the node build.

Feel free to use the `gear` binary from the `target` directory. It is not a perfect solution, but it is acceptable, in my opinion.

gclient/src/node/mod.rs
DennisInSky: It si fine for this particular case as it is called just once. But have turned it into reentrant here and in the place where I copy pasted it from

DennisInSky:
> > @shamilsan Any ideas on how to add such? Using of this requires an existing binary for the node, but I do not think we want to make gclient dependent on the node build.
> 
> Feel free to use the `gear` binary from the `target` directory. It is not a perfect solution, but it is acceptable, in my opinion.

Have done it that way.

gclient/src/node/result.rs
```diff
@@ -0,0 +1,9 @@
+#[derive(Debug, thiserror::Error)]
```
breathx, line 1: I really see no reason in so many almost empty files. result mod / port mod may be just moved inside mod.rs

breathx (approved):

PR #2219, opened 2023-02-10
Use injected program addresses for send-calls

### LINKED ISSUE #2184: Fuzz inter-program communication within node-loader
### Problem to Solve

`wasm-gen` can receive as an input an implementor of the `Context` trait from `node-loader`, which by `Context::get_program` gives `ProgramId`, that will be a static defined address of the receiver of messages sent by this generated program.

Very likely, that inter-program communication is regulated by the `wasm-gen` config, such that in some cases we will have programs sending messages to other programs, but in other cases to `msg::source`.

gshep:
Resolves #2184 

@gear-tech/dev

### DIFF
--- a/utils/wasm-gen/src/lib.rs
+++ b/utils/wasm-gen/src/lib.rs
@@ -204,6 +204,30 @@ impl GearConfig {
     }
 }
 
+// Module and an optional index of gr_debug syscall.
+struct ModuleWithDebug {
+    module: Module,
+    debug_syscall_index: Option<u32>,
+}
+
+impl From<Module> for ModuleWithDebug {
+    fn from(module: Module) -> Self {
+        Self {
+            module,
+            debug_syscall_index: None,
+        }
+    }
+}
+
+impl From<(Module, Option<u32>)> for ModuleWithDebug {
+    fn from((module, debug_syscall_index): (Module, Option<u32>)) -> Self {
+        Self {
+            module,
+            debug_syscall_index,
+        }
+    }
+}
+
 pub fn default_swarm_config(u: &mut Unstructured, gear_config: &GearConfig) -> SwarmConfig {
     let mut cfg: SwarmConfig = u.arbitrary().unwrap();
 
@@ -662,16 +686,17 @@ impl<'a> WasmGen<'a> {
         )
     }
 
-    pub fn insert_sys_calls(&mut self, module: Module, memory_pages: WasmPageCount) -> Module {
+    pub fn insert_sys_calls(&mut self, module: Module, memory_pages: WasmPageCount) -> ModuleWithDebug {
         let code_size = if let Some(code) = module.code_section() {
             code.bodies()
                 .iter()
                 .fold(0, |sum, body| sum + body.code().elements().len())
         } else {
-            return module;
+            return module.into();
         };
 
         let mut source_call_index = None;
+        let mut debug_call_index = None;
 
         let import_count = module.import_count(ImportCountType::Function);
 
@@ -724,6 +749,10 @@ impl<'a> WasmGen<'a> {
                 source_call_index = Some(call_index);
             }
 
+            if name == SysCallName::Debug {
+                debug_call_index = Some(call_index);
+            }
+
             self.calls_indexes
                 .push(FuncIdx::Import((import_count + i) as u32));
             syscall_data.insert(
@@ -747,7 +776,7 @@ impl<'a> WasmGen<'a> {
             }
         }
 
-        module
+        (module, debug_call_index).into()
     }
 
     fn build_call_instructions(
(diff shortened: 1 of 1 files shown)


utils/call-gen/src/lib.rs
```diff
@@ -44,8 +45,12 @@ pub enum GearCall {
     UploadCode(UploadCodeArgs),
 }
 
-pub fn generate_gear_program<Rng: CallGenRng>(seed: u64) -> Vec<u8> {
+pub fn generate_gear_program<Rng: CallGenRng>(
+    existing_programs: Vec<ProgramId>,
```
SabaunT, line 49: due to `NonEmpty` type usage in the crate to state existing programs/codes, I'd state in the comment, that call to function accepts empty program vector too

utils/call-gen/src/upload_code.rs
```diff
@@ -51,8 +53,8 @@ impl TryFrom<GearCall> for UploadCodeArgs {
 
 impl UploadCodeArgs {
     /// Generates `pallet_gear::Pallet::<T>::upload_code` call arguments.
-    pub fn generate<Rng: CallGenRng>(code_seed: Seed) -> Self {
-        let code = crate::generate_gear_program::<Rng>(code_seed);
+    pub fn generate<Rng: CallGenRng>(existing_programs: Vec<ProgramId>, code_seed: Seed) -> Self {
```
SabaunT, line 56: same here. Maybe even better to call it `programs`

utils/call-gen/src/upload_program.rs
```diff
@@ -54,10 +56,15 @@ impl TryFrom<GearCall> for UploadProgramArgs {
 
 impl UploadProgramArgs {
     /// Generates `pallet_gear::Pallet::<T>::upload_program` call arguments.
-    pub fn generate<Rng: CallGenRng>(code_seed: Seed, rng_seed: Seed, gas_limit: u64) -> Self {
+    pub fn generate<Rng: CallGenRng>(
+        existing_programs: Vec<ProgramId>,
```
SabaunT, line 60: same

utils/wasm-gen/src/lib.rs
```diff
@@ -793,38 +835,58 @@ impl<'a> WasmGen<'a> {
...
-        // pass the offset as the first argument to the send-call
-        instructions.push(Instruction::I32Const(offset as i32));
+                let memory_size = memory_pages.memory_size();
+                let upper_limit = memory_size.saturating_sub(MEMORY_VALUE_SIZE);
+                let offset = self
+                    .u
+                    .int_in_range(0..=upper_limit)
+                    .expect("build_call_instructions: Unstructured::int_in_range failed");
+
+                // call msg::source (gr_source) with a memory offset
```
SabaunT, line 849: is it actual?

SabaunT (approved):
Seems `gear-wasm-gen` becomes more complex and has to be refactored and well documented.

utils/wasm-gen/src/lib.rs
gshep: yep, it is a possible execution path

utils/call-gen/src/lib.rs
```diff
@@ -44,8 +45,12 @@ pub enum GearCall {
     UploadCode(UploadCodeArgs),
 }
 
-pub fn generate_gear_program<Rng: CallGenRng>(seed: u64) -> Vec<u8> {
+/// Function generates WASM-binary of a Gear program with the
+/// specified `seed`. `programs` may specify addresses which
+/// can be used for send-calls.
+pub fn generate_gear_program<Rng: CallGenRng>(programs: Vec<ProgramId>, seed: u64) -> Vec<u8> {
```
breathx, line 51:
```suggestion
pub fn generate_gear_program<Rng: CallGenRng>(seed: u64, programs: Vec<ProgramId>) -> Vec<u8> {
```
For me seed seems more basic and important here

utils/call-gen/src/upload_code.rs
```diff
@@ -51,8 +53,8 @@ impl TryFrom<GearCall> for UploadCodeArgs {
 
 impl UploadCodeArgs {
     /// Generates `pallet_gear::Pallet::<T>::upload_code` call arguments.
-    pub fn generate<Rng: CallGenRng>(code_seed: Seed) -> Self {
-        let code = crate::generate_gear_program::<Rng>(code_seed);
+    pub fn generate<Rng: CallGenRng>(programs: Vec<ProgramId>, code_seed: Seed) -> Self {
```
breathx, line 56: same about args ordering all over the place

utils/call-gen/src/upload_code.rs
```diff
@@ -18,6 +18,8 @@
 
 //! Upload code args generator.
 
+use gear_core::ids::ProgramId;
+
```
breathx, line 22: Excess newlines in each import section
```suggestion
```

utils/wasm-gen/src/lib.rs
```diff
@@ -794,38 +836,58 @@ impl<'a> WasmGen<'a> {
...
-    pub fn make_print_test_info(&mut self, mut module: Module) -> Module {
-        let text = if let Some(text) = &self.config.print_test_info {
-            text
-        } else {
-            return module;
+    pub fn make_print_test_info(&mut self, result: ModuleWithDebug) -> Module {
+        let text = match &self.config.print_test_info {
+            Some(text) => text,
+            None => return result.module,
         };
```
breathx, line 877:
```suggestion
        let Some(text) = &self.config.print_test_info  else {
                return result.module
        };
```

breathx (approved):
Shouldn't we add some percentage of inserting valid addresses to keep program -> user interactions with random ids?

gshep:
> Shouldn't we add some percentage of inserting valid addresses to keep program -> user interactions with random ids?

Currently the generator won't create programs with send-calls to other actors if the input actor list is empty. So it is a caller responsibility to control generating. Anyway this feature could be implemented later if required.

PR #2240, opened 2023-02-12
Use u32::MAX as special ptr on backend

### LINKED ISSUE #2027: change special ptr in syscalls from i32::MAX to u32::MAX
### Problem to Solve

Currently we have special ptr equal to i32::MAX for syscalls with value_ptr. We should change it to u32::MAX. But this is change of user interface for smart contracts, so we cannot do it without storage resets. And all contracts rebuild.

### Possible Solution

-

### Notes

_No response_

breathx:
Resolves #2027


shamilsan (approved):
👍

ark0f (approved):

PR #2244, opened 2023-02-13
Fix fuzzing tests being non-reproducible

### LINKED ISSUE #2170: Gear program generator is not deterministic
### Problem
Run on Linux

```rust
fn generate_gear_program<Rng: LoaderRng>(seed: Seed) -> Vec<u8> {
    let mut rng = Rng::seed_from_u64(seed);

    let mut buf = vec![0; 100_000];
    rng.fill_bytes(&mut buf);

    let mut u = Unstructured::new(&buf);

    let mut config = gear_wasm_gen::GearConfig::new_normal();
    config.print_test_info = Some(format!("Gear program seed = '{seed}'"));

    gear_wasm_gen::gen_gear_program_code(&mut u, config)
}

#[test]
fn code_generation_is_deterministic() {
    use rand::rngs::SmallRng;

    let seed = 11943721913250886906;

    let first_code = generate_gear_program::<SmallRng>(seed);
    let second_code = generate_gear_program::<SmallRng>(seed);

    assert_eq!(first_code, second_code); // panics
}
```

SabaunT:
Resolves #2170.

Bug was basically with using `HashMap`/`HashSet` as during the iteration these types do not give guarantee on the order, so for the same input seed we will receive different iteration order of the same data.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3854,6 +3854,7 @@ dependencies = [
  "arbitrary",
  "gear-wasm-instrument",
  "indicatif",
+ "proptest",
  "rand 0.8.5",
  "wasm-smith",
  "wasmparser 0.99.0",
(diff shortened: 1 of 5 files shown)


breathx (approved):

gshep (approved):
Good catch! 💯

PR #2254, opened 2023-02-15
Switch to substrate v0.9.38

ukint-vs:
@gear-tech/dev

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -116,8 +116,8 @@ codegen-units = 1
 
 [patch.crates-io]
 # Two dependencies below go for `subxt v0.25.0`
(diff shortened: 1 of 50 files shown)


Cargo.toml
```diff
@@ -83,7 +83,8 @@ members = [
     "utils/call-gen",
     "utils/utils",
     "utils/bags-thresholds",
-    "utils/integral-tests",
+    # TODO: Remove gov deps
```
breathx, line 86: Isn't solved yet?

breathx (approved):

Cargo.toml
ekovalev: onto it shortly

ukint-vs: Part of this tests has been moved to staking-rewards pallet. The rest should be rewritten by @ekovalev later.

ekovalev (approved):

PR #2268, opened 2023-02-20
Fix `demo-meta` is being rebuilt each time

ark0f:


utils/wasm-builder/test-program/build.rs
```diff
@@ -16,6 +16,11 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+use std::fs;
+
 fn main() {
     gear_wasm_builder::build();
+
+    // to be built by other tests
+    fs::write("src/rebuild_test.rs", "").unwrap();
```
shamilsan, line 25: Why so complicated? Can't we just use `println!("cargo:rerun-if-...")`?

utils/wasm-builder/src/smart_fs.rs
```diff
@@ -0,0 +1,67 @@
...
+    assert_linear_comparison_size(metadata.len());
+    assert_linear_comparison_size(contents.len() as u64);
+
+    let old_data = fs::read(path)?;
+    let new_data = contents;
+    if old_data != new_data {
+        return Ok(true);
+    }
+
+    Ok(false)
```
shamilsan, line 55:
```suggestion
    Ok(old_data != new_data);
```

utils/wasm-builder/src/smart_fs.rs
```diff
@@ -0,0 +1,67 @@
...
+
+    if metadata.len() != contents.len() as u64 {
+        return Ok(true);
+    }
+
+    assert_linear_comparison_size(metadata.len());
+    assert_linear_comparison_size(contents.len() as u64);
+
+    let old_data = fs::read(path)?;
+    let new_data = contents;
```
shamilsan, line 50: Useless assign.
```suggestion
```

utils/wasm-builder/test-program/build.rs
ark0f: Cargo instruction will not help because `test-program` always includes the file

utils/wasm-builder/src/smart_fs.rs
ark0f: Simpler to understand

shamilsan: Rename `old_data` to `old_contents,` and there is no need to assign for "understanding".

utils/wasm-builder/src/smart_fs.rs
```diff
@@ -0,0 +1,67 @@
...
+    let Ok(metadata) = fs::metadata(path) else {
+        return Ok(true);
+    };
+
+    if metadata.len() != contents.len() as u64 {
+        return Ok(true);
+    }
+
+    assert_linear_comparison_size(metadata.len());
+    assert_linear_comparison_size(contents.len() as u64);
```
shamilsan, line 47: Why two asserts if we check `len`s equality earlier?

ark0f: Because we check each length with constant

shamilsan: If A == B and B == C, then no need to compare whether A == C (see transitiveness property).

We checked that `metadata.len() == contents.len()`. So it is sufficient to test only one of them against `LINEAR_COMPARISON_FILE_SIZE`.

shamilsan (approved):

PR #2305, opened 2023-02-27
Cross-compile Windows on CI

ark0f:


lazy-pages/src/init_flag.rs
```diff
@@ -0,0 +1,71 @@
...
+
+#[cfg(not(test))]
+pub use not_tests::*;
+
+#[cfg(test)]
+mod tests {
+    use crate::InitError;
+    use std::sync::Mutex;
+
+    #[cfg(test)]
```
shamilsan, line 48: Why do we need `#[cfg(test)]` inside `#[cfg(test)]`?

gclient/tests/perf.rs
```diff
@@ -219,7 +219,7 @@ async fn send_messages(api: &GearApi, progs: &HashMap<&str, ProgramId>) -> Resul
 
     println!("Message count: {}", messages.len());
 
-    assert!(block_gas_limit == 0);
+    assert_eq!(block_gas_limit, 0);
```
shamilsan, line 222: You may disable this test by removing this line at the moment as it occurred to be too sensitive to Wasm optimizations.

shamilsan (approved):
👍

ukint-vs (approved):

PR #2318, opened 2023-02-28
Refactor  gsdk

DennisInSky:
This PR fixes usage of the Deref trait. Apparently, having own funcs around &self for a struct implementing the Deref trait is considered an anti-pattern. If there are such functions, they should have a signature like pub fn some_func(signer: &Signer, ...) so that conflicts can be avoided.

@reviewer-or-team

### DIFF
--- a/gsdk/src/signer/mod.rs
+++ b/gsdk/src/signer/mod.rs
@@ -23,8 +23,6 @@ use crate::{
     Api,
 };
 pub use pair_signer::PairSigner;
-use std::ops::{Deref, DerefMut};
-
 use sp_core::{crypto::Ss58Codec, sr25519::Pair, Pair as PairT};
 use sp_runtime::AccountId32;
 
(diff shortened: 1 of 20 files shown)


gsdk/src/signer/mod.rs
```diff
@@ -37,7 +35,7 @@ mod utils;
 pub struct Signer {
```
DennisInSky, line 35: I would probably think of renaming this struct into something like `SignedApi` or `TxApi` as its main purpose is executing signed transactions whereas the `api()` func could be renamed into something `unsigned()`. In that case any client code would look like
```rust
let signed_api: SignedApi = ...;

let source = signed_api.source();

let gprog = signed_api.unsigned().gprog(some_prog_id);

```

shamilsan: I agree that current names are a little bit confusing.

Also, to me, this usage looks weird:

```rust
let signer = Api::new(...).signer(...);
let api = signer.api();
```

@clearloop What do you think about renaming? Let's choose the most appropriate names but in separate PR.

shamilsan (approved):
👍

breathx (approved):

gsdk/src/signer/mod.rs
breathx: to be considered in future

PR #2331, opened 2023-03-02
Add program migration functionality

### LINKED ISSUE #2055: Add a possibility to save and restore program's memory pages in `gtest`/`gclient`
### Problem to Solve

Sometimes it is helpful to have some not initial state of the program. E.g., you can perform some long during operation once and use its results in tests multiple times.

### Possible Solution

Add to `gtest`/`gclient`:

- Function for saving the whole program's memory to file: `save_memory(program_id, file_path)`
- Function for restoring the program's memory from a file: `restore_memory(program_id, file_path)`

Also: 

- Consider using a text format for the file to keep it in the version control system.
- Don't keep empty memory blocks.
- File format should be interchangeable between `gtest` and `gclient`.

### Notes

_No response_

DennisInSky:
Resolves #2055

### DIFF
--- a/gcli/src/cmd/program.rs
+++ b/gcli/src/cmd/program.rs
@@ -76,7 +76,7 @@ impl Program {
         // Get program
         let program = api.gprog(ProgramId(pid.0)).await?;
         let code_id = program.code_hash;
-        let code = api.gcode(code_id.0).await?;
+        let code = api.code_storage(code_id.0).await?;
         let pages = api.gpages(ProgramId(pid.0), &program).await?;
 
         // Query state
--- a/gcli/src/cmd/program.rs
+++ b/gcli/src/cmd/program.rs
@@ -19,7 +19,7 @@
 //! Command `program`.
 use crate::{metadata::Metadata, result::Result, utils};
 use clap::Parser;
-use gsdk::{ext::sp_core::H256, metadata::runtime_types::gear_core::ids::ProgramId, Api};
+use gsdk::{ext::sp_core::H256, Api};
 use std::{fs, path::PathBuf};
 
 /// Read program state, etc.
@@ -74,10 +74,10 @@ impl Program {
         } = self.action.clone();
 
         // Get program
-        let program = api.gprog(ProgramId(pid.0)).await?;
+        let program = api.gprog(pid.0.into()).await?;
         let code_id = program.code_hash;
-        let code = api.code_storage(code_id.0).await?;
-        let pages = api.gpages(ProgramId(pid.0), &program).await?;
+        let code = api.code_storage(code_id.0.into()).await?;
+        let pages = api.gpages(pid.0.into(), &program).await?;
 
         // Query state
         let state = Metadata::read(
--- a/gcli/src/cmd/program.rs
+++ b/gcli/src/cmd/program.rs
@@ -19,7 +19,7 @@
 //! Command `program`.
 use crate::{metadata::Metadata, result::Result, utils};
 use clap::Parser;
-use gsdk::{ext::sp_core::H256, Api};
+use gsdk::{ext::sp_core::H256, metadata::runtime_types::gear_core::ids::ProgramId, Api};
 use std::{fs, path::PathBuf};
 
 /// Read program state, etc.
@@ -74,10 +74,10 @@ impl Program {
         } = self.action.clone();
 
         // Get program
-        let program = api.gprog(pid).await?;
+        let program = api.gprog(ProgramId(pid.0)).await?;
         let code_id = program.code_hash;
-        let code = api.code_storage(code_id.0).await?;
-        let pages = api.gpages(pid, &program).await?;
+        let code = api.gcode(code_id.0).await?;
+        let pages = api.gpages(ProgramId(pid.0), &program).await?;
 
         // Query state
         let state = Metadata::read(
--- a/gclient/src/api/calls.rs
+++ b/gclient/src/api/calls.rs
@@ -38,7 +38,9 @@ use gsdk::{
         utility::Event as UtilityEvent,
         Event,
     },
+    Error as GsdkError,
 };
+use hex::ToHex;
 use parity_scale_codec::Encode;
 use std::{collections::BTreeMap, path::Path};
 
@@ -212,17 +214,35 @@ impl GearApi {
         src_program_id: ProgramId,
         dest_node_api: &GearApi,
     ) -> Result<ProgramId> {
-        // TODO: Ideally we have to check program is not already there (via gprog, for
-        // example). Do we want to override it? If yes, then we have to check
-        // migration is done to another node (network). How?
-        // Another solution might be that we crate a brand new ProgramId which would
-        // allow us to migrate program even withing the same node/network
+        if dest_node_api.0.api().gprog(src_program_id).await.is_ok() {
+            return Err(Error::ProgramAlreadyExists(
+                src_program_id.as_ref().encode_hex(),
+            ));
+        }
 
         let dest_program_id = src_program_id;
 
         // Collect data from the source program
-        let src_free_balance = self.free_balance(src_program_id).await?;
-        let src_reserved_balance = self.reserved_balance(src_program_id).await?;
+        let src_free_balance = self.free_balance(src_program_id).await.map_or_else(
+            |e| {
+                if let Error::GearSDK(GsdkError::StorageNotFound) = e {
+                    Ok(0)
+                } else {
+                    Err(e)
+                }
+            },
+            |v| Ok(v),
+        )?;
+        let src_reserved_balance = self.reserved_balance(src_program_id).await.map_or_else(
+            |e| {
+                if let Error::GearSDK(GsdkError::StorageNotFound) = e {
+                    Ok(0)
+                } else {
+                    Err(e)
+                }
+            },
+            |v| Ok(v),
+        )?;
 
         let src_program = self.0.api().gprog(src_program_id).await?;
 
--- a/gclient/src/api/calls.rs
+++ b/gclient/src/api/calls.rs
@@ -232,19 +232,27 @@ impl GearApi {
             .gpages(src_program_id.into(), &src_program)
             .await?;
 
-        let src_code_len = self.0.api().gcode_len(src_program.code_hash.into()).await?;
+        let src_code_len = self
+            .0
+            .api()
+            .code_len_storage(src_program.code_hash.into())
+            .await?;
 
-        let src_code = self.0.api().gcode(src_program.code_hash.into()).await?;
+        let src_code = self
+            .0
+            .api()
+            .code_storage(src_program.code_hash.into())
+            .await?;
 
         // Apply data to the target program
         tgt_node_api
             .0
-            .set_gcode_len(src_program.code_hash.0, src_code_len)
+            .set_code_len_storage(src_program.code_hash.0, src_code_len)
             .await?;
 
         tgt_node_api
             .0
-            .set_gcode(src_program.code_hash.0, &src_code)
+            .set_code_storage(src_program.code_hash.0, &src_code)
             .await?;
 
         tgt_node_api
--- a/gclient/src/api/calls.rs
+++ b/gclient/src/api/calls.rs
@@ -17,7 +17,7 @@
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
 use super::{GearApi, Result};
-use crate::{utils, Error};
+use crate::{api::storage::account_id::IntoAccountId32, utils, Error};
 use gear_core::ids::*;
 use gsdk::{
     ext::{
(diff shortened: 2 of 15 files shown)


gclient/src/api/calls.rs
```diff
@@ -204,6 +204,74 @@ impl GearApi {
...
+    pub async fn migrate_program(
+        &self,
+        src_program_id: ProgramId,
+        tgt_node_api: &GearApi,
+    ) -> Result<ProgramId> {
+        // TODO: Ideally we have to check program is not already there (via gprog, for
+        // example). Do we want to override it? If yes, then we have to check
+        // migration is done to another node (network). How?
+        // Another solution might be that we crate a brand new ProgramId which would
+        // allow us to migrate program even withing the same node/network
```
DennisInSky, line 219: What are your thoughts about this? Either I have to make a check that we migrate to another node or I can generate a brand-new ProgramId and check if the code if already there then skip its migration and just clone the program

gsdk/src/storage.rs
```diff
@@ -64,7 +64,7 @@ impl Api {
     }
 
     /// Get program pages from program id.
-    pub async fn program_pages(&self, pid: H256) -> Result<types::GearPages> {
+    pub async fn program_pages(&self, pid: ProgramId) -> Result<types::GearPages> {
```
DennisInSky, line 67: I think it is better to have more specific types whenever it is possible rather than abstract H256 or arrays

gsdk/src/storage.rs
```diff
@@ -172,7 +172,7 @@ impl Api {
 // pallet-gear-program
 impl Api {
     /// Get `InstrumentedCode` by `code_hash`
-    pub async fn code_storage(&self, code_hash: [u8; 32]) -> Result<InstrumentedCode> {
+    pub async fn gcode(&self, code_hash: [u8; 32]) -> Result<InstrumentedCode> {
```
DennisInSky, line 175: Renamed for better consistency - we have had gprog and gpages thus gcode

gsdk/src/signer/calls.rs
```diff
@@ -19,16 +19,27 @@
...
+        sp_weights::weight_v2::Weight,
+    },
     signer::Signer,
-    types::{InBlock, TxStatus},
+    types::{self, InBlock, TxStatus},
+    Error,
 };
 use anyhow::anyhow;
 use async_recursion::async_recursion;
 use gear_core::ids::{CodeId, MessageId, ProgramId};
```
DennisInSky, line 35: Inconsistent usage of IDs throughout this crate confuses me:
- in some places it uses crate::metadata::gear_core::ids::{ProgramId, etc} whereas in this place it uses gear_core::{ProgramId, etc}
- at the same time the crate exposes some methods which use this similar abstractions from different crates (see comment below)
Why can't we use `gear_core::ids` everywhere? Why do we need `crate::metadata::gear_core::ids` at all? It causes a lot of confusing conversions.

gsdk/src/signer/calls.rs
```diff
@@ -194,10 +205,102 @@ impl Signer {
...
+            program_pages_to_set.push((addr, value));
+        }
+        self.set_storage(&program_pages_to_set).await
+    }
+
+    /// Writes `ActiveProgram` into storage at `program_id`
+    pub async fn set_gprog(
+        &self,
+        program_id: ProgramId,
+        program: ActiveProgram,
```
DennisInSky, line 292: This is an example where `ProgramId` comes from the `gear_core` crate whereas `ActiveProgram` comes from this crate. At the same time the complimentary `gprog` method uses both `ProgramId` and `ActiveProgram` from this crate

SabaunT (commented):
The solution reminds me of our past pallet tests, when we mocked a lot state transitions, by setting directly some values to storage instead of calling extrinsics, which will do this for us the proper way. If you emulate in one function logic done by a set of processed extrinsics, it increases complexity of the code maintenance.
Also the functionality is questionable for complicated test-cases, for example when the storage of the program is not completely changed by some message in waitlist.

gclient/src/api/calls.rs
DennisInSky: As far as I am aware, this call is meant to be used for testing purposes by our community when they want to take a program from test/prod net and bring it into their dev net (#2055).

gsdk/src/storage.rs
DennisInSky: What kind of efficiency do we get here when we use `H256` in lieu of `ProgramId`. I do not see any except for confusion.

gsdk/src/storage.rs
DennisInSky: Alright, I got the idea and will rename it back.

gsdk/src/signer/calls.rs
DennisInSky: Right, I think we have to crate another issue and adjust the code so it makes use of `gear_core` as much as possible. Taking account of the fact this code is for our eco-system, the code has to be nice, clean and consistent as everything publicly facing

DennisInSky:
> The solution reminds me of our past pallet tests, when we mocked a lot state transitions, by setting directly some values to storage instead of calling extrinsics, which will do this for us the proper way. If you emulate in one function logic done by a set of processed extrinsics, it increases complexity of the code maintenance. Also the functionality is questionable for complicated test-cases, for example when the storage of the program is not completely changed by some message in waitlist.

I am not going to defend this solution, but as far as I get, this feature is not meant to be used for prod and its main purpose is to allow contract developers to migrate a program from test/prod into their own dev environment for some testing. If we'd had extrinsics allowing uploading a program with its state, I would've definitely used them. The question about delayed messages (messages sitting in waitlist) is still to be a subject for discussion as in general there is more general issue - different timeline (chain height) on source and target networks. @breathx mentioned that we can "move clocks" on the target node into future to make up for the source one. In addition to this we can prohibit migration if the target node is in the future compared to the source one.

gsdk/src/storage.rs
```diff
@@ -181,12 +181,22 @@ impl Api {
         self.fetch_storage(&addr).await
     }
 
+    /// Get `InstrumentedCode` length by its `code_hash`
+    pub async fn code_len_storage(&self, code_hash: [u8; 32]) -> Result<u32> {
```
breathx, line 185:
```suggestion
    pub async fn code_len_storage(&self, code_id: CodeId) -> Result<u32> {
```

gsdk/src/storage.rs
breathx: I see no efficiency improvements to use H256 as well
The only thing that for user lib we need use gear_core::ProgramId, while for gsdk we can go with metadata_generated::ProgramId

gsdk/src/signer/calls.rs
```diff
@@ -194,6 +205,101 @@ impl Signer {
...
+                "GearProgram",
+                "MemoryPageStorage",
+                vec![
+                    subxt::dynamic::Value::from_bytes(program_id.as_ref()),
+                    subxt::dynamic::Value::u128(*program_page.0 as u128),
+                ],
+            );
+            let page_buf_inner = gear_core::memory::PageBufInner::try_from(program_page.1.clone())
+                .map_err(|_| Error::PageBroken(*program_page.0, program_id.encode_hex()))?;
+            let value = gear_core::memory::PageBuf::from_inner(page_buf_inner);
```
breathx, line 282:
```suggestion
            let page_buf_inner = PageBufInner::try_from(program_page.1.clone())
                .map_err(|_| Error::PageBroken(*program_page.0, program_id.encode_hex()))?;
            let value = PageBuf::from_inner(page_buf_inner);
```
Import types directly and recheck if here proper constructor for such cases

gsdk/src/signer/calls.rs
breathx: Previously we had auto generated extrinsics and calls which were using metadata::ids::*
Due to the fact that we have now dynamic api, seems reasonable to use gear_core::ids everywhere (furthermore, publicly reimport it in gsdk/gclient maybe?)

gclient/src/api/calls.rs
breathx: In case of existance program on target destination just return error
It will require querying by program id at the begging of the function

breathx: > I assume this `migrate_program` call is designed for users, the problem is, users are not allowed to run the sudo calls

This fn will be used for local node, where sudo calls exist

> program migration maybe require us to add a new function in pallet-gear that allows us to submit program with state

Unbelievable api, we won't add something like that cause otherwise any smart-contract platforms become meaningless

gcli/src/cmd/program.rs
```diff
@@ -74,10 +74,10 @@ impl Program {
         } = self.action.clone();
 
         // Get program
-        let program = api.gprog(pid).await?;
+        let program = api.gprog(ProgramId(pid.0)).await?;
```
breathx, line 77: Lets deal with core ids

breathx (changes requested):
Requesting changes to prevent merging before solving all of the conversations properly.

Otherwise, lgtm 👍🏻

breathx:
> The solution reminds me of our past pallet tests, when we mocked a lot state transitions, by setting directly some values to storage instead of calling extrinsics, which will do this for us the proper way. If you emulate in one function logic done by a set of processed extrinsics, it increases complexity of the code maintenance.
> Also the functionality is questionable for complicated test-cases, for example when the storage of the program is not completely changed by some message in waitlist.

You're right, seems like our good old tests :D
But we had a lot of conversation with ecosystem and decided that we don't need any stuff related to waitlist and bns (at least for now)

breathx:
@DennisInSky Just realised that there is one important thing to be also added within the PR: migrate GasTree nodes representing reservations inside the program (specified in ActiveProgram structure). Don't forget to reserve proper amount of balance from GasNode origins to avoid post-execution panics. Test for reservation sending within the PR would be also good!

gsdk/src/storage.rs
DennisInSky: I would go with `gear_core::*` in gsdk as well as this dependency is already there and as @clearloop metioned it will allow us to get rid of at least some types duplicated in metadata which in turn will minimize maintenance burden

gclient/tests/node.rs
```diff
@@ -1,44 +1,128 @@
...
-    let node_2 = Node::try_from_path(GEAR_PATH).expect("Unable to instantiate dev node 2.");
+    let node_1 = Node::try_from_path(GEAR_PATH).expect("Unable to intsantiate dev node 1");
+    let node_2 = Node::try_from_path(GEAR_PATH).expect("Unable to instantiate dev node 2");
     let salt = gclient::now_in_micros().to_le_bytes();
 
     // The assumption is that it is not allowed to load the same code with the same
     // salt to the same node twice
-    upload_program_to_node(&node_1, WASM_BINARY, &salt).await;
-    upload_program_to_node(&node_2, WASM_BINARY, &salt).await;
+    upload_program_to_node(&node_1, demo_mul_by_const::WASM_BINARY, &salt, Some(42u64)).await;
```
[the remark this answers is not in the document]
DennisInSky: This won't work for the used dapp. It requires initialization payload

gsdk/src/signer/calls.rs
DennisInSky: @breathx I have imported the types, but I am not sure what you mean by `proper ctor` - there are no ctors for these types which take an array, there are only conversion methods which I use here

DennisInSky:
> @DennisInSky Just realised that there is one important thing to be also added within the PR: migrate GasTree nodes representing reservations inside the program (specified in ActiveProgram structure). Don't forget to reserve proper amount of balance from GasNode origins to avoid post-execution panics. Test for reservation sending within the PR would be also good!

@breathx I think I have addressed all the comments in this PR except for this one. Do you really want me to do this as a part of this PR or I can do it in the next one. Anyway, I have some questions about it so I would rather merge this code in.

gclient/src/api/calls.rs
```diff
@@ -204,6 +204,78 @@ impl GearApi {
             .await
     }
 
+    /// Migrates an active program identified by `src_program_id` onto another
+    /// node identified by `tgt_node_api` and returns the migrated program
+    /// identifier.
+    pub async fn migrate_program(
+        &self,
+        src_program_id: ProgramId,
+        tgt_node_api: &GearApi,
```
shamilsan, line 213: Usually, we use pair `src`/`dest`.
```suggestion
        dest_node_api: &GearApi,
```

gclient/src/api/calls.rs
```diff
@@ -204,6 +204,78 @@ impl GearApi {
...
+        let src_reserved_balance = self.reserved_balance(src_program_id).await?;
+
+        let src_program = self.0.api().gprog(src_program_id).await?;
+
+        let src_program_pages = self.0.api().gpages(src_program_id, &src_program).await?;
+
+        let src_code_len = self
+            .0
+            .api()
+            .code_len_storage(src_program.code_hash.0.into())
```
shamilsan, line 234: It is better to get code hash one time and keep it in a variable.

NikVolf:
Programs should preserve their id-s I think (or have a way to preserve)

Otherwise interacting programs will break

DennisInSky:
> Programs should preserve their id-s I think (or have a way to preserve)
> 
> Otherwise interacting programs will break

It's been preserved. There is even assertion of this fact in the test.

breathx (approved):
Waiting further discussed improvements 👍🏻

gclient/src/api/calls.rs
```diff
@@ -204,6 +204,72 @@ impl GearApi {
...
+    pub async fn migrate_program(
+        &self,
+        src_program_id: ProgramId,
+        dest_node_api: &GearApi,
+    ) -> Result<ProgramId> {
+        // TODO: Ideally we have to check program is not already there (via gprog, for
+        // example). Do we want to override it? If yes, then we have to check
+        // migration is done to another node (network). How?
+        // Another solution might be that we crate a brand new ProgramId which would
+        // allow us to migrate program even withing the same node/network
```
shamilsan, line 219: Have discussed, need to add an issue and reference to it here if the check will not be added to this PR.

shamilsan (approved):

gclient/src/api/calls.rs
DennisInSky: @shamilsan , @breathx I have fixed it in this PR. Please check this commit https://github.com/gear-tech/gear/pull/2331/commits/02f9036677434f3f733dbeef5ee1ff3e1fb6d49d. Found and fixed one bug while was doing this.

gclient/src/api/calls.rs
```diff
@@ -204,6 +206,90 @@ impl GearApi {
...
+        let src_free_balance = self.free_balance(src_program_id).await.map_or_else(
+            |e| {
+                if let Error::GearSDK(GsdkError::StorageNotFound) = e {
+                    Ok(0)
+                } else {
+                    Err(e)
+                }
+            },
+            |v| Ok(v),
+        )?;
```
shamilsan, line 235: Could we use `or_else` here?
```suggestion
        let src_free_balance = self.free_balance(src_program_id).await.or_else(
            |e| {
                if let Error::GearSDK(GsdkError::StorageNotFound) = e {
                    Ok(0)
                } else {
                    Err(e)
                }
            })?;
```

shamilsan (approved):

PR #2333, opened 2023-03-03
Fix `Result` aliases and typos

shamilsan:
- Fixed custom `Result` type aliases that allow using one `Result` type with both custom error and default one.
- Fixed some typos.

This PR allows avoiding such types of two `Result`s using:

https://github.com/gear-dapps/app/blob/master/src/contract.rs#L4


gshep (approved):
:+1:

PR #2266, opened 2023-02-18
return old small syscall weights

grishasobol:
Also fix problem with too big gas charging in send(reply)_input.
Add `CountersOwner` trait, which allows to operate with `ext` gas counters.
Move `STACK_END_EXPORT_NAME` to `gear-wasm-instruments` crate.

PR #2336, opened 2023-03-04
Fix `make pre-commit` and reveal broken tests.

### LINKED ISSUE #1998: Test no refund happens when a new sys-call is called except for `free`
### Problem to Solve

We must be sure that when execution ends, global gas counters (gas/allowance) are the _freshest_ stores of gas info, not the gas in the `Ext`.

### Possible Solution

Tests should be done on real backends for all sys-calls. Would suggest to do it after some simplification refactoring between core-processor and backend levels.

### Notes

_No response_

### LINKED ISSUE #2259: Rewrite free no excess refund test
### Problem to Solve

We now charge gas only on backend-side, none of costs of syscall thrown into ext level

### Possible Solution

`gear_core_processor::test_free_no_refund()`

### Notes

_No response_

breathx:
Resolves #2259.
Post factum resolves #1998 and problems occurred before #2266 merged.
Furthermore, fixes `make pre-commit` command and speed up clippy due to fixing incompatibility with `substrate-wasm-builder` monitored compilation env vars.
Deleted `gcli/Makefile` as redundant one.


SabaunT (approved):
👍

shamilsan:
Discussed with @grishasobol 
What do you think about adding to CI/pre-commit a check on whether the runtime requires new imports? A similar test as we do before runtime upgrade and release publishing?

breathx:
> Discussed with @grishasobol
> What do you think about adding to CI/pre-commit a check on whether the runtime requires new imports? A similar test as we do before runtime upgrade and release publishing?

We had similar conversation. Seems reasonable for me, lets do with separate PR

PR #2345, opened 2023-03-06
Create `.meta.wasm` only if `#[metawasm]` used

breathx:


examples/binaries/meta/src/lib.rs
```diff
@@ -26,8 +26,9 @@ mod code {
     include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
 }
 
+// TODO: delete once moved gcli on new reading state approach.
```
shamilsan, line 29: Do we have a task for it?

shamilsan (approved):

examples/binaries/meta/src/lib.rs
breathx: Will create it

PR #2346, opened 2023-03-06
Add test with the batch of messages filling the whole block

### LINKED ISSUE #2300: Test the network with the batch of messages filling the whole block
We need to write a script to see whether the network can process a batch of messages that fills the block in terms of weight.

- All programs should be different.
- Upload N programs with different code hashes.
- Send N messages (1 per program) to fill the whole block in terms of weight.

shamilsan:
Resolves #2300

TODO:
- [x] Add more examples


gclient/tests/perf.rs
```diff
@@ -0,0 +1,265 @@
...
+        block_gas_limit = block_gas_limit.saturating_sub(gas_info.burned);
+        if block_gas_limit == 0 {
+            break;
+        }
+        messages.push((prog_id, payload, gas_info.min_limit, 0));
+    }
+
+    println!("Message count: {}", messages.len());
+
+    assert!(block_gas_limit == 0); // TODO: Uncomment after implementation
```
gshep, line 222: Sorry comment is confusing a bit. And imho it is better to not place comments aside )

gshep (approved):
perfect! :100:

gclient/tests/perf.rs
shamilsan: Ah, forgot to remove it after uncommenting the line itself 🙈 
Thanks!

PR #2354, opened 2023-03-07
Update Schedule module. Charge for bytes len in `reply_*`.

ukint-vs:
- Switch from u64 to Weight in `HostFnWeights`.
- Add `instr_call_per_local` bench and weights.
- Impl dirty hack to include i64_const in template for weights.
- Update macros.
- Update weights and template

Add charging for length in `gr_reservation_reply_push` and `gr_reply_commit`

@gear-tech/dev


Cargo.lock
```diff
@@ -12341,7 +12341,7 @@ checksum = "97fee6b57c6a41524a810daee9286c02d7752c4253064d0b05472833a438f675"
 dependencies = [
  "cfg-if",
  "digest 0.10.6",
- "rand 0.8.5",
+ "rand 0.7.3",
```
breathx, line 12344: Why was that dumped?

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -1329,64 +1375,84 @@ benchmarks! {
...
 
-    // w_instr_call_indirect_per_param = w_bench - w_i64const
+    // w_instr_call_indirect_per_param = w_bench - 1 * w_param
     // Calling a function indirectly causes it to go through a thunk function whose runtime
     // linearly depend on the amount of parameters to this function.
+    // Please note that this is not necessary with a direct call.
     instr_call_indirect_per_param {
         let p in 0 .. T::Schedule::get().limits.parameters;
         let num_elements = T::Schedule::get().limits.table_size;
+        use self::code::TableSegment;
```
breathx, line 1447: mv above

pallets/gear/src/lib.rs
```diff
@@ -1004,28 +1004,38 @@ pub mod pallet {
             let pages_config = PagesConfig {
                 max_pages: schedule.limits.memory_pages.into(),
                 lazy_pages_weights: LazyPagesWeights {
-                    signal_read: CostPerPage::new(schedule.memory_weights.lazy_pages_read),
-                    signal_write: CostPerPage::new(schedule.memory_weights.lazy_pages_write),
+                    signal_read: CostPerPage::new(
```
breathx, line 1007: can't we hide ref_time for some macro mb? Why the schedule should use exactly Weight type?

breathx: seems carelessly for me

pallets/gear/src/schedule.rs
```diff
@@ -36,10 +37,11 @@ use sp_std::{marker::PhantomData, vec::Vec};
 /// How many API calls are executed in a single batch. The reason for increasing the amount
 /// of API calls in batches (per benchmark component increase) is so that the linear regression
 /// has an easier time determining the contribution of that component.
-pub const API_BENCHMARK_BATCH_SIZE: u32 = 100;
+pub const API_BENCHMARK_BATCH_SIZE: u32 = 80;
```
breathx, line 40: why do we dump these amounts of calls?

breathx: What about to go with 20 / 50 as is done in pallet contracts?

breathx (approved):
👍🏻

breathx:
Did you try node-loader locally?

pallets/gear/src/benchmarking/mod.rs
ukint-vs: should be after ranges

pallets/gear/src/schedule.rs
ukint-vs: > What about to go with 20 / 50 as is done in pallet contracts?

They have 80/100 in constants. 100 for instructions don't have time to warm up lazy pages.

pallets/gear/src/benchmarking/syscalls.rs
```diff
@@ -835,6 +835,76 @@ where
...
+                // rid_value ptr
+                Instruction::I32Const(rid_value_offset as i32),
+                // payload ptr
+                Instruction::I32Const(payload_offset),
+                // payload len
+                Instruction::I32Const(payload_len),
+                // delay
+                Instruction::I32Const(10),
+                // err_mid ptr
+                Instruction::I32Const(err_mid_offset as i32),
```
ark0f, line 873: Possibly `i32.const` must be used instead of counter in other benches too

pallets/gear/src/benchmarking/code.rs
```diff
@@ -319,6 +309,21 @@ where
         .into()
     }
 
+    /// Same as `dummy` but with maximum sized linear memory and a dummy section of specified size.
+    pub fn dummy_with_bytes(dummy_bytes: u32) -> Self {
```
ark0f, line 313: Unused function

pallets/gear/src/benchmarking/code.rs
```diff
@@ -85,11 +85,6 @@ pub struct ModuleDefinition {
     /// The amount of I64 arguments the aux function should have.
     pub aux_arg_num: u32,
     pub aux_res: Option<ValueType>,
-    /// If set to true the stack height limiter is injected into the the module. This is
-    /// needed for instruction debugging because the cost of executing the stack height
-    /// instrumentation should be included in the costs for the individual instructions
-    /// that cause more metering code (only call).
-    pub inject_stack_metering: bool,
```
ark0f, line 92: Why it has been removed?

ark0f (changes requested):

core-backend/sandbox/src/funcs.rs
```diff
@@ -520,7 +520,7 @@ where
 
         ctx.run_fallible::<_, _, LengthWithHash>(
             err_mid_ptr,
-            RuntimeCosts::ReservationReply,
```
grishasobol, line 523: gr_reply / gr_reply_wgas should also be charged for payload len

pallets/gear/src/schedule.rs
```diff
@@ -792,8 +830,8 @@ impl<'a, T: Config> gas_metering::Rules for ScheduleRules<'a, T> {
         let max_params = self.schedule.limits.parameters;
 
         let weight = match *instruction {
-            End | Unreachable | Return | Else | Block(_) | Loop(_) | Nop | Drop => 0,
-            I32Const(_) | I64Const(_) => w.i64const,
+            End | Unreachable | Return | Else => 0,
```
grishasobol, line 833: I had set 0 cost for loop, nop and etc instructions to zero, because they 100% removed in wasmer, so we shouldn's set weight for them.

pallets/gear/src/schedule.rs
```diff
@@ -521,18 +548,24 @@ macro_rules! cost_byte_batched {
     };
 }
 
+macro_rules! to_weight {
```
grishasobol, line 551: what the reason to use weight instead of u64? and to release this as macros, if there is no places where we se proof size? we can use simple function here I think.

gcli/Cargo.toml
```diff
@@ -27,7 +27,7 @@ hex = "0.4.3"
 jsonrpsee = { version = "^0.16", features = [ "http-client", "ws-client" ] }
 keyring = "1.2.1"
 lazy_static = "1.4.0"
-libp2p = { version = "0.51.0", features = [ "identify" ], optional = true }
+libp2p = { version = "0.51.1", features = [ "identify" ], optional = true }
```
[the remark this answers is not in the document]
ukint-vs: Updated Cargo.lock

pallets/gear/src/benchmarking/code.rs
ukint-vs: No need

pallets/gear/src/schedule.rs
ukint-vs: > what the reason to use weight instead of u64? and to release this as macros, if there is no places where we se proof size? we can use simple function here I think.

we probably want to use proof_size in the future

pallets/gear/src/benchmarking/syscalls.rs
ukint-vs: We can't use `body::plain()` in other benches. It won't be accurate. `ERR_LEN_SIZE` is 4 bytes and it can't exceed even 1 gear page.

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -1203,7 +1214,7 @@ benchmarks! {
         sbox.invoke();
     }
 
-    // w_select = w_bench - 2 * w_i64const
+    // w_select = w_bench - 4 * w_param
```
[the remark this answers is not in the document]
ukint-vs: This is how weight is calculated.

pallets/gear/src/benchmarking/code.rs
ark0f: `inject_stack_metering()` is unused and should be removed then

pallets/gear/src/schedule.rs
```diff
@@ -605,69 +647,59 @@ impl Default for Limits {
...
+            call_indirect: cost_instr!(instr_call_indirect, 1),
+            call_indirect_per_param: cost_instr!(instr_call_indirect_per_param, 1),
+            call_per_local: cost_instr!(instr_call_per_local, 1),
+            local_get: cost_instr!(instr_local_get, 0),
+            local_set: cost_instr!(instr_local_set, 1),
+            local_tee: cost_instr!(instr_local_tee, 1),
+            global_get: cost_instr!(instr_global_get, 0),
+            global_set: cost_instr!(instr_global_set, 1),
+            memory_current: cost_instr!(instr_memory_current, 1),
+            i64clz: cost_instr!(instr_i64clz, 2),
```
grishasobol, line 671: Please check all this numbers: we should ignore `Drop` and `I32Const` because this instructions are optimised by wasmer and in has weight 0 often.

grishasobol (changes requested):

pallets/gear/src/schedule.rs
```diff
@@ -605,69 +647,59 @@ impl Default for Limits {
...
+            local_set: cost_instr!(instr_local_set, 1),
+            local_tee: cost_instr!(instr_local_tee, 1),
+            global_get: cost_instr!(instr_global_get, 0),
+            global_set: cost_instr!(instr_global_set, 1),
+            memory_current: cost_instr!(instr_memory_current, 1),
+            i64clz: cost_instr!(instr_i64clz, 1),
+            i64ctz: cost_instr!(instr_i64ctz, 1),
+            i64popcnt: cost_instr!(instr_i64popcnt, 1),
+            i64eqz: cost_instr!(instr_i64eqz, 1),
+            i64extendsi32: cost_instr!(instr_i64extendsi32, 1),
```
grishasobol, line 675: for i64extendsi32 and i64extendui32 number must be 0 because it uses only `drop` for result and `i32const` for argument. Both will be optimised and removed (i32const often when const argument can be inserted in instruction logic)

PR #2361, opened 2023-03-08
Revert "Build WASM projects in their own unified directory (#1126)"

ark0f:
This reverts commit 5fdcf478

PR #2362, opened 2023-03-08
feat(wasm-builder): introduce wasm-opt as a feature

clearloop:
Ref https://github.com/gear-tech/gear/pull/2361

- [x] revert [d894f76](https://github.com/gear-tech/gear/pull/2362/commits/d894f76c418a31c2b69e9fc9ce1869343ab2c48b)
- [x] introduce feature `wasm-opt` for `wasm-builder`
- [x] test case `btree`

@gear-tech/dev

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -82,8 +82,8 @@ jobs:
         run: |
           sudo apt update
           sudo apt install -y git clang curl libssl-dev llvm libudev-dev cmake protobuf-compiler wget bzip2
-          sudo wget -c https://github.com/WebAssembly/binaryen/releases/download/version_105/binaryen-version_105-x86_64-linux.tar.gz -O - | sudo tar -xz -C .
-          sudo cp binaryen-version_105/bin/wasm-opt /usr/bin/
+          sudo wget -c https://github.com/WebAssembly/binaryen/releases/download/version_111/binaryen-version_111-x86_64-linux.tar.gz -O - | sudo tar -xz -C .
+          sudo cp binaryen-version_111/bin/wasm-opt /usr/bin/
           sudo wget -c https://github.com/paritytech/cachepot/releases/download/0.1.0-rc.2/cachepot-0.1.0-rc.2-x86_64-unknown-linux-musl.tar.gz -O - | sudo tar -xz -C .
           sudo cp cachepot-0.1.0-rc.2-x86_64-unknown-linux-musl/cachepot /usr/bin/
           sudo chmod +x /usr/bin/cachepot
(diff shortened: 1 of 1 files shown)


utils/wasm-proc/src/main.rs
```diff
@@ -188,14 +188,17 @@ fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-        let res = gear_wasm_builder::optimize::optimize_wasm(file.clone(), "s", true)?;
-
-        log::info!(
-            "wasm-opt: {} {} Kb -> {} Kb",
-            res.dest_wasm.display(),
-            res.original_size,
-            res.optimized_size
-        );
+
+        #[cfg(feature = "wasm-opt")]
```
gshep, line 192: that is unnecessary imho since `gear_wasm_builder` is able to optimize in both cases.

gshep (approved):
:+1:

utils/wasm-proc/src/main.rs
shamilsan: I'd use `wasm-opt` in `wasm-proc` by default as it is the once-building tool.

shamilsan: > wondering why we annotated these lines before

I guess it was done to prevent the developer from installing `binaryen`. But if we use the `wasm-opt` crate by default in `wasm-proc`, we'll solve the problem.

gshep: nope it is due to the issue with the new binaryen - https://github.com/gear-tech/gear/pull/1972/files#r1045663991

The issue relates to signing-extensions whatsoever, @clearloop found the corresponding flags for wasm-opt library:  https://github.com/gear-tech/gear/commit/52e4cc5b95efddbd37c4320f97a208639f94df1e#diff-4d80c505a5a2697acc24e5218d2722d7c0148c70a7ac164fb5224ad9d0ed51d7R185

gshep: So frankly speaking we also need to set corresponding flag for the `wasm-opt` binary

utils/wasm-builder/Cargo.toml
```diff
@@ -17,9 +17,11 @@ thiserror = "1.0.39"
...
 which = "4.4.0"
+colored = "2.0.0"
 gmeta = { path = "../../gmeta", default-features = false }
 gear-wasm-instrument = { path = "../wasm-instrument", default-features = false }
-wasm-opt = "0.111.0"
+wasm-opt = { version = "0.111.0", optional = true }
 
 [features]
 metawasm = ["gmeta/codegen"]
+wasm-opt = ["dep:wasm-opt"]
```
ark0f, line 27: The line is unnecessary

ark0f (approved):

PR #2363, opened 2023-03-08
prune locks map in gstd

grishasobol:
without this patch locks map will grows forever for any contract with asyncs 

Change internal map for locks back to BTreeMap, because I think is more probably, that internal map for messages is small, so it's more effective to use BTreeMap than HashMap for small data

### DIFF
--- a/gstd/src/async_runtime/futures.rs
+++ b/gstd/src/async_runtime/futures.rs
@@ -68,6 +68,7 @@ where
 
     if Pin::new(&mut task.future).poll(&mut cx).is_ready() {
         super::futures().remove(&msg_id);
+        super::locks().remove_message_entry(msg_id);
     } else {
         super::locks().wait(msg_id);
     }
--- a/gstd/src/async_runtime/locks.rs
+++ b/gstd/src/async_runtime/locks.rs
@@ -20,9 +20,8 @@
 use crate::{
     config::WaitType,
     errors::{ContractError, Result},
-    exec, Config, MessageId,
+    exec, BTreeMap, Config, MessageId,
 };
-use alloc::collections::BTreeMap;
 use core::cmp::Ordering;
 use hashbrown::HashMap;
 
@@ -157,7 +156,7 @@ impl LocksMap {
         // Locks with `deadline < now`, they should already been removed since
         // the node will trigger timeout when locks reaching their deadline.
         //
-        // Locks with `deadline == now`, they will be removed in the following pollings.
+        // Locks with `deadline <= now`, they will be removed in the following polling.
         let now = exec::block_height();
         map.iter()
             .filter_map(|(_, lock)| (lock.deadline() > now).then_some(lock))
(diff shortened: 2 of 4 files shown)


gstd/src/lib.rs
```diff
@@ -31,6 +31,8 @@
 #[cfg(target_arch = "wasm32")]
 extern crate galloc;
 
+extern crate alloc;
```
breathx, line 34: alloc and btreemap is already exported in prelude mod

gstd/src/async_runtime/locks.rs
```diff
@@ -158,15 +157,13 @@ impl LocksMap {
         // Locks with `deadline < now`, they should already been removed since
         // the node will trigger timeout when locks reaching their deadline.
         //
-        // Locks with `deadline == now`, they will be removed in the following
-        // pollings.
+        // Locks with `deadline == now`, they will be removed in the following pollings.
```
breathx, line 160:
```suggestion
        // Locks with `deadline <= now`, they will be removed in the following polling.
```

gstd/src/async_runtime/futures.rs
```diff
@@ -68,6 +68,7 @@ where
 
     if Pin::new(&mut task.future).poll(&mut cx).is_ready() {
         super::futures().remove(&msg_id);
+        super::locks().remove_message_entry(msg_id);
```
breathx, line 71: One more place to remove lock
https://github.com/gear-tech/gear/blob/7c0ae01484e77bf4d5f5d7dcd830829b68011263/gstd/src/async_runtime/mod.rs#L58-L60

gstd/src/async_runtime/locks.rs
```diff
@@ -181,6 +178,16 @@ impl LocksMap {
...
     }
 
+    pub fn remove_message_entry(&mut self, message_id: MessageId) {
+        self.0.remove(&message_id);
+        // Question for reviewers: how map for message can be not empty at this
+        // point ?!!! .map(|locks| {
+        //     if !locks.is_empty() {
+        //         // unreachable!("Wanna to remove locks map for message, but
+        // the map is not empty: {:?}", locks);     }
+        // });
```
breathx, line 188: It can't so the panic on `None` removal res seems ok

breathx (changes requested):

gstd/src/lib.rs
grishasobol: ok removed

gstd/src/async_runtime/locks.rs
grishasobol: done

gstd/src/async_runtime/futures.rs
grishasobol: added

gstd/src/async_runtime/locks.rs
grishasobol: This panic is appeared some times, so we should discuss it more

breathx: discussed

breathx (approved):

PR #2364, opened 2023-03-09
gclient: Add dev Vara node spawning

shamilsan:
- Made it possible to spawn a Vara node in the dev-mode.

@gear-tech/dev

### DIFF
--- a/gclient/src/node/mod.rs
+++ b/gclient/src/node/mod.rs
@@ -36,16 +36,45 @@ pub struct Node {
 }
 
 impl Node {
-    /// Instantiates a node runnin in development mode (--dev) via spawning its
-    /// process using the specified `path` on a randomly picked port. Waits for
-    /// the node getting initialized before returning it to a caller.
-    /// Ideally, the node's binary should be downloaded by means of CI pipeline
-    /// from https://get.gear.rs./
+    /// Instantiates a Gear node running in development mode (`--dev`) via
+    /// spawning its process using the specified `path` on a randomly picked
+    /// port. Waits for the node getting initialized before returning it to
+    /// a caller. Ideally, the node's binary should be downloaded by means
+    /// of CI pipeline from <https://get.gear.rs>.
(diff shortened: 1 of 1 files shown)


gclient/src/node/mod.rs
```diff
@@ -36,16 +36,45 @@ pub struct Node {
...
+    pub fn ws_address(&self) -> &WSAddress {
+        &self.ws_address
+    }
+
+    fn try_node_from_path(path: impl AsRef<OsStr>, args: Vec<&str>) -> Result<Self> {
         let port = port::pick();
         let port_string = port.to_string();
 
-        let args = vec!["--ws-port", &port_string, "--tmp", "--dev"];
+        let mut args = args;
```
gshep, line 75: looks like there is no need for mutable rebinding. Argument can be marked `mut` directly in paratemeters list

gshep (approved):
:+1:

gclient/src/node/mod.rs
shamilsan: I don't like `mut` in the function signature when it is not required there by the function contract as it looks a little bit confusing in my opinion.

shamilsan: But after reading this comment I've changed my mind: https://stackoverflow.com/questions/31388644/whats-the-idiomatic-way-to-pass-by-mutable-value#comment50755158_31388745

shamilsan: But I get another error 🙈 

```
error[E0597]: `port_string` does not live long enough
  --> gclient/src/node/mod.rs:77:19
   |
71 |     fn try_node_from_path(path: impl AsRef<OsStr>, mut args: Vec<&str>) -> Result<Self> {
   |                                                                  - let's call the lifetime of this reference `'1`
...
77 |         args.push(&port_string);
   |         ----------^^^^^^^^^^^^-
   |         |         |
   |         |         borrowed value does not live long enough
   |         argument requires that `port_string` is borrowed for `'1`
...
93 |     }
   |     - `port_string` dropped here while still borrowed
```

So let's keep it as is.

PR #2377, opened 2023-03-11
pack cargo registry and set env

ukint-vs:
Resolves # .

-
-
-

@reviewer-or-team

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -10,7 +10,10 @@ on:
   workflow_dispatch:
 
 env:
+  CARGO_INCREMENTAL: 0
+  CARGO_REGISTRIES_CRATES_IO_PROTOCOL: sparse
   CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: short
   TERM: xterm-256color
   NIGHTLY_TOOLCHAIN_VERSION: ${{ secrets.NIGHTLY_TOOLCHAIN_VERSION }}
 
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yaml
```diff
@@ -94,6 +97,7 @@ jobs:
         continue-on-error: true
         run: |
           tar -xf /root/cache/check_cache_${{ github.base_ref }}.tar -C /
+          tar -xf /root/cache/check_cargo_registry_${{ github.base_ref }}.tar -C /
```
breathx, line 100:
```suggestion
          tar -xf /root/cache/check_cargo_registry_${{ github.base_ref }}.tar -C
```

utils/node-loader/src/batch_pool/api/mod.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::redundant_async_block)]
```
breathx, line 1: Is it solvable issue?

breathx (approved):

utils/node-loader/src/batch_pool/api/mod.rs
ukint-vs: Clippy suggestion is not working.

.github/workflows/CI.yaml
ukint-vs: Why

PR #2379, opened 2023-03-11
[gclient] Get all messages from the mailbox

### LINKED ISSUE #2325: gclient: Get all messages from the mailbox
### Problem to Solve

E.g., we need to get a message from the mailbox that a program has sent. We need a message identifier, but we can't obtain it.

### Possible Solution

~Need to add a method to `EventProcessor` or `EventListener` that will process events and find the `UserMessageSent` event and return its details. It is similar to the `EventProcessor::reply_bytes_on` function but should capture arbitrary messages.~

Add a function to get all messages from the mailbox.

### Notes

_No response_

ukint-vs:
Resolves #2325.

@gear-tech/dev


gclient/src/api/storage/mod.rs
```diff
@@ -58,6 +58,13 @@ impl GearApi {
         Ok(data.map(|(m, i)| (m.into(), i)))
     }
 
+    /// Get up to `count` messages from the mailbox.
+    pub async fn get_all_mailbox(&self, count: u32) -> Result<Vec<(StoredMessage, Interval<u32>)>> {
+        let data: Vec<(stored::StoredMessage, Interval<u32>)> =
```
shamilsan, line 63: No need to be so explicit.
```suggestion
        let data =
```

gsdk/src/storage.rs
```diff
@@ -286,6 +286,26 @@ impl Api {
 
         Ok(mailbox)
     }
+
+    /// Get up to `count` messages from the mailbox
+    pub async fn all_mailbox(&self, count: u32) -> Result<Vec<(StoredMessage, Interval<u32>)>> {
```
shamilsan, line 291: I'd rename this method to the `mailbox` and current `mailbox` to the `account_mailbox`.

gsdk/src/storage.rs
```diff
@@ -286,6 +286,26 @@ impl Api {
...
+        let storage = self.storage().at(None).await?;
+        let query_key =
+            storage_address_root_bytes(&subxt::dynamic::storage_root("GearMessenger", "Mailbox"));
+
+        let keys = storage.fetch_keys(&query_key, count, None).await?;
+
+        let mut mailbox: Vec<(StoredMessage, Interval<u32>)> = vec![];
```
shamilsan, line 298: Type can be inferred below.
```suggestion
        let mut mailbox = vec![];
```

gsdk/src/storage.rs
```diff
@@ -286,6 +286,26 @@ impl Api {
...
+        for key in keys.into_iter() {
```
shamilsan, line 299: No need to convert `Vec` to `Iterator`.
```suggestion
        for key in keys {
```

gsdk/src/storage.rs
shamilsan, line 291: This method is too similar to the `mailbox` method. Better would be to generalize them two.
```suggestion
    /// Get up to `count` messages from the mailbox.
    pub async fn all_mailbox(&self, count: u32) -> Result<Vec<(StoredMessage, Interval<u32>)>> {
```

gclient/src/api/storage/mod.rs
```diff
@@ -58,6 +58,13 @@ impl GearApi {
         Ok(data.map(|(m, i)| (m.into(), i)))
     }
 
+    /// Get up to `count` messages from the mailbox.
+    pub async fn get_all_mailbox(&self, count: u32) -> Result<Vec<(StoredMessage, Interval<u32>)>> {
```
shamilsan, line 62: Need to review all function names, as they looks inconsistent now. As an option:

- `get_from_mailbox` -> `get_mailbox_message`
- `get_from_account_mailbox` -> `get_account_mailbox_message`
- `get_all_mailbox` -> `get_mailbox_messages`

gclient/tests/upload.rs
```diff
@@ -113,3 +115,55 @@ async fn alloc_zero_pages() -> Result<()> {
...
+    let (messages, _hash) = api.send_message_bytes_batch(messages).await?;
+
+    let (message_id, _hash) = messages.last().unwrap().as_ref().unwrap();
+
+    assert!(listener.message_processed(*message_id).await?.succeed());
+
+    let mailbox = api.get_all_mailbox(15).await?;
+
+    // Check that all messages is in mailbox
+    assert_eq!(mailbox.len(), 5);
```
shamilsan, line 166: We may additionally test that they are all `PONG`s I guess.

gsdk/src/storage.rs
shamilsan: Also, we need to filter messages by self account ID as a destination. We are interested in our messages instead of someone else's. We may provide two functions in `gclient` as we do it for one message. E.g., `get_mailbox_messages` (destination is signer) and `get_account_mailbox_messages` (destination is set explicitly as a parameter).

shamilsan (approved):
🔥

gclient/tests/upload.rs
```diff
@@ -27,6 +27,8 @@ const PATHS: [&str; 2] = [
     "../target/wat-examples/inf_recursion.wasm",
 ];
 
+const PING_PATH: &str = "../target/wasm32-unknown-unknown/release/demo_ping.opt.wasm";
```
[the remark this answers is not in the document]
DennisInSky: I meant referencing the binary this way
```rust
demo_mul_by_const::WASM_BINARY
```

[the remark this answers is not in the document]
ukint-vs: > I meant referencing the binary this way instead of using path string
> 
> ```rust
> demo_mul_by_const::WASM_BINARY
> ```

This can only be done with `examples/binaries` as they have the `WASM_BINARY` constant.

PR #2384, opened 2023-03-13
Support migration of a program with gas reservations

DennisInSky:
@reviewer-or-team


pallets/gear/src/lib.rs
```diff
@@ -188,8 +188,15 @@ pub mod pallet {
         /// Balances management trait for gas/value migrations.
         type Currency: LockableCurrency<Self::AccountId> + ReservableCurrency<Self::AccountId>;
 
+        /// Multiplier used for conversion from gas into balance.
+        #[pallet::constant]
+        type GasToBalanceMultiplier: Get<BalanceOf<Self>>;
```
DennisInSky, line 193: This has got exposed, but in fact I do not need it for the implemented functionality

DennisInSky:
With regards to the gsdk methods reading values from storage, @breathx you wanted us to introduce the `block_hash` param for each of them. How are we gonna do this:
1. A pair of methods like `something_from_storage` + `something_from_storage_at` (pretty much like I have done in this PR)
2. Just a single method taking the `block_hash` param which replaces the old methods

gsdk/src/storage.rs
```diff
@@ -73,10 +86,19 @@ impl Api {
 impl Api {
     /// Get account info by address
     pub async fn info(&self, address: &str) -> Result<AccountInfo<u32, AccountData<u128>>> {
+        self.info_at(address, None).await
+    }
+
+    /// Get account info by address at specified block
```
shamilsan, line 92:
```suggestion
    /// Get account info by address at specified block.
```

gsdk/src/storage.rs
```diff
@@ -138,6 +160,37 @@ impl Api {
     }
 }
 
+// pallet-gas
+impl Api {
+    /// Get Gear gas nodes by their ids.
+    pub async fn gas_nodes(
```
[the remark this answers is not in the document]
breathx: to auto generate similar boilerplate code or somehow figure out the way to provide some special closure that will run fn with such arg (no idea atm, we can consider later)

pallets/gear/src/lib.rs
breathx: As it become useless I'd remove the stuff cause it may cause some compatibility problems in future for JS side, metadata etc, as less thing we throw in config as better it is

breathx: See no reason to keep it now

breathx (approved):

gclient/src/api/calls.rs
```diff
@@ -208,10 +211,12 @@ impl GearApi {
...
     /// Migrates an active program identified by `src_program_id` onto another
     /// node identified by `dest_node_api` and returns the migrated program
-    /// identifier.
+    /// identifier. All source program data is taken at the time of
+    /// `src_block_hash` if it is specified or the most recent one.
     pub async fn migrate_program(
         &self,
         src_program_id: ProgramId,
+        src_block_hash: Option<H256>,
         dest_node_api: &GearApi,
```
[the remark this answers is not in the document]
DennisInSky: I would prefer to keep it as it is. The reason is that you have params groupped so you read it `I want to migrate src_program_id at block scr_block hash onto dst_node_api`. The other methods you use as example have a bit different semantics.

gclient/src/api/calls.rs
```diff
@@ -220,45 +225,126 @@ impl GearApi {
...
-            }
-        })?;
-        let src_reserved_balance = self.reserved_balance(src_program_id).await.or_else(|e| {
+        let src_program_account_data = self
+            .account_data_at(src_program_id, src_block_hash)
+            .await
+            .or_else(|e| {
             if let Error::GearSDK(GsdkError::StorageNotFound) = e {
-                Ok(0)
+                Ok(AccountData {
```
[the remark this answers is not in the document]
DennisInSky: Unfortunatelly it does not have it

PR #2396, opened 2023-03-14
Add checking that gear stack end pointer does not overlap

### LINKED ISSUE #2334: Reject binaries with incorrect stack pointer
### File Location(s)

_No response_

### Proposal

Program binaries export `__gear_stack_end` pointer if there is one. Reject the binaries that have the stack pointer overlapping data section(s).

gshep:
Resolves #2334

@gear-tech/dev

### DIFF
--- a/core/src/code.rs
+++ b/core/src/code.rs
@@ -25,15 +25,17 @@ use crate::{
 };
 use alloc::{collections::BTreeSet, vec::Vec};
 use codec::{Decode, Encode};
+use core::ops::ControlFlow;
 use gear_wasm_instrument::{
     parity_wasm::{
         self,
-        elements::{Internal, Module},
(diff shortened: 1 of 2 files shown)


ukint-vs (approved):

core/src/code.rs
```diff
@@ -68,6 +70,72 @@ fn get_exports(
     Ok(exports)
 }
 
+fn get_stack_end_init_code(module: &Module) -> Option<&[Instruction]> {
+    let Some(section) = module.export_section() else {
+        return None;
+    };
```
shamilsan, line 76: You can use `?` with `Option`.
```suggestion
    let section = module.export_section()?;
```

breathx (approved):

PR #2406, opened 2023-03-15
Add benchmarks for i32 operations

### LINKED ISSUE #1815: BENCHMARKING: investigate and make if need separate benches for 32-bit instructions
### Problem to Solve

Currently we use the same weights for 64 bits wasm instructions and for their 32 bit corresponding analogs. But actually 32 bit instructions can be cheaper, because 32 bit constants/literal arguments are smaller and also they have more possibility to be optimised. 

### Possible Solution

Investigate and make if need.

### Notes

_No response_

DennisInSky:
Resolves #1815 
@reviewer-or-team


pallets/gear/src/benchmarking/code.rs
```diff
@@ -371,12 +371,24 @@ where
     }
 
     pub fn unary_instr(instr: Instruction, repeat: u32) -> Self {
```
DennisInSky, line 373: Can rename this into unary_instr_64 if it would look better

pallets/gear/src/benchmarking/code.rs
```diff
@@ -387,12 +399,24 @@ where
     }
 
     pub fn binary_instr(instr: Instruction, repeat: u32) -> Self {
```
DennisInSky, line 401: Can rename this into `binary_instr_64` if it would look better

pallets/gear/src/benchmarking/code.rs
ukint-vs: sure

pallets/gear/src/benchmarking/code.rs
```diff
@@ -371,12 +371,24 @@ where
     }
 
     pub fn unary_instr(instr: Instruction, repeat: u32) -> Self {
-        use body::DynInstr::{RandomI64Repeated, Regular};
+        Self::unary_instr_for_arch(instr, true, repeat)
+    }
+
+    pub fn unary_instr_32(instr: Instruction, repeat: u32) -> Self {
+        Self::unary_instr_for_arch(instr, false, repeat)
```
gshep, line 378: I would not be obsessive with primitives: introduce new `enum Arch { Amd64, X86 }` and implement `From` trait for `DynInstr` (or custom function if the latter is impossible).

DennisInSky: The latter would be nice, but in reality it will require matchiing all the `Instruction` variants which I am not keen on. Will go with a enum as a param instead of bool

DennisInSky: But even with the enum? Does it worth introducing it in this private code?

gshep: From my point of view it is worth since improves code readability however I don't insist ).
What concerns conversion function I thought there is no need to match all `Instruction` : it should convert from `Arch` enum to `DynInstr`, something like this:
```rust
fn to_dyn_instr(arch: &Arch, count: i32) -> body::DynInstr {
    match arch {
        Arch::Amd64 => body::DynInstr::RandomI64Repeated(count),
        Arch::X86 => body::DynInstr::RandomI32Repeated(count),
    }
}
```

DennisInSky: Ahh... I see what you mean. Nice one. Done. Thanks

pallets/gear/src/benchmarking/code.rs
```diff
@@ -600,3 +616,17 @@ where
 {
     T::Schedule::get().limits.memory_pages
 }
+
+enum BitWidth {
```
breathx, line 620: some comments would be also good for any newcomers

pallets/gear/src/weights.rs
```diff
@@ -959,6 +990,12 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
             // Standard Error: 63_779
             .saturating_add(Weight::from_ref_time(9_522_432).saturating_mul(r.into()))
     }
+    /// The range of component `r` is `[0, 50]`.
+    fn instr_i32load(r: u32, ) -> Weight {
```
breathx, line 994: could you copy weights from x64 analogs for newly added weights or do proper benches on special hardware we use.
these seem to be benched on your local machine, isn't it?

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -1279,6 +1298,25 @@ benchmarks! {
         sbox.invoke();
     }
 
+    // w_store = w_bench - w_i64const
```
breathx, line 1301:
```suggestion
    // w_store = w_bench - w_i32const
```

breathx (changes requested):
Please go to `pallets/gear/src/schedule` and properly use newly added weights for:
- `struct InstructionWeight` and it's `impl Default` realisation
- `impl gas_metering::Rules for ScheduleRules`

breathx:
Otherwise well done btw!

breathx (approved):

grishasobol (changes requested):

runtime/gear/src/weights/pallet_gear.rs
```diff
@@ -969,6 +1010,16 @@ impl<T: frame_system::Config> pallet_gear::WeightInfo for SubstrateWeight<T> {
...
     }
+    /// The range of component `r` is `[50, 500]`.
+    fn instr_i32store(r: u32, ) -> Weight {
+        // Proof Size summary in bytes:
+        //  Measured:  `0`
+        //  Estimated: `0`
+        // Minimum execution time: 11_146_392 nanoseconds.
+        Weight::from_ref_time(11_214_000_333)
+            // Standard Error: 214_016
+            .saturating_add(Weight::from_ref_time(15_493_976).saturating_mul(r.into()))
```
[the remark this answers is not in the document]
breathx: will be replaced with proper weights on next benchmarking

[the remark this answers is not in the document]
breathx: feel free to push the weights from our benchmarking hardware directly to this PR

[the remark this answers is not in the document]
grishasobol: ok, I agree with changes them in next PR, but first I will check how weights are changed, then will approve this one

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -1261,6 +1261,25 @@ benchmarks! {
         sbox.invoke();
     }
 
+    // w_load = w_bench
+    instr_i32load {
+        let r in 0 .. INSTR_BENCHMARK_BATCHES;
```
grishasobol, line 1266: repeats diapason must be like in i64load

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -1279,6 +1298,25 @@ benchmarks! {
         sbox.invoke();
     }
 
+    // w_store = w_bench - w_i32const
+    instr_i32store {
+        let r in 0 .. INSTR_BENCHMARK_BATCHES;
```
grishasobol, line 1303: repeats diapason must be like in i64store

runtime/gear/src/weights/pallet_gear.rs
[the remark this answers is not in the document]
grishasobol: ok looks very good. Some instructions have really big perf gain

grishasobol (approved):

PR #2410, opened 2023-03-16
Get rid of gcli dependency in node

DennisInSky:
@reviewer-or-team

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3599,7 +3599,6 @@ dependencies = [
  "frame-benchmarking-cli",
  "frame-system",
  "futures",
- "gcli",
  "gear-runtime",
  "gear-runtime-primitives",
  "gear-runtime-test-cli",
--- a/node/cli/Cargo.toml
+++ b/node/cli/Cargo.toml
@@ -52,14 +52,11 @@ frame-benchmarking = { version = "4.0.0-dev", git = "https://github.com/gear-tec
 frame-benchmarking-cli = { version = "4.0.0-dev", git = "https://github.com/gear-tech/substrate.git", branch = "gear-polkadot-v0.9.38", optional = true }
 try-runtime-cli = { version = "0.10.0-dev", git = "https://github.com/gear-tech/substrate.git", branch = "gear-polkadot-v0.9.38", optional = true }
 
-# Program CLI
-gcli = { path = "../../gcli", optional = true }
-
 [build-dependencies]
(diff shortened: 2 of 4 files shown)


ukint-vs:
Remove here also https://github.com/gear-tech/gear/blob/ec972b88ce176a765497095e5bad0ab19d69f1e1/node/cli/src/cli.rs#L109

ukint-vs (approved):

DennisInSky:
> Remove here also
> 
> https://github.com/gear-tech/gear/blob/ec972b88ce176a765497095e5bad0ab19d69f1e1/node/cli/src/cli.rs#L109

nice catch 🔥

gshep (approved):
:+1:

shamilsan:
We are to decide whether we provide `gcli` functionality as a `gear` subcommand (preferable for me) or as a separate CLI tool. Why we've decided to make it separate? @clearloop

DennisInSky:
Personally I do not see any good reason for having functionality of one independent tool incorporated into another especially when that functionality is intended for interacting with the tool itself. As it comes to concern about the functionality being consistent with the main tool (node), I do not see why a person can't build it from the same repo the build the main tool. Anyway they will need to build the main tool via custom build (-F cli) which is pretty much the same as running a separate build of the gcli.
Moreover, I would thought of opposite dependency especially when it comes to testing part - gli tests as well as gclient ones rely on the fact the node has been built.

shamilsan:
> it was original a separated CLI, like `subkey` https://github.com/paritytech/substrate/tree/master/bin/utils/subkey, the subcommand key in node cli shares the same logic of subkey just like gcli, we registered it as subcommand of our node cli after moving `gprogram` into this repo

So I vote for leaving the `gcli` subcommand (but maybe to rename it to `cli` as `gear gcli ...` looks over `g`ed).

DennisInSky:
> > We are to decide whether we provide `gcli` functionality as a `gear` subcommand (preferable for me) or as a separate CLI tool. Why we've decided to make it separate? @clearloop
> 
> it was original a separated CLI, like `subkey` https://github.com/paritytech/substrate/tree/master/bin/utils/subkey, the subcommand key in node cli shares the same logic of subkey just like gcli, we registered it as subcommand of our node cli after moving `gprogram` into this repo

As far as I understand, the main difference between those is that the `subkey` functionality does not interact with the `node` whereas the `gcli` functionality is about interacting with the node itself. But even then, I do not see why functionality about working with keys has to be a part of the node and can't be a separate tool. As mentioned in the other post, to my point of view toolset is better than a God tool

PR #2436, opened 2023-03-21
[vara] 3 sec per block

ukint-vs:
Resolves # .

-
-
-

@gear-tech/dev

### DIFF
--- a/runtime/vara/src/lib.rs
+++ b/runtime/vara/src/lib.rs
@@ -433,12 +433,12 @@ impl pallet_gear_staking_rewards::Config for Runtime {
 
 // TODO: review staking parameters - currently copying Kusama
 parameter_types! {
-    // Six sessions in an era (4 hours)
+    // Six sessions in an era (12 hours)
     pub const SessionsPerEra: sp_staking::SessionIndex = 6;
     // 42 eras for unbonding (7 days)
-    pub const BondingDuration: sp_staking::EraIndex = 42;
+    pub const BondingDuration: sp_staking::EraIndex = 14;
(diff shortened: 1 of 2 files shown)


runtime/vara/src/lib.rs
```diff
@@ -433,12 +433,12 @@ impl pallet_gear_staking_rewards::Config for Runtime {
 
 // TODO: review staking parameters - currently copying Kusama
 parameter_types! {
-    // Six sessions in an era (4 hours)
+    // Six sessions in an era (12 hours)
     pub const SessionsPerEra: sp_staking::SessionIndex = 6;
     // 42 eras for unbonding (7 days)
```
shamilsan, line 438: Should we fix this comment too?

ukint-vs: @ekovalev

ekovalev: `// 42 eras for unbonding (7 days) ` is obviously misleading given 12 hours epoch duration.
I guess @shamilsan is referring to this comment

ekovalev:
How about the vesting schedules in the chain spec?
We calculate 1 year period after which vesting kicks in based on the old block length.
@NikVolf , will you generate a new spec accordingly?

runtime/vara/src/lib.rs
shamilsan: > `// 42 eras for unbonding (7 days) ` is obviously misleading given 12 hours era duration. I guess @shamilsan is referring to this comment

Exactly.

NikVolf (approved):

PR #2439, opened 2023-03-21
[gclient] Get rid of GearApiWithNode

### LINKED ISSUE #2397: gclient: `GearApiWithNode` doesn't support login operation
### File Location(s)

gclient/src/api/mod.rs

### Proposal

`GearApiWithNode` can't provide `GearApi::with` method as it implements `Deref` that can use `&self` only instead of `self`.
Also, it is quite inconvenient to have two structures (`GearApiWithNode` and `GearApi`).

Consider to combine the node functionality in `GearApi`.

DennisInSky:
Resolves #2397  .

`Node` has been moved insied `GearApi` as an option

@reviewer-or-team

### DIFF
--- a/gclient/src/api/mod.rs
+++ b/gclient/src/api/mod.rs
@@ -83,7 +83,7 @@ impl GearApi {
     /// random port number. The node process uses a binary specified by the
     /// `path` param. Ideally, the binary should be downloaded by means of CI pipeline from <https://get.gear.rs>.
     pub async fn dev_from_path(path: impl AsRef<OsStr>) -> Result<Self> {
-        let node = Node::try_node_from_path(path, vec!["--dev"])?;
+        let node = Node::try_from_path(path, vec!["--dev"])?;
         let api = Self::init(node.ws_address().clone()).await?;
         Ok(Self(api.0, Some(Arc::new(node))))
     }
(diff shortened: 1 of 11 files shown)


gclient/src/node/mod.rs
```diff
@@ -25,44 +25,23 @@ use std::{
...
-        )
-    }
-
     /// Returns Web Socket address the node is listening to.
     pub fn ws_address(&self) -> &WSAddress {
         &self.ws_address
     }
 
-    fn try_node_from_path(path: impl AsRef<OsStr>, args: Vec<&str>) -> Result<Self> {
+    pub fn try_node_from_path(path: impl AsRef<OsStr>, args: Vec<&str>) -> Result<Self> {
```
shamilsan, line 44: Maybe just `try_from_path`?

gclient/src/api/mod.rs
```diff
@@ -28,11 +28,11 @@ use crate::{
 };
 use error::*;
 use gsdk::{ext::sp_runtime::AccountId32, signer::Signer, Api};
-use std::{marker::PhantomData, ops::Deref};
+use std::{ffi::OsStr, sync::Arc};
 
 /// The API instance contains methods to access the node.
 #[derive(Clone)]
-pub struct GearApi(Signer);
+pub struct GearApi(Signer, Option<Arc<Node>>);
```
shamilsan, line 35: I'd change it to struct
```suggestion
pub struct GearApi {
    signer: Signer, 
    node: Option<Arc<Node>>
}
```

shamilsan (approved):
I like it 👍

gclient/src/api/mod.rs
DennisInSky: @shamilsan are you sure about this. I have done it locally, but it imposes a lot of changes.

shamilsan: As it doesn't affect external interface (only internal implementation), we can do it later, ok.

PR #2489, opened 2023-03-27
Stabilize fuzzer docker

SabaunT:
Running fuzzer with docker was not successful - current version resulted in OS errors related to file access. Possibly that was because the directory which stored fuzzing seeds was removed by fuzzer, also it was mounted on the host machine. Later fuzzer re-created it in accordance to internal logic. That workflow seems to be buggy.

That was edited in the current PR, which introduces changes that doesn't remove fuzzing seeds directory.

TODO:
- [ ] Testing on the reference machine by @sergeyfilyanin

### DIFF
--- a/docker/runtime-fuzzer/Dockerfile
+++ b/docker/runtime-fuzzer/Dockerfile
@@ -17,7 +17,7 @@ RUN rustup update nightly && rustup target add wasm32-unknown-unknown --toolchai
 RUN rustup default nightly
 
 # Clone gear repo
-RUN git clone https://github.com/gear-tech/gear.git
+RUN git clone --branch st-stabilize-fuzzer-docker https://github.com/gear-tech/gear.git
 
 # Install cargo-fuzz
 RUN cd gear
--- a/utils/runtime-fuzzer/fuzz/fuzz_targets/main.rs
+++ b/utils/runtime-fuzzer/fuzz/fuzz_targets/main.rs
@@ -22,7 +22,7 @@ use libfuzzer_sys::fuzz_target;
 use once_cell::sync::OnceCell;
 use std::{
     fs::{self, OpenOptions},
-    io::Write,
+    io::{Write, Result as IoResult, Error as IoError},
     path::Path,
 };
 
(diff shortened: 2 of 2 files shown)


docker/runtime-fuzzer/Dockerfile
```diff
@@ -17,7 +17,7 @@ RUN rustup update nightly && rustup target add wasm32-unknown-unknown --toolchai
 RUN rustup default nightly
 
 # Clone gear repo
-RUN git clone https://github.com/gear-tech/gear.git
+RUN git clone --branch st-stabilize-fuzzer-docker https://github.com/gear-tech/gear.git
```
SabaunT, line 20: undo after tests

utils/runtime-fuzzer/fuzz/fuzz_targets/main.rs
```diff
@@ -44,21 +44,31 @@ fuzz_target!(|seed: u64| {
...
-fn dump_seed(seed: u64) -> Result<(), String> {
+fn dump_seed(seed: u64) -> IoResult<()> {
     let seeds_file = RUN_INTIALIZED.get_or_try_init(|| {
         let seeds_dir = Path::new(SEEDS_STORE_DIR);
-        if seeds_dir.exists() {
-            fs::remove_dir_all(seeds_dir).map_err(|e| e.to_string())?;
+        match seeds_dir.exists() {
+            true => clear_dir(seeds_dir)?,
+            false => fs::create_dir_all(seeds_dir)?,
         }
```
shamilsan, line 53: `match` for `bool` is overkill in my opinion. I'd use the old good `if`-`else`.

utils/runtime-fuzzer/fuzz/fuzz_targets/main.rs
```diff
@@ -44,21 +44,31 @@ fuzz_target!(|seed: u64| {
...
         .and_then(|mut file| writeln!(file, "{seed}"))
-        .map_err(|e| e.to_string())
+}
+
+fn clear_dir(path: &Path) -> IoResult<()> {
+    for dir_entry in fs::read_dir(path)? {
+        let dir_entry = dir_entry?;
+        let path = dir_entry.path();
+
+        fs::remove_file(path)?;
```
shamilsan, line 70: We can write it in one line I guess.
```suggestion
        fs::remove_file(dir_entry?.path())?;
```

shamilsan (approved):

utils/runtime-fuzzer/fuzz/fuzz_targets/main.rs
gshep: `match` is shorter

SabaunT: discussed.

utils/runtime-fuzzer/fuzz/fuzz_targets/main.rs
SabaunT: done.

PR #2490, opened 2023-03-27
change fallible sys-calls checkers in `runtime-benchmarks`

grishasobol:
1) remove fallible sys-calls checks in benchmarks verify.
2) add fallible sys-call checks inside benchmark wasm code:
```wat
block
  ;; ... move sys-call params to stack
  call $sys_call
  i32.load ($res_offset)
  i32.eqz
  br_if 0
  unreachable
end
```
if result is not 0 then unreachable instruction happens.
This checkers will be inserted only if `runtime-benchmark-checkers` feature is set.

3) __Very important__. Add stack end export generation in benchmarks and set it to all memory size by default. Currently this means that for all sys-call and instructions benchmarks all memory is stack, so no interruptions occurs during benchmark execution. This excludes contribution of lazy-pages handling, storage access and memory protections. 

4) increase measurement accuracy for many benchmarks.

5) benchmarks refactoring


ukint-vs (approved):
Great🔥

PR #2503, opened 2023-03-30
Performance improvements

ukint-vs:
- Bench reservations correctly
- Use `HashMap::with_capacity`
- Use `TryInto` in `blake2b` hash generation.
- Use [mimalloc_rust](https://github.com/purpleprotocol/mimalloc_rust) [mimalloc](https://github.com/microsoft/mimalloc) as default allocator. (major gain in runtime interfaces, especially with set_global)
- Optimise some things in lazy pages.
- Add profiling profile in `Cargo.toml`

@gear-tech/dev


lazy-pages/src/signal.rs
```diff
@@ -64,15 +63,16 @@ unsafe fn user_signal_handler_internal(
...
+        .wasm_mem_addr
+        .ok_or_else(|| Error::WasmMemAddrIsNotSet)?;
 
     if native_addr < wasm_mem_addr {
         return Err(Error::OutOfWasmMemoryAccess);
     }
 
-    let offset =
-        u32::try_from(native_addr - wasm_mem_addr).map_err(|_| Error::OutOfWasmMemoryAccess)?;
+    let offset = (native_addr - wasm_mem_addr) as u32;
```
grishasobol, line 75: Why did you do that? This moment is very important.

lazy-pages/src/host_func.rs
```diff
@@ -102,28 +102,31 @@ impl<'a> AccessHandler for HostFuncAccessHandler<'a> {
...
+                .offset
+                .checked_add(access.size.saturating_sub(1))
+                .ok_or_else(|| Error::OutOfWasmMemoryAccess)?;
+
+            let start = (access.offset / page_size) * page_size;
+            let end = (last_byte / page_size) * page_size;
+            let mut offset = start;
+            while offset <= end {
+                pages.insert(GearPageNumber::from_offset(ctx, offset));
+                offset += page_size;
```
grishasobol, line 125: Where is checked add ? It is also very important  case

lazy-pages/src/host_func.rs
```diff
@@ -102,28 +102,31 @@ impl<'a> AccessHandler for HostFuncAccessHandler<'a> {
...
-            .ok_or(Error::OutOfWasmMemoryAccess)?;
-
-        let page_size = GearPageNumber::size(ctx);
-        let mut offset = access.offset;
-        while offset <= last_byte {
-            set.insert(GearPageNumber::from_offset(ctx, offset));
-            offset = match offset.checked_add(page_size) {
-                Some(offset) => (offset / page_size) * page_size,
-                None => break,
+    pages: &mut BTreeSet<GearPageNumber>,
```
grishasobol, line 105: I don't think return result thru `&mut` param is more effective way. Also it's less rust styled way)) so ...

grishasobol: I don't think so, because this function will be 100% inlined, so this is not important to do things like this

lazy-pages/src/host_func.rs
```diff
@@ -88,10 +88,10 @@ impl<'a> AccessHandler for HostFuncAccessHandler<'a> {
         pages: Self::Pages,
         mut process_one: impl FnMut(GearPageNumber) -> Result<(), Error>,
     ) -> Result<(), Error> {
-        for page in pages {
-            process_one(page)?;
-        }
-        Ok(())
+        pages.iter().try_for_each(|page| -> Result<(), Error> {
```
grishasobol, line 91: whether this one works really faster? Because it looks much less understandable. I cannot see any reasons to do like this

grishasobol (changes requested):

lazy-pages/src/host_func.rs
ukint-vs: https://doc.rust-lang.org/book/ch13-04-performance.html

lazy-pages/src/signal.rs
ukint-vs: Missed a moment with `usize`.

lazy-pages/src/host_func.rs
ukint-vs: Added

pallets/gear/src/weights.rs
```diff
@@ -19,13 +19,13 @@
 //! Autogenerated weights for pallet_gear
 //!
 //! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 4.0.0-dev
-//! DATE: 2023-03-31, STEPS: `20`, REPEAT: `40`, LOW RANGE: `[]`, HIGH RANGE: `[]`
+//! DATE: 2023-04-04, STEPS: `50`, REPEAT: `20`, LOW RANGE: `[]`, HIGH RANGE: `[]`
 //! WORST CASE MAP SIZE: `1000000`
-//! CPU: `Intel(R) Xeon(R) Platinum 8375C CPU @ 2.90GHz`
+//! HOSTNAME: `ip-172-31-10-238`, CPU: `Intel(R) Xeon(R) Platinum 8375C CPU @ 2.90GHz`
```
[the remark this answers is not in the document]
ukint-vs: Updated

lazy-pages/src/host_func.rs
ukint-vs: Why u need to create empty set inside fn?

grishasobol (approved):

PR #2511, opened 2023-04-04
Move gas charging inside alloc

### LINKED ISSUE #2337: move charging for grow inside alloc function, so that we can skip refunding
### File Location
/core-processor/src/ext.rs

### Proposal

see title

DennisInSky:
Resolves #2337 

@reviewer-or-team

### DIFF
--- a/core-processor/src/ext.rs
+++ b/core-processor/src/ext.rs
@@ -388,8 +388,8 @@ impl Ext {
     }
 
     fn charge_gas_if_enough(
-        gas_counter: &mut gear_core::gas::GasCounter,
-        gas_allowance_counter: &mut gear_core::gas::GasAllowanceCounter,
+        gas_counter: &mut GasCounter,
+        gas_allowance_counter: &mut GasAllowanceCounter,
         amount: u64,
     ) -> Result<(), ChargeError> {
         if gas_counter.charge_if_enough(amount) != ChargeResult::Enough {
--- a/core-processor/src/ext.rs
+++ b/core-processor/src/ext.rs
@@ -32,7 +32,7 @@ use gear_core::{
     env::Ext as EnvExt,
     gas::{
         ChargeError, ChargeResult, CountersOwner, GasAllowanceCounter, GasAmount, GasCounter,
-        GasLeft, GasRefunder, Token, ValueCounter,
+        GasLeft, Token, ValueCounter,
     },
     ids::{CodeId, MessageId, ProgramId, ReservationId},
     memory::{
@@ -357,20 +357,6 @@ impl Ext {
         }
     }
 
-    fn check_charge_results(
-        &mut self,
-        common_charge: ChargeResult,
-        allowance_charge: ChargeResult,
-    ) -> Result<(), ChargeError> {
-        use ChargeResult::*;
-
-        match (common_charge, allowance_charge) {
-            (NotEnough, _) => Err(ChargeError::GasLimitExceeded),
-            (Enough, NotEnough) => Err(ChargeError::GasAllowanceExceeded),
-            (Enough, Enough) => Ok(()),
-        }
-    }
-
     fn charge_sending_fee(&mut self, delay: u32) -> Result<(), ChargeError> {
         if delay == 0 {
             self.charge_gas_if_enough(self.context.message_context.settings().sending_fee())
@@ -400,6 +386,24 @@ impl Ext {
         }
         Ok(())
     }
+
+    fn charge_gas_if_enough(
+        gas_counter: &mut gear_core::gas::GasCounter,
+        gas_allowance_counter: &mut gear_core::gas::GasAllowanceCounter,
+        amount: u64,
+    ) -> Result<(), ChargeError> {
+        if gas_counter.charge_if_enough(amount) != ChargeResult::Enough {
+            return Err(ChargeError::GasLimitExceeded);
+        }
+        if gas_allowance_counter.charge_if_enough(amount) != ChargeResult::Enough {
+            if gas_counter.refund(amount) != ChargeResult::Enough {
+                // We have just charged `amount` from `self.gas_counter`, so this must be correct.
+                unreachable!("Cannot refund {amount} for `gas_counter`");
+            }
+            return Err(ChargeError::GasAllowanceExceeded);
+        }
+        Ok(())
+    }
 }
 
 impl CountersOwner for Ext {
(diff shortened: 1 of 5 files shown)


core-processor/src/ext.rs
```diff
@@ -872,18 +876,16 @@ impl Ext {
...
-
-        // Refund gas if memory has not been grown to `pages` size.
-        gas_refunder
-            .refund(self, self.context.page_costs.mem_grow.calc(not_grown))
-            .unwrap_or_else(|err| unreachable!("Alloc implementation error: {err}"))?;
-
+        let AllocInfo { page, .. } =
+            self.context
+                .allocations_context
+                .alloc::<G>(pages, mem, |pages| {
```
shamilsan, line 882: I don't know whether it is better (if feasible) to implement a charger in `alloc` as a second generic instead of closure. In theory, static dispatch would be a little bit more efficient. But maybe it is not worth messing up.

shamilsan (approved):
Looks good to me.

core-processor/src/ext.rs
DennisInSky: Isn't it a static dispatch? The `alloc` method takes impl FnOnce which is a generic, doesn't it? It does take a function pointer.

core/src/memory.rs
```diff
@@ -543,6 +543,10 @@ pub enum AllocError {
     /// outside additionally allocated for this program.
     #[display(fmt = "Page {_0} cannot be freed by the current program")]
     InvalidFree(u32),
+    /// Gas charge error
+    #[from]
+    #[display(fmt = "{_0}")]
+    GasChargeError(ChargeError),
```
ark0f, line 549:
```suggestion
    Charge(ChargeError),
```
Consistency with other error variants

core-processor/src/ext.rs
```diff
@@ -416,17 +426,11 @@ impl CountersOwner for Ext {
...
-        }
-        if self.context.gas_allowance_counter.charge_if_enough(amount) != ChargeResult::Enough {
-            if self.context.gas_counter.refund(amount) != ChargeResult::Enough {
-                // We have just charged `amount` from `self.gas_counter`, so this must be correct.
-                unreachable!("Cannot refund {amount} for `gas_counter`");
-            }
-            return Err(ChargeError::GasAllowanceExceeded);
-        }
-        Ok(())
+        Ext::charge_gas_if_enough(
```
ark0f, line 429: I'd inline the function

core-processor/src/ext.rs
```diff
@@ -872,18 +876,16 @@ impl Ext {
...
-        // Refund gas if memory has not been grown to `pages` size.
-        gas_refunder
-            .refund(self, self.context.page_costs.mem_grow.calc(not_grown))
-            .unwrap_or_else(|err| unreachable!("Alloc implementation error: {err}"))?;
-
+        let AllocInfo { page, .. } =
+            self.context
+                .allocations_context
+                .alloc::<G>(pages, mem, |pages| {
+                    Ext::charge_gas_if_enough(
```
ark0f, line 883: Use `CountersOwner::charge_gas_if_enough` instead

core/src/memory.rs
```diff
@@ -584,6 +588,7 @@ impl AllocationsContext {
         &mut self,
         pages: WasmPage,
         mem: &mut impl Memory,
+        charge_gas_for_grow: impl FnOnce(WasmPage) -> Result<(), ChargeError>,
```
ark0f, line 591: Consider
```suggestion
        counters: &mut impl CountersOwner,
```

ark0f (changes requested):

core/src/memory.rs
DennisInSky: That was the idea at the very beginning, but the issue is that the `self` has already got shared borrow  and passing it down requires the second shared borrow which is prohibited. Even though if it was possible to do so, the charging requires info about the amount to charge which is calculated based on the structures residing in the `core-processor` whereas the `alloc` sits in the `core`. Hopefully this answers your questions with regards to inlining and using `CountersOwner::charge_gas_if_enough` directly

core-processor/src/ext.rs
DennisInSky: This is what I've found in an earlier edition of the Book
https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/first-edition/closures.html#:~:text=In%20many%20languages%2C%20closures%20are,often%20take%20closures%20as%20arguments.
Unfortunatelly, I can't find the same in the most recent edition.

ark0f (approved):

core/src/memory.rs
```diff
@@ -584,6 +588,7 @@ impl AllocationsContext {
...
     ) -> Result<AllocInfo, AllocError> {
```
grishasobol, line 592: The only one reason to return AllocInfo instead of pure allocated page is to return gas. So this must be removed

core/src/gas.rs
```diff
@@ -301,47 +301,6 @@ pub trait CountersOwner {
...
-
-/// [GasRefunder] error, in case somebody tries to refund more than was charged.
-#[derive(Debug, derive_more::Display)]
-#[display(fmt = "Trying to refund {charged}, but {refunded} was charged")]
-pub struct RefundError {
-    charged: u64,
-    refunded: u64,
-}
-
-impl GasRefunder {
```
grishasobol, line 318: Gas refunder can be used in other cases. I think we must left it as is

grishasobol (changes requested):

core/src/gas.rs
DennisInSky: There is no any good reason to keep unsued code in the codebase. It contradicts to the YAGNI. One can restore it from the source control if they need it.

core-processor/src/ext.rs
```diff
@@ -400,14 +386,38 @@ impl Ext {
         }
         Ok(())
     }
+
+    fn charge_gas_if_enough(
```
breathx, line 390: can't find the reason this to be inside Ext impl without self param

core-processor/src/ext.rs
```diff
@@ -872,19 +876,16 @@ impl Ext {
...
-            .unwrap_or_else(|err| unreachable!("Alloc implementation error: {err}"))?;
-
-        Ok(page)
+        self.context
+            .allocations_context
+            .alloc::<G>(pages, mem, |pages| {
+                Ext::charge_gas_if_enough(
+                    &mut self.context.gas_counter,
+                    &mut self.context.gas_allowance_counter,
+                    self.context.page_costs.mem_grow.calc(pages),
```
breathx, line 885: it's amount of gas to be charged for growed `pages` amount of pages, but this count is differ in case. provide tests for it as well please

breathx (changes requested):

core-processor/src/ext.rs
DennisInSky: This method has been made "static" on purpose as I can't use `&mut self` in the closure on line 882. I can use just `GasCounter` and `GasAllowanceCounter` from the `self`.

breathx: I'm exactly about static usage: why not just fn somewhere around but nvm

breathx (approved):

breathx:
@grishasobol take one more look pls

grishasobol (approved):

PR #2513, opened 2023-04-05
gtest: Do not validate dispatch for delayed messages

### LINKED ISSUE #2392: `gtest` unexpected behaviour with the SFT contract
### Problem

The SFT contract is one of the few that use delayed messages. For some reason, if test executes any SFT's action that produces delayed message (e.g. Mint), and then uses the `System::spend_blocks()` function, test fails with the following message:
```sh
thread 'test' panicked at 'Sending messages allowed only from users id', /home/user/.cargo/git/checkouts/gear-bdc2b7d48ef3acbd/2c3dc71/gtest/src/manager.rs:335:13
```
There was no such behavior before.

### Steps

1. Initialize SFT.
2. Execute some action that produces delayed message (e.g. Mint).
3. Call `System::spend_blocks()`.
4. See the `Sending messages allowed only from users id` error.

### Possible Solution

_No response_

### Notes

_No response_

### Relevant Log Output

_No response_

DennisInSky:
Resolves #2392 .

- We do not want to validate disptaches taken from the delayed message queue. They had been validated before they were put in it

@reviewer-or-team


shamilsan (approved):
You was fater than me 😄

shamilsan:
Is it possible to add a test for the issue?

breathx (approved):

PR #2514, opened 2023-04-05
backend-sandbox: Use syscall_trace! in liue of sys_trace!

### LINKED ISSUE #2352: use `syscall_trace!` in sandbox backend
### File Location(s)

core-backend/sandbox/src/func.rs

### Proposal

see title

DennisInSky:
Resolves #2352  .

@reviewer-or-team

### DIFF
--- a/core-backend/sandbox/src/funcs.rs
+++ b/core-backend/sandbox/src/funcs.rs
@@ -26,8 +26,8 @@ use codec::Encode;
 use core::{convert::TryInto, marker::PhantomData};
 use gear_backend_common::{
     memory::{MemoryAccessError, MemoryAccessRecorder, MemoryOwner},
-    ActorTerminationReason, BackendAllocExtError, BackendExt, BackendExtError, BackendState,
-    TrapExplanation, PTR_SPECIAL,
+    syscall_trace, ActorTerminationReason, BackendAllocExtError, BackendExt, BackendExtError,
+    BackendState, TrapExplanation, PTR_SPECIAL,
 };
 use gear_core::{
     buffer::RuntimeBuffer,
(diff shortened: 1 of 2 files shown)


core-backend/sandbox/src/funcs.rs
```diff
@@ -96,7 +84,7 @@ where
 
     /// Fallible `gr_send` syscall.
     pub fn send(ctx: &mut Runtime<E>, args: &[Value]) -> SyscallOutput {
-        sys_trace!(target: "syscalls", "send, args = {}", args_to_str(args));
+        syscall_trace!("send", args_to_str(args));
```
grishasobol, line 87: we need to use it exactly in the same way as in wasmi backend

grishasobol (changes requested):

core-backend/sandbox/src/funcs.rs
DennisInSky: I can't use it exactly the same way as it is in the wasmi backend. Args parsing here is done via `args.iter().read_X()?;` and can return an erro thus one can miss log entry. If we don't care about it, I can pass successfully parsed args into the `syscall_trace`

shamilsan (approved):
Looks better!

core-backend/sandbox/src/funcs.rs
grishasobol: i think it's not a big problem that args.iter().read_X()? can return an error. It can return an error only in case syscall signature is incorrect, and this is supposed to be impossible. Actually it's better to add panic here instead returning error

core-backend/sandbox/src/funcs.rs
```diff
@@ -804,9 +919,12 @@ where
 
     /// Fallible `gr_system_reserve_gas` syscall.
     pub fn system_reserve_gas(ctx: &mut Runtime<E>, args: &[Value]) -> SyscallOutput {
-        sys_trace!(target: "syscalls", "system_reserve_gas, args = {}", args_to_str(args));
+        let (gas, err_len_ptr) = args
+            .iter()
+            .read_2()
+            .expect("There are 2 arguments expected");
```
grishasobol, line 925: please move this panics inside helper `read_*` functions. Also we use 
```rust
.unwrap_or_else(|| unreachable!("..."))
```
instead of `expect` to unify format of panics exprs inside runtime.

PR #2517, opened 2023-04-06
gear-common: Return error in case of imbalance application leads to overflow

### LINKED ISSUE #2142: Make apply_to method for positive/negative imbalances returning Result
### File Location(s)

common/src/gas_provider/negative_imbalance.rs
common/src/gas_provider/positive_imbalance.rs

### Proposal

Make the `apply_to` method returning Result instead of producing debug messages in case of overflow
https://github.com/gear-tech/gear/blob/ace76b6c7b5a275aeb92503d015091a20cb5f3ee/common/src/gas_provider/negative_imbalance.rs#L43
https://github.com/gear-tech/gear/blob/ace76b6c7b5a275aeb92503d015091a20cb5f3ee/common/src/gas_provider/positive_imbalance.rs#L41

DennisInSky:
Resolves #2142  .

@reviewer-or-team

### DIFF
--- a/common/src/gas_provider/internal.rs
+++ b/common/src/gas_provider/internal.rs
@@ -520,7 +520,12 @@ where
         let positive_imbalance = PositiveImbalance::new(amount);
 
         // Update Total in storage
-        TotalValue::mutate(|total| positive_imbalance.apply_to(total));
+        TotalValue::mutate(|total| {
+            positive_imbalance.apply_to(total).map_err(|_| {
+                *total = None;
+                InternalError::total_value_is_overflowed()
+            })
+        })?;
 
         Ok(positive_imbalance)
     }
@@ -652,7 +657,12 @@ where
 
         // Update Total in storage
         if let Some((negative_imbalance, _)) = res.as_ref() {
-            TotalValue::mutate(|total| negative_imbalance.apply_to(total));
+            TotalValue::mutate(|total| {
+                negative_imbalance.apply_to(total).map_err(|_| {
+                    *total = None;
+                    InternalError::total_value_is_underflowed()
+                })
+            })?;
         }
 
         Ok(res)
@@ -694,7 +704,12 @@ where
         let negative_imbalance = NegativeImbalance::new(amount);
 
         // Update Total in storage
-        TotalValue::mutate(|total| negative_imbalance.apply_to(total));
+        TotalValue::mutate(|total| {
+            negative_imbalance.apply_to(total).map_err(|_| {
+                *total = None;
+                InternalError::total_value_is_underflowed()
+            })
+        })?;
 
         Ok(negative_imbalance)
     }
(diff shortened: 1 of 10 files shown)


gclient/src/api/calls.rs
```diff
@@ -355,6 +360,22 @@ impl GearApi {
...
+                    Err(e)
+                }
+            })?;
+
+        dest_node_api
+            .0
+            .set_total_issuance(
+                dest_gas_total_issuance.saturating_add(src_program_reserved_gas_total),
+            )
+            .await?;
```
DennisInSky, line 377: @breathx I do not like this piece of code as well as the above one. It is unsafe in term of atomicity as well as the above one where we update balances on the dest node. The base value which is used for later update can change between 2 calls. For the source node we have solved similar issues using some fixed block number, but I can't make up anything what we can do for the destionation node

common/src/gas_provider/internal.rs
```diff
@@ -520,7 +520,12 @@ where
         let positive_imbalance = PositiveImbalance::new(amount);
 
         // Update Total in storage
-        TotalValue::mutate(|total| positive_imbalance.apply_to(total));
+        TotalValue::mutate(|total| {
+            positive_imbalance.apply_to(total).map_err(|_| {
+                *total = None;
```
breathx, line 525: should we set None here?

common/src/gas_provider/mod.rs
```diff
@@ -306,5 +306,13 @@ pub trait Imbalance {
...
-    fn apply_to(&self, amount: &mut Option<Self::Balance>);
+    fn apply_to(&self, amount: &mut Option<Self::Balance>) -> Result<(), ImbalanceError>;
+}
+
+/// Represents errors returned by via the [Imbalance] trait.
+#[derive(Debug, PartialEq)]
+pub enum ImbalanceError {
+    /// Indicates the imbalance value causes amount value overflowing
+    /// when applied to the latter.
+    Overflow,
```
breathx, line 317: we have different methods about overflow and underflow but the only one error variant: seems strange
furthermore, for case with single error variant possible we can use just empty struct as a flag

common/src/gas_provider/negative_imbalance.rs
```diff
@@ -37,18 +37,13 @@ impl<Balance: BalanceTrait> Imbalance for NegativeImbalance<Balance> {
...
-        let new_value = amount.unwrap_or_else(Zero::zero);
-        if self.0 > new_value {
-            log::debug!(
-                target: "essential",
-                "Unaccounted gas detected: burnt {:?}, known total supply was {:?}.",
-                self.0,
-                *amount
-            )
+    fn apply_to(&self, amount: &mut Option<Balance>) -> Result<(), ImbalanceError> {
+        let amount_value = amount.unwrap_or_else(Zero::zero);
```
breathx, line 41: checked ops seems more familiar

gclient/src/api/calls.rs
breathx: seems any "mutate" fns are impossible here so nothing to do :(

breathx (approved):

PR #2521, opened 2023-04-06
docs(gstd): add deprecated attribute for macro `metadata!`

clearloop:
Resolves # .

in case some libraries are still using this, go to new-meta

@gear-tech/dev

### DIFF
--- a/examples/async-init/src/lib.rs
+++ b/examples/async-init/src/lib.rs
@@ -1,4 +1,5 @@
 #![no_std]
+#![allow(deprecated)]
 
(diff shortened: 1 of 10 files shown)


gstd/src/macros/metadata.rs
```diff
@@ -119,6 +119,10 @@ macro_rules! metadata {
 
 #[cfg(not(doc))]
 #[allow(missing_docs)]
+#[deprecated(
```
breathx, line 122: Allow all deprecated usages in examples/* crates

breathx (approved):

PR #2525, opened 2023-04-08
Extend locking in the Gas Tree

### LINKED ISSUE #1286: Expand `GasTree::lock` functionality
### Problem to Solve

After #1274.

- Need to figure out where such mechanic may be also useful, and add ability to unlock after consume, lock named etc.
- Update proptests following final version of locking mechanic

ekovalev:
Partly resolves #1286.

Current implementation allows adding locks to a tree node (by summing up locked amounts) but can't tell between different locks added at different times. This makes difficult to handle situations when a message needs to be placed in several storages (for instance, wait list and delayed dispatch etc.): every time a message is charged for storage the entire unlocked amount will the released as if there were no other potential storages involved.

Another issue currently overlooked is that if the cost of storage has changed in the meantime via runtime upgrade or a message has overstayed in a storage (e.g. wait list) due to the queue congestion the amount actually charged may differ from what had been locked and then released potentially creating a risk of the `InsufficientBalance` error.

This PR introduces "named" locks for different storages types so that locking and unlocking is done independently from other storages.

Besides, some refactoring of the GasTree interface has been done:
- functions related to locking and reservation of gas in gas tree nodes have been collected as separate traits (`LockableTree` and `ReservableTree`, respectively) to "offload" interfaces not having to do with usual tree operations from the `Tree` trait definition. This change is, however, questionable because in our project the only tree we have is both reservable and lockable;
- some housekeeping in terms of tree node identification has been done. Over time we ended up having a handful of types (`Key`, `MapKey`, `MapReservationKey`, `NodeIdOf<>` etc) to identify a node of a tree. This has caused confusion and could create even more confusion as we go ahead adding new node types in the future. To mitigate this the `NodeId` type has been made generic and not dependent on the node type. This didn't affect the gas tree api at all.

TODO: these changes are consistent with the existing tests (including the prop tests), however, we don't have specific tests to check the "many simultaneous storages" use case. Adding such tests would be useful.


common/src/gas_provider/property_tests/mod.rs
```diff
@@ -558,7 +558,7 @@ proptest! {
                 GasTreeAction::Lock(from, amount) => {
                     let &from = NonEmpty::from_slice(&node_ids).expect("always has a tree root").ring_get(from);
 
-                    if let Err(e) = Gas::lock(from, amount) {
+                    if let Err(e) = Gas::lock(LockIdentifier::Waitlist, from, amount) {
```
SabaunT, line 561: I'd make these random too

SabaunT (approved):
Excellent work! :+1: 

However I'd wait for comments from other devs

SabaunT:
> TODO: these changes are consistent with the existing tests (including the prop tests), however, we don't have specific tests to check the "many simultaneous storages" use case. Adding such tests would be useful.

We must definetely provide tests for that. I would love to see some high level tests using the `pallets/gear/src/internal.rs` logic of managing different message storages.

common/src/gas_provider/internal.rs
```diff
@@ -414,41 +395,41 @@ where
         }
 
         // Detect inner from `reserve`.
-        let new_node = match new_node_key {
-            NodeCreationKey::Cut(_) => {
+        let new_node = match node_type {
+            NodeTypeWithValue::Cut => {
                 let id = Self::get_external(key)?;
                 GasNode::Cut {
```
breathx, line 401: for code optimisation we can provide some "default" constructor for GasNode

common/src/gas_provider/node.rs
```diff
@@ -81,7 +108,7 @@ pub enum GasNode<ExternalId: Clone, Id: Clone, Balance: Zero + Clone> {
     External {
         id: ExternalId,
         value: Balance,
-        lock: Balance,
+        lock: [Balance; 4],
```
breathx, line 111:
```suggestion
        lock: NodeLock<Balance>,
```

common/src/gas_provider/node.rs
```diff
@@ -70,6 +71,32 @@ impl<T> From<ReservationId> for GasNodeId<T, ReservationId> {
     }
 }
 
+pub type NodeLock<Balance> = [Balance; 4];
```
breathx, line 74: impl Zero maybe?

pallets/gear/src/internal.rs
```diff
@@ -302,11 +303,21 @@ where
...
+    /// a message must lock some funds as a deposit to "book" the storage until some time
+    /// in the future. Regardless whether storing costs or the safety margin have changed
+    /// in the meantime (due to storage migration) a message can't be charged more than
+    /// deposited upfront.
+    ///
+    /// `id` - parameter convertible to the respective gas node id;
+    /// `hold_interval` - determines the time interval to charge rent for;
+    /// `cost` - rental cost per time unit (block);
+    /// `max_charge` - if set, contains the deposit locked at the time of starting storage rental.
     pub(crate) fn charge_for_hold(
```
breathx, line 316: Seems we were discussing here ability to unlock funds within the fn: we can depend cost on lock type as well

breathx: max charge would be queried in place tho

pallets/gear/src/internal.rs
```diff
@@ -488,6 +506,7 @@ where
             mailboxed.id(),
             hold_interval,
             CostsPerBlockOf::<T>::mailbox(),
+            None,
```
breathx, line 509: won't there be similar max amount? ref to comment on charge for hold fn

breathx (changes requested):

common/src/gas_provider/node.rs
breathx: obvious thing. newtype pattern will help

pallets/gear/src/internal.rs
ekovalev: Right. Can pull it inside the function. That would require default locking of the entire available gas for `Cut` nodes (which is currently not done yet). Anyway, we've discussed it already.

pallets/gear/src/internal.rs
ekovalev: Yes. The reason (as explained before) is that we haven't locked anything for `Cut` nodes because the only possible use case for this type of node is to pay mailbox rent. For the sake of uniformity will add locking for this node type, as well.

common/src/gas_provider/node.rs
ekovalev: This idea has also come across my mind. Just didn't want to complicate things unnecessarily - that's why tried to get away with the `Default` instead. Anyway, I agree that implementing `Zero` for the lock type (whatever it is) would be useful

pallets/gear/src/internal.rs
```diff
@@ -315,25 +325,45 @@ where
...
+        let duration: u64 = paid_until
             .saturating_sub(hold_interval.start)
             .unique_saturated_into();
 
+        // Cost per block based on the storage used for holding (identifiable by `lock_id`)
+        let cost = match lock_id {
+            LockIdentifier::Mailbox => CostsPerBlockOf::<T>::mailbox(),
+            LockIdentifier::Waitlist => CostsPerBlockOf::<T>::waitlist(),
+            LockIdentifier::Reservation => CostsPerBlockOf::<T>::reservation(),
+            _ => CostsPerBlockOf::<T>::dispatch_stash(),
```
breathx, line 350: Bad practice for future changes
```suggestion
            LockIdentifier::DispatchStash => CostsPerBlockOf::<T>::dispatch_stash(),
```

common/src/gas_provider/internal.rs
```diff
@@ -592,17 +526,19 @@ where
...
 
         if node.is_consumed() {
             return Err(InternalError::node_was_consumed().into());
         }
 
-        if !node.lock().is_zero() {
-            return Err(InternalError::consumed_with_lock().into());
-        }
+        // Check if at least one lock has not been released
+        ensure!(
```
breathx, line 538: I'd remove foreign macros here to support the idea one day remove substrate deps here

common/src/gas_provider/internal.rs
```diff
@@ -805,7 +797,7 @@ where
             ancestor_node
         };
 
-        let node_lock = node.lock_mut();
+        let node_lock = &mut node.lock_mut()[id];
```
breathx, line 800:
```suggestion
        let node_lock = node.lock_mut()[id];
```
won't it work?

pallets/gear/src/tests.rs
```diff
@@ -587,7 +587,7 @@ fn delayed_send_user_message_payment() {
 fn delayed_send_user_message_with_reservation() {
```
breathx, line 587: should be solved within the PR
https://github.com/gear-tech/gear/blob/f087d0e3762281124c09fb23d804fb0a2d984048/pallets/gear/src/tests.rs#L678

pallets/gear/src/internal.rs
```diff
@@ -302,11 +303,20 @@ where
         }
     }
 
-    /// Charges for holding in some storage.
+    /// Charges for holding in some storage. In order to be placed into a holding storage
```
breathx, line 306: Adjust comment and say that unlock occurs

pallets/gear/src/internal.rs
```diff
@@ -315,25 +325,45 @@ where
...
+            _ => CostsPerBlockOf::<T>::dispatch_stash(),
+        };
+
+        // Unlocking all funds that have been locked for using the respective storage
+        let mut prepaid = GasHandlerOf::<T>::unlock_all(lock_id, id)
+            .unwrap_or_else(|e| unreachable!("GasTree corrupted! {:?}", e));
+
+        // In case gas hasn't been locked before putting a message in storage (for whatever reason)
+        // set the `prepaid` to the maximum possible value.
+        if prepaid.is_zero() {
```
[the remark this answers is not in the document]
breathx: otherwise it will trigger panic in other place that wouldn't be such readable

pallets/gear/src/internal.rs
```diff
@@ -315,25 +325,45 @@ where
...
+            prepaid = <CostsPerBlockOf<T> as SchedulingCostsPerBlock>::Cost::MAX;
+        }
+
         // Amount of gas to charge for holding.
-        let amount = duration.saturating_mul(cost);
+        // Note: the amount being charged is capped with the `prepaid` to handle potential
```
breathx, line 364: make sure unlock isn't used anywhere else except the file

pallets/gear/src/internal.rs
```diff
@@ -631,7 +649,7 @@ where
             let hold = delay_hold.min(maximal_hold);
 
             // Locking funds for holding.
-            GasHandlerOf::<T>::lock(dispatch.id(), hold.lock())
+            GasHandlerOf::<T>::lock(LockIdentifier::DispatchStash, dispatch.id(), hold.lock())
```
breathx, line 652: HoldBound may contain LockIdentifier to be consistent with usage within GasHandler::lock 
```suggestion
            GasHandlerOf::<T>::lock(dispatch.id(), hold.lock_type(), hold.lock_amount())
```

breathx (changes requested):

common/src/gas_provider/node.rs
```diff
@@ -70,6 +75,72 @@ impl<T> From<ReservationId> for GasNodeId<T, ReservationId> {
...
+
+    fn is_zero(&self) -> bool {
+        self.0.iter().all(|x| x.is_zero())
+    }
+}
+
+impl<Balance: Add + Zero + Copy> Add<Self> for NodeLock<Balance> {
+    type Output = Self;
+
+    fn add(self, other: Self) -> Self::Output {
```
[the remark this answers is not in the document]
ekovalev: We don't use this function anywhere, it's just a requirement that a `Zero` type has to be `Add`, hence such inefficient implementation. Anyway, yours looks good - will go with that one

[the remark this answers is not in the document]
ekovalev: > Furthermore, I'd go with saturating_add naming (as I used in playground op)

This would require making the `Balance` type parameter a `frame_support::traits::Saturating` everywhere to make the code compile while the only thing we want here is to implement the `Zero` trait. Left unchecked edition for the meantime - as I said, this function is never used in the code anyway.

pallets/gear/src/internal.rs
```diff
@@ -370,8 +403,10 @@ where
         }
 
         // Locking funds for holding.
-        GasHandlerOf::<T>::lock(dispatch.id(), hold.lock())
-            .unwrap_or_else(|e| unreachable!("GasTree corrupted! {:?}", e));
+        if let Some(lock_id) = hold.lock_id() {
```
breathx, line 406: here and in similar places as well

breathx (changes requested):
Seems great!
The last but not least thing now is to write storage migrations for the changes and test them locally until #2584 merged

pallets/gear/src/internal.rs
```diff
@@ -29,35 +30,44 @@ use common::{
...
-#[derive(Clone, Debug, Decode, Encode, PartialEq, Eq, PartialOrd, Ord)]
-pub(crate) struct HoldBoundCost<T: Config>(SchedulingCostOf<T>);
+/// `HoldBound<T>` builder
+#[derive(Clone, Debug)]
+pub(crate) struct HoldBoundBuilder<T: Config> {
+    storage_type: StorageType,
+    cost: SchedulingCostOf<T>,
+}
 
 #[allow(unused)]
```
ark0f, line 55: I guess it's obsolete?

pallets/gear/src/internal.rs
```diff
@@ -631,7 +655,10 @@ where
             let hold = delay_hold.min(maximal_hold);
 
             // Locking funds for holding.
-            GasHandlerOf::<T>::lock(dispatch.id(), hold.lock())
+            let lock_id = hold.lock_id().unwrap_or_else(|| {
```
ark0f, line 658: Consider remove `lock_id` function as it only used in cases like this - `StorageType` is few lines above

ark0f (changes requested):

common/src/gas_provider/internal.rs
```diff
@@ -881,22 +873,51 @@ where
...
+    Error: From<InternalError>,
+    ExternalId: Clone,
+    NodeId: Copy,
+    StorageMap: MapStorage<Key = NodeId, Value = GasNode<ExternalId, NodeId, Balance>>,
+{
     fn reserve(
-        key: Self::Key,
-        new_key: Self::ReservationKey,
+        key: impl Into<Self::NodeId>,
+        new_key: impl Into<Self::NodeId>,
```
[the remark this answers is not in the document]
ekovalev: > Consider something like
> 
> To keep type-safety like it was before

Not sure it's necessary: all we care about here at the Tree API level is that we require something that can uniquely identify the node being created and be used as the key in the respective `StorageMap`. The fact that for gas reservation nodes their `Id`s are derived from some `Reservation` id (and not message id) is merely an implementation detail

common/src/gas_provider/node.rs
```diff
@@ -70,6 +73,59 @@ impl<T> From<ReservationId> for GasNodeId<T, ReservationId> {
     }
 }
 
+#[derive(Clone, Copy, Decode, Encode, Debug, Default, PartialEq, Eq, TypeInfo, MaxEncodedLen)]
+#[codec(crate = codec)]
+#[scale_info(crate = scale_info)]
+pub struct NodeLock<Balance>([Balance; cardinality::<LockId>()]);
```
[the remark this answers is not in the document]
ekovalev: Not convinced it'd be appropriate: in `lockable.rs` we define an API of a lockable tree whereas here we define a concrete type to represent a node. It's all in the same crate anyway, after all.

common/src/gas_provider/node.rs
```diff
@@ -70,6 +73,59 @@ impl<T> From<ReservationId> for GasNodeId<T, ReservationId> {
...
+
+        for (i, elem) in inner.iter_mut().enumerate() {
+            *elem = *elem + other[i];
+        }
+
+        Self(inner)
+    }
+}
+
+#[cfg(feature = "extended-lock")]
```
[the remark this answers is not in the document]
ekovalev: It's only used in tests. Couldn't think of any use case (at least, existing ones) where we would want to sum up the locked amount across all lock types.

pallets/gear/src/internal.rs
ekovalev: > Consider create `LockId`, then convert into `StorageType` to avoid these `unwrap_or_else`

That would invert the flow of reasoning. It's not the `LockId` that determines the `StorageType` which, in turn, determines the `HoldBound` but the other way around: we create a `HoldBound` for a specific `StorageType` which may (or may not in theory) have an associated `LockId`. In reality though any storage we are currently charging for would have a lock therefore this `_or_else` part will never execute.

gsdk/src/metadata/mod.rs
```diff
@@ -634,6 +634,8 @@ pub mod runtime_types {
                     pub spec_refs: ::core::primitive::u32,
                     pub unspec_refs: ::core::primitive::u32,
                 }
+                // Making use of zero cost abstraction over the inner array representing locks
```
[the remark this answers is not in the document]
ekovalev: Well, the `gclient` `node` tests work fine: the lock data is decoded correctly. But I agree that maybe it's better to use the same `new_type` pattern to abstract away the inner representation here, as well.

pallets/gear/src/internal.rs
ekovalev: Seems, there still is a currently unused method in the `HoldBoundBuilder` implementation. Hence leaving as is for now.

common/src/gas_provider/node.rs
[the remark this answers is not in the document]
breathx: I'd just leave it as it is. Less features is better. Or use cfg[test] for such case

runtime/vara/src/migrations.rs
```diff
@@ -14,4 +15,4 @@ impl OnRuntimeUpgrade for SessionValidatorSetMigration {
     }
 }
 
-pub type Migrations = (SessionValidatorSetMigration,);
+pub type Migrations = (SessionValidatorSetMigration, GasTreeMigration<Runtime>);
```
breathx, line 18: Why should we specify all migrations in runtime directly?

pallets/gas/src/migrations.rs
```diff
@@ -0,0 +1,680 @@
...
+                        _ => LockId::Reservation,
+                    }
+                };
+
+                let mut translated = 0_u64;
+                GasNodes::<T>::translate::<OldGasNode<AccountIdOf<T>, NodeId, Balance>, _>(
+                    |key, old_value| {
+                        let mut new_lock = NodeLock::<Balance>::zero();
+                        let new_value = match old_value {
+                            OldGasNode::External {
```
breathx, line 182:
```suggestion
                            v1::GasNode::External {
```
Seems more readable for me taking in account that there will be more migrations in the file soon

pallets/gas/src/migrations.rs
```diff
@@ -0,0 +1,680 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use crate::{AccountIdOf, Balance, Config, GasNodes, Pallet};
+
+pub mod v1 {
```
breathx, line 21: the only thing maybe we can import here already existing mock runtime ?

breathx (approved):
👍🏻

common/src/gas_provider/node.rs
```diff
@@ -70,6 +74,52 @@ impl<T> From<ReservationId> for GasNodeId<T, ReservationId> {
     }
 }
 
+#[derive(Clone, Copy, Decode, Encode, Debug, Default, PartialEq, Eq, TypeInfo, MaxEncodedLen)]
+#[codec(crate = codec)]
+#[scale_info(crate = scale_info)]
+pub struct NodeLock<Balance>([Balance; 4]);
+
+const_assert!(cardinality::<LockId>() == 4);
```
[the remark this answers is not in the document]
ark0f: Should we leave const assertion to be sure `NodeLock` hasn't changed so migration is not required?

pallets/gas/src/migrations.rs
ekovalev: Done

pallets/gas/src/migrations.rs
ekovalev: > Seems more readable for me taking in account that there will be more migrations in the file soon

Yeah, but the problem is that we are importing the post-migration `GasNode` from `gear-common` into the namespace, as well - in any case we'd need to use different names for these two types.

runtime/vara/src/migrations.rs
ekovalev: The block executor needs an implementation of the `OnRuntimeUpgrade` trait to apply migrations

ark0f (approved):

PR #2529, opened 2023-04-10
Increase fuzzer performance

SabaunT:
After each run the events store grows bigger, so iterating on it becomes very expensive. Optimize that by removing events after processing them.


gshep (approved):
Great! :100:

PR #2543, opened 2023-04-13
Disable `lto` in dev profile for `wasm-builder`

ukint-vs:
lto breaks clippy and rust-lld in debug profile with latest nightly.

@gear-tech/dev


SabaunT (approved):

ukint-vs:
> great work!

thanks for the tip!

PR #2546, opened 2023-04-15
Check `Sudid` event in gclient

ark0f:
Also add `set_code` and `set_code_without_checks` calls

### DIFF
--- a/gclient/src/api/mod.rs
+++ b/gclient/src/api/mod.rs
@@ -108,6 +108,11 @@ impl GearApi {
         }
     }
 
+    /// Returns Web Socket address the node is listening to.
+    pub fn ws_address(&self) -> &WSAddress {
+        self.1.as_ref().unwrap().ws_address()
+    }
+
     /// Create and init a new `GearApi` instance that will be used with the
     /// public Gear testnet.
     pub async fn gear() -> Result<Self> {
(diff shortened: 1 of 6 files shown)


gclient/tests/sudo.rs
```diff
@@ -0,0 +1,30 @@
...
+
+#[tokio::test]
+async fn set_code_failed() {
+    let api = GearApi::dev_from_path("../target/release/gear")
+        .await
+        .unwrap();
+    let _err = api
+        .set_code_by_path("../target/release/wbuild/gear-runtime/gear_runtime.wasm")
+        .await
+        .unwrap_err();
```
DennisInSky, line 29: Do you want to make any assertions here?

DennisInSky (approved):

gclient/tests/sudo.rs
breathx: Agree @DennisInSky

gclient/src/api/mod.rs
```diff
@@ -108,6 +108,11 @@ impl GearApi {
         }
     }
 
+    /// Returns Web Socket address the node is listening to.
+    pub fn ws_address(&self) -> &WSAddress {
+        self.1.as_ref().unwrap().ws_address()
```
breathx, line 113:
```suggestion
        self.1.as_ref().expect("msg").ws_address()
```

breathx (approved):

PR #2547, opened 2023-04-16
feat!: Auto replies

### LINKED ISSUE #1676: Invariant: every message should get a reply
If program is voluntary does not create a reply message, it should be generated by the runtime

### LINKED ISSUE #1741: Send error reply to program A that program B message is out of rent in waitlist
### Problem to Solve

- Program A sends message to program B
- Program B calls `wait()` so message is in waitlist
- Message being removed from waitlist after some blocks for being in storage

So should we send error reply to program A and how?

### Possible Solution

.

### Notes

_No response_

### LINKED ISSUE #1742: Send signal to program when message is removed from mailbox
### Problem to Solve

Should we send signal to program itself when message being removed from mailbox

### Possible Solution

.

### Notes

_No response_

breathx:
Resolves #1676, resolves #1741 and resolves #1742.


gshep (approved):
Great work! :100:

PR #2549, opened 2023-04-17
feat(utils): switch to `clap` arg parser instead of `structopt`

### LINKED ISSUE #2442: Use clap instead of structopt in utils
### File Location(s)

https://github.com/gear-tech/gear/blob/a2f6ac69cd73a911494f8a5a78f00e08c3d1a8f9/utils/node-loader/Cargo.toml#L31

--- 

- gear-node-loader
- runtime-fuzzer
- wasm-info

### Proposal

most of our CLI are using clap directly for now since it just support proc-macro as well last years, so using clap instead of structopt here as well for syncing dependencies

tltsutltsu:
Resolves #2442.

Changes made due to the transition:
- In `wasm-info`, the shorthand for `--hex` has been changed to `-H` from `-h`. This has been done because it conflicts with the shorthand for help in current version of clap. Another option is to disable both `--help` and `-h`, but this requires further discussion.

Rework of PR #2510

### DIFF
--- a/utils/runtime-fuzzer/Cargo.toml
+++ b/utils/runtime-fuzzer/Cargo.toml
@@ -16,8 +16,7 @@ rand = { workspace = true, features = ["small_rng"] }
 
 # Temporary deps for the reproducing crash script until #2313 is implemented
 #
-# TODO: use clap instead
-structopt = "0.3.26"
+clap = { workspace = true, features = ["derive"] }
 
 gear-call-gen.workspace = true
 runtime-primitives.workspace = true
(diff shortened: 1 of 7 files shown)


utils/runtime-fuzzer/src/bin/reproduce_fuzz.rs
```diff
@@ -25,24 +25,24 @@
...
     path::PathBuf,
 };
-use structopt::StructOpt;
 
-#[derive(Debug, StructOpt)]
+#[derive(Debug, Parser)]
 pub struct Params {
     /// Path to the file, which contains seeds from previously run fuzzer.
-    #[structopt(short = "p", long, parse(from_os_str))]
+    #[arg(long)]
```
breathx, line 38:
```suggestion
    #[arg(short, long)]
```

utils/runtime-fuzzer/Cargo.toml
```diff
@@ -16,8 +16,7 @@ rand = { workspace = true, features = ["small_rng"] }
 
 # Temporary deps for the reproducing crash script until #2313 is implemented
 #
```
breathx, line 18:
```suggestion
```

breathx (approved):

PR #2565, opened 2023-04-19
wasm-gen: Generate stub for send_from_reservations

DennisInSky:
Resolves # .

-
-
-

@reviewer-or-team

### DIFF
--- a/utils/wasm-gen/src/lib.rs
+++ b/utils/wasm-gen/src/lib.rs
@@ -728,25 +728,61 @@ impl<'a> WasmGen<'a> {
     pub fn insert_sys_calls(
         &mut self,
         builder: ModuleBuilderWithData,
+        syscall_data: &BTreeMap<SysCallName, SyscallData>,
         memory_pages: WasmPageCount,
     ) -> ModuleWithDebug {
-        if builder.code_size == 0 {
-            return builder.into();
-        }
-
         let ModuleBuilderWithData {
-            module_builder: mut builder,
+            module_builder: builder,
             offsets,
             last_offset,
-            import_count,
-            code_size,
+            ..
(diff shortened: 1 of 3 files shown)


utils/wasm-gen/src/lib.rs
```diff
@@ -788,63 +855,157 @@ impl<'a> WasmGen<'a> {
...
+            Instruction::I64Load(3, 24),
+            Instruction::I64Store(3, 24),
+            Instruction::I32Const(pid_value_ptr),
+            Instruction::GetLocal(0),
+            Instruction::I64Load(3, 32),
+            Instruction::I64Store(3, 32),
+            Instruction::I32Const(pid_value_ptr),
+            Instruction::GetLocal(0),
+            Instruction::I64Load(3, 40),
+            Instruction::I64Store(3, 40),
```
DennisInSky, line 954: This memory copying does not look nice. I can think of making some sort of macro for this. Other thoughts?

gshep: you have to resolve import indices/etc in that case

gshep (approved):
LGTM!

DennisInSky:
> `call` means internal calls in wasm in general, so mb combine syscalls and combocalls ( mb better naming instead, extcalls? ) to calls is not proper, so imo
> 
> 1. wat file for combocalls
> 2. create a new file for combocalls
> 3. split shared functions in `lib.rs` and make it simple

I do not quite understand why it should be called a combocall. Google does not suggest any reasonbale results and as such I consider this terminology made up and confusing. Any call is just a call. We just distinguish a subset of them which are improted ones (they are not even obliged to a sys call)

PR #2579, opened 2023-04-23
fix(node): slow down grandpa finality.

ukint-vs:
Resolves # .

Slower grandpa finality to reduce CPU usage.

@gear-tech/dev


grishasobol:
How this change will increase blocks amount between produced and finalised blocks?

ukint-vs:
> How this change will increase blocks amount between produced and finalised blocks?

https://github.com/paritytech/polkadot/issues/5464

grishasobol (approved):

PR #2591, opened 2023-04-26
feat(gsdk): squash node wrappers to gsdk

### LINKED ISSUE #2481: Move node wrapper into gsdk with testing feature
### Problem to Solve

duplicated code in 

https://github.com/gear-tech/gear/blob/master/gclient/src/node/mod.rs

and 

https://github.com/gear-tech/gear/blob/master/gcli/tests/common/node.rs

### Possible Solution

well we can just deduplicate the code to gsdk

### Notes

_No response_

clearloop:
Resolves #2481 

- [x] rename node wrapper from `gcli/common` to `gsdk/testing`
- [x] rename node wrapper from `gclient/node` to `gsdk/testing`
- [ ] 

@gear-tech/dev


gcli/tests/common/traits.rs
```diff
@@ -28,3 +30,10 @@ impl Convert<String> for Vec<u8> {
         String::from_utf8_lossy(self).to_string()
     }
 }
+
+/// Run node.
+pub trait NodeExt {
```
DennisInSky, line 35: -> `NodeRunner` ?

gcli/tests/common/traits.rs
```diff
@@ -28,3 +30,10 @@ impl Convert<String> for Vec<u8> {
...
+    fn run(&self, args: Args) -> Result<Output>;
+
+    fn ws(&self) -> String;
```
DennisInSky, line 38: Doc for this as it does not look quite obvious?

gsdk/src/testing/node.rs
```diff
@@ -1,46 +1,43 @@
...
-pub(crate) mod result;
-pub(crate) mod ws;
 
 /// A struct representing a node running on local PC.
 #[derive(Debug)]
-pub(crate) struct Node {
+pub struct Node {
     process: Child,
-    ws_address: WSAddress,
+    address: SocketAddrV4,
```
DennisInSky, line 31: Could it be that we might want to interact with the node via http and ws both?

DennisInSky:
What are the benefits of craming testing functionality into gsdk using conditional compilation compared to having the same functionality in a separate crate like gsdk_testing which can be used in gclient and gcli. Wondering, but it does not sound that gcli uses gsdk for testing purposes while the feature is called testing. I guess the same is valid for gclient

gclient/src/api/error.rs
```diff
@@ -89,4 +89,10 @@ pub enum Error {
     /// Occurs when node spawining failed.
     #[error(transparent)]
     Node(#[from] NodeError),
+    /// Occurs when parsing websocket domain failed.
+    #[error("Failed to parse WebSocket domain.")]
+    IncorrectWSDomain,
```
DennisInSky, line 94: Should be somthing like `WSDomainIncorrect` so it is consistent with errors like `TimestampNotFound`, `MaxDepthReached`

gcli/tests/common/traits.rs
```diff
@@ -28,3 +30,23 @@ impl Convert<String> for Vec<u8> {
...
+    /// let args = Args::new("upload")
+    ///              .flag("--code-only")
+    ///              .program(env::wasm_bin("demo_meta.opt.wasm"));
+    /// let output = node.run(args)
+    ///
+    /// // ...
+    /// ```
+    fn run(&self, args: Args) -> Result<Output>;
+
+    fn ws(&self) -> String;
```
DennisInSky, line 51: imo, just `ws` is not descriptive enough

DennisInSky (approved):

PR #2594, opened 2023-04-27
add script for rust nightly version local pin

grishasobol:


scripts/pin-rust-nightly.sh
```diff
@@ -0,0 +1,25 @@
...
+    exit 1
+fi
+
+pin_date=$1
+os_name="$(uname)"
+
+if [ "$os_name" = "Darwin" ]; then
+    rustup toolchain install nightly-$pin_date --component llvm-tools-preview
+    rustup target add wasm32-unknown-unknown --toolchain nightly-$pin_date
+    rm -rf ~/.rustup/toolchains/nightly-aarch64-apple-darwin
```
shamilsan, line 16: This will not work on x84 macOS. I'd better use target-triple info from the compiler itself. E.g.:

```shell
❯ rustc -Vv | grep "host: " | sed "s/^host: \(.*\)$/\1/"
aarch64-apple-darwin
```

grishasobol: ok add identifying

scripts/pin-rust-nightly.sh
```diff
@@ -0,0 +1,28 @@
...
+    rustup target add wasm32-unknown-unknown --toolchain nightly-$pin_date
+    rm -rf ~/.rustup/toolchains/nightly-$suffix
+    mv ~/.rustup/toolchains/nightly-$pin_date-$suffix ~/.rustup/toolchains/nightly-$suffix
+elif [ "$os_name" = "Linux" ]; then
+    rustup toolchain install nightly-$pin_date --component llvm-tools-preview
+    rustup target add wasm32-unknown-unknown --toolchain nightly-$pin_date
+    rm -rf ~/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu
+    mv ~/.rustup/toolchains/nightly-$pin_date-x86_64-unknown-linux-gnu ~/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu
+else
+    echo "Unknown operating system"
```
shamilsan, line 27: No need to process Linux separately due to autodetect.

shamilsan (commented):
I'd add it to the Makefile. E.g. just added similar thing here:

https://github.com/gear-dapps/app/blob/master/Makefile#L3-L4
https://github.com/gear-dapps/app/blob/master/Makefile#L17-L22

scripts/pin-rust-nightly.sh
grishasobol: Ok, changed in sh file. But I think using `init` `Makefile` is a little not correct way for core team. Because we will have different not connected nightly version in CI

shamilsan (approved):

PR #2595, opened 2023-04-27
feat!: Add new RI to pre process and access memory simultaneously 

grishasobol:


grishasobol:
Cannot be done presently because must be done with colaboration of new host sandbox  realization

PR #2599, opened 2023-04-27
build(wasm-builder)!: Pass program build features to wasm-builder

### LINKED ISSUE #2407: Support features of programs in wasm-builder
### Problem to Solve

https://github.com/gear-tech/gear/blob/97525093e88e44954e1c8d56470c1163ce5f018c/utils/wasm-builder/src/cargo_command.rs#L71

`CargoCommand::run(&self)` ignores the features provided in the original command

### Possible Solution

insert features as well in https://github.com/gear-tech/gear/blob/97525093e88e44954e1c8d56470c1163ce5f018c/utils/wasm-builder/src/cargo_command.rs#L36

### Notes

_No response_

DennisInSky:
Resolves #2407  .

It might be a breaking change for some users.

@reviewer-or-team


utils/wasm-builder/src/wasm_project.rs
```diff
@@ -127,11 +130,43 @@ impl WasmProject {
         &self.profile
     }
 
+    /// Return features available in the generated `Cargo.toml`.
+    pub fn features(&self) -> &[String] {
+        self.features
+            .as_ref()
+            .expect("Run project generation first")
```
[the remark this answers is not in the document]
DennisInSky: `First` is exactly what it is meant to be. It is done similar to https://github.com/gear-tech/gear/blob/master/utils/wasm-builder/src/wasm_project.rs#L234
This code contains a lot of other calls of the `expect` function.

utils/wasm-builder/src/cargo_command.rs
```diff
@@ -67,6 +69,11 @@ impl CargoCommand {
         self.profile = profile;
     }
 
+    /// Set features.
+    pub fn set_features(&mut self, features: &[String]) {
```
shamilsan, line 73: `&[&str]` is more general, but not critical.
```suggestion
    pub fn set_features(&mut self, features: &[&str]) {
```

utils/wasm-builder/src/lib.rs
```diff
@@ -39,6 +40,7 @@ pub const TARGET: &str = env!("TARGET");
 pub struct WasmBuilder {
     wasm_project: WasmProject,
     cargo: CargoCommand,
+    excluded_features: Vec<&'static str>,
```
shamilsan, line 43: I'd rename it as `excluded` may confuse. Maybe `scoped_features`? With some description as a doc-comment.

utils/wasm-builder/src/lib.rs
```diff
@@ -79,21 +90,57 @@ impl WasmBuilder {
...
+            .filter(|feature| {
+                // Omit the `default` feature because the generated project excludes it
+                feature != "default" && !features_to_exclude.contains(&feature.as_str())
+            })
+            .collect::<Vec<String>>();
+        wasm_project.match_features(&features)
+    }
+
+    /// Force cargo to rebuild the build script next time it is invoked (e.g. when a feature is added or removed).
+    fn force_build_script_rebuild_on_next_run() {
```
shamilsan, line 135: The function name is too long.
```suggestion
    fn refresh_build_script() {
```

utils/wasm-builder/src/lib.rs
```diff
@@ -102,17 +149,42 @@ impl Default for WasmBuilder {
...
+const FEATURES_TO_EXCLUDE_BY_DEFAULT: &[&str] = &["std"];
+
 /// Shorthand function to be used in `build.rs`.
 pub fn build() {
-    WasmBuilder::new().build();
+    build_custom(FEATURES_TO_EXCLUDE_BY_DEFAULT)
+}
+
+/// Shorthand function similar to the [`build`] one, but allowing some customizations.
+pub fn build_custom(features_to_exclude: &[&'static str]) {
```
shamilsan, line 162: Too common function name.
```suggestion
pub fn build_with_scoped_features(scoped_features: &[&'static str]) {
```

shamilsan (approved):

examples/binaries/proxy/Cargo.toml
```diff
@@ -7,7 +7,7 @@ license = "GPL-3.0"
 workspace = "../../../"
 
 [dependencies]
-gstd = { workspace = true, features = ["debug"] }
+gstd = { workspace = true }
```
DennisInSky, line 10: I am going to the same for all examples right before mering to the `master`

utils/wasm-builder/src/lib.rs
DennisInSky: Imo, `scoped` says even less, and frunkly nothing to me. I can suggest `omitted`, `ignored`. Any other preferences?

utils/wasm-builder/src/lib.rs
DennisInSky: The idea behind this name is making it as a fully customisable shortcut, i.e. if anything else needs to be added as a customisation, it will be added as yet another param of this function. It eliminates the need of making up some weird function names. You won't be able to come up with a set of shortcuts which cover all possible use cases.

DennisInSky:
> lgtm! but for the testing part, I think we can test it with https://github.com/gear-tech/gear/tree/master/utils/wasm-builder/test-program

@clearloop can you please elaborate on it?

utils/wasm-builder/src/lib.rs
DennisInSky: I prefer a longer but more descriptive name if I can't come up with a shorter one as it seems more confusing

utils/wasm-builder/src/lib.rs
shamilsan: Maybe `skipped` but it is pretty the same as `excluded` :)

utils/wasm-builder/src/lib.rs
shamilsan: Any addition to this function will break backward compatibility. Is it a sense to make such kind of function?

DennisInSky: Nice catch!!! I will scrap it away then. Let people use underlaying builder directly in case of customisations.

examples/binaries/proxy/Cargo.toml
SabaunT: still need this here, maybe do it here: #2602

DennisInSky: @SabaunT What do you mean? Why do we need building with the `debug` feature all the time?

PR #2604, opened 2023-05-01
feat(gsdk): introduce proc-macro for generating short storage queries

### LINKED ISSUE #2484: Come up with an idea of simplifying gsdk calls related to getting info from storage
### File Location(s)

gsdk/src/storage.rs

### Proposal

Accessing storage requires 2 calls: one for getting info at the specified block and another for getting the same info, but at the most recent one. Basically the latter is a call of the former with `None` as block hash.

clearloop:
Resolves #2484 

introduced proc-macro `short`

```rust
/// imdocs at specified block.
#[short]
pub async fn storage_query_at(&self, a: T, block_hash: Option<H256>) -> R {
  // ...
}

// ----> will generate

/// imdocs
pub async fn storage_query(&self, a: T) -> R {
  self.storage_query_at(a, None).await
}
```

@gear-tech/dev

### DIFF
--- a/gsdk/codegen/src/storage.rs
+++ b/gsdk/codegen/src/storage.rs
@@ -65,13 +65,13 @@ impl StorageQueryBuilder {
     }
 
     /// Build storage query for the latest state.
-    fn full(&self) -> ItemFn {
-        let mut full = self.0.clone();
+    fn latest(&self) -> ItemFn {
+        let mut latest = self.0.clone();
 
         // reset function docs.
         //
         // - remove `at specified block` suffix.
-        full.attrs.iter_mut().for_each(|attr| {
+        latest.attrs.iter_mut().for_each(|attr| {
             if let Meta::NameValue(MetaNameValue {
                 value:
                     Expr::Lit(ExprLit {
@@ -88,15 +88,15 @@ impl StorageQueryBuilder {
         // reset function name.
         //
         // - `storage_query_at` -> `storage_query`
-        full.sig.ident = Ident::new(
-            full.sig.ident.to_string().trim_end_matches(AT_SUFFIX),
-            full.sig.ident.span(),
+        latest.sig.ident = Ident::new(
+            latest.sig.ident.to_string().trim_end_matches(AT_SUFFIX),
+            latest.sig.ident.span(),
         );
 
         // reset function inputs.
         //
         // - remove `block_hash: Option<H256>`
-        full.sig.inputs = Punctuated::from_iter(full.sig.inputs.into_iter().filter(|v| {
+        latest.sig.inputs = Punctuated::from_iter(latest.sig.inputs.into_iter().filter(|v| {
             if let FnArg::Typed(PatType { ty, .. }) = v {
                 return !ty
                     .to_token_stream()
@@ -117,7 +117,7 @@ impl StorageQueryBuilder {
         // ```
         {
             let fn_at = &self.0.sig.ident;
-            let args = full
+            let args = latest
                 .sig
                 .inputs
                 .iter()
@@ -133,21 +133,21 @@ impl StorageQueryBuilder {
                 })
                 .collect::<Vec<Ident>>();
 
-            full.block.stmts = parse_quote! {
+            latest.block.stmts = parse_quote! {
                 self.#fn_at(#(#args,)* None).await
             };
         }
 
-        full
+        latest
     }
 
     /// Build all storage queries.
     pub fn build(&self) -> TokenStream {
-        let (at, full) = (self.at(), self.full());
+        let (at, latest) = (self.at(), self.latest());
         quote! {
             #at
 
-            #full
+            #latest
         }
         .into()
     }


breathx:
Can we do it this way:
```rust
/// DOCS.
#[short]
pub async fn storage_query_at(&self, a: T, block_hash: Option<H256>) -> R { / .. / }

// ----
// EXPANDS INTO
// ----


async fn _storage_query_at(&self, a: T, block_hash: Option<H256>) -> R { / .. / }

/// DOCS
pub async fn storage_query(&self, a: T) -> R {
    _storage_query_at(None).await
}

/// DOCS as specific block_hash
pub async fn storage_query(&self, a: T, block_hash: H256) -> R {
    _storage_query_at(Some(block_hash)).await
}
```
See no reason to keep Option field in `*_at` fns

DennisInSky:
I do not quite like the macro name. It lacks of being descriptive enough. What do you think about `storage_read_func` or `storage_read_func_essence` denoting that this is a piece of code used for generating storage access funcs?

DennisInSky:
Another confusing thing wrt to naming is the `derive` in the name of the macro crate. As they say here https://doc.rust-lang.org/rust-by-example/trait/derive.html the Derive is about some trait implementation where you say which trait one wants implemented

DennisInSky:
> I do not quite like the macro name. It lacks of being descriptive enough. What do you think about `storage_read_func` or `storage_read_func_essence` denoting that this is a piece of code used for generating storage access funcs?

Maybe something like `read_storage_func_body` will be the most appropriate

gsdk/Cargo.toml
```diff
@@ -26,3 +26,4 @@ subxt.workspace = true
 thiserror.workspace = true
 sp-runtime = { workspace = true, features = [ "std" ] }
 sp-core.workspace = true
+gsdk-codegen = { path = "./codegen" }
```
breathx, line 29:
```suggestion
gsdk-codegen = { path = "codegen" }
```

gsdk/codegen/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "gsdk-codegen"
+version = "0.1.0"
+edition = "2021"
```
breathx, line 4:
```suggestion
[package]
name = "gsdk-codegen"
version = "0.1.0"
authors = ["Gear Technologies"]
edition = "2021"
license = "GPL-3.0"
```

gsdk/codegen/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use proc_macro::TokenStream;
+use syn::ItemFn;
+
+mod storage;
+
+/// Builds fn `query_storage_at` to
```
breathx, line 24: I'd add code sample from our GitHub conversation here

gsdk/codegen/src/storage.rs
```diff
@@ -0,0 +1,131 @@
...
+
+impl StorageQueryBuilder {
+    /// Build storage query from full blocks.
+    fn full(&self) -> ItemFn {
+        let mut short = self.0.clone();
+
+        // reset function docs.
+        //
+        // - remove `at specified block` suffix.
+        short.attrs.iter_mut().find_map(|attr| {
```
breathx, line 43: Seems `find_map` usage is incorrect here. Try using `for_each`

DennisInSky:
> > Maybe something like `read_storage_func_body` will be the most appropriate
> 
> but this macro does not just generate a body, but functions with different usages, mark the generated functions' usage(s) make more sense tbh
> 
> since `fetch_at` means fetch storage at block, `fetch` means fetch storage from full blocks, so the full path of this macro could be `gsdk_codegen::storage::fetch_full`
> 
> or just mark `storage_fetch` and the docs will show the details

My understanding that this macro usage is more like an attribute especially in variant suggested by @breathx, i.e. with that macro one says "this is how I want data to be fetched, please generate proper set of funcs based on this"

gsdk/codegen/src/storage.rs
```diff
@@ -0,0 +1,131 @@
...
+            }
+
+            None
+        });
+
+        // reset function name.
+        //
+        // - `storage_query_at` -> `storage_query`
+        short.sig.ident = Ident::new(
+            short.sig.ident.to_string().trim_end_matches(FULL_SUFFIX),
```
breathx, line 67: Don't we want to add some validation here for ident that it end with `_at` ? Same for comment and args

gsdk/codegen/src/storage.rs
```diff
@@ -0,0 +1,131 @@
...
+                self.#fn_at(#(#args,)* None).await
+            };
+        }
+
+        short
+    }
+
+    /// Build all storage queries.
+    pub fn build(&self) -> TokenStream {
+        let (at, full) = (self.0.clone(), self.full());
```
breathx, line 117: I'd request to write original signature using Option<H256> and placing into impl within macro expand

breathx (changes requested):

gsdk/codegen/src/storage.rs
```diff
@@ -0,0 +1,202 @@
...
+        stmts.push(parse_quote! {
+            let block_hash = #ident.into();
+        });
+        at.block.stmts = [stmts, at.block.stmts].concat();
+
+        at
+    }
+
+    /// Build storage query for full blocks.
+    fn full(&self) -> ItemFn {
```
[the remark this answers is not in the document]
DennisInSky: `at_latest`, `latest` works fine to my pov

ark0f:
I noticed that your changes have resulted in an increase of 300 lines, which seems excessive for one-time boilerplate. Can you please help me to understand reasons behind these changes?

NikVolf:
@breathx can you re-review?

PR #2617, opened 2023-05-04
refactor(runtime): change saturating to checked in `MAXIMUM_BLOCK_WEIGHT`

### LINKED ISSUE #2021: Change saturating to checked with static_assertion check
https://github.com/gear-tech/gear/blob/80d087190d9883fdffc322bc43119a80909baff2/runtime/vara/src/lib.rs#L120-L122

tltsutltsu:
Resolves #2021.

@gear-tech/dev

### DIFF
--- a/runtime/gear/src/lib.rs
+++ b/runtime/gear/src/lib.rs
@@ -16,7 +16,6 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
-#![feature(const_option)]
 #![cfg_attr(not(feature = "std"), no_std)]
 // `construct_runtime!` does a lot of recursion and requires us to increase the limit to 256.
 #![recursion_limit = "256"]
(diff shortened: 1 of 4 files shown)


runtime/gear/src/lib.rs
```diff
@@ -16,6 +16,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+#![feature(const_option)]
```
grishasobol, line 19: Can we use stable rust to compile after that feature enabled ?

grishasobol: Unfortunately we cannot then use this solution. May be  better to change `saturating_div` to common `div`. Do you know the difference between them. Actually I think they have no difference))

runtime/gear/src/lib.rs
```diff
@@ -119,11 +121,19 @@ pub const BABE_GENESIS_EPOCH_CONFIG: sp_consensus_babe::BabeEpochConfiguration =
         allowed_slots: sp_consensus_babe::AllowedSlots::PrimaryAndSecondaryPlainSlots,
     };
 
+// We'll check that WEIGHT_REF_TIME_PER_SECOND is not overflows so we can use
```
shamilsan, line 124:
```suggestion
// We'll check that WEIGHT_REF_TIME_PER_SECOND does not overflow, so we can use
```

runtime/vara/src/lib.rs
```diff
@@ -136,13 +138,23 @@ pub const BABE_GENESIS_EPOCH_CONFIG: sp_consensus_babe::BabeEpochConfiguration =
         allowed_slots: sp_consensus_babe::AllowedSlots::PrimaryAndSecondaryPlainSlots,
     };
 
+// We'll check that WEIGHT_REF_TIME_PER_SECOND is not overflows so we can use checked_mul and
+// checked_div instead of saturating ones.
+const_assert!(WEIGHT_REF_TIME_PER_SECOND
+    .checked_div(3)
+    .expect("cannot divide WEIGHT_REF_TIME_PER_SECOND by 3")
```
shamilsan, line 145: We can avoid `expect` here if replacing this `const_assert` with two. So, we can omit the `const_option` feature.

runtime/gear/src/lib.rs
tltsutltsu: Removed with @shamilsan advice

runtime/gear/src/lib.rs
```diff
@@ -119,11 +121,19 @@ pub const BABE_GENESIS_EPOCH_CONFIG: sp_consensus_babe::BabeEpochConfiguration =
...
+
 /// We allow for 1/3 of block time for computations, with maximum proof size.
 ///
 /// It's 1/3 sec for gear runtime with 1 second block duration.
-const MAXIMUM_BLOCK_WEIGHT: Weight =
-    Weight::from_parts(WEIGHT_REF_TIME_PER_SECOND.saturating_div(3), u64::MAX);
+const MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_parts(
+    WEIGHT_REF_TIME_PER_SECOND
+        .checked_div(3)
+        .expect("cannot divide WEIGHT_REF_TIME_PER_SECOND by 3"),
```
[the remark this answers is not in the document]
tltsutltsu: Yes

runtime/vara/src/lib.rs
tltsutltsu: Thanks! Done

grishasobol (approved):

PR #2620, opened 2023-05-04
Bump version for new release

NikVolf:


shamilsan (approved):

gcli/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "gcli"
-version = "0.1.5"
+version = "0.1.6"
```
shamilsan, line 3: Our node version is here: https://github.com/gear-tech/gear/blob/master/node/cli/Cargo.toml#LL3C1-L3C18

And the next version is `0.1.5`

NikVolf: Unfortunately we already released 0.1.5, so we skip 0.1.5 in package spec

`gcli` I believe should also be bumped synchrnously with `node/cli`

PR #2624, opened 2023-05-05
Implement save/load of program memory state in gtest/gclient

### LINKED ISSUE #2055: Add a possibility to save and restore program's memory pages in `gtest`/`gclient`

mertwole:
Resolves #2055 

@breathx


common/src/memory_dump.rs
```diff
@@ -0,0 +1,98 @@
+// This file is part of Gear.
+
+// Copyright (C) 2021-2022 Gear Technologies Inc.
```
shamilsan, line 3:
```suggestion
// Copyright (C) 2023 Gear Technologies Inc.
```

common/src/memory_dump.rs
```diff
@@ -0,0 +1,98 @@
...
+
+impl MemoryPageDump {
+    pub fn new(page_number: GearPage, page_data: PageBuf) -> MemoryPageDump {
+        let mut data_vec = vec![];
+        page_data.encode_to(&mut data_vec);
+        let data = if data_vec.iter().all(|&byte| byte == 0) {
+            None
+        } else {
+            Some(hex::encode(data_vec))
+        };
```
shamilsan, line 39: Might be shorter:
```suggestion
        let data = data_vec.iter().any(|&byte| byte != 0).then(|| hex::encode(data_vec));
```

common/src/memory_dump.rs
```diff
@@ -0,0 +1,98 @@
...
+
+#[derive(Serialize, Deserialize, Default)]
+pub struct ProgramMemoryDump {
+    pub balance: u128,
+    pub reserved_balance: u128,
+    pub pages: Vec<MemoryPageDump>,
+}
+
+impl ProgramMemoryDump {
+    pub fn save_to_file<P: AsRef<Path>>(&self, path: P) {
```
shamilsan, line 69: `impl` is more readable for single generic.
```suggestion
    pub fn save_to_file(&self, path: impl AsRef<Path>) {
```

shamilsan (commented):
Need to fix test too

ukint-vs (approved):

PR #2643, opened 2023-05-11
chore(gsdk): move rpc tests from gcli to gsdk

### LINKED ISSUE #2481: Move node wrapper into gsdk with testing feature

clearloop:
Resolves #2481 

- [x] clean outdated code in gcli
- [x] move rpc from gsdk to gcli

@gear-tech/dev

### DIFF
--- a/gsdk/tests/rpc.rs
+++ b/gsdk/tests/rpc.rs
@@ -15,7 +15,7 @@
 //
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
-#![cfg(test)] // workaround for using `testing` module.
+#![cfg(feature = "testing")]
 
 use gear_core::ids::{CodeId, ProgramId};
 use gsdk::{
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -240,6 +240,9 @@ jobs:
       - name: "Test: Gear workspace"
         run: ./scripts/gear.sh test gear --exclude gclient --exclude gcli --features pallet-gear-debug/lazy-pages --release --locked
 
+      - name: "Test: `gsdk`"
+        run: ./scripts/gear.sh test gsdk --release --locked --retries 3
+
       - name: "Test: `gcli`"
         run: ./scripts/gear.sh test gcli --release --locked --retries 3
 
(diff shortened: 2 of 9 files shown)


gsdk/Cargo.toml
```diff
@@ -30,5 +30,11 @@ sp-core.workspace = true
 # deps for features `testing`
 rand = { workspace = true, optional = true }
 
+[dev-dependencies]
+rand = { workspace = true }
+demo-waiter.workspace = true
```
DennisInSky, line 35: Can we use `= { workspace = true }` everywhere instead of `.workspace = true`? Even in this single file there are 2 different ways of it

gsdk/tests/rpc.rs
```diff
@@ -15,24 +15,72 @@
...
+
+/// Generate program id from code id and salt
+fn program_id(bin: &[u8], salt: &[u8]) -> ProgramId {
+    ProgramId::generate(CodeId::generate(bin), salt)
+}
+
+/// Websocket address of the node.
+fn ws(node: &Node) -> String {
+    format!("ws://{}", node.address())
+}
```
DennisInSky, line 42: Do you want to move this into `tests/common/mod.rs` and use it in the `NodeExec` impl as well as somewhere else

DennisInSky: Maybe `program_id` also can be moved into the common place for possible re-use throughout the tests

gsdk/Cargo.toml
shamilsan: There is a short `dep.workspace = true` entry in the Rust book: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#inheriting-a-dependency-from-a-workspace

But I vote too for the consistent style across the whole manifest.

DennisInSky: I would go with more unified approach. Moreover, my Better TOML does not like the shortend version :)

shamilsan: Let's use `<dep>.workspace = true` everywhere.

PR #2644, opened 2023-05-11
feat(gstd): Implement ReservationManager in gstd

### LINKED ISSUE #1585: Implement `UnspentReservationPool` in `gstd` to ease gas reservations management
### Problem to Solve

After #1505, `gcore` and `gstd` users will have an opportunity to reserve gas from the message on their programs. The downside of the implementation is that programs have to count saved amount of gas along with reservations ids, because programs have no possibility to know, how much did they reserved. 

### Possible Solution

The idea is to make some kind of reservations manager, which abstracts the crate's user from reservation Ids and such. The simplest version of the suggesting entity just reserves the gas and unreserves it with a `amount: u64` argument instead of reservation id. Under the hood the type must effectively free existing reservations. By effectiveness I mean it should cause less read/writes in node's storage.

mertwole:
Resolves #1585


examples/reservation-manager/src/lib.rs
```diff
@@ -0,0 +1,66 @@
+#![no_std]
+
+use gstd::{msg, prelude::*, ReservationManager};
+
+static mut RESERVATION_MANAGER: Option<ReservationManager> = None;
+
+#[no_mangle]
+extern "C" fn handle() {
+    let new_msg = String::from_utf8(msg::load_bytes().expect("Failed to load payload bytes"))
```
StackOverflowExcept1on, line 9: We recently found encoding issue. Please replace this code with the following: https://github.com/gear-dapps/ping/commit/4001429180e0aabe2888ecdd22f8d82a564f0edc
See https://github.com/gear-tech/gear-js/issues/1280

NikVolf:
Better just call it `Reservations` in gstd

gstd/src/common/reservation_manager.rs
```diff
@@ -0,0 +1,155 @@
...
+        let valid_until = duration.saturating_add(block_height);
+
+        let insert_to = self.reservations.partition_point(|reservation| {
+            reservation.amount < amount && reservation.valid_until < valid_until
+        });
+
+        // self.reservations will allocate new memory.
+        if self.reservations.capacity() == self.reservations.len() {
+            self.cleanup(block_height);
+        }
```
SabaunT, line 107: @mertwole @grishasobol can we be sure that allocating a x2 memory is more expensive than calling host function (`exec::block_height`) + "shrinking" value owned by `self.reservations`?

gstd/src/common/reservation_manager.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// equal to `amount`
+    ///
+    /// If such a reservation is found, [`ReservationId`](ReservationId) and
+    /// reserved amount are returned
+    ///
+    /// # See also
+    /// - [`ReservationId`](ReservationId) is used to reserve and unreserve gas
+    ///   amount
+    /// for program execution later.
+    pub fn try_take_reservation(&mut self, amount: u64) -> Option<(ReservationId, u64)> {
```
SabaunT, line 131: This funtion must return `Result`, with `Err(less_than_amount)` if there's no value `>= amount`

gstd/src/common/reservation_manager.rs
```diff
@@ -0,0 +1,155 @@
...
+    /// Reserve the `amount` of gas for further usage.
+    ///
+    /// `duration` is the block count within which the reserve must be used.
+    pub fn reserve(&mut self, amount: u64, duration: u32) -> Result {
+        let id = ReservationId::reserve(amount, duration)?;
+        let block_height = exec::block_height();
+        let valid_until = duration.saturating_add(block_height);
+
+        let insert_to = self.reservations.partition_point(|reservation| {
+            reservation.amount < amount && reservation.valid_until < valid_until
```
SabaunT, line 101: what's the property of the layout of reservations? I guess we can't say that `reserve` stores reservations in the sorted order.

SabaunT (changes requested):
Please make the underlying structure of the data optimal as much as possible for the insertions and deletions.

gstd/src/common/reservations.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// Create a new [`Reservations`] struct
+    pub fn new() -> Reservations {
+        Reservations {
+            reservations: vec![],
+        }
+    }
+
+    /// Reserve the `amount` of gas for further usage.
+    ///
+    /// `duration` is the block count within which the reserve must be used.
```
SabaunT, line 118: please add some info on the details of the underlying algorithm and it's effectiveness (it's O(logN) as far as I understand).

gstd/src/common/reservations.rs
```diff
@@ -0,0 +1,215 @@
...
+///         }
+///     }
+/// }
+/// ```
+///
+/// # See also
+/// - [`ReservationId`](ReservationId) is used to reserve and unreserve gas
+/// for program execution later.
+#[derive(Default)]
+pub struct Reservations {
```
SabaunT, line 104: the functionality must be thoroughly tested. It includes unit-tests and property tests.

SabaunT (changes requested):
Everything is going nice, keep it going!

gstd/src/common/reservations.rs
```diff
@@ -0,0 +1,215 @@
...
+use crate::{errors::Result, exec, prelude::*};
+
+struct Reservation {
+    id: ReservationId,
+    amount: u64,
+    valid_until: u32,
+}
+
+/// Result of [`Reservations::try_take_reservation`](Reservations::try_take_reservation)
+pub enum TakeReservationResult {
```
breathx, line 29: Can't we just go with `-> Option<Reservation>` for api methods?
And reservation method would has get fns + api to send message with it

gstd/src/common/reservations.rs
```diff
@@ -0,0 +1,215 @@
...
+/// }
+/// ```
+///
+/// # See also
+/// - [`ReservationId`](ReservationId) is used to reserve and unreserve gas
+/// for program execution later.
+#[derive(Default)]
+pub struct Reservations {
+    reservations: Vec<Reservation>,
+}
```
breathx, line 106: Seems more neat for me
```suggestion
pub struct Reservations(Vec<Reservation>);
```

gstd/src/common/reservations.rs
```diff
@@ -0,0 +1,215 @@
...
+
+impl Reservations {
+    /// Create a new [`Reservations`] struct
+    pub fn new() -> Reservations {
+        Reservations {
+            reservations: vec![],
+        }
+    }
```
breathx, line 114:
```suggestion
    pub const fn new() -> Self {
        Self(Vec::new())
    }
```
This will allow us to use directly:
```rust
static mut RS: Reservations = Reservations::new();
```

SabaunT (approved):

examples/binaries/reservation-manager/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "demo-reservation-manager"
+version = "0.1.0"
+authors.workspace = true
+edition = "2021"
+license = "GPL-3.0"
+workspace = "../../../"
+
+[dependencies]
+codec = { package = "parity-scale-codec", version = "3.4.0", default-features = false, features = ["derive"] }
```
breathx, line 10: Use from workspace

examples/binaries/reservation-manager/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "demo-reservation-manager"
+version = "0.1.0"
+authors.workspace = true
+edition = "2021"
```
breathx, line 5: •Workspace

examples/binaries/reservation-manager/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+codec = { package = "parity-scale-codec", version = "3.4.0", default-features = false, features = ["derive"] }
+gstd.workspace = true
+
+[build-dependencies]
+gear-wasm-builder.workspace = true
+
+[dev-dependencies]
+gtest.workspace = true
+
+[lib]
```
breathx, line 19:
```suggestion
```

gclient/Cargo.toml
```diff
@@ -47,5 +47,6 @@ demo-program-factory.workspace = true
 demo-proxy = { workspace = true, features = ["std"] }
 demo-proxy-relay.workspace = true
 demo-proxy-with-gas.workspace = true
+demo-reservation-manager.workspace = true
 demo-reserve-gas.workspace = true
 gstd = { workspace = true, features = ["debug"] }
```
breathx, line 52: Newline at the end of file

gclient/tests/reservation_manager.rs
```diff
@@ -0,0 +1,140 @@
+// This file is part of Gear.
```
breathx, line 1: Why do we need gclient test for that? Pallet one would be enough

gstd/src/common/reservations.rs
```diff
@@ -0,0 +1,328 @@
...
+    ///
+    /// Also, when the underlying `Vec` will allocate new memory
+    /// the attempt to clean expired reservations occurs to avoid memory
+    /// allocations.
+    pub fn reserve(&mut self, amount: u64, duration: u32) -> Result<()> {
+        let id = ReservationId::reserve(amount, duration)?;
+        let block_height = exec::block_height();
+        let valid_until = duration.saturating_add(block_height);
+
+        let insert_range_start = self
```
breathx, line 126: Why not just use Btreeset?

gstd/src/common/reservations.rs
```diff
@@ -0,0 +1,328 @@
...
+        } else {
+            match insert_to {
+                Ok(pos) => pos + 1,
+                Err(pos) => pos,
+            }
+        };
+
+        // self.0 will allocate new memory.
+        if self.0.capacity() == self.0.len() {
+            self.cleanup(block_height);
```
breathx, line 146: Any cleanup should occur only in-place when you query some reservation and looking for it using bsearch or manually

breathx: Or by schedule: e.g. each 500 blocks on first interaction with manager

gstd/src/common/reservations.rs
```diff
@@ -0,0 +1,328 @@
...
+    /// All the reservations are sorted by their amount and then by the time
+    /// when they'll expire (both are in ascending order) in underlying `Vec`,
+    /// so when one's trying to take reservation, reservation with the least
+    /// possible amount is found and if it's already expired, the search to the
+    /// end of underlying `Vec` occurs. After that, all the expired
+    /// reservations that were found in process of search are cleaned out.
+    ///
+    /// # See also
+    /// - [`ReservationId`] is used to reserve and unreserve gas amount
+    /// for program execution later.
```
breathx, line 181: Seems this fn doesn’t pay attention on diff of expiration for two +- similar reservations

gstd/src/common/reservations.rs
```diff
@@ -0,0 +1,328 @@
...
+        self.0.drain(search_from..);
+
+        None
+    }
+
+    /// Returns an amount of the stored reservations
+    /// that aren't expired at this time.
+    ///
+    /// Executes for O(N) where N is amount of stored reservations.
+    pub fn count_valid(&self) -> usize {
```
breathx, line 213: Do we really need this?

gstd/src/common/reservations.rs
breathx: Still I’d also return bn here or even Reservation structure itself

examples/binaries/reservation-manager/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+
+#[derive(Debug, Encode, Decode)]
+pub enum SendMessageResult {
+    ReservationFound,
+    ReservationNotFound,
+}
+
+static mut RESERVATIONS: Reservations = Reservations::new();
+
+#[no_mangle]
```
breathx, line 46: Introduce mod wasm under cfg not std

examples/binaries/reservation-manager/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+                        msg::send_bytes_from_reservation(
+                            id,
+                            msg::source(),
+                            SendMessageResult::ReservationFound.encode(),
+                            0,
+                        )
+                        .expect("Failed to send message from reservation");
+                    }
+                    None => {
+                        msg::send_bytes(
```
breathx, line 70: Shouldn’t we panic here for test purposes?

breathx (changes requested):

gstd/src/common/reservations.rs
SabaunT: well, current implementation actually performs on-query invalidation for the range of values in `unreserve`.

As invalidation done on the, imho, right level (during `unreserve`), I don't see any problems with including it in the `reserve`. What's more, it safes from unneeded allocations.

gstd/src/common/reservations.rs
SabaunT: I guess business logic can rely on valid amount of reservations. There is no way to count valid ones from the API user level.

gstd/src/common/reservations.rs
SabaunT: Using `BTreeSet` as an only storage for reservations is impossible:
1. it has incovenient API for updating (not inserting) values.
2. we query and insert types more complex than just a value.

However, we can do some research on optimising the solution. Currently I believe we have the simplest solution working close to O(logN).

SabaunT (commented):
@breathx I left some comments to your review.

gstd/src/common/reservations.rs
breathx: Ok, anyway to be restructured and introduces in wide trait approach with different strategies some time soon

breathx:
Fix please minor things left from review @mertwole

examples/binaries/reservation-manager/src/lib.rs
mertwole: But we may need to check that we aren't getting reservation if there's no suitable reservations

breathx: > But we may need to check that we aren't getting reservation if there's no suitable reservations

asserting_error with specific panic message

gstd/src/common/reservations.rs
```diff
@@ -0,0 +1,368 @@
...
+
+        Ok(Self {
+            id: ReservationId::reserve(amount, duration)?,
+            amount,
+            valid_until: duration.saturating_add(block_height),
+        })
+    }
+
+    /// Converts this [`Reservation`] into [`ReservationId`].
+    pub fn into_id(self) -> ReservationId {
```
breathx, line 60: impl ReservationId from Reservation && provide getter by ref instead of just consume + derive some primitives for Reservation

gstd/src/common/reservations.rs
```diff
@@ -0,0 +1,368 @@
...
+///             .expect("failed to reserve gas");
+///         RESERVATIONS
+///             .reserve(50_000, 30)
+///             .expect("failed to reserve gas");
+///     }
+/// }
+///
+/// extern "C" fn handle() {
+///     let reservation = unsafe { RESERVATIONS.try_take_reservation(100_000) };
+///     match reservation {
```
breathx, line 114: same matching

gstd/src/common/reservations.rs
```diff
@@ -0,0 +1,366 @@
...
+///             .reserve(100_000, 100)
+///             .expect("failed to reserve gas");
+///         RESERVATIONS
+///             .reserve(50_000, 30)
+///             .expect("failed to reserve gas");
+///     }
+/// }
+///
+/// extern "C" fn handle() {
+///     let reservation = unsafe { RESERVATIONS.try_take_reservation(100_000) };
```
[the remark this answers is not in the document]
breathx: add nomangle

pallets/gear/src/tests.rs
```diff
@@ -12128,3 +12128,91 @@ fn calculate_gas_fails_when_calculation_limit_exceeded() {
...
+                }
+                _ => false,
+            });
+
+            assert!(expected_message_received);
+
+            System::reset_events();
+        };
+
+        let assert_failed = move |action| {
```
breathx, line 12180: messing up naming, assert_failed already exists in the scope

breathx: I'd go with inroducing some `fn scenario()`, see module entries for such "pattern"

gstd/src/common/reservations.rs
```diff
@@ -0,0 +1,368 @@
...
+///         }
+///     }
+/// }
+/// ```
+///
+/// # See also
+/// - [`ReservationId`](ReservationId) is used to reserve and unreserve gas
+/// for program execution later.
+/// - [`Reservation`] stores some additional data along with `ReservationId`.
+#[derive(Default)]
```
breathx, line 135: here typical derives as well, please

breathx (approved):

PR #2647, opened 2023-05-11
feat(gear,vara): Add proxy and multisig pallets to both vara and gear runtimes

ekovalev:
Add two pallets to both runtimes:
- pallet-proxy
- pallet-multisig.

The `pallet-proxy` in `Gear` runtime lacks proxy types for governance, staking etc. On the other hand, it has the `SudoBalances` instance filter for the `ProxyType` there that doesn't make much sense in `Vara` (hence not included).

The runtime constants are still subject to review and can be changed at later time if necessary.

### DIFF
--- a/runtime/gear/src/lib.rs
+++ b/runtime/gear/src/lib.rs
@@ -26,11 +26,13 @@ include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
 
 use frame_support::weights::ConstantMultiplier;
 pub use frame_support::{
+    codec::{Decode, Encode, MaxEncodedLen},
     construct_runtime,
     dispatch::{DispatchClass, WeighData},
     parameter_types,
     traits::{
-        ConstU128, ConstU32, Contains, FindAuthor, KeyOwnerProofSystem, Randomness, StorageInfo,
+        ConstU128, ConstU32, Contains, FindAuthor, InstanceFilter, KeyOwnerProofSystem, Randomness,
+        StorageInfo,
     },
     weights::{
         constants::{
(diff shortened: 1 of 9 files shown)


runtime/gear/src/lib.rs
```diff
@@ -442,6 +545,8 @@ construct_runtime!(
         TransactionPayment: pallet_transaction_payment = 6,
         Session: pallet_session = 7,
         Utility: pallet_utility = 8,
+        Proxy: pallet_proxy = 9,
```
ukint-vs, line 548: pls set 10 12 so that it is possible to add the necessary pallets between them.

runtime/vara/src/constants.rs
```diff
@@ -29,8 +29,10 @@ pub mod currency {
     pub const CENTS: Balance = DOLLARS / 100; // 200_000_000_000
     pub const MILLICENTS: Balance = CENTS / 1_000; // 200_000_000
 
-    /// Function that defines runtime constants used in voting
+    /// Helper function to calculate various deposits for using pallets' storage
     pub const fn deposit(items: u32, bytes: u32) -> Balance {
+        // TODO: review numbers. Current values are defaults from Substrate. For reference,
```
ukint-vs, line 34: make a task not to forget about all those todos

ekovalev: Done

runtime/gear/src/lib.rs
ekovalev: As discussed offline, leaving the discriminants as is for the meantime as they don't seem to ever require reordering.

ukint-vs (approved):

PR #2627, opened 2023-05-05
refactor(backend-common): Avoid memory accesses on zero size reads

DennisInSky:
Partially resolves #2095.

- do not access memory on zero size reads

It is not quite clear what needs to be changed/optimised in [lazy-pages](https://github.com/gear-tech/gear/blob/master/lazy-pages/src/host_func.rs#L112) with regards to it. @grishasobol can you please clarify?

@reviewer-or-team

PR #2648, opened 2023-05-11
refactor(backend-common): Avoid memory accesses on zero size writes

### LINKED ISSUE #2095: Take in account zero size memory accesses in sys-calls.
### Problem to Solve

Currently we can get for example zero size payload in sys-calls. Even in this case this accesses will be handled as common access and host function to get memory will be called and also this access will be pre-processed in lazy-pages.
We need to make separate more fast handling for this accesses in sys-calls.

See also TODO in lazy-pages/src/lib.rs

DennisInSky:
Resolves #2095.

In addition to taking account of zero size reads (#2627), this PR takes care of zero size writes

@reviewer-or-team


SabaunT (approved):
nice

shamilsan (approved):
Flawless as always

PR #2658, opened 2023-05-15
test(ci): Run runtime upgrade check

ark0f:
Note that summary migrations weight exceeds `max_block` parameter

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -275,9 +275,13 @@ jobs:
       - name: "Test: Sys-calls Wasmi integrity"
         run: ./scripts/gear.sh test syscalls --release
         
+      - name: "Test: `try-runtime` feature tests"
+        run: |
+          cargo test --workspace --features try-runtime --release --locked
+
       - name: "Test: Try runtime migrations"
         run: |
-          cargo build -p gear-cli --release --features try-runtime
+          cargo build -p gear-cli --features try-runtime --release --locked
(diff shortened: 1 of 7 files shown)


.github/workflows/CI.yaml
```diff
@@ -274,6 +274,13 @@ jobs:
 
       - name: "Test: Sys-calls Wasmi integrity"
         run: ./scripts/gear.sh test syscalls --release
+        
+      - name: "Test: Try runtime migrations"
+        run: |
+          cargo build -p gear-cli --release --features try-runtime
+          ./target/release/gear try-runtime --dev --runtime ./target/release/wbuild/gear-runtime/gear_runtime.wasm on-runtime-upgrade --checks live --uri wss://rpc-node.gear-tech.io:443
```
breathx, line 281:
```suggestion
          ./target/release/gear try-runtime --dev --checks live --uri wss://rpc-node.gear-tech.io:443
```
This will result in using currently built runtime, isn't it?

breathx (approved):

breathx:
> Note that summary migrations weight exceeds `max_block` parameter

Why? Seems we have not so many data in staging net?

gshep (approved):
Don't we want to have `cargo test ... -F try-runtime`  checking?

.github/workflows/CI.yaml
ark0f: `--runtime existing`:
```
Error: Input("given runtime is NOT compiled with try-runtime feature!")
```

ark0f:
> > Note that summary migrations weight exceeds `max_block` parameter
> 
> Why? Seems we have not so many data in staging net?

Seems we have. I see weight increase by 0.45% after 2 days

PR #2661, opened 2023-05-17
chore(spec-test): Obsolete some spec-tests and examples [2]

breathx:


shamilsan (approved):
It's about time

ukint-vs (approved):

grishasobol (changes requested):
set this because of merge once green. I will review

examples/panicker/src/lib.rs
grishasobol, line 1: this test is a simple demonstration of how panic works. If some developer will make error which cause problems in panic handling, then it will be much more faster and simpler to identify problem using this test than in more complex, but less unit tests

PR #2667, opened 2023-05-17
feat(gsdk): conditional generating client API

### LINKED ISSUE #2618: Generate runtimes types for client API
### Problem to Solve

subxt-codgen generates `runtime types`, `calls`, ... in one file, but we actually only need runtime types in our implementation since our API is dynamic

### Possible Solution

- ?fork `subxt-codgen`, make it possible to only generate the code we need
- add a CI updating the types on clicking the workflow

### Notes

_No response_

clearloop:
Resolves #2618 

- [x] introduce `gsdk/api-gen`
  - [x] get metadata from runtime wasm
  - [x] format generated code
- [x] clean metadata

now for generating the client API

```
GSDK_API_GEN=true cargo build -p gsdk --release
```

@gear-tech/dev


shamilsan (approved):
👍

PR #2678, opened 2023-05-22
refactor(runtime-fuzzer): Make gas limit variable in runtime-fuzzer

### LINKED ISSUE #2328: Make gas limit in the fuzzer a randomly generated value
this is about `runtime-fuzzer`

mertwole:
Resolves #2328


SabaunT:
Actually Im thinking of closing this, becase what can be tested with a low gas value, is actually tested with a max gas value and too expensive program. This case is something common while fuzzer works, as I see from logs. What do you think?

@mertwole @breathx @grishasobol

SabaunT:
due to core-audit/#22 error I think different gas limit values is great for testing.

breathx (changes requested):
Such range will mostly run into small gas amount. I'd make it random, but with much more changes close to default gas limit.
and 100% left border should be > 0

SabaunT:
> 

Makes sense

breathx (approved):

PR #2687, opened 2023-05-28
feat!: Deny replies with delay

breathx:


shamilsan (approved):
Goodbye to delayed reply 😢

PR #2689, opened 2023-05-29
feat(demo): Introduce `demo-constructor`

breathx:


grishasobol (commented):
For more simple implementation I prefer to make like this: 
in init we send Request Enum which tells the contract in which mode it must work.
in handle contract works in this mod.

For example for ping:
1) init - saves in static variable that in handle we must work like Ping
2) handle - we send "PING", in handle contract match static variable to identify work mod and process "PING" as ping used to.

The same logic for others. Using this scheme we can send data in handle in free format.
Also this simplify test and reduce probability of problems because of side effects.

breathx:
> For more simple implementation I prefer to make like this: in init we send Request Enum which tells the contract in which mode it must work. in handle contract works in this mod.
> 
> For example for ping:
> 
> 1. init - saves in static variable that in handle we must work like Ping
> 2. handle - we send "PING", in handle contract match static variable to identify work mod and process "PING" as ping used to.
> 
> The same logic for others. Using this scheme we can send data in handle in free format. Also this simplify test and reduce probability of problems because of side effects.

For sure. Its just first imply. Further it will accept commands to run on each handle etc

examples/binaries/constructor/src/lib.rs
```diff
@@ -0,0 +1,197 @@
...
+        Arg::New(value)
+    }
+}
+
+#[derive(Clone, Debug, Decode, Encode)]
+pub enum Call {
+    Store(Key),
+    StoreVec(Key),
+    Source,
+    Send(Arg<[u8; 32]>, Arg<Vec<u8>>, Option<u64>, u128, u32),
```
shamilsan, line 59: Maybe better to use struct here?

breathx: Can't see reason for it.
It much more comfortable to just write as fn args. Furthermore, there is Calls builder now, that is preferred to be used

examples/binaries/constructor/src/arg.rs
```diff
@@ -0,0 +1,102 @@
+use alloc::{
+    string::{String, ToString},
+    vec::Vec,
+};
+use parity_scale_codec::{Codec, Decode, Encode};
+
+#[derive(Clone, Debug, Decode, Encode)]
+pub enum Arg<T: 'static + Clone + Codec> {
```
shamilsan, line 8: I guess some doc-comments would be appreciated.

examples/binaries/constructor/src/arg.rs
```diff
@@ -0,0 +1,102 @@
...
+
+impl From<[u8; 32]> for Arg<Vec<u8>> {
+    fn from(hash: [u8; 32]) -> Self {
+        Arg::New(hash.encode())
+    }
+}
+
+impl From<&'static str> for Arg<Vec<u8>> {
+    fn from(key: &'static str) -> Self {
+        Self::get(key)
```
shamilsan, line 57: Some conversions are to `Arg::New`, some to `Arg::Get`. Isn't it confusing? I'd like to see more use cases.

examples/binaries/constructor/src/builder.rs
```diff
@@ -0,0 +1,183 @@
...
+impl Calls {
+    pub fn builder() -> Self {
+        Default::default()
+    }
+
+    pub(crate) fn calls(self) -> Vec<Call> {
+        self.0
+    }
+
+    pub fn push(mut self, call: Call) -> Self {
```
shamilsan, line 37: `push` can be confused with `Vec::push` (as we impl Deref). I'd rename it. Maybe `add` or just `call`.

examples/binaries/constructor/src/builder.rs
```diff
@@ -0,0 +1,183 @@
...
+#[derive(Default, Debug, Clone)]
+pub struct Calls(Vec<Call>);
+
+impl From<Vec<Call>> for Calls {
+    fn from(calls: Vec<Call>) -> Self {
+        Self(calls)
+    }
+}
+
+impl Deref for Calls {
```
shamilsan, line 15: Maybe it is worth to impl `DerefMut` too, to use some `Vec`'s methods?

shamilsan (approved):
Good start 👍 
Waiting for more use cases and some in-code comments.

examples/binaries/constructor/src/builder.rs
```diff
@@ -0,0 +1,183 @@
...
+    pub(crate) fn calls(self) -> Vec<Call> {
+        self.0
+    }
+
+    pub fn push(mut self, call: Call) -> Self {
+        self.0.push(call);
+        self
+    }
+
+    pub fn push_many<const N: usize>(mut self, calls: [Call; N]) -> Self {
```
SabaunT, line 42:
```suggestion
    pub fn push_many<const N: usize>(mut self, calls: impl Iterator<Item = Call>) -> Self {
```

SabaunT (approved):
It's good as a first iteration for the constructor. I guess the set of desired changes will be clear later, when we remove simple demos one by one.

examples/binaries/constructor/src/arg.rs
breathx: Added minimalistic ones

examples/binaries/constructor/src/arg.rs
breathx: everything from &'static str leads to get call. You give exactly the key from storage.
Other impulse that lead to "new" is wrappers for empty array and hash array (32 bytes) to vec conversion. nothing else

examples/binaries/constructor/src/builder.rs
breathx: renamed

examples/binaries/constructor/src/builder.rs
breathx: Implemented deref only for encode_like trait bound. I don't want inner access here at all, but deref also allows to call for example .len, so its okay without mutable impl for me

examples/binaries/constructor/src/builder.rs
breathx: kept both
for me seems more neat using just arrays [val, val], instead of [val, val].iter()

examples/binaries/constructor/src/scheme/mod.rs
```diff
@@ -0,0 +1,44 @@
+use crate::{Call, Calls};
+use alloc::vec::Vec;
+use parity_scale_codec::{Decode, Encode};
+
+#[derive(Encode, Decode, Clone, Debug)]
+pub enum Scheme {
+    Direct(Vec<Call>),
+    Predefined(Vec<Call>, Vec<Call>, Vec<Call>),
```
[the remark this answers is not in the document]
breathx: Added docs here

examples/binaries/constructor/src/builder.rs
SabaunT: iterator is a more general approach, very likely to be in such api

PR #2690, opened 2023-05-29
feat: add assembler-script support in wasm-proc

grishasobol:
To support assembly-script in gear, make mod for wasm-proc where `_start` function calls in each export function and all mutable globals except stack_end are moved to programs data section.
This command will create `code.opt.wasm`, which can be successfully using as gear program.
```
./wasm-proc --assembly-script code.wasm
```

### DIFF
--- a/gtest/src/program.rs
+++ b/gtest/src/program.rs
@@ -275,7 +275,7 @@ impl<'a> Program<'a> {
             let mut optimizer = Optimizer::new(path).expect("Failed to create optimizer");
 
             // Ignore result, because it's not important.
-            let _ = optimizer.insert_stack_end_export();
+            _ = optimizer.insert_stack_end_export();
 
             optimizer.strip_custom_sections();
             optimizer
--- a/gtest/src/program.rs
+++ b/gtest/src/program.rs
@@ -273,7 +273,10 @@ impl<'a> Program<'a> {
 
         let opt_code = if !is_opt {
             let mut optimizer = Optimizer::new(path).expect("Failed to create optimizer");
-            optimizer.insert_stack_and_export();
+
+            // Ignore result, because it's not important.
+            let _ = optimizer.insert_stack_end_export();
+
             optimizer.strip_custom_sections();
             optimizer
                 .optimize(OptType::Opt)
--- a/utils/wasm-builder/src/stack_end.rs
+++ b/utils/wasm-builder/src/stack_end.rs
@@ -16,6 +16,7 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+use crate::optimize;
 use gear_wasm_instrument::STACK_END_EXPORT_NAME;
 use pwasm_utils::parity_wasm::{
     builder,
@@ -37,7 +38,7 @@ fn get_global_index(module_bytes: &[u8], name_predicate: impl Fn(&str) -> bool) 
             }
             _ => None,
         })
-        .flat_map(|sub_section| sub_section)
+        .flatten()
         .filter_map(|res| res.ok())
         .filter_map(|name| match name {
             Name::Global(m) => Some(m),
@@ -56,12 +57,14 @@ fn get_global_index(module_bytes: &[u8], name_predicate: impl Fn(&str) -> bool) 
 ///
 /// This export can be used in runtime to identify the end of stack memory
 /// and skip its uploading to the storage.
-pub fn insert_stack_end_export(
-    module_bytes: &[u8],
-    module: &mut Module,
-) -> Result<(), &'static str> {
+pub fn insert_stack_end_export(module: &mut Module) -> Result<(), &'static str> {
+    let module_bytes = module
+        .clone()
+        .to_bytes()
+        .map_err(|_| "cannot get code from module")?;
+
     let stack_pointer_index =
-        get_global_index(module_bytes, |name| name.ends_with("__stack_pointer"))
+        get_global_index(&module_bytes, |name| name.ends_with("__stack_pointer"))
             .ok_or("has no stack pointer global")?;
 
     let glob_section = module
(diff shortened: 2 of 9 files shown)


utils/wasm-proc/src/main.rs
```diff
@@ -99,11 +99,14 @@ const RT_ALLOWED_IMPORTS: [&str; 62] = [
...
 #[derive(Debug, clap::Parser)]
 struct Args {
-    /// Don't create gear stack end export
+    /// Insert gear stack end export, enabled by default.
+    #[arg(long, default_value = "true")]
+    insert_stack_end: bool,
+
     #[arg(long)]
-    skip_stack_end: bool,
+    as_script: bool,
```
gshep, line 107: `assembly_script` looks better

utils/wasm-builder/src/stack_end.rs
```diff
@@ -99,11 +81,323 @@ pub fn insert_stack_end_export(
...
+        global_data_offset += data.len() as u32;
+    }
+
+    // Insert new data section for globals initial values
+    own_module = builder::from_module(own_module)
+        .data()
+        .offset(Instruction::I32Const(data_end_offset as i32))
+        .value(new_data_in_section)
+        .build()
+        .build();
```
gshep, line 250: If I understand that correctly you place globals in lower part of the stack, i.e. by extending the last data segment. It will work for most cases however I didn't find update of the stack size itself.

utils/wasm-proc/src/main.rs
grishasobol: Ok done

ukint-vs (approved):

gtest/src/program.rs
```diff
@@ -273,7 +273,10 @@ impl<'a> Program<'a> {
 
         let opt_code = if !is_opt {
             let mut optimizer = Optimizer::new(path).expect("Failed to create optimizer");
-            optimizer.insert_stack_and_export();
+
+            // Ignore result, because it's not important.
+            let _ = optimizer.insert_stack_end_export();
```
shamilsan, line 278: Underscore can be without `let` but up to you.
```suggestion
            _ = optimizer.insert_stack_end_export();
```

utils/wasm-builder/src/stack_end.rs
```diff
@@ -99,11 +81,323 @@ pub fn insert_stack_end_export(
...
+        }
+    } else {
+        return Ok(());
+    };
+
+    for export_name in optimize::FUNC_EXPORTS {
+        let Some(export) = module
+            .export_section()
+            .ok_or("Cannot find export section")?
+            .entries()
```
shamilsan, line 110: Do we need to call this chain every time inside a loop?

utils/wasm-builder/src/stack_end.rs
```diff
@@ -99,11 +81,323 @@ pub fn insert_stack_end_export(
...
+    {
+        if !global.global_type().is_mutable() {
+            continue;
+        }
+        if index == data_end_index as usize {
+            continue;
+        }
+        if index == stack_pointer_index as usize {
+            continue;
+        }
```
shamilsan, line 171: Why not `if condition1 || condition2 || condition3 { continue }`.

utils/wasm-builder/src/stack_end.rs
```diff
@@ -99,11 +81,323 @@ pub fn insert_stack_end_export(
...
+            .bodies_mut()
+            .iter_mut()
+            .flat_map(|body| body.code_mut().elements_mut().iter_mut())
+        {
+            let global_index = u32::try_from(index).expect("Global index bigger than u32");
+            if *instr == Instruction::GetGlobal(global_index) {
+                *instr = Instruction::Call(get_global_function_index);
+            } else if *instr == Instruction::SetGlobal(global_index) {
+                *instr = Instruction::Call(set_global_function_index);
+            }
```
shamilsan, line 237: Why not expressions?

```suggestion
            *instr = if *instr == Instruction::GetGlobal(global_index) {
                Instruction::Call(get_global_function_index)
            } else if *instr == Instruction::SetGlobal(global_index) {
                Instruction::Call(set_global_function_index)
            }
```

utils/wasm-builder/src/stack_end.rs
```diff
@@ -99,11 +81,323 @@ pub fn insert_stack_end_export(
...
+                Some(NameSectionReader::new(r.data(), r.data_offset()))
+            }
+            _ => None,
+        })
+        .flatten()
+        .filter_map(|res| res.ok())
+        .filter_map(|name| match name {
+            Name::Global(m) => Some(m),
+            _ => None,
+        })
```
shamilsan, line 293: At least this two `filter_map`s could be replaced by one.
```suggestion
        .filter_map(|name| match name {
            Ok(Name::Global(m)) => Some(m),
            _ => None,
        })
```

shamilsan (commented):
Too fragile and not expandable approach (when we will add support for another languages, we are to change it again).

utils/wasm-builder/src/stack_end.rs
grishasobol: Yes, because `module` is used as mut and const inside loop. And this check is not expensive.

utils/wasm-builder/src/stack_end.rs
grishasobol: Because instr must be changed not in all cases

utils/wasm-builder/src/stack_end.rs
grishasobol: Done

gtest/src/program.rs
grishasobol: Done

utils/wasm-builder/src/stack_end.rs
grishasobol: DOne

PR #2698, opened 2023-05-30
feat(wasm-builder): generate `metahash` function in programs

### LINKED ISSUE #2360: Consider generating `metahash` function in programs automatically
`metahash` function is a boilerplate from a program to a program:

```rust
#[no_mangle]
extern "C" fn metahash() {
    let metahash: [u8; 32] = include!("../.metahash");
    msg::reply(metahash, 0).expect("Failed to share metahash");
}
```
Usage of `metahash()`:
- gear-dapps: https://github.com/search?q=org%3Agear-dapps+metahash%28%29&type=code
- whole github: https://github.com/search?q=%22extern+%5C%22C%5C%22+fn+metahash%28%29%22&type=code

Also, it depends on an artifact generated before the build (it is hard to read).

Consider autogenerating this function.

StackOverflowExcept1on:
Resolves #2360

### DIFF
--- a/utils/wasm-builder/src/wasm_project.rs
+++ b/utils/wasm-builder/src/wasm_project.rs
@@ -198,17 +198,12 @@ impl WasmProject {
 
         fs::write(self.manifest_path(), toml::to_string_pretty(&cargo_toml)?)?;
 
-        let src_dir = self.out_dir.join("src");
-        fs::create_dir_all(&src_dir)?;
-        fs::write(
-            src_dir.join("lib.rs"),
-            "#![no_std] pub use orig_project::*;",
-        )?;
-
         // Copy original `Cargo.lock` if any
         let from_lock = self.original_dir.join("Cargo.lock");
         let to_lock = self.out_dir.join("Cargo.lock");
         let _ = fs::copy(from_lock, to_lock);
+
+        let mut source_code = "#![no_std] pub use orig_project::*;\n".to_owned();
 
         // Write metadata
         if let Some(metadata) = &self.project_type.metadata() {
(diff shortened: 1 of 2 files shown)


utils/wasm-builder/src/wasm_project.rs
```diff
@@ -220,15 +215,71 @@ impl WasmProject {
...
+            payload: *const BufferStart,
+            len: Length,
+            value: *const Value,
+            _delay: BlockNumber,
+            err_mid: *mut LengthWithHash,
+        );
+    }}
+}}
+
+use fake_gsys::*;
```
shamilsan, line 258: I don't know whether it can yield some namespace conflicts with `orig_project::*`, but I'd use it explicitly: `fake_gsys::gr_reply`, `fake_gsys::LengthWithHash`, etc.

shamilsan (approved):
Looks great to me 👍

utils/wasm-builder/src/wasm_project.rs
```diff
@@ -220,15 +215,69 @@ impl WasmProject {
...
-            let wasm_meta_hash_path = self.original_dir.join(".metahash");
 
             smart_fs::write_metadata(wasm_meta_path, metadata)
                 .context("unable to write `*.meta.txt`")?;
 
-            smart_fs::write(wasm_meta_hash_path, format!("{:?}", metadata.hash()))
-                .context("unable to write `.metahash`")?;
+            source_code = format!(
+                r#"{source_code}
+#[allow(improper_ctypes)]
```
breathx, line 224: the only thing I consider here is to simplify this mod. can't see the reason for so many lines of code on purpose

breathx (approved):

utils/wasm-builder/src/wasm_project.rs
shamilsan: We can avoid type aliases and some consts indeed.

breathx: repr(c) type may be also skipped because we don't assert result further. just [u8; 36] const array filled with zeroes to keep it on stack and throw ptr safely

PR #2699, opened 2023-05-30
feat(gstd): Introduce async exec::sleep_for method

### LINKED ISSUE #349: Time-limited variants of locking primitives

DennisInSky:
Resolves #349.

@reviewer-or-team


gstd/src/exec/mod.rs
```diff
@@ -0,0 +1,33 @@
...
+//!
+//! Wraps methods from [`gcore::exec`](https://docs.gear.rs/gcore/exec/)
+//! for receiving details about the current execution and controlling it.
+
+pub use basic::*;
+pub use gcore::exec::{
+    block_height, block_timestamp, gas_available, leave, random, system_reserve_gas,
+    value_available, wait, wait_for, wait_up_to,
+};
+pub use r#async::*;
```
shamilsan, line 30: Frankly speaking, I'm not too fond of such kind of names 😄 
I'd rename it to something less "keywordish". Maybe `asynchronous` or `async_io`.

gstd/src/exec/async.rs
```diff
@@ -0,0 +1,69 @@
...
+};
+use gcore::exec;
+
+/// Delays message execution in asynchronous way for the specified number of
+/// blocks. It works pretty much like the `exec::wait_for` function, but allows
+/// to continue execution after the delay in the same handler. It is worth
+/// mentioning that the program state gets persisted inside the call, and the
+/// execution resumes with potentially different state.
+pub fn delay_for(block_count: u32) -> impl Future<Output = ()> {
+    MessageDelayFuture::new(msg::id(), exec::block_height().saturating_add(block_count))
```
shamilsan, line 33: Should we check whether `block_count` is zero and do nothing?

shamilsan (approved):
Looks good to me 👍

gstd/src/exec/mod.rs
shamilsan: Agree as I see this module is not public, so let's keep it as is.

gstd/src/exec/async.rs
```diff
@@ -0,0 +1,69 @@
...
+            async_runtime::locks().lock(
+                self.msg_id,
+                self.msg_id,
+                async_runtime::Lock::exactly(self.till_block_number - current_block_number)
+                    .expect("Unreachable: never fails with block count > 0"),
+            );
+            Poll::Pending
+        } else {
+            Poll::Ready(())
+        }
```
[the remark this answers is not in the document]
shamilsan: Comparing `current_block_number < self.till_block_number` looks more robust to me, as it works when some blocks are skipped.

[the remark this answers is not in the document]
shamilsan: Ah, sorry, misunderstood, your code is similar. But I don't see the difference except that `block_count` may be equal to zero in your case.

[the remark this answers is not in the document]
DennisInSky: We need the `self.till_block_number - current_block_number` to be exactly > 0 otherwise the `Lock::exactly` returns an error. That's why it is safe to write `.expect("Unreachable: never fails with block count > 0"),`

gstd/src/exec/async.rs
shamilsan: Of course this is up to you 😄

DennisInSky:
> imo the original issue is asking for `error/return` logic for waiting messages which requires checking the message status, so we may not have to introduce new messages, besides, `delay_for` + `wakeup` is similar with `up_to` but only have checks in the contract end
> 
> hmm, as I remember we may have error handling logic for these but didn't resolve them as `timeout`

Ah... I see what you mean. To be honest, I did not read the issue, I just took @breathx word, and this is how he described it. After reading it, I see that the idea was to have an extra param for our set of `xxx_for_reply` functions allowing wait for it no longer than specified amount of blocks which is different to what is implemented.
Apparently, I just have to remove `resolves #349` and leave this PR as implementing the `delay_for` functionality. The issue itself has to addressed separately. @breathx

grishasobol (approved):

NikVolf:
@DennisInSky resolve conflicts?

gstd/src/exec/mod.rs
```diff
@@ -0,0 +1,33 @@
+// This file is part of Gear.
+
+// Copyright (C) Gear Technologies Inc.
```
breathx, line 3: copyright year everywhere

gstd/src/exec/basic.rs
```diff
@@ -86,7 +76,7 @@ pub fn exit(inheritor_id: ActorId) -> ! {
 
 /// Resume previously paused message handling.
 ///
-/// Suppose a message has been paused using the [`wait`] function,
+/// Suppose a message has been paused using the [`gcore::exec::wait`] function,
```
breathx, line 79: ref to gcore doc in gstd seems weird
```suggestion
/// Suppose a message has been paused using the [`crate::exec::wait`] function,
```

gstd/src/exec/async.rs
```diff
@@ -0,0 +1,69 @@
...
+
+use crate::{async_runtime, msg, MessageId};
+use core::{
+    future::Future,
+    pin::Pin,
+    task::{Context, Poll},
+};
+use gcore::exec;
+
+/// Delays message execution in asynchronous way for the specified number of
```
breathx, line 27: Great docs, thank you!

gstd/src/exec/async.rs
```diff
@@ -0,0 +1,69 @@
...
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
```
breathx, line 18: mod docs

gstd/src/exec/async.rs
```diff
@@ -0,0 +1,69 @@
...
+    }
+}
+
+impl Future for MessageDelayFuture {
+    type Output = ();
+
+    fn poll(self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let current_block_number = exec::block_height();
+
+        async_runtime::locks().remove(self.msg_id, self.msg_id);
```
breathx, line 56: what if we introduce for locks special lock id? 
I suppose that this will work incorrect for following case (please provide test for it):
*here goes some pseudo code*
```
let long = delay_for(10);
let short = delay_for(8);
join(long,short).await
```

await will wake us up in 8, isn't it or even end up with panic?
just because if we poll similar lock ids it will rewrite each other?

gstd/src/exec/async.rs
```diff
@@ -0,0 +1,69 @@
...
+    task::{Context, Poll},
+};
+use gcore::exec;
+
+/// Delays message execution in asynchronous way for the specified number of
+/// blocks. It works pretty much like the `exec::wait_for` function, but allows
+/// to continue execution after the delay in the same handler. It is worth
+/// mentioning that the program state gets persisted inside the call, and the
+/// execution resumes with potentially different state.
+pub fn delay_for(block_count: u32) -> impl Future<Output = ()> {
```
breathx, line 32: what about just `sleep`?

breathx: otherwise could be messed up with delayed sendings

pallets/gear/src/tests.rs
```diff
@@ -8018,6 +8018,108 @@ fn call_forbidden_function() {
     });
 }
 
+#[test]
+fn async_delay_for() {
```
breathx, line 8022: tests to be reviewed after fixes

breathx (changes requested):

shamilsan:
@DennisInSky please merge the master branch.
@breathx please re-review.

pallets/gear/src/tests.rs
```diff
@@ -8537,6 +8538,270 @@ fn call_forbidden_function() {
...
+            EMPTY_PAYLOAD.to_vec(),
+            BlockGasLimitOf::<Test>::get(),
+            0,
+        )
+        .expect("Failed to upload Waiter");
+        let waiter_prog_id = get_last_program_id();
+        run_to_next_block(None);
+
+        // Helper functions (collapse the block)
+        let (send_command_to_waiter, assert_waiter_single_reply) = {
```
breathx, line 8569: why tuple here?

gstd/src/exec/async.rs
```diff
@@ -0,0 +1,66 @@
...
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//! Module for asynchronous execution of messages.
```
breathx, line 19: It's more about async execution control?

breathx (approved):

PR #2712, opened 2023-06-01
fix(weights): Fix `SessionValidatorSetMigration` weight

### LINKED ISSUE #2696: Migrations exceed block weight
### Problem to Solve

Current migrations make up 127% of block weight. Possibly next runtime upgrade will fail

### Possible Solution

Something like migration pallet with steps in `on_initialize`

### Notes

_No response_

ark0f:
Resolves #2696

### DIFF
--- a/runtime/common/src/migration.rs
+++ b/runtime/common/src/migration.rs
@@ -1,0 +1,50 @@
+// This file is part of Gear.
+//
+// Copyright (C) 2023 Gear Technologies Inc.
+// SPDX-License-Identifier: GPL-3.0-or-later WITH Classpath-exception-2.0
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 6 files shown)


ark0f:
Now weight is just 38.75%

runtime/common/src/migration.rs
```diff
@@ -0,0 +1,63 @@
...
+            1 // read pallet session validators
+                + 1 // read validator set
+                + 1, // read approved validator set
+        );
+
+        let session_validators = pallet_session::Pallet::<T>::validators();
+        let validator_set = validator_set::Validators::<T>::get();
+        let approved_validator_set = validator_set::ApprovedValidators::<T>::get();
+
+        if session_validators == validator_set && session_validators == approved_validator_set {
```
ukint-vs, line 47: Can you add checks that the current runtime version is under v140?

ukint-vs (approved):
Nice.

runtime/common/src/migration.rs
ark0f: We don't need it because condition will be true for `spec_version >= 140`

ukint-vs: It is also be true after v140. We can't run this twice.

PR #2716, opened 2023-06-02
perf(wasm-builder): Improve wasm-builder performance

DennisInSky:
Improves performance of building wasms by the means of:
1. Re-running build script instead of re-building it
2. Re-building artifacts like optimised wasm, wasm_binary, toml, etc. only when there are actual changes

@reviewer-or-team


utils/wasm-builder/src/smart_fs.rs
```diff
@@ -22,10 +22,34 @@
 
 use anyhow::Result;
 use gmeta::MetadataRepr;
-use std::{fs, path::Path};
+use std::{fs, io::ErrorKind, path::Path};
 
 const LINEAR_COMPARISON_FILE_SIZE: u64 = 4096;
 
+pub fn copy_if_newer(from: impl AsRef<Path>, to: impl AsRef<Path>) -> Result<bool> {
```
shamilsan, line 29: Should this function be `pub` or `pub(crate)`?

shamilsan (approved):

utils/wasm-builder/src/smart_fs.rs
breathx, line 29: shouldn't it be just `smart_fs::copy` ?

utils/wasm-builder/src/wasm_project.rs
```diff
@@ -354,10 +356,20 @@ pub const WASM_EXPORTS: &[&str] = &{:?};
             .context("unable to write `wasm_binary.rs`")?;
         }
 
-        Ok(())
+        self.force_rerun_on_next_run(&from_path)
```
breathx, line 359: much more common pattern
```suggestion
        let _ = self.force_rerun_on_next_run(&from_path)?;
        
        Ok(())
```

breathx (approved):

utils/wasm-builder/src/wasm_project.rs
DennisInSky: Maybe, it is a more common pattern, but definitely not for clippy. It does not like it ;)

shamilsan: You might remove `let _`, but I vote for your code as it is shorter.

```rust
// Just in case for clippy. But expression-based code is preferable to me.
self.force_rerun_on_next_run(&from_path)?;
Ok(())
```

PR #2719, opened 2023-06-05
fix(wasm-builder): detect incorrect code at early stage

### LINKED ISSUE #2703: Using floating point type in smart contracts should be detected at early stage
### Problem

Check the minimal app with the `f32` type: https://github.com/gear-dapps/app/tree/sy-f32

It compiles and passes internal tests with `gtest` successfully (`make test`) but fails on `gclient`-based tests (`make full-test`) with a not too informative error message `Internal error: upload_program failed with '<wasm:stripped>`.

### Steps

1. `git clone https://github.com/gear-dapps/app.git && cd app && git checkout sy-f32`
2. `GEAR_NODE_PATH=/usr/local/bin/gear make test` -> OK
3. `GEAR_NODE_PATH=/usr/local/bin/gear make full-test` -> FAILED

### Possible Solution

The best solution is to catch the floating point type in `gear-wasm-builder` and `wasm-proc`. Also, we are to detect it in `gtest` as a second frontier.

StackOverflowExcept1on:
Resolves #2703

@gear-tech/dev

### DIFF
--- a/utils/wasm-instrument/src/rules.rs
+++ b/utils/wasm-instrument/src/rules.rs
@@ -139,11 +139,9 @@ impl Rules for CustomConstantCostRules {
             | I64Rotl
             | I32Rotr
             | I64Rotr
-            | SignExt(_) => (),
-            _ => return None,
-        };
-
-        Some(self.instruction_cost)
+            | SignExt(_) => Some(self.instruction_cost),
+            _ => None,
+        }
     }
 
     fn memory_grow_cost(&self) -> MemoryGrowCost {
--- a/utils/wasm-instrument/src/rules.rs
+++ b/utils/wasm-instrument/src/rules.rs
@@ -1,0 +1,156 @@
+use core::num::NonZeroU32;
+use wasm_instrument::{
+    gas_metering::{MemoryGrowCost, Rules},
+    parity_wasm::elements::{self, Instruction},
+};
+
+pub struct CustomConstantCostRules {
+    instruction_cost: u32,
+    memory_grow_cost: u32,
+    call_per_local_cost: u32,
+}
+
+impl CustomConstantCostRules {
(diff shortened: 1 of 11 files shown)


grishasobol (approved):

StackOverflowExcept1on:
I think this can be improved because unsupported instructions like `f32.*` and `f64.*` are just a special case of this problem. For example, the user may forget to add `init` or `handle` function. 
```rust
#[derive(Debug, PartialEq, Eq, derive_more::Display)]
pub enum CodeError {
    /// The provided code doesn't contain the required `init` or `handle` export function.
    #[display(fmt = "Required export function `init` or `handle` not found")]
    RequiredExportFnNotFound,
}
```
Now this check does not catch a potential error.

utils/wasm-instrument/src/rules.rs
```diff
@@ -0,0 +1,154 @@
+use core::num::NonZeroU32;
+use wasm_instrument::{
+    gas_metering::{MemoryGrowCost, Rules},
+    parity_wasm::elements::{self, Instruction},
+};
+
+pub struct CustomConstantCostRules {
```
gshep, line 7: this looks like a code duplication however the rules need to be in sync with the `Schedule`. Although it would require a bit of `Schedule` refactoring it doesn't worth it IMO.

So at the very least the unit test in `Schedule` crate/module could be written that both `instruction_cost` impls return `Some`/None respectively for the specific instruction(s).

StackOverflowExcept1on:
CI failed due to binary called `demo-out-of-memory`. Seems rustc over-optimized it and removed `memory (import "env" "memory")`

utils/wasm-instrument/src/rules.rs
StackOverflowExcept1on: @gshep should be fine now

shamilsan (approved):
Excellent 🔥

PR #2720, opened 2023-06-05
feat(pallet): Provision flag for external node on gas pallet

breathx:
First of two steps to introduce reply provision


common/src/gas_provider/mod.rs
```diff
@@ -190,9 +190,19 @@ pub trait Tree {
...
+    fn create_provision(
+        key: impl Into<Self::NodeId>,
+        new_key: impl Into<Self::NodeId>,
+        amount: Self::Balance,
+    ) -> Result<(), Self::Error>;
+
     /// Return bool, defining does node exist.
     fn exists(key: impl Into<Self::NodeId>) -> bool;
 
+    /// Return bool, defining does node exist and is external.
```
SabaunT, line 203:
```suggestion
    /// Returns bool, defining does node exist and is external with provision.
```

SabaunT (changes requested):

SabaunT (approved):

common/src/gas_provider/internal.rs
```diff
@@ -744,10 +744,40 @@ where
         )
     }
 
+    fn create_provision(
```
ark0f, line 747:
```suggestion
    fn create_supply(
```
Why don't use standard words?

pallets/gas/src/migrations/v2.rs
```diff
@@ -0,0 +1,231 @@
...
+                };
+
+                unhashed::put(&key[..], &node);
+            });
+
+            // run migration from v1 to v2.
+            let weight = MigrateToV2::<Test>::on_runtime_upgrade();
+            assert_ne!(weight.ref_time(), 0);
+
+            for (_key, node) in GasNodes::<Test>::iter() {
```
ark0f, line 224: Also add assertion there are as many nodes as message ids

common/src/gas_provider/mod.rs
ark0f, line 1: Add prop tests for new functions

ark0f (changes requested):

common/src/gas_provider/internal.rs
breathx: discussed. name would be create_deposit because gas tree has no exact chain with messages and replies etc

ark0f (approved):

PR #2721, opened 2023-06-05
refactor(core-backend): move syscall implementations to common

### LINKED ISSUE #2067: core-backend refactoring
### File Location(s)

core-backend/sandbox
core-backend/wasmi

### Proposal

Currently we have two core-backend implementations for sp-sandbox and wasmi. Generally big parts of code from these  implementations can be reused in both:
1) Sys calls internal implementation
2) errors
3) runtimes (partly)

mertwole:
Partially resolves #2067


SabaunT (changes requested):
Nice job!
Please refer to comments and issues found.

core-backend/common/src/funcs.rs
```diff
@@ -0,0 +1,659 @@
...
+            }
+            Err(err) => {
+                log::trace!("Alloc failed: {err}");
+                u32::MAX
+            }
+        };
+        Ok(page)
+    }
+
+    pub fn free(ctx: &mut R, page_no: u32) -> Result<i32, R::Error> {
```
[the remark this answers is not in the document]
mertwole: the `page` is calculated before `run_any` call and the error is propagated. But this kind of behavior couldn't be achieved through this macro

SabaunT (approved):
lgtm, but I'd wait for some more reviews.

core-backend/common/src/funcs.rs
```diff
@@ -0,0 +1,659 @@
...
+        syscall_trace!("free", page_no);
+
+        let page = WasmPage::new(page_no).map_err(|_| R::unreachable_error_code())?;
```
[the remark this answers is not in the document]
SabaunT: I guess we must move it inside the `ctx.run_any`, because we receive an input, which fails validation. That's a common error situation. Otherwise we will have `termination_reason` be `Success` with `HostError`/`Trap` as a result of calling the wasm entry point

core-backend/codegen/src/lib.rs
ark0f, line 1: Consider write tests for macro as we do in gstd

core-backend/common/src/funcs.rs
```diff
@@ -0,0 +1,660 @@
...
+
+impl<E, R> FuncsHandler<E, R>
+where
+    E: BackendExt + 'static,
+    E::Error: BackendExtError,
+    E::AllocError: BackendAllocExtError<ExtError = E::Error>,
+    R: Runtime<E>,
+{
+    /// !!! Usage warning: make sure to do it before any other read/write,
+    /// because it may contain register read.
```
ark0f, line 57:
```suggestion
    /// because it may contain registered read.
```

core-backend/common/src/funcs.rs
```diff
@@ -0,0 +1,660 @@
...
+            ctx.ext_mut()
+                .reply_commit(ReplyPacket::new(Default::default(), value))
+        };
+
+        f().map_err(Into::into)
+    }
+
+    #[host(fallible, cost = RuntimeCosts::ReplyPushInput, err_len = LengthBytes)]
+    pub fn reply_push_input(ctx: &mut R, offset: u32, len: u32) -> Result<(), R::Error> {
+        // Charge for `len` inside `reply_push_input`
```
ark0f, line 401:
```suggestion
        // Charge for `len` is inside `reply_push_input`
```

core-backend/common/src/runtime.rs
```diff
@@ -0,0 +1,59 @@
...
+    memory::{MemoryAccessError, MemoryAccessRecorder, MemoryOwner, WasmMemoryWrite},
+    BackendExt, BackendState, TerminationReason,
+};
+use gear_core::{costs::RuntimeCosts, gas::GasLeft, memory::WasmPage};
+use gear_core_errors::ExtError;
+
+pub trait Runtime<E: BackendExt>: MemoryOwner + MemoryAccessRecorder + BackendState {
+    type Error;
+
+    fn unreachable_error_code() -> Self::Error;
```
ark0f, line 31:
```suggestion
    fn unreachable_error() -> Self::Error;
```

ark0f (changes requested):

core-backend/common/src/funcs.rs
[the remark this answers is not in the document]
grishasobol: The correct behaviour is return `1` from `free` in this case. So, we must move it inside closure and match it with `res`

core-backend/codegen/src/lib.rs
mertwole: But the macros work is fully tested with pallet tests as they use many of the syscalls

ark0f (approved):

breathx (approved):

PR #2724, opened 2023-06-05
feat(tests, cli): CLI tool for executing blocks against the live chain state

ekovalev:
A custom tool to replay a block against the live chain state (at some height, or the latest finalized block).
Essentially, a simplification of the `gear try-runtime` command that hides away the complexity of the latter and also allows to:
- not enable the `try-runtime` feature during node/runtime build;
- use the native runtime to apply a block, that is run extrinsics in a debugger, if needed.

See the `README.md` for more details.


NikVolf:
seems useful

@ukint-vs can you try this and review?

ekovalev:
> Im getting `ERROR runtime: panicked at 'Bad input data provided to execute_block: Codec error', runtime/gear/src/lib.rs:651:1` every time when I try to execute block with some `pallet_gear` tx in local dev network.

The reason for errors may have been found: since we are using the `Core` `runtime_api` (specifically, `Executive::execute_block(block)`), the final checks are also run by this runtime api implementation (the `executive` pallet). In case we don't have extrinsics in a block these checks go through, but if we run extrinsics, the state resulting root doesn't match the one downloaded from the live chain.

The only option I can think of now is to enable the `try-runtime` feature and use the `try_execute_block` runtime api call instead (which allows to skip the state root checks). This would mean we will have to only use the native runtime (because the downloaded one from the live chain will obviously not have the `try-runtime` feature enabled).

breathx:
> The reason for errors may have been found: since we are using the `Core` `runtime_api` (specifically, `Executive::execute_block(block)`), the final checks are also run by this runtime api implementation (the `executive` pallet). In case we don't have extrinsics in a block these checks go through, but if we run extrinsics, the resulting state root doesn't match the one downloaded from the live chain.
> 
> The only option I can think of now is to enable the `try-runtime` feature and use the `try_execute_block` runtime api call instead (which allows to skip the state root checks). This would mean we will have to only use the native runtime (because the downloaded one from the live chain will obviously not have the `try-runtime` feature enabled).

Special RPC node with the feature wouldn't be the solution?

utils/remote-ext-tests/replay-block/src/main.rs
```diff
@@ -0,0 +1,200 @@
...
+    #[arg(short, long, num_args = 1..)]
+    pallet: Vec<String>,
+
+    /// Fetch the child-keys as well.
+    ///
+    /// Default is `false`, if specific `--pallets` are specified, `true` otherwise. In other
+    /// words, if you scrape the whole state the child tree data is included out of the box.
+    /// Otherwise, it must be enabled explicitly using this flag.
+    #[arg(long)]
+    child_tree: bool,
```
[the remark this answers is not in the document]
ekovalev: Done.

utils/remote-ext-tests/replay-block/README.md
```diff
@@ -0,0 +1,159 @@
...
+        ```bash
+        gear try-runtime --chain=vara --runtime vara_runtime.compact.compressed.wasm execute-block live --uri wss://archive-rpc.vara-network.io:443
+        ```
+
+    * at block `$HASH`
+
+        ```bash
+        export HASH=0x8dc1e32576c1ad4e28dc141769576efdbc19d0170d427b69edb2261cfc36e905
+
+        gear try-runtime --chain=vara --runtime vara_runtime.compact.compressed.wasm execute-block live --uri wss://archive-rpc.vara-network.io:443 --at "$HASH"
```
breathx, line 50: can't we use block number here?

utils/remote-ext-tests/replay-block/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+
+    /// Sets a custom logging filter. Syntax is `<target>=<level>`, e.g. -lsync=debug.
+    ///
+    /// Log levels (least to most verbose) are error, warn, info, debug, and trace.
+    /// By default, all targets log `info`. The global log level can be set with `-l<level>`.
+    #[arg(short = 'l', long, value_name = "LOG_PATTERN", num_args = 1..)]
```
breathx, line 94:
```suggestion
    #[arg(short = 'l', long, value_name = "NODE_LOG", num_args = 0..)]
```

breathx: Same as RUST_LOG

utils/remote-ext-tests/replay-block/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+                "Block is not provided, setting it to the latest finalized head: {:?}",
+                height
+            );
+
+            BlockHashOrNumber::Hash(height)
+        }
+    };
+
+    let (current_number, current_hash) = match block_hash_or_num {
+        BlockHashOrNumber::Number(n) => (n, block_number_to_hash::<Block>(&rpc, n).await?),
```
breathx, line 145: I see that number is possible so state it in the docs please

utils/remote-ext-tests/replay-block/src/main.rs
```diff
@@ -0,0 +1,221 @@
...
+use sp_runtime::{
+    generic::SignedBlock,
+    traits::{Block as BlockT, Header as HeaderT, One},
+};
+use sp_state_machine::ExecutionStrategy;
+use std::fmt::Debug;
+use substrate_rpc_client::{ws_client, ChainApi};
+use util::*;
+
+pub(crate) const LOG_TARGET: &str = "remote-ext::cli";
```
breathx, line 40: too complicated name for target for me, it's not as comfort to use as could be with for example `gear-replay`
And I'd then consider crate renaming into something similar and not such general just because it supports concrete functionality of us as separate and the only one project

utils/remote-ext-tests/replay-block/src/parse.rs
```diff
@@ -0,0 +1,67 @@
...
+    if let Some(pos) = block_hash.chars().position(|c| !c.is_ascii_hexdigit()) {
+        Err(offset + pos)
+    } else {
+        Ok(block_hash.into())
+    }
+}
+
+pub(crate) fn url(s: &str) -> Result<String, &'static str> {
+    if s.starts_with("ws://") || s.starts_with("wss://") {
+        // could use Url crate as well, but lets keep it simple for now.
```
breathx, line 39: Open an issue maybe and ref it here

utils/remote-ext-tests/replay-block/src/parse.rs
```diff
@@ -0,0 +1,67 @@
...
+
+pub(crate) fn block(block_hash_or_number: &str) -> Result<BlockHashOrNumber<Block>, String> {
+    if let Ok(block_number) = block_hash_or_number.parse::<NumberFor<Block>>() {
+        Ok(BlockHashOrNumber::Number(block_number))
+    } else {
+        let block_hash = hash(block_hash_or_number).map_err(|e| {
+            format!(
+                "Expected block hash or number, found illegal hex character at position {}",
+                e,
+            )
```
breathx, line 54:
```suggestion
            format!("Expected block hash or number, found illegal hex character at position {e}")
```
Everywhere where possible please

utils/remote-ext-tests/replay-block/src/util.rs
```diff
@@ -0,0 +1,236 @@
...
+            instantiation_strategy: WasmtimeInstantiationStrategy::RecreateInstanceCopyOnWrite,
+        },
+        heap_pages,
+        max_runtime_instances,
+        None,
+        runtime_cache_size,
+    )
+}
+
+#[allow(dead_code)]
```
breathx, line 87: deadcode?

breathx (approved):
Awesome

utils/remote-ext-tests/replay-block/src/parse.rs
ekovalev: It's a leftover from the original `try-runtime`. I better remove the comment completely.

utils/remote-ext-tests/replay-block/README.md
ekovalev: this refers to the original `try-runtime` CLI. In our tool we do support both the hash and block number

utils/remote-ext-tests/replay-block/src/main.rs
ekovalev: The `README.md` has the full description. It'a just the first part refers to the in-built `try-runtime` node subcommand (and how it can be used for the similar purpose).

utils/remote-ext-tests/replay-block/src/util.rs
ekovalev: It's not used anymore, kept it in case we need something like this in the future (parse a string literal into a block hash). However, it's quite trivial anyway - I think it can be safely removed for now

utils/remote-ext-tests/replay-block/src/main.rs
ekovalev: Why `num_args = 0..`? Do we want to allow a dangling `-l` flag with no target?

utils/remote-ext-tests/replay-block/src/parse.rs
breathx: > It's a leftover from the original `try-runtime`. I better remove this comment completely.

lol

utils/remote-ext-tests/replay-block/src/main.rs
breathx: > Why `num_args = 0..`? Do we want to allow a dangling `-l` flag with no target?

for example if we will throw some env var like `-l LOG_ENV` so why not

breathx:
Waiting for last minor fixes @ekovalev and review from @ukint-vs

ukint-vs (approved):

breathx:
@ekovalev merge if ready

PR #2730, opened 2023-06-07
chore: use Rust toolchain override

shamilsan:
Better to use the standard Rust toolchain override (e.g. `rust-tolchain.toml`) instead of the manual toolchain pinning.

- Added `rust-tolchain.toml` with `nightly-2023-04-25` version.
- Removed CI steps with nightly toolchain pinning.


ukint-vs (approved):
🔥

PR #2740, opened 2023-06-11
Properly handle status code

breathx:
This PR is preparation for different status codes used for auto-replies.
Now all assertions for success of status code's reply will be done by first byte


core/src/message/common.rs
```diff
@@ -179,7 +179,8 @@ pub enum MessageDetails {
 impl MessageDetails {
     /// Returns bool defining if message is error reply.
     pub fn is_error_reply(&self) -> bool {
-        self.is_reply_details() && self.status_code() != 0
+        // TODO: issue #2739.
```
ark0f, line 182: Maybe this issue must be solved first?

breathx: The thing is that within the issue we need to introduce the type in some shared space for both core and user libs (to be considered to avoid unnecessary deps, but atm I think that it may be some new crate for core-primitives maybe). Furthermore, it will require update of pallet entities like pallet::Event (that will require big changes from js side and mostly sure that no, but maybe our own migrations).
So it's better do such work that require attention a bit later and implement this ugly stuff now to work with status codes correctly

ark0f (approved):

Issue #1829, opened 2022-11-12
Redesign `gstd`


ark0f:
### Problem to Solve

`gstd` seems to need redesign with functions like `msg::send_delayed_from_reservation_and_wait_for_reply`

### Possible Solution

.

shamilsan:
We could introduce many message types (all combinations):

```rust
struct MessageValue {
    payload: ...,
    value: ...,
}

struct MessageGasValue {
    message_value: MessageValue,
    gas: ...,
}

struct MessageGasValueDest {
    message_gas_value: MessageGasValue,
    dest: ...,
}
```

Also, we need to implement all combinations of methods to transit from one type to another.

```rust
impl MessageValue {
    pub fn gas(self, limit: u64) -> MessageGasValue {
       ...
    }
}
```

Finally, we are to implement send and reply functions for every type.

```rust
impl MessageGasValueDestDelayReservation {
    pub fn send(self) -> ... {
        msg::send_delayed_with_gas_from_reservation(...)
    }
}
```

We don't need to write all these types and implementations manually but use macros or code generation instead.

So sending a message yields a chain of function calls:

```rust
msg::payload("Hello!")
    .to(dest_id)
    .with_gas(1_000_000)
    .delayed(100)
    .from_reservation(rsv_id)
    .send()
    .expect("Unable to send");
```

### Pros

- Zero cost abstraction

### Cons

- Not so easy to implement
- Ugly source code and docs

FrodoBaggins74524:
Use TryFrom -> https://doc.rust-lang.org/stable/std/convert/trait.TryFrom.html

StackOverflowExcept1on:
The solution is located here: https://github.com/gear-foundation/gstd-fluent

PR #2746, opened 2023-06-11
feat!(gstd): Add reply deposit for each async *_for_reply fn

breathx:
TODO:
- [x] Add test after #2745


shamilsan (approved):
@StackOverflowExcept1on Please consider introduced `reply_deposit` while dealing with #1829.

NikVolf:
needs merge

PR #2747, opened 2023-06-11
feat: restrict mutable globals to be export in gear

grishasobol:
restrict mutable globals to be export in gear on instrumentation. Exception is GEAR_STACK_END but only temporary.  In nearest feature GEAR_STACK_END will point to self const export.

### DIFF
--- a/core/src/code.rs
+++ b/core/src/code.rs
@@ -23,12 +23,11 @@ use crate::{
     memory::{PageU32Size, WasmPage},
     message::{DispatchKind, WasmEntry},
 };
-use alloc::{collections::BTreeSet, vec::Vec};
-use core::ops::ControlFlow;
+use alloc::{collections::BTreeSet, vec, vec::Vec};
 use gear_wasm_instrument::{
     parity_wasm::{
         self,
-        elements::{Instruction, Internal, Module},
+        elements::{ExportEntry, GlobalEntry, GlobalType, InitExpr, Instruction, Internal, Module},
     },
     wasm_instrument::{
         self,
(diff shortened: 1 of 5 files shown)


pallets/gear-program/src/mock.rs
```diff
@@ -40,8 +40,11 @@ type AccountId = u64;
 type BlockNumber = u64;
 type Balance = u128;
 
+#[allow(unused)]
```
breathx, line 43: Annoys me too. Better to `#![allow(unused)]` whole mock mod, because it's only test tooling

grishasobol: Done

NikVolf (approved):
As CI seems fine, Rust does not produce public mutable globals in the binaries?

core/src/code.rs
```diff
@@ -202,6 +208,44 @@ pub struct Code {
     instruction_weights_version: u32,
 }
 
+fn check_for_mut_global_exports(module: &Module) -> Result<(), CodeError> {
+    let global_exports_indexes = module
```
NikVolf, line 212: please no, don't put it in the protocol 

just make it const in instrumentantion or whatever

NikVolf (changes requested):
shouldn't put magic global names in the protocol

grishasobol:
> As CI seems fine, Rust does not produce public mutable globals in the binaries?
Yes

core/src/code.rs
grishasobol: Do you mean make __gear_stack_end export to const global?

grishasobol:
> shouldn't put magic global names in the protocol

This const is already in protocol since 2022 February =)

NikVolf (approved):
other than that make this injected stack pointer non-mutable

core/src/code.rs
grishasobol: Done

ukint-vs (approved):

gshep (approved):
💯

PR #2751, opened 2023-06-12
remove meta txt files

grishasobol:
This files are necessary for some tests, but this tests also require `gcli` or `gear` binaries which must be pre-built, so we can use the same logic for *.meta.txt which appears during demo built.
`make pre-commit` works well


examples/binaries/fungible-token/demo_fungible_token.meta.txt
shamilsan, line 1: Should we add them to `.gitignore` so they do not reappear?

shamilsan (approved):

breathx (approved):

PR #2754, opened 2023-06-13
refactor(gtest): Check the types for read_state_using_wasm args

### LINKED ISSUE #2702: gtest: Improve reading state using Wasm
### Problem to Solve

We have two functions for reading the program's state using some custom Wasm: [`Program::read_state_bytes_using_wasm`](https://docs.gear.rs/gtest/struct.Program.html#method.read_state_bytes_using_wasm) and [`Program::read_state_using_wasm`](https://docs.gear.rs/gtest/struct.Program.html#method.read_state_using_wasm). We have some inconvenience with them:

- State conversion functions signature is not verified, so we can mix up arguments and get no error, just inconsistent results.
- It is not quite convenient to pass multiple arguments:
    ```rust
    prog.read_state_bytes_using_wasm("my_state_func", WASM, Some((arg1, arg2, arg3).encode()));
    ```

### Possible Solution

Maybe we could add some macro for passing arguments that forms a tuple from arguments, encode it, and wraps it to `Some`:

```rust
prog.read_state_bytes_using_wasm("my_state_convert", WASM, state_args!(arg1, arg2, arg3));
```

Also, this macro could extract the function signature from Wasm and compare it to the arguments provided, but I don't know whether it is feasible. We may need to find a more appropriate solution.

### Notes

_No response_

mertwole:
Resolves #2702


gtest/src/program.rs
```diff
@@ -172,6 +173,19 @@ impl From<&str> for ProgramIdWrapper {
     }
 }
 
+#[macro_export]
+macro_rules! read_state_wasm_args {
```
shamilsan, line 177: I'd shorten this macro name. `state_args` is ok as for me.

examples/binaries/new-meta/tests/read_state.rs
```diff
@@ -61,6 +65,32 @@ fn read_state_bytes_with_parameterized_wasm_func_returns_transformed_state() {
     assert_eq!(expected_state, actual_state);
 }
 
+#[test]
+fn read_state_bytes_with_two_args_wasm_func_returns_transformed_state() {
+    let system = System::new();
+    let program = initialize_current_program(&system);
+    const WALLET_BY_NAME_AND_SURNAME_FUNC_NAME: &str = "wallet_by_name_and_surname";
```
shamilsan, line 72: Too long name for local const.
```suggestion
    const FUNC_NAME: &str = "wallet_by_name_and_surname";
```

gtest/src/manager.rs
```diff
@@ -475,8 +478,62 @@ impl ExtManager {
...
+                None,
+                None,
+                None,
+                vec![],
+                u64::MAX,
+                self.block_info,
+            )
+            .map_err(TestError::ReadMetadataError)?;
+        let metadata =
+            MetawasmData::decode(&mut &*metadata).expect("fn metadata() returned incorrect data");
```
shamilsan, line 494: Why so complex? `&*metadata`

gtest/src/manager.rs
```diff
@@ -475,8 +478,62 @@ impl ExtManager {
...
+
+        let func = metadata
+            .funcs
+            .get(fn_name)
+            .ok_or(TestError::ReadMetadataError(format!(
+                "{fn_name} not found in metadata"
+            )))?;
+        let mut types_registry = PortableRegistry::decode(&mut &*metadata.registry)
```
shamilsan, line 502: Also, why `&*metadata.registry`?

gtest/src/manager.rs
```diff
@@ -475,8 +478,62 @@ impl ExtManager {
...
+            .expect("fn metadata() returned incorrect data");
+        let mapping = types_registry.retain(|id| Some(id) == func.input);
+        let func_input = func.input.map(|inp| {
+            types_registry.resolve(*mapping.get(&inp).expect("Entry not found in mapping"))
+        });
+
+        fn new_metatype<T: 'static + TypeInfo>(_: &T) -> MetaType {
```
shamilsan, line 509: What is this one-line function for?

gtest/src/program.rs
```diff
@@ -434,15 +448,43 @@ impl<'a> Program<'a> {
...
-    pub fn read_state_bytes_using_wasm(
+    ///
+    /// # Usage
+    /// You can pass arguments as `Option<(arg1, arg2, ...)>` or by using
+    /// [`read_state_wasm_args`] macro.
+    ///
+    /// # Examples
+    ///
+    /// Read state bytes with no arguments passed to wasm.
+    /// ```ignore
```
shamilsan, line 459: `ignore`d examples have a risk of being stale. I'd replace several examples with one compilable example.

gtest/src/manager.rs
```diff
@@ -475,8 +478,62 @@ impl ExtManager {
             .into_parts()
             .0;
 
-        // The `metawasm` macro knows how to decode this as a tuple
-        let mut mapping_code_payload = argument.unwrap_or_default();
+        let metadata =
```
shamilsan, line 481: It would be great to add some comments for these steps of getting function signatures.

gtest/src/manager.rs
mertwole: `decode` takes `&mut Input` as parameter and `Input` is only implemented for `&[u8]` so we need to get `&mut &[u8]` from `Vec<u8>`

shamilsan: I'd use [`as_mut_slice`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.as_mut_slice) instead of deref + ref.

mertwole: `as_slice` you mean?

shamilsan (approved):
Looks great to me 🔥

gtest/src/manager.rs
```diff
@@ -461,12 +464,12 @@ impl ExtManager {
...
     }
 
-    pub(crate) fn read_state_bytes_using_wasm(
+    pub(crate) fn read_state_bytes_using_wasm<E: Encode + TypeInfo + 'static>(
         &mut self,
         program_id: &ProgramId,
         fn_name: &str,
         wasm: Vec<u8>,
-        argument: Option<Vec<u8>>,
+        args: Option<E>,
```
breathx, line 472: this fn is not codec-compatible, avoid it here or introduce another

gtest/src/program.rs
```diff
@@ -172,6 +173,19 @@ impl From<&str> for ProgramIdWrapper {
     }
 }
 
+#[macro_export]
+macro_rules! state_args {
+    () => {
+        Option::<()>::None
```
breathx, line 179: disagree with unit type here. as I think you did it to satisfy ur own : Encode bound, but it's incorrect, go with `Option::<Vec<u8>>::None` instead

gtest/src/manager.rs
```diff
@@ -475,8 +478,62 @@ impl ExtManager {
             .into_parts()
             .0;
 
-        // The `metawasm` macro knows how to decode this as a tuple
-        let mut mapping_code_payload = argument.unwrap_or_default();
+        // Call fn metadata() that's generated by `metawasm` macro and get function
```
breathx, line 481: can't see any reason to validate args, especially this way
please, remove it and that error aswell

breathx (changes requested):

gtest/src/manager.rs
mertwole: But `E: Encode` or what do you mean?

gtest/src/program.rs
mertwole: but `state_args!` constructs `Option<(arg1, arg2, ...)> where arg1: Encode, arg2: Encode, ...` so the `()` is here for consistency

gtest/src/manager.rs
mertwole: #2702 
image

breathx: > #2702
> image

Yeah, I see. Anyway I disagree with the initial task formulae then. Not your bad

gtest/src/program.rs
```diff
@@ -441,15 +516,61 @@ impl<'a> Program<'a> {
...
     pub fn read_state_using_wasm<E: Encode, D: Decode>(
         &self,
         fn_name: &str,
         wasm: Vec<u8>,
         argument: Option<E>,
     ) -> Result<D> {
-        let argument_bytes = argument.map(|arg| arg.encode());
-        let state_bytes = self.read_state_bytes_using_wasm(fn_name, wasm, argument_bytes)?;
-        D::decode(&mut state_bytes.as_ref()).map_err(Into::into)
+        let res = self.manager.borrow_mut().read_state_bytes_using_wasm(
```
breathx, line 567: why not to proxy to self.non_encodable as in was before?

examples/new-meta/tests/read_state.rs
```diff
@@ -38,18 +38,18 @@ fn read_state_bytes_with_wasm_func_returns_transformed_state() {
...
         name: "OtherName".into(),
     };
 
     let actual_state = program
         .read_state_bytes_using_wasm(
-            WALLET_BY_PERSON_FUNC_NAME,
+            FUNC_NAME,
             META_WASM_V2.to_vec(),
-            Some(other_person.encode()),
+            state_args_encoded!(other_person.clone()),
```
breathx, line 52:
```suggestion
            state_args_encoded!(&other_person),
```

breathx (approved):

PR #2755, opened 2023-06-13
refactor(gclient): avoid async calls inside into_iter() loop

ukint-vs:
Since awaiting inside into_iter() loop can't be `Send` we can fetch mailbox messages before and filter.

@gear-tech/dev

### DIFF
--- a/gclient/src/api/calls.rs
+++ b/gclient/src/api/calls.rs
@@ -533,14 +533,14 @@ impl GearApi {
         &self,
         args: impl IntoIterator<Item = MessageId> + Clone,
     ) -> Result<(Vec<Result<u128>>, H256)> {
-        let message_ids: Vec<MessageId> = args.clone().into_iter().collect();
+        let message_ids: Vec<_> = args.clone().into_iter().collect();
 
         let messages = futures::future::try_join_all(
             message_ids.iter().map(|mid| self.get_mailbox_message(*mid)),
         )
         .await?;
 
-        let mut values: BTreeMap<MessageId, u128> = messages
+        let mut values: BTreeMap<_, _> = messages
             .into_iter()
             .flatten()
             .map(|(msg, _interval)| (msg.id(), msg.value()))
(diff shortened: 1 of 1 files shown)


gclient/src/api/calls.rs
```diff
@@ -753,15 +754,17 @@ impl GearApi {
...
+            args.clone().into_iter().map(|(mid, _, _, _)| mid).collect();
+        let mailbox_messages = self
+            .get_mailbox_messages(100)
+            .await?
+            .into_iter()
+            .map(|(message, _interval)| (message.id(), message.value()));
+        let mut values: BTreeMap<MessageId, Option<u128>> = mailbox_messages
+            .filter(|(mid, _value)| message_ids.contains(mid))
+            .map(|(mid, value)| (mid, Some(value)))
+            .collect();
```
breathx, line 767:
```suggestion
        let message_ids: Vec<_> =
            args.clone().into_iter().map(|(mid, _, _, _)| mid).collect();
        let mailbox_messages = self
            .get_mailbox_messages(message_ids.len()) // maybe? or x5 something like that
            .await?
            .into_iter()
            .map(|(message, _interval)| (message.id(), message.value()));
        let mut values: BTreeMap<_, _> = mailbox_messages
            .filter(|(mid, _value)| message_ids.contains(mid))
            .collect();
```

breathx (changes requested):
This take of "all" messages from mailbox isn't optimal and may cause errors often. I'd go with query one-by-one anyway

gclient/src/api/calls.rs
```diff
@@ -533,15 +533,18 @@ impl GearApi {
...
-        let mut values = BTreeMap::new();
-        for message_id in args.clone().into_iter() {
-            values.insert(
-                message_id,
-                self.get_mailbox_message(message_id)
-                    .await?
-                    .map(|(message, _interval)| message.value()),
-            );
-        }
+        let message_ids: Vec<MessageId> = args.clone().into_iter().collect();
```
breathx, line 536: for all changes auto-guessing types still valid 
```suggestion
        let message_ids: Vec<_> = args.clone().into_iter().collect();
//      or
//      let message_ids = args.clone().into_iter().collect::<Vec<_>>();
```

breathx (approved):
Awesome!

SabaunT (approved):

PR #2757, opened 2023-06-13
refactor(gsdk): Make sign_and_submit_then_watch `Send`

ukint-vs:
Awaiting result inside of async recursion can't be `Send`.

Think about how we can match the error outside of recursion.

@gear-tech/dev

### DIFF
--- a/gsdk/src/signer/calls.rs
+++ b/gsdk/src/signer/calls.rs
@@ -375,7 +375,7 @@ impl Signer {
     }
 
     /// Wrapper for submit and watch with nonce.
-    async fn sign_and_submit<'a>(
+    async fn sign_and_submit_then_watch<'a>(
         &self,
         tx: &DynamicTxPayload<'a>,
     ) -> Result<TxProgressT, SubxtError> {
@@ -398,7 +398,7 @@ impl Signer {
         use subxt::tx::TxStatus::*;
 
         let before = self.balance().await?;
-        let mut process = self.sign_and_submit(&tx).await?;
+        let mut process = self.sign_and_submit_then_watch(&tx).await?;
 
         // Get extrinsic details.
         let (pallet, name) = (tx.pallet_name(), tx.call_name());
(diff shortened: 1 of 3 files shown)


gsdk/src/signer/calls.rs
```diff
@@ -55,7 +55,8 @@ use subxt::{
 type TxProgressT = TxProgress<GearConfig, OnlineClient<GearConfig>>;
 type EventsResult = Result<ExtrinsicEvents<GearConfig>, Error>;
 
-const ERRORS_REQUIRE_RETRYING: [&str; 2] = ["Connection reset by peer", "Connection refused"];
+// TODO: Somehow match error outside of async recursion in `sign_and_submit_then_watch`
```
breathx, line 58: For me this issue should be solved within the PR.
Consider avoid async recursion at all by introducing some loop maybe, or by throwing waker for async runtime and introducing custom poll result?

breathx (changes requested):

gsdk/src/signer/calls.rs
ukint-vs: We need to completely remove retries because this is a userspace issue. You need to establish a connection to the rpc node before submitting tx.

ukint-vs: CI works

gsdk/src/signer/calls.rs
```diff
@@ -377,14 +374,12 @@ impl Signer {
         }
     }
 
-    /// Wrapper for submit and watch with error handling.
-    #[async_recursion(?Send)]
-    async fn sign_and_submit_then_watch<'a>(
+    /// Wrapper for submit and watch with nonce.
+    async fn sign_and_submit<'a>(
```
breathx, line 378: why without "watch" now?

breathx (approved):

gsdk/src/signer/calls.rs
ukint-vs: missed subxt naming.

shamilsan (approved):
I like it!
➕3️⃣ ➖2️⃣6️⃣

PR #2758, opened 2023-06-13
fix(core-backend): continue to adopt `LimitedStr`

### LINKED ISSUE #2098: use `TrimmedString` for global names in `GlobalsCtx`
### File Location(s)

[core-backend/common/src/lazy_pages.rs](https://github.com/gear-tech/gear/pull/1988/files/55e7748b7428483d5594328b4ad0295e2b312d5f#diff-37fb51e19dec925a94d0609cc0723eabd2b3d8d7759aaad92e883ec2b786e63a)

### Proposal

__IMPORTANT__: need to implement `TryFrom` for `TrimmedString`, using `From` is RESTRICTED!!!!

tltsutltsu:
Resolves #2098

The `LimitedStr<'a>` type is now used to replace `TrimmedString` and to access global variables. It provides the following API:
```rust
/// This method returns `Result<LimitedStr, &'static str>` and checks the maximum length of the input string
LimitedStr::try_from("&'a str");

/// This method returns `LimitedStr` and trims the string to the last char
LimitedStr::from(String::new());

/// This method is used to create `LimitedStr` from `&str`
LimitedStr::from_small_str("global name");
```

@gear-tech/dev


runtime-interface/src/lib.rs
```diff
@@ -72,8 +73,7 @@ impl PassBy for LazyPagesProgramContext {
 #[codec(crate = codec)]
 pub struct LazyPagesRuntimeContext {
     pub page_sizes: Vec<u32>,
-    // TODO: considering change global name types to `TrimmedString` (issue #2098)
```
breathx, line 75: Need to figure out does this changes invalidates runtime interfaces
(check that LimitedStr encodes like String)

breathx (changes requested):
Seems ok, the only I care about is compatibility

breathx:
You can ask @grishasobol for help in such research

core-backend/common/src/utils.rs
```diff
@@ -87,25 +58,55 @@ fn smart_truncate(s: &mut String, max_bytes: usize) {
...
-    const INIT_ERROR_MSG: &'static str = concat!(
+    const INIT_ERROR_MSG: &str = concat!(
         "String must be less than ",
         stringify!(TRIMMED_MAX_LEN),
         " bytes."
     );
 
-    pub fn new(s: &'a str) -> Result<Self, &'static str> {
+    #[track_caller]
+    pub const fn from_small_str(s: &'a str) -> Self {
```
grishasobol, line 81: If we really wanna use panic here, then use unwrap_or_else, as it is common way. Also, why do we need to specify str lifetime ?

Actually I'm not in glad with using panic here. Better to return an error, and make unwrap_or_else in places where we really wanna panic in wrong case. And you have already try_from

grishasobol (changes requested):

core-backend/common/src/utils.rs
StackOverflowExcept1on: We cannot use `try_from().expect("msg")` in constant context. The main point of `LimitedStr::from_small_str` is optimization for global names.
```rust
/// zero cost
const GLOBAL_NAMES: &[LimitedStr<'static>] = &[
    LimitedStr::from_small_str(GLOBAL_NAME_GAS),
];

/// `try_from` isn't const fn, `except(...)` isn't stabilized yet in const fns
const GLOBAL: LimitedStr<'static> = LimitedStr::try_from(GLOBAL_NAME_GAS).except("failed")
```

grishasobol: Ok agree

grishasobol (commented):
I wanna to investigate this case more deeply, so will not remove `request changes` currently. Please wait 1-2 days

core-backend/common/src/utils.rs
```diff
@@ -87,25 +58,55 @@ fn smart_truncate(s: &mut String, max_bytes: usize) {
...
+/// Wrapped string to fit `core-backend::TRIMMED_MAX_LEN` amount of bytes.
+///
+/// The `Cow` is used to avoid allocating a new `String` when the `LimitedStr` is
+/// created from a `&str`.
+///
+/// Plain `str` is not used because it can't be properly encoded/decoded via scale codec.
+#[derive(
+    TypeInfo, Encode, Decode, Debug, Clone, derive_more::Display, PartialEq, Eq, PartialOrd, Ord,
+)]
+#[codec(crate = scale)]
```
grishasobol, line 70: Could you explain me: why do we need this #[codec(crate = scale)] ?

grishasobol (approved):

core-backend/common/src/utils.rs
breathx: > Could you explain me: why do we need this #[codec(crate = scale)] ?

It allows don't depend from parity_scale_codec, but at the same time use their derive macros from `pub extern crate codec` of other libs.
That caused by internal implementation of these derive macros.

core-backend/common/src/utils.rs
```diff
@@ -87,25 +58,55 @@ fn smart_truncate(s: &mut String, max_bytes: usize) {
...
+        Self(Cow::Borrowed(s))
+    }
+
+    pub fn as_str(&self) -> &str {
+        self.0.as_ref()
+    }
+}
+
+impl<'a> TryFrom<&'a str> for LimitedStr<'a> {
+    type Error = &'static str;
```
ark0f, line 95: I suggest
```suggestion
    type Error = LimitedStrTryFromError;
```

because `&'static str` is not convenient to match and work with in general

core-backend/common/src/utils.rs
grishasobol: But.. as I can see scale is parity scale codec:
https://github.com/paritytech/scale-info/blob/cc97f4c6496acf0f92d9caa45856331f094a7f38/Cargo.toml#L23

So, why we need to specify this, if it's by default

core-backend/common/src/utils.rs
```diff
@@ -87,25 +58,58 @@ fn smart_truncate(s: &mut String, max_bytes: usize) {
     }
 }
 
-#[derive(Debug, Copy, Clone, derive_more::Display)]
-pub struct LimitedStr<'a>(&'a str);
+/// Wrapped string to fit `core-backend::TRIMMED_MAX_LEN` amount of bytes.
```
breathx, line 61:
```suggestion
/// Wrapped string to fit `core_backend::TRIMMED_MAX_LEN` amount of bytes.
```

breathx (approved):
Fetch master please

PR #2760, opened 2023-06-13
feat(node): Implement RPC method returning WASM runtime version

DennisInSky:
The implemented RPC method returns WASM runtime version which includes the commit hash it was built from.
It has exactly the same format as the one returned by the `system_version` RPC method - `0.1.4-e2dcdb474e7`
The method supports an optional `at` param allowing to obtain the WASM runtime version at specified block number.
I do not know whether we want to return native runtime version as a part of this call, but in fact it is exactly the same value as the one returned by the `system_version` RPC method.

@gear-tech/dev


DennisInSky:
> returned by the `system_version` RPC method - `0.1.4-e2dcdb474e7`

We want to be like this
returned by the system_version RPC method - 0.1.4-e2dcdb474e7

node/service/src/rpc/runtime_info.rs
```diff
@@ -0,0 +1,94 @@
...
+};
+use sc_client_api::StorageProvider;
+use sc_executor_common::runtime_blob::RuntimeBlob;
+use sp_blockchain::HeaderBackend;
+use sp_runtime::traits::Block as BlockT;
+use std::{marker::PhantomData, sync::Arc};
+
+#[rpc(server)]
+pub(crate) trait RuntimeInfo<BlockHash> {
+    #[method(name = "gear_wasmRuntimeVersion")]
```
DennisInSky, line 33: Probably, it is worth renaming this into something like
`runtime_wasmVersion`
`runtime_wasmRuntimeVersion`
any other options are welcome

breathx: RuntimeVersion refers to https://github.com/gear-tech/gear/blob/4c345bc4270982cad4a695d1b910c6277f4498b6/runtime/gear/src/lib.rs#L106C1-L107 so it isn't good for me

breathx: furthermore, it's not just `gear_` stuff so `runtime_` prefix is preferable

breathx: maybe `runtime_rev`/`runtime_revision`/`runtime_blobRev`?

node/service/src/rpc/runtime_info.rs
```diff
@@ -0,0 +1,94 @@
...
+        let at = at.unwrap_or_else(|| self.client.info().best_hash);
+
+        let wasm_runtime_data = self
+            .client
+            .storage(
+                at,
+                &sp_storage::StorageKey(sp_core::storage::well_known_keys::CODE.into()),
+            )
+            .map_err(map_err_into_rpc_err)?;
+        let Some(wasm_runtime_data) = wasm_runtime_data else { return Ok("Not available".into()); };
```
breathx, line 70: Why Ok then? xd 
```suggestion
        let Some(wasm_runtime_data) = wasm_runtime_data else { return Err("Not available".into()); };
```

runtime/gear/src/lib.rs
```diff
@@ -98,6 +98,10 @@ mod migrations;
 // Weights used in the runtime.
 mod weights;
 
+#[link_section = "wasm_runtime_version"]
+static _RUNTIME_VERSION: [u8; const_str::to_byte_array!(env!("SUBSTRATE_CLI_IMPL_VERSION")).len()] =
```
breathx, line 102: is this name only one available? logically incorrect for me

breathx, line 102: u can just &str here and it will be utf-8 encoded as bytes in blob, isn't it?

breathx (changes requested):
Comments below and it'd great to introduce some test. Maybe gclient one? Or just sh script with curl request + CI step

runtime/gear/src/lib.rs
DennisInSky: Using references with named sections doesn't work with the wasm target. It requires it to be a simple fixed size array. I would not do all that trible dancing on my own.

runtime/vara/src/lib.rs
```diff
@@ -116,6 +116,10 @@ pub use extensions::DisableValueTransfers;
 
 mod migrations;
 
+#[link_section = "wasm_runtime_version"]
+static _RUNTIME_VERSION: [u8; const_str::to_byte_array!(env!("SUBSTRATE_CLI_IMPL_VERSION")).len()] =
```
shamilsan, line 120: Could we use `env!().len()` here, or do you find it unsafe for some Unicode cases? Anyway, it will be detected in compile time.
```suggestion
static _RUNTIME_VERSION: [u8; env!("SUBSTRATE_CLI_IMPL_VERSION").len()] =
```

DennisInSky: I think we can, but it to me it looks more consistent and more accurate. Do not see why we might want to go with `env!(...).len()`

shamilsan: It is shorter and more readable. We may use `env!("...").as_bytes().len()` to be more precise.

grishasobol (approved):

NikVolf:
@breathx @clearloop pls re-review

DennisInSky:
Added some testing

> * Get versions from the official node from different blocks
>   * the result is always the same ( hardcode versions for tests )

I do not think it is worth adding hardcoded versions in tests as they will change next time the node runtime gets upgraded which will break the test w/o any good reason. I guess we have some rules about release branches. In that case we can read its commit version and check against the official node.

> * Get versions from the different (official) nodes, ( or mb random peers ( if possible ) of the official node and the official node) from different blocks
>   

Do we have any means for connecting to different official nodes?

>   * compare the results
> * ( advanced, mb in another PR ) launch a local testnet with different wasm runtime set up, run `set_code` for serval nodes, check the result of this PRC

What are the options for doing this? The only one which comes to my mind is downloading some old node binary from gear, spin it up and set its code to the one built from the current repo, then check runtime version gets updated. Is it what you mean?

DennisInSky:
> lgtm!
> 
> plz one more test for the official (default) node mb, that would be really helpful for testing this API in production

How useful will it be, if the only option is to hardcode what it returns? I.e. it will break the test next time the official node runtime gets upgraded. This is not how the tests should work. I am pretty sure this version will be shown in the idea so it will be a good test for the official nodes

DennisInSky:
> Let's think about a situation in practice that we have missed in the current tests, now our chain is stuck at block 100, after tracing the error logs, we found there is something strange at block 50, during 50 ~ 100, we had several runtime upgrades, so now, **we want to know what's the WASM runtime version at block 50 in a specified node**, if the node client has the WASM runtime at block 50 in storage, version at 50 must be different to the version result at block 100, or the rpc will panic because of storage not found
> 
> > it will break the test next time the official node runtime gets upgraded.
> 
> the test still works after runtime upgrades because when we do runtime upgrades, we just upgrade the runtime but not destroying the storage, when client request to the wasm runtime version from previous blocks, the result should be as before, for example, the following code should work whatever how many runtime upgrades we have
> 
> ```rust
> const RUNTIME_VERSION_AT_BLOCK_999: &str = "xxx";
> // whatever how many runtime upgrades we have done, the test still works
> assert_eq!(request_runtime_version_at(Some(999)), RUNTIME_VERSION_AT_BLOCK_999)
> 
> // # when will it fails?
> //
> // 1, we restart our chain from genesis, this happens only if we are attacked or bcz of some big decisions
> // we decide to restart our network from genesis
> //
> // 2. we have cleaned the storage of our official node( since it is official node, imo we will never clean it xd)
> // there will be a panic from the rpc implementation beacuse of storage not found.
> ```
> 
> https://github.com/gear-tech/gear/blob/27fa917c4f377d1c1cce85acc403d19e646c00c9/node/service/src/rpc/runtime_info.rs#L73

Still not sure it will be helpful, but I have added a test against the `vara` node. At the moment it returns `No method` error which is fine.

NikVolf:
> Comments below and it'd great to introduce some test. Maybe gclient one? Or just sh script with curl request + CI step

resolve?

Issue #2512, opened 2023-04-04
Reverse gas model on Gear protocol


breathx:
### Problem to Solve

There is common problem of Web3 applications in need of getting somewhere money for even initial chain interactions, but workaround exists and appears as somebody's else gas and fee expenses.

### Smart contract solution
Create some program that will represent some "gas station" allowing usage of reserved gas for specified users.
`+` Fully on-chain
`-` Requires a lot of interactions
`-` Requires user to pay fee and minimal gas limit to interact the program (Doesn't totally solve initial problem)

### Extrinsic solution
Create some extrinsic that will be signed and send from user A and accept user B identifier and call that he wants to execute on-chain by accepting it's encoded and signed (for verification of call) data.
`+` Properly solves funds management problem
`+` More flexible and wide solution
`+` Doesn't require excess program calls: so will be executed faster due to absence of second queue waiting
`+ -` May be a bit unusual for substrate-around developers and users
`+ -` Requires a lot of documentation and considers about who should receive unused funds (mostly about just implementation)

### Fund-related solution
Allow on the protocol level delegate ability to use some funds as it realised for fungible tokens to avoid just granting for fees.

SabaunT:
#2777 has introduced that somehow, right? Maybe we should close that.

breathx:
Solved via the vouchers system

PR #2777, opened 2023-06-15
feat(node): Add `gear-voucher` pallet

ekovalev:
### Release notes

This change aims to provide an API for contract owners (or anyone who wishes to do it) to "sponsor" users by allocating some funds that can only be used to send messages to a particular program and which will cover both transaction fees and the gas cost (in case `value` is being sent, as well, it will have to be charged to the user's personal account).
A new pallet - `gear-voucher`, is introduced (the name owes to an obvious similarity between the described use-case and a "payment voucher" one gets in a usual retail store): a "voucher" is created for a user by sending the `pallet_gear_voucher::issue()` extrinsic that would create a keyless account in the `Balances` pallet unique for a `(account_id, program_id)` pair and transfer the voucher amount to it.

Besides, two more extrinsics have been added to the `gear` pallet:
- sending messages is done via the dedicated `send_message_with_voucher` extrinsic;
- sending replies using vouchers is also supported by means of the `send_reply_with_voucher` extrinsic.

When a user sends the `send_message_with_voucher` extrinsic to a program, the transaction fees are attempted to be charged to the keyless account corresponding to the `(user, program)` pair. If no such account exists or if it has been depleted, the entire call is invalidated and fails. In case of successful validation, the gas tree attached to the message to the program will have the voucher account as its `external origin` so that upon the message processing completion the gas leftovers are refunded back to the voucher and not the message sender (the user), thereby preventing potential abuse.

After a voucher has been fully redeemed, the `Balances` pallet will take care of removing the voucher account from the storage.

No additional storage is needed by the proposed implementation of the `gear-voucher` pallet thereby introducing very little overhead in the runtime.


NikVolf:
`reply_with_voucher also required` i think

ukint-vs (approved):

pallets/gear/src/lib.rs
```diff
@@ -1727,6 +1738,102 @@ pub mod pallet {
...
+                    })?;
+
+            // # Safety
+            //
+            // The `unreachable` clause is due to the `message_id` being newly generated
+            // with `Self::next_message_id`.
+            //
+            // Note: using the `voucher_id` as the external parent of the gas node in order for
+            // the leftover being refunded back to the voucher account and not the user's account.
+            GasHandlerOf::<T>::create(voucher_id, message.id(), gas_limit)
```
breathx, line 1821: Self::create

breathx: You can wrap all this functionality into one function and reuse for both extrinsics due to the fact that it differs only with account id owner of gas node

SabaunT:
I guess this https://github.com/gear-tech/gear/issues/2512 is resolved?

@breathx @ekovalev @NikVolf

NikVolf:
> gr_origin

i guess then we need to make an optional field to go along with origin node in the storage

not sure who added `gr_origin` to syscalls though, it's should not be required -- programs supposed to be agnostic to origin.

ekovalev:
> > gr_origin
> 
> i guess then we need to make an optional field to go along with origin node in the storage
> 
> not sure who added `gr_origin` to syscalls though, it's should not be required -- programs supposed to be agnostic to origin.

Agree that this syscall seems to violate some actor model assumptions (or, rather, impose additional restrictions). The actor processing a message shouldn't care whether the message sender is the "original" caller or not. No possible message handling logic should be based on any such assumption.

Anyway, for the meantime I'll go ahead changing the `ExternalOrigin` type for a the gas tree from plain `AccountId` to `(AccountId, Option<AccountId>)` where the "optional" field is the funds source/refund destination in case it is different from the sender's account.

shamilsan:
> Agree that this syscall seems to violate some actor model assumptions (or, rather, impose additional restrictions). The actor processing a message shouldn't care whether the message sender is the "original" caller or not. No possible message handling logic should be based on any such assumption.

This function had been added by the ecosystem's request and is used at the moment: https://github.com/search?q=org%3Agear-dapps+exec%3A%3Aorigin&type=code

SabaunT (approved):

NikVolf:
> > Agree that this syscall seems to violate some actor model assumptions (or, rather, impose additional restrictions). The actor processing a message shouldn't care whether the message sender is the "original" caller or not. No possible message handling logic should be based on any such assumption.
> 
> This function had been added by the ecosystem's request and is used at the moment: https://github.com/search?q=org%3Agear-dapps+exec%3A%3Aorigin&type=code

we don't blindly add what ecosystem wants without considering design

i suggest we remove it, and there should be away around it
and as i see, those examples are very poorly designed and implemented

pallets/gear-voucher/src/lib.rs
```diff
@@ -0,0 +1,205 @@
...
+//! - Deriving a unique keyless account id for a pair (user, program) to hold allocated funds.
+//!
+//! ## Interface
+//!
+//! ### Dispatchable Functions
+//!
+//! * `issue` - Issue an `amount` tokens worth voucher for a `user` to be used to pay fees and gas when sending messages to `program`.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+#![allow(clippy::items_after_test_module)]
```
NikVolf, line 47: why?

ekovalev: Otherwise clippy would complain. This is also the case for some other pallets/crates. I frankly fail to see the pattern: in some cases clippy check goes through, in some cases it does not.

PR #2779, opened 2023-06-15
refactor(gstd)!: Fix poor error naming in gstd

NikVolf:
we have no notion of `Contract`

global error can have no predicate at all

### DIFF
--- a/examples/binaries/reserve-gas/src/lib.rs
+++ b/examples/binaries/reserve-gas/src/lib.rs
@@ -66,7 +66,7 @@ pub type BlockCount = u32;
 mod wasm {
     use super::*;
     use gstd::{
-        errors::{ContractError, ExtError, ReservationError},
+        errors::{Error, ExtError, ReservationError},
         exec, msg,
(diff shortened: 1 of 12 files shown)


shamilsan (approved):

examples/binaries/reserve-gas/src/lib.rs
```diff
@@ -66,7 +66,7 @@ pub type BlockCount = u32;
 mod wasm {
     use super::*;
     use gstd::{
```
breathx, line 68: What about addition in gstd `pub use errors::Error` to be able to use `gstd::Error`

breathx: Same for `gstd::Result`

breathx (approved):

examples/binaries/reserve-gas/src/lib.rs
NikVolf: in separate pr sure

PR #2783, opened 2023-06-15
refactor(core-backend): rename generics `core-backend` & `core-processor` so they emphasize their essence

### LINKED ISSUE #2654: Rename all generics in the `core-backend` so they emphasize their nature/purposes
### Proposal

For example, `EP` can be changed to `EntryPoint`, `PrepMem` to `MemoryPreparationError`, `T` in `Environment::execute` to `MemoryPreparationError`.

It's hard to navigate in the code with short unclear generics names.

SabaunT:
Resolves #2654 

This PR actually **renames** only several traits/types:
- `EnvironmentExecutionError` -> `EnvironmentError` (as environment actually doesn't execute, errors aren't much related to execution).
- `ProcessorError` -> `ExtError` (`ProcessorExt` trait can have `ProcessorExt` error, but there's no such associated type in this trait. What's more there's no additional helping context with "Processor*" prefix to the error, especially if we look at the type definition). Same for `ProcessorAllocError` -> `ExtAllocError`.

All other changes are **name extensions** (instead of short names or one-letter generic names, which made it hard to understand the code and definitions of some types):
- generic `EP` -> EntryPoint
- generic `A` in some functions of `core_processor::execute` -> `ProcessorExt`
- trait `Ext` -> `Externalities`
- trait `ProcessorExt` -> `ProcessorExternalities`
- generic `PrepMem` (and just `T` in some places) -> `PrepareMemoryError`
- generic `E` -> `Ext`. That was the most important as `E` is used in the same crate/package for `Environment`, some error and `Externalities` too.

Some rules used for changes:
- Traits are given full names
- Generics, associated types and common types can have either full or short name (like `Externalities` trait implementor is `Ext`).

### DIFF
--- a/core-backend/common/src/lib.rs
+++ b/core-backend/common/src/lib.rs
@@ -233,6 +233,7 @@ pub struct ExtInfo {
     pub context_store: ContextStore,
 }
 
+/// Extended externalities that can manage gas counters.
 pub trait BackendExternalities: Externalities + CountersOwner {
     fn into_ext_info(self, memory: &impl Memory) -> Result<ExtInfo, MemoryError>;
 
@@ -246,23 +247,27 @@ pub trait BackendExternalities: Externalities + CountersOwner {
     ) -> Result<(), ProcessAccessError>;
 }
 
+/// A trait for conversion of the externalities API error to `TerminationReason`.
 pub trait BackendExternalitiesError: Clone + Sized {
     fn into_termination_reason(self) -> TerminationReason;
 }
 
 // TODO: consider to remove this trait and use Result<Result<Page, AllocError>, GasError> instead #2571
+/// A trait for conversion of the externalities memory management error to api error.
+///
+/// If the conversion fails, then `Self` is returned in the `Err` variant.
 pub trait BackendAllocExternalitiesError: Sized {
     type ExtError: BackendExternalitiesError;
 
     fn into_backend_error(self) -> Result<Self::ExtError, Self>;
 }
 
-pub struct BackendReport<MemWrap, Ext>
+pub struct BackendReport<EnvMem, Ext>
 where
     Ext: Externalities,
 {
     pub termination_reason: TerminationReason,
-    pub memory_wrap: MemWrap,
+    pub memory_wrap: EnvMem,
     pub ext: Ext,
 }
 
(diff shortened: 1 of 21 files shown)


gshep (approved):
👍

core-backend/common/src/lib.rs
```diff
@@ -246,83 +247,99 @@ pub trait BackendExt: EnvExt + CountersOwner {
...
-        entry_point: EP,
+        entry_point: EntryPoint,
         entries: BTreeSet<DispatchKind>,
         mem_size: WasmPage,
-    ) -> Result<Self, EnvironmentExecutionError<Self::Error, Infallible>>;
+    ) -> Result<Self, EnvironmentError<Self::SystemError, Infallible>>;
 
     /// Run instance setup starting at `entry_point` - wasm export function name.
-    fn execute<F, T>(self, pre_execution_handler: F) -> EnvironmentExecutionResult<T, Self, EP>
+    fn execute<PrepareMemory, PrepareMemoryError>(
```
shamilsan, line 332: As for me, we could leave the `F` and `E` generics here as well-known and readable within the context.
```suggestion
    fn execute<F, E>(
```

core-backend/common/src/lib.rs
```diff
@@ -356,10 +373,10 @@ pub trait BackendState {
     }
 
     /// Process alloc function result
-    fn process_alloc_func_result<T: Sized, E: BackendAllocExtError>(
+    fn process_alloc_func_result<T: Sized, ExtAllocError: BackendAllocExternalitiesError>(
```
shamilsan, line 376: <T, E> is also quite a stable pattern for generics returning `Result`.

core-backend/sandbox/src/env.rs
```diff
@@ -339,12 +342,19 @@ where
         }
     }
 
-    fn execute<F, T>(self, pre_execution_handler: F) -> EnvironmentExecutionResult<T, Self, EP>
+    fn execute<PrepareMemory, PrepareMemoryError>(
```
shamilsan, line 345: Also, I'd leave `<F, E>`:
```suggestion
    fn execute<F, E>(
```

core-backend/sandbox/src/env.rs
```diff
@@ -384,7 +394,7 @@ where
             unreachable!("We cannot use sandbox backend in std environment currently");
         };
 
-        match pre_execution_handler(&mut runtime.memory, stack_end, globals_config) {
+        match prepare_memory(&mut runtime.memory, stack_end, globals_config) {
```
shamilsan, line 397: What if we add some additional functionality to this function?

shamilsan (approved):
👍

core-backend/sandbox/src/env.rs
SabaunT: Currently it's a handler, which returns `PrepareMemoryError`, so renaming that seems fine. If complex functionality is introduced, not only name, but the closure signature will change.

core-backend/common/src/lib.rs
SabaunT: I can leave `F`, but I'd like to move from `E` as it was used for errors, environment, ext. what do you think?

core-backend/common/src/lib.rs
SabaunT: In this module we have another method, which accepts `Result` with a generic error type. Although `<T, E>` is a suitable pattern, I'd move from short generics in this crate/package as it lowers readability.

core-backend/common/src/lib.rs
SabaunT: agreed to resolve

core-backend/common/src/lib.rs
SabaunT: agreed to resolve

core-backend/sandbox/src/env.rs
SabaunT: agreed to resolve

PR #2789, opened 2023-06-16
refactoring(core): remove gas counter from `read`

### LINKED ISSUE #2380: consider removing GasLeft from `fn read` in environment externalities
### Problem to Solve

remove GasLeft return from `fn read` result:
```rust
fn read(&mut self, at: u32, len: u32) -> Result<(&[u8], GasLeft), Self::Error>;
```

SabaunT:
Resolves #2380

Introduced changes:
- [x] Rename `LimitedVec::get*` to `LimitedVec::inner*`
- [x] Rename messages method `payload` to `payload_bytes` as they return buffer, not the `Payload` itself. Also the reason for renaming is that `IncomingMessage` now returns `&mut Payload`.
- [x] Introduce type safe mechanics for optimized the `gr_read`. When `gr_read` is called the ownership over payload is taken, but must it be returned back to the message, because other payload usage is impossible. We now have type-level guarantees and compilation error guarding the proper implementation of the `read` in `core-backend::common::funcs`.

### DIFF
--- a/core-backend/common/src/funcs.rs
+++ b/core-backend/common/src/funcs.rs
@@ -30,7 +30,7 @@ use gear_backend_codegen::host;
 use gear_core::{
     buffer::RuntimeBuffer,
     costs::RuntimeCosts,
-    env::{DropPayloadLockBoundResult, Externalities},
+    env::{DropPayloadLockBound, Externalities},
     memory::{PageU32Size, WasmPage},
     message::{HandlePacket, InitPacket, ReplyPacket},
 };
@@ -174,14 +174,14 @@ where
 
     #[host(fallible, cost = RuntimeCosts::Read, err_len = LengthBytes)]
     pub fn read(ctx: &mut R, at: u32, len: u32, buffer_ptr: u32) -> Result<(), R::Error> {
-        let payload_holder = ctx.ext_mut().lock_payload(at, len)?;
-        payload_holder
-            .drop_with_job::<MemoryAccessError, _>(|payload_to_slice| {
+        let payload_lock = ctx.ext_mut().lock_payload(at, len)?;
+        payload_lock
+            .unlock_with_job::<MemoryAccessError, _>(|payload_slice| {
                 let write_buffer = ctx.register_write(buffer_ptr, len);
-                let write_res = ctx.write(write_buffer, payload_to_slice.to_slice());
-                let reclaim_res = ctx.ext_mut().unlock_payload(payload_to_slice.into_holder());
+                let write_res = ctx.write(write_buffer, payload_slice.as_slice());
+                let reclaim_res = ctx.ext_mut().unlock_payload(payload_slice.into_lock());
 
-                DropPayloadLockBoundResult::from((reclaim_res, write_res))
+                DropPayloadLockBound::from((reclaim_res, write_res))
             })
             .into_inner()
             .map_err(Into::into)
--- a/core-backend/common/src/funcs.rs
+++ b/core-backend/common/src/funcs.rs
@@ -174,10 +174,17 @@ where
 
     #[host(fallible, cost = RuntimeCosts::Read, err_len = LengthBytes)]
     pub fn read(ctx: &mut R, at: u32, len: u32, buffer_ptr: u32) -> Result<(), R::Error> {
-        let buffer = ctx.ext_mut().read_inner(at, len)?.to_vec();
+        let payload_holder = ctx.ext_mut().hold_payload(at, len)?;
+        payload_holder
+            .release_with_job(|payload_in_range| {
+                let write_buffer = ctx.register_write(buffer_ptr, len);
+                let res = ctx.write(write_buffer, payload_in_range.to_slice());
+                ctx.ext_mut()
+                    .reclaim_payload(payload_in_range.into_holder());
 
-        let write_buffer = ctx.register_write(buffer_ptr, len);
-        ctx.write(write_buffer, &buffer).map_err(Into::into)
+                res
+            })
+            .map_err(Into::into)
     }
 
     #[host(cost = RuntimeCosts::Size)]
--- a/core-backend/common/src/funcs.rs
+++ b/core-backend/common/src/funcs.rs
@@ -30,7 +30,7 @@ use gear_backend_codegen::host;
 use gear_core::{
     buffer::RuntimeBuffer,
     costs::RuntimeCosts,
-    env::{Externalities, ReleaseBoundResult},
+    env::{Externalities, UsePayloadHolderBoundResult},
     memory::{PageU32Size, WasmPage},
     message::{HandlePacket, InitPacket, ReplyPacket},
 };
@@ -183,7 +183,7 @@ where
                     .ext_mut()
                     .reclaim_payload(payload_to_slice.into_holder());
 
-                ReleaseBoundResult::from((reclaim_res, write_res))
+                UsePayloadHolderBoundResult::from((reclaim_res, write_res))
             })
             .into_inner()
             .map_err(Into::into)
--- a/core-backend/common/src/funcs.rs
+++ b/core-backend/common/src/funcs.rs
@@ -30,7 +30,7 @@ use gear_backend_codegen::host;
 use gear_core::{
     buffer::RuntimeBuffer,
     costs::RuntimeCosts,
-    env::Externalities,
+    env::{Externalities, ReleaseBoundResult},
     memory::{PageU32Size, WasmPage},
     message::{HandlePacket, InitPacket, ReplyPacket},
 };
@@ -176,14 +176,16 @@ where
     pub fn read(ctx: &mut R, at: u32, len: u32, buffer_ptr: u32) -> Result<(), R::Error> {
         let payload_holder = ctx.ext_mut().hold_payload(at, len)?;
         payload_holder
-            .release_with_job(|payload_in_range| {
+            .release_with_job::<MemoryAccessError, _>(|payload_to_slice| {
                 let write_buffer = ctx.register_write(buffer_ptr, len);
-                let res = ctx.write(write_buffer, payload_in_range.to_slice());
-                ctx.ext_mut()
-                    .reclaim_payload(payload_in_range.into_holder());
+                let write_res = ctx.write(write_buffer, payload_to_slice.to_slice());
+                let reclaim_res = ctx
+                    .ext_mut()
+                    .reclaim_payload(payload_to_slice.into_holder());
 
-                res
+                ReleaseBoundResult::from((reclaim_res, write_res))
             })
+            .into_inner()
             .map_err(Into::into)
     }
 
--- a/core-backend/common/src/funcs.rs
+++ b/core-backend/common/src/funcs.rs
@@ -176,7 +176,7 @@ where
     pub fn read(ctx: &mut R, at: u32, len: u32, buffer_ptr: u32) -> Result<(), R::Error> {
         let payload_lock = ctx.ext_mut().lock_payload(at, len)?;
         payload_lock
-            .unlock_with_job::<MemoryAccessError, _>(|payload_slice| {
+            .drop_with::<MemoryAccessError, _>(|payload_slice| {
                 let write_buffer = ctx.register_write(buffer_ptr, len);
                 let write_res = ctx.write(write_buffer, payload_slice.as_slice());
                 let reclaim_res = ctx.ext_mut().unlock_payload(payload_slice.into_lock());
--- a/core-backend/common/src/funcs.rs
+++ b/core-backend/common/src/funcs.rs
@@ -176,7 +176,7 @@ where
 
     #[host(fallible, cost = RuntimeCosts::Read, err_len = LengthBytes)]
     pub fn read(ctx: &mut R, at: u32, len: u32, buffer_ptr: u32) -> Result<(), R::Error> {
-        let buffer = ctx.ext_mut().read(at, len)?.to_vec();
+        let buffer = ctx.ext_mut().read_inner(at, len)?.to_vec();
 
         let write_buffer = ctx.register_write(buffer_ptr, len);
         ctx.write(write_buffer, &buffer).map_err(Into::into)
--- a/core-backend/common/src/funcs.rs
+++ b/core-backend/common/src/funcs.rs
@@ -176,14 +176,10 @@ where
 
     #[host(fallible, cost = RuntimeCosts::Read, err_len = LengthBytes)]
     pub fn read(ctx: &mut R, at: u32, len: u32, buffer_ptr: u32) -> Result<(), R::Error> {
-        let (buffer, mut gas_left) = ctx.ext_mut().read(at, len)?;
-        let buffer = buffer.to_vec();
+        let buffer = ctx.ext_mut().read(at, len)?.to_vec();
 
         let write_buffer = ctx.register_write(buffer_ptr, len);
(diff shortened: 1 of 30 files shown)


core-backend/common/src/funcs.rs
```diff
@@ -176,14 +176,10 @@ where
...
-        let (buffer, mut gas_left) = ctx.ext_mut().read(at, len)?;
-        let buffer = buffer.to_vec();
+        let buffer = ctx.ext_mut().read(at, len)?.to_vec();
 
         let write_buffer = ctx.register_write(buffer_ptr, len);
-        ctx.memory_manager_write(write_buffer, &buffer, &mut gas_left)?;
-
-        ctx.ext_mut().set_gas_left(gas_left);
-        Ok(())
+        ctx.write(write_buffer, &buffer).map_err(Into::into)
```
SabaunT, line 182: PR is ready for the review, but in draft until @grishasobol looks at it.

grishasobol (changes requested):
Blocked until not solved problem with `to_vec` in read syscall

core-backend/common/src/funcs.rs
```diff
@@ -174,14 +174,19 @@ where
...
-        ctx.ext_mut().set_gas_left(gas_left);
-        Ok(())
+        let payload_holder = ctx.ext_mut().lend_payload(at, len)?;
+        payload_holder
+            .use_with_job::<MemoryAccessError, _>(|payload_to_slice| {
+                let write_buffer = ctx.register_write(buffer_ptr, len);
+                let write_res = ctx.write(write_buffer, payload_to_slice.to_slice());
+                let reclaim_res = ctx
+                    .ext_mut()
+                    .reclaim_payload(payload_to_slice.into_holder());
```
grishasobol, line 184:
```suggestion
                 let mut aa = MessageContext::new(Default::default(), Default::default(), ContextSettings::new(0, 0, 0, 0, 0, 0));
                let mut bb = PayloadSliceHolder::try_new((0, 0), &mut aa).unwrap();
                let reclaim_res = ReclaimBoundResult::from((&mut aa, &mut bb));
```
This code is also compiles well but obviously incorrect.

grishasobol: Lol, interesting  but tests goes well))) may be need to add double read test

grishasobol (changes requested):

core-backend/common/src/funcs.rs
SabaunT: Definitely need this test

SabaunT: > ```suggestion
> 
>                  let mut aa = MessageContext::new(Default::default(), Default::default(), ContextSettings::new(0, 0, 0, 0, 0, 0));
> 
>                 let mut bb = PayloadSliceHolder::try_new((0, 0), &mut aa).unwrap();
> 
>                 let reclaim_res = ReclaimBoundResult::from((&mut aa, &mut bb));
> 
> ```
> 
> This code is also compiles well but obviously incorrect.

There's no defence against writing such wrong stuff at all. It violates everything. It's not the reason that current implementation is unsafe. We can do the same stuff in every syscall implementation. To have a proper security against that, we must make some types private.

grishasobol: Yes, but what the reason in such complex realisation if it does not guaranty compilation time safety. Better to use take_payload() and set_payload() carefully. You can also create PayloadHandler with panic in drop. Or there is also solution for compilation time, but it is a little strange maybe =) this one https://stackoverflow.com/questions/71694803/rust-type-that-requires-manual-drop

shamilsan:
The solution looks overcomplicated to me. Also, I'd rename `holder`s to `guard`s as the approach itself resembles `Mutex`/`MutexGuard`, as I see.

Agree with `LimitedVec::get*` -> `LimitedVec::inner*` and `payload` ->  `payload_bytes`.

core-backend/common/src/funcs.rs
SabaunT: > Yes, but what the reason in such complex realisation

1. That's the price of introducing type-safe approach. You literally create types for "every" step. As a result, you get code safe from **unintentional** wrong usage (when you forget to call something or use some type and etc.), which will error in compile time.

>  if it does not guaranty compilation time safety.

2. I guess you misunderstand what is a compile time safety feature. Basically, you use types to **thoroughly describe** your business logic. Wrong usage of types will mean a wrong implementation of business logic and you will understand that either from compilation error (which is the best case) or from tests (which is good too, but worse). What you have done in comment example, is that you have abused fully the business logic of `read`. Following this practise I can abuse business logic of every sys-call and then say, that implementation is not so safe, so why to make any "_complex realisation_".
3. Current type-safe implementation guarantees no **unintentional** errors, when you forget things like here https://github.com/gear-tech/gear/commit/c1d951afadeebc9f9490661ea04fb35af2122977 (return from funcs before setting gas), or like calling `to_vec` in `read` here https://github.com/gear-tech/gear/commit/c1d951afadeebc9f9490661ea04fb35af2122977, or like we had before refactoring the `wasmi` situations when termination reason caused gas counter not being updated. These are **unintentional** errors, because developer "forgets" doing something. They were in our code base for a long time. We can't rely much on reviews when there are vague errors from too much implicit business logic not guarded by types. But, again, example you provided is a full abuse of business logic: you can't and not expected to create a `MessageContext` there, it's unarguably wrong.
4. You can make a fair point that `MessageContext` then mustn't be instantiated so easy with just call to `new` from anywhere. Yes, it must be possible only to create it from `execute_wasm` method. I decided not to implement this constraint as it was a bit too much, because no-one with a common sense will think to commit to `master` such a code, that creates an additional `MessageContext` in a wrong place, when this type is hidden under `ctx` param of the function.

> Better to use take_payload() and set_payload() carefully.

5. Again, "carefully" is a wrong approach now. I gave you 3 examples of code in "master". Were dev not careful during that? I guess they were. Mistakes happen. But not guarantee they won't happen further. But with strong types we get more guarantees for a safer usage of externalities methods.

> You can also create PayloadHandler with panic in drop

7. Implementing such runtime approach or, for example, returning error from `into_ext_info` if payload was taken and not set again, will require returning a `System` error from backend, which will be considered as a panic on `pallet_gear`.

> Or there is also solution for compilation time, but it is a little strange maybe =) this one https://stackoverflow.com/questions/71694803/rust-type-that-requires-manual-drop

`ManuallyDrop` brings associated risks. It's actually making solution more complex. Complexity is not with 3-4 more additional types, but with what they do. Do you find, what introduced in this PR types do more risky and complex than `ManuallyDrop`?

core/src/buffer.rs
```diff
@@ -111,12 +111,12 @@ impl<T: Clone + Default, E: Default, const N: usize> LimitedVec<T, E, N> {
     }
 
     /// Returns ref to the internal data.
-    pub fn get(&self) -> &[T] {
```
ark0f, line 114: I guess naming was correct:

> The get naming is used only when there is a single and obvious thing that could reasonably be gotten by a getter. For example [Cell::get](https://doc.rust-lang.org/std/cell/struct.Cell.html#method.get) accesses the content of a Cell.

From: https://rust-lang.github.io/api-guidelines/naming.html

core/src/env.rs
```diff
@@ -19,25 +19,155 @@
...
+/// This is actually done to prevent a user of the type from holding payload of the
+/// message, which actually moves it, from forgetting to release it back, because
+/// if access to the slice buffer was granted directly from the holder, the type user
+/// could have written the data to memory and then have dropped the holder. As a result
+/// the executing message payload wouldn't have been returned. So [`PayloadSliceHolder::use_with_job`]
+/// is a kind of scope-guard for the data and the [`PayloadToSlice`] is a data access guard.
+///
+/// For more usage info read [`Self::use_with_job`] docs.
+
+pub struct PayloadSliceHolder {
```
ark0f, line 50:
```suggestion
pub struct PayloadHolder {
```

core/src/env.rs
```diff
@@ -19,25 +19,155 @@
...
+    ///
+    /// We ensure that the payload is released back by returning the [`UsePayloadHolderBoundResult`]
+    /// from the `job`. This type can actually be instantiated only from tuple of two:
+    /// [`ReclaimBoundResult`] and some result with err variant type to be `JobErr`.
+    /// The first is returned from [`Externalities::reclaim_payload`], so it means that
+    /// that payload was reclaimed by the original owner. The other result stores actual
+    /// error of the `Job` as it could have called fallible actions inside it. So,
+    /// [`UsePayloadHolderBoundResult`] gives an opportunity to store the actual result of the job,
+    /// but also gives guarantee that payload was reclaimed.
+    pub fn use_with_job<JobErr, Job>(mut self, mut job: Job) -> UsePayloadHolderBoundResult<JobErr>
```
ark0f, line 99:
```suggestion
    pub fn with<JobErr, Job>(mut self, mut job: Job) -> UsePayloadHolderBoundResult<JobErr>
```

core/src/env.rs
```diff
@@ -177,9 +307,23 @@ pub trait Externalities {
...
+    /// means, that the payload value in the currently executed message will become
+    /// "de-allocated" or just zeroed. To prevent from the risk of payload being
+    /// not "returned" back to the message a [`Externalities::reclaim_payload`] is
+    /// introduced. For more info, read docs to [`PayloadSliceHolder`], [`UsePayloadHolderBoundResult`],
+    /// [`ReclaimBoundResult`], [`PayloadToSlice`] types and their methods.
+    fn lend_payload(&mut self, at: u32, len: u32) -> Result<PayloadSliceHolder, Self::Error>;
+
+    /// Reclaims ownership from the payload holder over previously taken payload from the
+    /// currently executing message.
+    fn reclaim_payload(&mut self, payload_holder: &mut PayloadSliceHolder) -> ReclaimBoundResult;
```
ark0f, line 326:
```suggestion
    fn reclaim_payload(&mut self, payload_to_slice: PayloadToSlice) -> ReclaimBoundResult;
```

core/src/env.rs
```diff
@@ -19,25 +19,155 @@
...
+}
+
+/// Result of calling a `job` within [`PayloadSliceHolder::use_with_job`].
+///
+/// This is a "bound" type which means it's main purpose is to give
+/// some type-level guarantees. More precisely, it gives guarantee
+/// that payload value was reclaimed by the owner. Also it stores the error
+/// of the `job`, which gives opportunity to handle the actual job's runtime
+/// error, but not bound wrappers.
+pub struct UsePayloadHolderBoundResult<JobError> {
```
ark0f, line 139:
```suggestion
pub struct WithPayloadHolderResult<JobError> {
```

core/src/env.rs
```diff
@@ -19,25 +19,155 @@
...
+        UsePayloadHolderBoundResult { job_result }
+    }
+}
+
+/// Result of calling [`Externalities::reclaim_payload`].
+///
+/// This is a "bound" type which means it doesn't store
+/// anything, but gives type-level guarantees that [`PayloadSliceHolder`]
+/// released the payload back to the message context.
+pub struct ReclaimBoundResult(());
```
ark0f, line 163:
```suggestion
pub struct PayloadReclaimResult(());
```

ark0f (changes requested):

core/src/env.rs
```diff
@@ -19,25 +19,155 @@
...
+        })
+    }
+
+    /// Releases back ownership of the held payload to the message context.
+    ///
+    /// The method actually performs [`mem::swap`] under the hood. It's supposed
+    /// to be called from [`Externalities::reclaim_payload`], implementor of which
+    /// owns provided message context.
+    pub fn release_back(&mut self, msg_ctx: &mut MessageContext) {
+        mem::swap(msg_ctx.payload_mut(), &mut self.payload);
```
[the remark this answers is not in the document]
SabaunT: Making private :+1:

core/src/env.rs
```diff
@@ -19,25 +19,155 @@
...
+/// is instantiated only inside the [`Self::use_with_job`] method.
+/// This is actually done to prevent a user of the type from holding payload of the
+/// message, which actually moves it, from forgetting to release it back, because
+/// if access to the slice buffer was granted directly from the holder, the type user
+/// could have written the data to memory and then have dropped the holder. As a result
+/// the executing message payload wouldn't have been returned. So [`PayloadSliceHolder::use_with_job`]
+/// is a kind of scope-guard for the data and the [`PayloadToSlice`] is a data access guard.
+///
+/// For more usage info read [`Self::use_with_job`] docs.
+
```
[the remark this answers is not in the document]
SabaunT: done

core/src/env.rs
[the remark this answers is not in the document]
SabaunT: done

core/src/env.rs
```diff
@@ -177,9 +307,23 @@ pub trait Externalities {
...
-    fn read(&mut self, at: u32, len: u32) -> Result<(&[u8], GasLeft), Self::Error>;
+    /// Takes ownership over payload of the executing message and
+    /// returns it in the wrapper [`PayloadSliceHolder`].
+    ///
+    /// Due to details of implementation of the runtime which executes gear
+    /// sys-calls inside wasm execution environment (either wasmi or sp_sandbox),
+    /// to prevent additional memory allocation on payload read op, we give
+    /// ownership over payload to the caller. Giving ownership over payload actually
+    /// means, that the payload value in the currently executed message will become
+    /// "de-allocated" or just zeroed. To prevent from the risk of payload being
```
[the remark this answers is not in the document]
SabaunT: done

core/src/env.rs
```diff
@@ -19,25 +19,155 @@
...
+
+/// A wrapper over mutable reference to [`PayloadSliceHolder`]
+/// which can give to the caller the slice of the held payload.
+///
+/// For more information read [`PayloadSliceHolder`] docs.
+pub struct PayloadToSlice<'a>(&'a mut PayloadSliceHolder);
+
+impl<'a> PayloadToSlice<'a> {
+    /// Returns slice of the held payload.
+    pub fn to_slice(&self) -> &[u8] {
```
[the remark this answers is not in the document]
SabaunT: done

core/src/env.rs
```diff
@@ -19,25 +19,155 @@
...
+        // Will not panic as range is checked.
+        &self.payload.inner()[start..end]
+    }
+}
+
+/// A wrapper over mutable reference to [`PayloadSliceHolder`]
+/// which can give to the caller the slice of the held payload.
+///
+/// For more information read [`PayloadSliceHolder`] docs.
+pub struct PayloadToSlice<'a>(&'a mut PayloadSliceHolder);
```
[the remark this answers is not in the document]
SabaunT: done

core/src/env.rs
SabaunT: done

shamilsan (approved):

core/src/env.rs
```diff
@@ -19,25 +19,152 @@
...
+            range: (start as usize, end as usize),
+        })
+    }
+
+    /// Releases back ownership of the locked payload to the message context.
+    ///
+    /// The method actually performs [`mem::swap`] under the hood. It's supposed
+    /// to be called from [`Externalities::unlock_payload`], implementor of which
+    /// owns provided message context.
+    fn unlock_back(&mut self, msg_ctx: &mut MessageContext) {
```
NikVolf, line 79:
```suggestion
    fn release(&mut self, msg_ctx: &mut MessageContext) {
```

grishasobol (approved):

PR #2790, opened 2023-06-16
feat(node-loader): Spawn batch pool in new thread

### LINKED ISSUE #1876: Implement "event-driven" design for the `gear-node-loader`
A draft design:
1. One "main" thread that subscribes to gear node events when the loader is started.
2. Required events, that are useful for some entities (for example, a batch of concrete gear calls), are received by applying filters on the subscription channel. Filters are applied by those entities.
3. Each entity that waits for the events, have an access to the subscription reading channel. After processing received events, if data mutations are required, the entity sends "journal note" (pretty same thing as we have in `core_processor`) to the main thread, which performs required mutations to the context.

ukint-vs:
Resolves #1876 .

- Use `EventsReciever` to process events.
- Use tokio timeout to catch stopped node.
- Spawn `BatchPool` in new thread.

@gear-tech/dev

### DIFF
--- a/utils/node-loader/src/batch_pool.rs
+++ b/utils/node-loader/src/batch_pool.rs
@@ -21,7 +21,7 @@ use std::{
 use tokio::{sync::broadcast::Receiver, time::Duration};
 use tracing::instrument;
 
-mod api;
+pub mod api;
 mod context;
 pub mod generators;
 mod report;
@@ -40,7 +40,12 @@ pub struct BatchPool<Rng: CallGenRng> {
 }
 
 impl<Rng: CallGenRng> BatchPool<Rng> {
-    fn new(api: GearApiFacade, batch_size: usize, pool_size: usize, rx: EventsReciever) -> Self {
+    pub fn new(
+        api: GearApiFacade,
+        batch_size: usize,
+        pool_size: usize,
+        rx: EventsReciever,
+    ) -> Self {
         Self {
             api,
             pool_size,
@@ -51,20 +56,12 @@ impl<Rng: CallGenRng> BatchPool<Rng> {
         }
     }
 
-    pub async fn run(params: LoadParams, rx: EventsReciever) -> Result<()> {
-        let api = GearApiFacade::try_new(params.node, params.user).await?;
-
-        let mut batch_pool = Self::new(
-            api.clone(),
-            params.batch_size,
-            params.workers,
-            rx.resubscribe(),
-        );
-
-        let run_pool_task = batch_pool.run_pool_loop(params.loader_seed, params.code_seed_type);
+    pub async fn run(mut self, params: LoadParams, rx: EventsReciever) -> Result<()> {
+        let gear_api = self.api.clone().into_gear_api();
+        let run_pool_task = self.run_pool_loop(params.loader_seed, params.code_seed_type);
         let inspect_crash_task = tokio::spawn(inspect_crash_events(rx));
         let renew_balance_task =
-            tokio::spawn(create_renew_balance_task(api.into_gear_api(), params.root).await?);
+            tokio::spawn(create_renew_balance_task(gear_api, params.root).await?);
 
         let run_result = tokio::select! {
             r = run_pool_task => r,
--- a/utils/node-loader/src/batch_pool.rs
+++ b/utils/node-loader/src/batch_pool.rs
@@ -6,7 +6,7 @@ use anyhow::{anyhow, Result};
 use api::GearApiFacade;
 use context::Context;
 use futures::{stream::FuturesUnordered, Future, StreamExt};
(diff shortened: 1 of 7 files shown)


utils/node-loader/src/batch_pool.rs
```diff
@@ -17,9 +17,11 @@ use report::{BatchRunReport, MailboxReport, Report};
 use std::{
     collections::{BTreeMap, BTreeSet},
     marker::PhantomData,
-    time::Duration,
 };
-use tokio::sync::broadcast::Receiver;
+use tokio::{
+    sync::broadcast::Receiver,
+    time::{sleep, timeout, Duration},
```
grishasobol, line 23: as I now this is not good practice to use functions like this. Better old variant with using tokio::time::sleep(...)

SabaunT (changes requested):
please wait for review

utils/node-loader/src/main.rs
```diff
@@ -61,5 +61,13 @@ async fn load_node(params: LoadParams) -> Result<()> {
         env!("CARGO_PKG_VERSION"),
     );
 
-    BatchPool::<SmallRng>::run(params, rx).await
+    let api = GearApiFacade::try_new(params.node.clone(), params.user.clone()).await?;
+
+    let batch_pool = BatchPool::<SmallRng>::new(
+        api.clone(),
```
SabaunT, line 67: `api` is unused after `batch_pool`. Any need for `clone`?

utils/node-loader/src/main.rs
```diff
@@ -61,5 +61,13 @@ async fn load_node(params: LoadParams) -> Result<()> {
...
+        params.batch_size,
+        params.workers,
+        rx.resubscribe(),
+    );
+    tokio::spawn(batch_pool.run(params, rx)).await?
```
SabaunT, line 72: Here we handle workers of the `batch_pool`, but what about the `listen_events`? That thread also returns an error, but it's not handled, as I see

utils/node-loader/src/batch_pool.rs
```diff
@@ -248,24 +257,55 @@ async fn process_events(
...
-            return Err(anyhow!(utils::EVENTS_TIMEOUT_ERR_STR));
-        }
+            let mut mailbox_from_events = utils::capture_mailbox_messages(&api, &mut v)
+                .await
+                .expect("always valid by definition");
+            mailbox_added.append(&mut mailbox_from_events);
+
+            tokio::task::block_in_place(|| {
+                tokio::runtime::Handle::current()
+                    .block_on(async { v.err_or_succeed_batch(messages.keys().copied()).await })
```
SabaunT, line 306: btw, we can re-implement that on the loader, because we now have subxt, gclient, gsdk as dependencies here. What's more, `err_or_succeed_batch` doesn't have inside it any tasks to wait. It will return value immediately without yielding control to other thread. So it's worth having own implementation without any sync.

utils/node-loader/src/batch_pool.rs
```diff
@@ -248,24 +257,55 @@ async fn process_events(
...
-                // `gclient::EventProcessor` implementation on `IntoIterator` implementor clones `self` without mutating
-                // it, although `proc` and `proc_many` take mutable reference on `self`.
-                let mut mailbox_from_events = utils::capture_mailbox_messages(&api, &mut v)
-                    .await
-                    .expect("always valid by definition");
-                mailbox_added.append(&mut mailbox_from_events);
-
-                v.err_or_succeed_batch(messages.keys().copied()).await
+        let mut events = rx.recv().await?;
+        while events.block_hash() != block_hash {
```
SabaunT, line 261: what's the purpose of this while loop? why can't we rely only on values from `r` below?

SabaunT (changes requested):
Need some clarifications and resolving given comments.

Do you have any suggestions how to support refactoring or any other changes? I mean, we should have some proper mechanics to be sure any changes won't affect sensitive implicit logic of the batch pool. Especially processing events. Maybe tests.

utils/node-loader/src/batch_pool.rs
ukint-vs: we need to get events since specific block hash. We are waiting for it here.

utils/node-loader/src/main.rs
ukint-vs: Done

utils/node-loader/src/batch_pool.rs
ukint-vs: DOne

grishasobol (approved):

SabaunT (approved):
Please add some clarification comments to most important parts of batch pool (like how events are processed, what's the general design for some mechanics and such).

We need that for further loader refactoring and tests.

PR #2801, opened 2023-06-16
fix(gear-wasm-instrument): correctly calculate cost of `gas_charge` itself

StackOverflowExcept1on:
```rust
fn gas_charge(gas_to_charge: u32) {}
```
This function compiles to
```wasm
i32.const $gas_to_charge ; <--- not charged now
call $gas_charge
```
There should be gas charge for this instruction too

### DIFF
--- a/utils/wasm-instrument/src/lib.rs
+++ b/utils/wasm-instrument/src/lib.rs
@@ -227,12 +227,17 @@ pub fn inject<R: Rules>(
         None => return Err(mbuilder.build()),
     };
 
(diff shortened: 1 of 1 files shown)


grishasobol (approved):

gshep (approved):
Good catch!

ukint-vs (approved):

gshep:
Sorry forgot to mention that after these changes all on-chain programs should be reinstrumented

ukint-vs:
> Sorry forgot to mention that after these changes all on-chain programs should be reinstrumented

Need to bump schedule version

PR #2805, opened 2023-06-17
refactor(gear-wasm-instrument): optimize `gas_charge` function

### LINKED ISSUE #1706: Optimize gas_charge function
### File Location(s)

utils/wasm-instrument/src/lib.rs

### Proposal

Current function body is a straightforward list of instructions. It could be optimized by storing value with gas to charge in a separate local, for example.

StackOverflowExcept1on:
Resolves #1706

Changes:
- bump version in schedule to force re-instrumentation
- make it easier to read the code that determines the cost of successful execution
- optimize the `gas_charge` function by using the `local.tee` instruction to store `total_gas_to_charge` in a local variable
  ```wasm
  i32.const 10
  local.tee 1 ;; <--- sets 10 to local variable with index 1 & pushes 10 onto the stack
  
  ;; for example 10 can be used as an argument
  call $func ;; func(10)
  ```
  The performance of the `gas_charge` function has increased by ~14%. (cost of calling this function: 47133 -> 40992, tested on gear-runtime).

@gear-tech/dev

### DIFF
--- a/utils/wasm-instrument/src/lib.rs
+++ b/utils/wasm-instrument/src/lib.rs
@@ -204,20 +204,20 @@ pub fn inject<R: Rules>(
     ];
 
     // determine cost for successful execution
-    let mut scopes = 0isize;
+    let mut block_of_code = false;
 
     let cost_blocks = match elements
         .iter()
         .filter(|instruction| match instruction {
             Instruction::If(_) => {
-                scopes += 1;
-                scopes == 1
+                block_of_code = true;
+                true
             }
             Instruction::End => {
-                scopes -= 1;
+                block_of_code = false;
                 false
             }
-            _ => scopes == 0,
+            _ => !block_of_code,
         })
         .try_fold(0u64, |cost, instruction| {
             rules
--- a/utils/wasm-instrument/src/lib.rs
+++ b/utils/wasm-instrument/src/lib.rs
@@ -238,7 +238,9 @@ pub fn inject<R: Rules>(
         None => return Err(mbuilder.build()),
     };
 
-    let cost = cost_push_arg + cost_call + cost_blocks;
+    let cost_local_var = rules.call_per_local_cost() as u64;
+
+    let cost = cost_push_arg + cost_call + cost_local_var + cost_blocks;
     // the cost is added to gas_to_charge which cannot
     // exceed u32::MAX value. This check ensures
     // there is no u64 overflow.
--- a/utils/wasm-instrument/src/lib.rs
+++ b/utils/wasm-instrument/src/lib.rs
@@ -149,18 +149,18 @@ pub fn inject<R: Rules>(
             .build(),
     );
 
-    // TODO: #1706
     let mut elements = vec![
         // check if there is enough gas
         Instruction::GetGlobal(gas_index),
-        // calculate gas_to_charge + cost_for_func
+        // total_gas_to_charge = gas_to_charge + cost_for_func
         // {
         Instruction::GetLocal(0),
         Instruction::I64ExtendUI32,
         Instruction::I64Const(i64::MAX),
         Instruction::I64Add,
+        Instruction::TeeLocal(1),
         // }
-        // if gas < (gas_to_charge + cost_for_func)
+        // if gas < total_gas_to_charge
         Instruction::I64LtU,
         Instruction::If(elements::BlockType::NoResult),
         Instruction::Call(out_of_gas_index),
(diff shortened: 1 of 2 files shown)


utils/wasm-instrument/src/lib.rs
```diff
@@ -260,6 +261,7 @@ pub fn inject<R: Rules>(
             .with_param(ValueType::I32)
             .build()
             .body()
+            .with_locals(vec![elements::Local::new(1, ValueType::I64)])
```
gshep, line 264: it is required to take a cost for defining local(s) into account

utils/wasm-instrument/src/lib.rs
```diff
@@ -213,17 +204,27 @@ pub fn inject<R: Rules>(
...
+            Instruction::If(_) => {
+                block_of_code = true;
+                true
+            }
+            Instruction::End => {
+                block_of_code = false;
+                false
+            }
+            _ => !block_of_code,
+        })
```
gshep, line 221: Sorry cannot agree with the edits. This cost calculation is hardwired to the function body so if it would be changed sometime then these lines should be adjusted accordingly the same way as with "magic constants".

gshep (approved):
looks good! 👍

ukint-vs (approved):

utils/wasm-instrument/src/lib.rs
```diff
@@ -237,7 +238,9 @@ pub fn inject<R: Rules>(
         None => return Err(mbuilder.build()),
     };
 
-    let cost = cost_push_arg + cost_call + cost_blocks;
+    let cost_local_var = rules.call_per_local_cost() as u64;
+
+    let cost = cost_push_arg + cost_call + cost_local_var + cost_blocks;
```
breathx, line 243: can we have a proof that it become cheaper?

utils/wasm-instrument/src/lib.rs
breathx: > Sorry cannot agree with the edits. This cost calculation is hardwired to the function body so if it would be changed sometime then these lines should be adjusted accordingly the same way as with "magic constants".

Why? It's more general than some previous skips etc.

utils/wasm-instrument/src/lib.rs
```diff
@@ -213,17 +204,27 @@ pub fn inject<R: Rules>(
...
+
     let cost_blocks = match elements
         .iter()
-        .take(7)
-        // block with update instructions
-        .chain(elements.iter().skip(10).take(7))
+        .filter(|instruction| match instruction {
+            Instruction::If(_) => {
+                block_of_code = true;
+                true
```
breathx, line 214: should we include if cost?

breathx (approved):

utils/wasm-instrument/src/lib.rs
breathx: > Yes, see how the previous implementation works with `.chain()`

But it may be initial issue xd

breathx: @gshep

utils/wasm-instrument/src/lib.rs
breathx: That's really nice performance improvement!

breathx:
Never force-push after reviews

utils/wasm-instrument/src/lib.rs
StackOverflowExcept1on: What do you mean by "how does this affect the fungible token"? This cost will be the same for all contracts. Because the `gas_charge` function is inserted at the end and always consists of the same instructions

utils/wasm-instrument/src/lib.rs
StackOverflowExcept1on: @breathx when `Instruction::If` is executed, a jump occurs either to an unreachable part of the code, or to the instruction after `Instruction::End`

utils/wasm-instrument/src/lib.rs
breathx: > What do you mean by "how does this affect the fungible token"? This cost will be the same for all contracts. Because the `gas_charge` function is inserted at the end and always consists of the same instructions

Sure, but what's wrong with some real example? Gas costs will differ between contract with 123012030 calls of gas charge and for example 50 :/

StackOverflowExcept1on:
CI is broken on the master. I think we can merge this anyway.

grishasobol (commented):
Whether we really need to globals for gas counters? See this one global work example:
1) Before execution we have `gl` and `ga`  (gas limit and gas allowance)
We set this counters in `ext` and set `min(gl, ga)` in wasm global.
2) During code execution can be two cases when we can decrease this global value:
2.1 In charge function (which you modify). If overflow occurs, then call function out_of_gas, which will check counters in `ext` and understand which one is overflowed (in case eq then suppose that it is `gl`).
2.2 In lazy-pages. After lazy-pages charging overflow occurs 3 cases can be:
2.2.1  Execution will be continued until the next  gas charging function call and then see `2.1`.
2.2.2 Execution will be continued until some sys-call call then see `3`.
2.2.3 Execution will be finished, then we will check lazy pages status and if it's overflow, then check counter like in `2.1`.
3) Inside syscall execution we can check which one `gl` or `ga` is minimum and understand which one value is inside wasm global. Then we will calculate a difference and subtract it from both counters, checking overflows.

utils/wasm-instrument/src/lib.rs
StackOverflowExcept1on: @breathx I am using `fungible-token::stress_transfer`. It shows the following results:
```
# this PR
Gas burned for one transfer operation = 1.3684201959330864 * 10^9. Calculated as geometric mean from 100 transfer operations.

# master
Gas burned for one transfer operation = 1.3721279099998007 * 10^9. Calculated as geometric mean from 100 transfer operations.
```

How it was tested:
```bash
cargo build --package gear-cli --release
cargo test --package demo-fungible-token --release stress_transfer -- --exact --include-ignored --nocapture
```

```diff
diff --git a/examples/binaries/fungible-token/tests/benchmarks.rs b/examples/binaries/fungible-token/tests/benchmarks.rs
index 10a255d8f..2ce34ad1d 100644
--- a/examples/binaries/fungible-token/tests/benchmarks.rs
+++ b/examples/binaries/fungible-token/tests/benchmarks.rs
@@ -23,7 +23,7 @@ use ft_io::*;
 use gclient::{EventProcessor, GearApi, Result};
 use gear_core::ids::{MessageId, ProgramId};
 use gstd::{vec, ActorId, Encode, Vec};
-use rand::Rng;
+use rand::{rngs::StdRng, Rng, SeedableRng};
 use statrs::statistics::Statistics;

 /// Path to the gear node binary.
@@ -206,7 +206,7 @@ async fn stress_test() -> Result<()> {
 #[ignore]
 #[tokio::test]
 async fn stress_transfer() -> Result<()> {
-    let mut rng = rand::thread_rng();
+    let mut rng = StdRng::seed_from_u64(0xdeadbeef);

     let api = GearApi::dev_from_path(GEAR_PATH).await?;
     // Use this code in comment for custom node run:
```

ukint-vs (approved):

PR #2827, opened 2023-06-19
feat(gtest): add some gtest logging

NikVolf:
also remove stupid duplicate dep


breathx (approved):
Don't we want to append gtest log target by default in env filter inside `fn init_logger?

breathx:
> > Don't we want to append gtest log target by default in env filter inside `fn init_logger?
> 
> i would've removed gwasm instead
> 
> debug logs should be shown when actively debugging

Than no reason to call `init_logger` for actual test without debug purpose

NikVolf:
changed all one log target which is already there

shamilsan (approved):

grishasobol (approved):

PR #2828, opened 2023-06-19
fix (gcore): fix read length in gcore

NikVolf:


ukint-vs (approved):

SabaunT (approved):

gcore/src/msg.rs
```diff
@@ -171,7 +171,7 @@ pub fn read_at(offset: usize, buffer: &mut [u8]) -> Result<()> {
         unsafe {
             gsys::gr_read(
                 offset as u32,
-                size as u32,
```
breathx, line 174: Call of `gcore::size()` must be removed from the beginning of func then as well as check for zero length adapted?

breathx (changes requested):

gcore/src/msg.rs
NikVolf: size is still used to return error if out of bounds

gcore/src/msg.rs
```diff
@@ -159,23 +159,24 @@ pub fn read(buffer: &mut [u8]) -> Result<()> {
 /// }
 /// ```
 pub fn read_at(offset: usize, buffer: &mut [u8]) -> Result<()> {
+    if buffer.is_empty() {
+        return SyscallError(0).into_result();
```
breathx, line 163: pay attention on that on merge @ark0f of your pr

breathx (approved):

PR #2838, opened 2023-06-19
base implementation of stack allocation usage in read, load and reply

grishasobol:
1) Add initial logic to use buffers allocated in program stack memory. This prevents from allocations in heap and from additional read/writes in heap memory, which leads to additional pages data read/writes from storage.
2) Implement read/load/reply function with stack allocation usages.
3) Apply with_read_on_stack for `demo-piggy-bank`.
4) Apply `load_on_stack` for `demo-btree` and `demo-fungble-token`.
5) Apply `reply_on_stack` for `demo-fungble-token`.

Performance command:
```
cargo test --release --package demo-fungible-token --test benchmarks -- stress_transfer --exact --nocapture --ignored
```
Measurement results:
```
Old burned gas amount per one transfer action in FT: 1.376363081119030 * 10^9
New burned gas amount per one transfer action in FT: 0.999131985815945 * 10^9
```


gstd/codegen/src/lib.rs
```diff
@@ -572,6 +573,86 @@ pub fn wait_create_program_for_reply(attr: TokenStream, item: TokenStream) -> To
...
+/// }
+/// ```
+///
+/// # IMPORTANT:
+/// 1) Input function must not have another attributes.
+/// 2) Input function return type must be `()`.
+/// 3) Because of `1)` cannot be used with async macros.
+///
+#[proc_macro_attribute]
+pub fn message_read_on_stack(_attr: TokenStream, item: TokenStream) -> TokenStream {
```
NikVolf, line 621: please no, I don't see any solid reason to pollute userspace api with macros here

grishasobol: Ok, just add it, because @LouiseMedova said, that for them it will be more comfortable to use macros like this.

NikVolf: This can be moved to more opinionated crate supported by ecosystem

grishasobol: removed

gstd/src/msg/encoded.rs
```diff
@@ -113,6 +131,34 @@ pub fn reply<E: Encode>(payload: E, value: u128) -> Result<MessageId> {
...
+            self.buffer[self.offset..end_offset].copy_from_slice(bytes);
+            self.offset = end_offset;
+        }
+    }
+
+    gcore::stack_buffer::with_byte_buffer(E::max_encoded_len(), |buffer| {
+        let mut output = ExternalBufferOutput { buffer, offset: 0 };
+        payload.encode_to(&mut output);
+        let ExternalBufferOutput { buffer, offset } = output;
+        super::reply_bytes(&buffer[..offset], value)
```
ukint-vs, line 158:
```suggestion
        super::reply_bytes(&output.buffer[..offset], value)
```

ukint-vs (approved):
Awesome

gstd/src/msg/encoded.rs
```diff
@@ -113,6 +131,34 @@ pub fn reply<E: Encode>(payload: E, value: u128) -> Result<MessageId> {
...
+                panic!("{}", err_log);
+            }
+            self.buffer[self.offset..end_offset].copy_from_slice(bytes);
+            self.offset = end_offset;
+        }
+    }
+
+    gcore::stack_buffer::with_byte_buffer(E::max_encoded_len(), |buffer| {
+        let mut output = ExternalBufferOutput { buffer, offset: 0 };
+        payload.encode_to(&mut output);
```
ark0f, line 156: Consider use `using_encoded` instead of `ExternBufferOutput`

ark0f (changes requested):

gcore/src/stack_buffer.rs
```diff
@@ -0,0 +1,55 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//! Stack allocations utils.
+
+use alloc::vec;
+
+#[inline(never)]
```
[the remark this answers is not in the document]
grishasobol: Done

gstd/src/msg/encoded.rs
```diff
@@ -113,6 +131,34 @@ pub fn reply<E: Encode>(payload: E, value: u128) -> Result<MessageId> {
...
+/// Buffer size for encoding is at least `E::max_encoded_len()`.
+pub fn reply_on_stack<E: Encode + MaxEncodedLen>(payload: E, value: u128) -> Result<MessageId> {
+    struct ExternalBufferOutput<'a> {
+        buffer: &'a mut [u8],
+        offset: usize,
+    }
+
+    impl<'a> Output for ExternalBufferOutput<'a> {
+        fn write(&mut self, bytes: &[u8]) {
+            let err_log = "Unexpected encoding behavior: too large input bytes size";
```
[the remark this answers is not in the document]
grishasobol: Done

ark0f (approved):

examples/binaries/stack-allocations/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "demo-stack-allocations"
+version = "0.1.0"
+authors.workspace = true
+edition = "2021"
+license = "GPL-3.0"
```
breathx, line 6: workspace

examples/binaries/stack-allocations/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+[dev-dependencies]
+gtest.workspace = true
+rand_pcg = { workspace = true }
+rand = "0.8"
+static_assertions.workspace = true
+
+[build-dependencies]
+gear-wasm-builder.workspace = true
+
+[lib]
```
breathx, line 22:
```suggestion
```

examples/binaries/stack-allocations/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+name = "demo-stack-allocations"
+version = "0.1.0"
+authors.workspace = true
+edition = "2021"
+license = "GPL-3.0"
+workspace = "../../../"
+
+[dependencies]
+gstd.workspace = true
+codec = { package = "parity-scale-codec", version = "3.4.0", default-features = false, features = ["derive"] }
```
breathx, line 11: workspace

breathx: and below

examples/binaries/stack-allocations/src/lib.rs
```diff
@@ -0,0 +1,169 @@
...
+
+    #[test]
+    fn stress() {
+        use Action::*;
+
+        const MAX_ACTIONS_AMOUNT: usize = 1000;
+        const MAX_NUMBER: u8 = 255;
+
+        // Check that check sum is less than u32::MAX
+        const_assert!(
```
breathx, line 121: no reason in const assertions inside function body

gcore/src/stack_buffer.rs
```diff
@@ -0,0 +1,54 @@
...
+    let mut buffer = [0u8; N];
+    let sub_buffer = &mut buffer[0..size];
+    f(sub_buffer)
+}
+
+/// Calls function `f` with provided byte buffer allocated on stack.
+/// ### IMPORTANT
+/// If buffer size is too big (currently bigger than 0x4000 bytes),
+/// then allocation will be on heap.
+/// If buffer is small enough to be allocated on stack, then real allocated
```
breathx, line 33: Why so?

gcore/src/stack_buffer.rs
```diff
@@ -0,0 +1,54 @@
...
+/// buffer size will be `size` aligned to upper power of 2.
+pub fn with_byte_buffer<T>(size: usize, f: impl FnOnce(&mut [u8]) -> T) -> T {
+    match size {
+        size if size <= 0x1 => with_byte_array::<_, 0x1>(size, f),
```
breathx, line 37: Simplify this to just if please
How to ref: https://doc.rust-lang.org/std/primitive.usize.html#method.next_power_of_two

gcore/src/lib.rs
```diff
@@ -67,6 +67,7 @@ pub mod errors;
 pub mod exec;
 pub mod msg;
 pub mod prog;
+pub mod stack_buffer;
```
breathx, line 70: How do you feel about this?
```suggestion
pub mod stack;
```

gstd/src/msg/encoded.rs
```diff
@@ -113,6 +125,34 @@ pub fn reply<E: Encode>(payload: E, value: u128) -> Result<MessageId> {
     super::reply_bytes(payload.encode(), value)
 }
 
+/// Same as [reply], but encodes payload to stack allocated buffer.
+/// Buffer size for encoding is at least `E::max_encoded_len()`.
+pub fn reply_on_stack<E: Encode + MaxEncodedLen>(payload: E, value: u128) -> Result<MessageId> {
```
breathx, line 130: Can't we avoid different methods and append this thing into already existing "reply" that require `Encode` because `encoded_size` must work without extra allocations https://github.com/paritytech/parity-scale-codec/blob/07c7e2c5a34da36684339710d46e2ea02a478dd6/src/codec.rs#L257

breathx: just branch logic depending on encoding size

gstd/src/msg/encoded.rs
```diff
@@ -113,6 +125,34 @@ pub fn reply<E: Encode>(payload: E, value: u128) -> Result<MessageId> {
...
+                panic!("{}", ERROR_LOG);
+            }
+            self.buffer[self.offset..end_offset].copy_from_slice(bytes);
+            self.offset = end_offset;
+        }
+    }
+
+    gcore::stack_buffer::with_byte_buffer(E::max_encoded_len(), |buffer| {
+        let mut output = ExternalBufferOutput { buffer, offset: 0 };
+        payload.encode_to(&mut output);
```
breathx, line 150: for max encoded len we can just encode to buffer, isn't it?

gstd/src/msg/encoded.rs
```diff
@@ -62,6 +62,18 @@ pub fn load<D: Decode>() -> Result<D> {
     D::decode(&mut super::load_bytes()?.as_ref()).map_err(Error::Decode)
 }
 
+/// Same as [`load`](self::load), but reads current message payload to allocated
+/// on stack buffer. Decoded object will be also on stack, but if it contains
+/// any fields, with heap allocations inside (for example vec), then
+/// this decoding may lead to heap allocations.
+pub fn load_on_stack<D: Decode>() -> Result<D> {
+    let wrapper = |read_result: Result<&mut [u8]>| -> Result<D> {
```
breathx, line 70: concat the code pls

breathx (approved):

breathx:
Furthermore, maybe just fail stack_buffer::with_bytes_array in case of overcoming 0x4000 size? 
To avoid implicit movement to heap

gcore/src/stack_buffer.rs
breathx: > Because stack is not a heap and we must restrict too big allocations

My question was about alignment, not limitations

gcore/src/stack_buffer.rs
breathx: [Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=201387c35e4a6203c4fd28c5b9c57929) just for you

gstd/src/msg/encoded.rs
breathx: > The problem is that each type which cannot be represented as MaxEncodedLen requires allocations. So, encoded_size must do additional allocation. We can skip this in future by using special self allocator for self output object.

But docs state other thing :/

gstd/src/msg/encoded.rs
breathx: Anyway can't get why this should be so complicated. Some offset and other stuff

examples/binaries/btree/src/lib.rs
```diff
@@ -52,20 +52,16 @@ mod wasm {
     use super::*;
 
     use alloc::collections::BTreeMap;
-    use gstd::{debug, msg};
+    use gstd::{debug, errors::Result, msg};
```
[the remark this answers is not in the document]
grishasobol: Done

examples/binaries/stack-allocations/Cargo.toml
grishasobol: Done

examples/binaries/stack-allocations/Cargo.toml
grishasobol: Cannot use workspace here - cannot understand why maybe some cargo problems

examples/binaries/stack-allocations/Cargo.toml
breathx: Resolve after pushed commits

examples/binaries/stack-allocations/src/lib.rs
breathx: ??

breathx: Replace with common assert or move above from function body

gstd/src/msg/encoded.rs
grishasobol: done

examples/binaries/stack-allocations/Cargo.toml
grishasobol: OK, this one works.
But codec also should work, cannot understand what the problem with it 
https://github.com/gear-tech/gear/blob/f76565dfb63b3e79ee1949cf3dd9360b2b88af12/gtest/Cargo.toml#L18

gstd/src/msg/encoded.rs
breathx: > However, this can not prevent allocations that some types are doing inside their own encoding.

It may be about custom Encode implementations
Need to check defaults

gcore/src/stack_buffer.rs
breathx: > Like in vector - when buffer is full, then increase buffer size approximately twice.

But should we follow such pattern for the case of const-size pre-allocated space?

examples/binaries/stack-allocations/src/lib.rs
grishasobol: it must be inside function body and that it

gstd/src/msg/encoded.rs
grishasobol: Ok, as I can see this comment is wrong. Actually in most cases this really will be allocation free operation.

grishasobol: I will add issue and TODO

grishasobol: TODO added

PR #2839, opened 2023-06-19
refactor(core-errors)!: New sys-call error representation

### LINKED ISSUE #2791: Refactor errors throwing
### File Location(s)

_No response_

### Proposal

Errors should be matched "concrete syscall error" => "u32"
- [ ] Write such matching
- [ ] Remove `gr_error` syscall
- [ ] Return from fallible syscalls error representation as u32 instead of occurred error len

ark0f:
Resolves #2791.


gcore/src/exec.rs
```diff
@@ -107,9 +106,9 @@ pub fn block_timestamp() -> u64 {
 /// }
 /// ```
 pub fn reply_deposit(message_id: MessageId, amount: u64) -> Result<()> {
-    let mut len = 0u32;
-    unsafe { gsys::gr_reply_deposit(message_id.as_ptr(), amount, &mut len as *mut u32) };
-    SyscallError(len).into_result()
+    let mut error_code = 0u32;
+    unsafe { gsys::gr_reply_deposit(message_id.as_ptr(), amount, &mut error_code as *mut u32) };
```
NikVolf, line 110:
```suggestion
    unsafe { gsys::gr_reply_deposit(message_id.as_ptr(), amount, &mut error_code as _) };
```

gcore/src/exec.rs
```diff
@@ -202,9 +201,9 @@ pub fn reserve_gas(amount: u64, duration: u32) -> Result<ReservationId> {
 ///
 /// - [`reserve_gas`] function reserves gas for further usage.
 pub fn system_reserve_gas(amount: u64) -> Result<()> {
-    let mut len = 0u32;
-    unsafe { gsys::gr_system_reserve_gas(amount, &mut len as *mut u32) };
-    SyscallError(len).into_result()
+    let mut error_code = 0u32;
```
grishasobol, line 204: As I can see in into_result implementation - we translate `0` code as Ok(()). But `0` is set by default in all cases. So, if by some reason node will not be able to write in `&error_code` memory chunk and by some unknown reasons program execution will not be stopped, then we will interpret this result as Ok(()). So, I think better to use some other default value.

core-errors/src/lib.rs
```diff
@@ -26,277 +26,354 @@ extern crate alloc;
...
+)]
 #[non_exhaustive]
-#[repr(u8)]
-pub enum MessageError {
-    /// Message has bigger then allowed one message size
-    #[display(fmt = "Max message size exceed")]
-    MaxMessageSizeExceed = 0,
+#[repr(u32)]
+pub enum ExecutionError {
+    /// An error occurs in attempt to wait duration greater than could be paid.
```
grishasobol, line 48: Is it right comment and info string? I can see this error may occur in many another cases

utils/wasm-proc/metadata-js/index.js
```diff
@@ -48,7 +48,6 @@ exports.getWasmMetadata = async (wasmBytes) => {
             gr_wait_for: () => { },
             gr_wait_up_to: () => { },
             gr_wake: () => { },
-            gr_error: () => { },
```
gshep, line 51: hmmm it doesn't relate to your PR but looks like the list is outdated

core-backend/common/src/funcs.rs
```diff
@@ -477,7 +474,7 @@ where
         gas: u64,
         err_mid_ptr: u32,
     ) -> Result<(), R::Error> {
-        ctx.run_fallible::<_, _, LengthBytes>(err_mid_ptr, RuntimeCosts::ReplyDeposit, |ctx| {
+        ctx.run_fallible::<_, _, ErrorBytes>(err_mid_ptr, RuntimeCosts::ReplyDeposit, |ctx| {
```
SabaunT, line 477: Not related to this line, but `reply_deposit` is not covered with `#[host..]` macro. Why?

gcore/src/exec.rs
SabaunT: isn't it a memory error, if you can't mutate error code?

gcore/src/exec.rs
```diff
@@ -107,9 +106,9 @@ pub fn block_timestamp() -> u64 {
 /// }
 /// ```
 pub fn reply_deposit(message_id: MessageId, amount: u64) -> Result<()> {
-    let mut len = 0u32;
-    unsafe { gsys::gr_reply_deposit(message_id.as_ptr(), amount, &mut len as *mut u32) };
-    SyscallError(len).into_result()
+    let mut error_code = 0u32;
```
SabaunT, line 109: Shouldn't it be `u32::MAX`?

`SyscallError` thinks that the result is successful only if it's `u32::MAX`. Also `reply_deposit` doesn't mutate value to `u32::MAX` if the execution was successful, as far as I see.

SabaunT (changes requested):
I guess @grishasobol 's suggestion to use another value to indicate a successful execution is good, but should be done in the following PR, as it's really hard to check it in the review of the current PR. Too much context to check.

gcore/src/exec.rs
NikVolf: 0 is well-esablished success code, please keep it as it is

gcore/src/exec.rs
NikVolf: No, 0 is used as a success code everywhere

NikVolf:
> I guess @grishasobol 's suggestion to use another value to indicate a successful execution is good, but should be done in the following PR, as it's really hard to check it in the review of the current PR. Too much context to check.

@SabaunT if no other suggestions, please re-review

gshep (approved):
lgtm! 👍

gcore/src/exec.rs
NikVolf: Shouldn't some test fail if it is incorrect?

gcore/src/errors.rs
```diff
@@ -34,8 +34,9 @@ pub struct SyscallError(pub(crate) u32);
 impl From<SyscallError> for Result<()> {
     fn from(value: SyscallError) -> Self {
         match value.0 {
-            0 => Ok(()),
-            _ => Err(value.get_err()?),
+            gsys::SUCCESS_ERROR_CODE => Ok(()),
+            code => Err(ExtError::decode(code)
```
breathx, line 38: same avoid "decode" here 
```suggestion
            code => Err(ExtError::decode(code)
```

breathx (approved):

gcore/src/exec.rs
SabaunT: didn't recognize that
https://github.com/gear-tech/gear/blob/623adb02b5d698b9b2d3597f70fa4cf9cb366935/gsys/src/lib.rs#L187
everything is ok

SabaunT (approved):

core-errors/src/lib.rs
```diff
@@ -26,277 +26,359 @@ extern crate alloc;
...
+            601 => Some(ProgramRentError::MaximumBlockCountPaid.into()),
+            //
+            0xffff => Some(ExtError::SyscallUsage),
+            _ => None,
+        }
+    }
+}
+
+#[cfg(feature = "codec")]
+impl Encode for ExtError {
```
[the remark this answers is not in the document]
ark0f: This bound is used in `with_string_payload` so user can get string

gsys/src/lib.rs
```diff
@@ -22,6 +22,14 @@
 
 use core::mem;
 
+/// Represents error code type.
+pub type ErrorCode = u32;
+
+/// Successful error code.
+///
+/// Set by default during error decoding to check sys-call implementation is sound.
+pub const SUCCESS_ERROR_CODE: ErrorCode = u32::MAX;
```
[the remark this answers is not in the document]
ark0f: Yes

PR #2843, opened 2023-06-20
refactor(gstd, galloc): Add ability to disable panic handler and/or global allocator

### LINKED ISSUE #2842: Benchmarks compilation fails if `gstd::ActorId` is used
### File Location(s)

examples/binaries/sys-calls/src/lib.rs

### Proposal

panic handler and global allocator should be gated by the corresponding feature(s).

the issue happens because both `sp-io` and `gstd` defines panic handler:
```
  error[E0152]: duplicate lang item in crate `gstd` (which `test_syscalls` depen
ds on): `panic_impl`.
    |
    = note: the lang item is first defined in crate `sp_io` (which `pallet_gear`
 depends on)
    = note: first definition in `sp_io` loaded from /home/user/devel/gear-tech/3
retreat/target/release/wbuild/gear-runtime/target/wasm32-unknown-unknown/release
/deps/libsp_io-9cfe96445d4c86e8.rmeta
    = note: second definition in `gstd` loaded from /home/user/devel/gear-tech/3
retreat/target/release/wbuild/gear-runtime/target/wasm32-unknown-unknown/release
/deps/libgstd-4104552f0716800f.rmeta

  For more information about this error, try `rustc --explain E0152`.
  warning: `pallet-gear` (lib) generated 1 warning
  error: could not compile `pallet-gear` (lib) due to previous error; 1 warning emitted

```

gshep:
Resolves #2842 

@gear-tech/dev

### DIFF
--- a/examples/binaries/sys-calls/src/lib.rs
+++ b/examples/binaries/sys-calls/src/lib.rs
@@ -21,6 +21,7 @@
 extern crate alloc;
 
 use codec::{Decode, Encode};
+use gstd::ActorId;
 
 #[cfg(feature = "wasm-wrapper")]
 mod code {
@@ -28,7 +29,6 @@ mod code {
 }
 
 type MessageId = [u8; 32];
-type ActorId = [u8; 32];
 
 #[cfg(feature = "wasm-wrapper")]
 pub use code::WASM_BINARY_OPT as WASM_BINARY;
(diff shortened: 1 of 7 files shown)


NikVolf (changes requested):
No, if you don't need panic handler, don't use gstd

And negative features are against cargo design

PR #2853, opened 2023-06-21
feat(gstd): optimize calls to const syscalls

### LINKED ISSUE #1467: Optimize calls to const syscalls
`gr_origin`
`gr_program_id`
`gr_size`
`gr_source`
`gr_value`

but caching them should occur per-message

StackOverflowExcept1on:
Resolves #1467

This is proof of concept for caching const syscalls. We probably need to remove the imports: `gr_is_getter_call`, `gr_set_getter_call` from the wasm binary instead of keeping dangling imports like we do now.

@gear-tech/dev

### DIFF
--- a/core-backend/sandbox/src/env.rs
+++ b/core-backend/sandbox/src/env.rs
@@ -260,6 +260,8 @@ where
 
         builder.add_func(Alloc, wrap_common_func!(FuncsHandler::alloc, (1) -> (1)));
         builder.add_func(Free, wrap_common_func!(FuncsHandler::free, (1) -> (1)));
+        builder.add_func(IsGetterCalled, wrap_common_func!(FuncsHandler::forbidden, () -> ()));
+        builder.add_func(SetGetterCalled, wrap_common_func!(FuncsHandler::forbidden, () -> ()));
     }
 }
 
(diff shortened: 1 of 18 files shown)


core-backend/sandbox/src/env.rs
```diff
@@ -260,6 +260,8 @@ where
 
         builder.add_func(Alloc, wrap_common_func!(FuncsHandler::alloc, (1) -> (1)));
         builder.add_func(Free, wrap_common_func!(FuncsHandler::free, (1) -> (1)));
+        builder.add_func(IsGetterCalled, wrap_common_func!(FuncsHandler::forbidden, () -> ()));
```
grishasobol, line 263: Why do we need this sys-calls support on back-end if they were removing on instrumentation stage ?

grishasobol (changes requested):
Blocked, until it's not discussed and synced with other possible implementations.

core-backend/sandbox/src/env.rs
StackOverflowExcept1on: I'm working on that. We probably need to add it to the list of forbidden functions. This causes the error right now: "Not all existing sys-calls were added to the module's env"

gshep:
To tell the truth such kind of caching could be done explicitly on the program side

PR #2857, opened 2023-06-25
chore: inherit version from workspace and bump to v0.2.1

### LINKED ISSUE #2448: Inherit versions from workspace for stable packages
### File Location(s)

_No response_

### Proposal

https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table

For example

```toml
[workspace]
version = "0.1.0"
```

```toml
[package]
name = "gstd"
version.workspace = true
```

ref https://github.com/gear-tech/gear/pull/2343

shamilsan:
Resolves #2448 

@gear-tech/dev

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3510,14 +3510,14 @@ dependencies = [
 
 [[package]]
 name = "galloc"
-version = "0.2.0"
+version = "0.2.1"
 dependencies = [
  "dlmalloc",
 ]
 
 [[package]]
 name = "gcli"
-version = "0.2.0"
+version = "0.2.1"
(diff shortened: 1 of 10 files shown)


StackOverflowExcept1on (approved):

shamilsan:
> we may not have to bump the version in this PR since no logic changes

We have two alternatives: 1) To bump the version right before release; 2) To bump it after the previous release.

I vote for the second option as we can forget to bump it as it was with [`v0.1.5`](https://get.gear.rs/#gear-v0-1-5). If we need to update the version number according to semver, we can change it before release again.

PR #2865, opened 2023-06-26
feat(wasm-builder): remap paths to hide username in debug builds

### LINKED ISSUE #2812: Consider to remove username (full path) from panic message
### Problem to Solve

/ .. /

### Possible Solution

/ .. /

### Notes

_No response_

StackOverflowExcept1on:
Resolves #2812

@gear-tech/dev

### DIFF
--- a/utils/wasm-builder/src/lib.rs
+++ b/utils/wasm-builder/src/lib.rs
@@ -17,10 +17,14 @@
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
 use crate::{cargo_command::CargoCommand, cargo_toolchain::Toolchain, wasm_project::WasmProject};
-use anyhow::Result;
+use anyhow::{Context, Result};
 use gmeta::{Metadata, MetadataRepr};
 use regex::Regex;
-use std::{env, path::Path, process};
+use std::{
+    env,
+    path::{Path, PathBuf},
+    process,
+};
 use wasm_project::ProjectType;
 
 mod builder_error;
@@ -103,6 +107,7 @@ impl WasmBuilder {
         let profile = if profile == "debug" { "dev" } else { profile };
         self.cargo.set_profile(profile.to_string());
         self.cargo.set_features(&self.enabled_features()?);
+        self.cargo.set_paths_to_remap(&self.paths_to_remap()?);
 
         self.cargo.run()?;
         self.wasm_project.postprocess()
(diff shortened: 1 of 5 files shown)


utils/wasm-builder/src/lib.rs
```diff
@@ -159,6 +164,25 @@ impl WasmBuilder {
...
+
+        let cargo_dir = std::env::var_os("CARGO_HOME")
+            .map(PathBuf::from)
+            .context("unable to get cargo home directory")?;
+
+        let cargo_checkouts_dir = cargo_dir.join("git").join("checkouts");
+
+        Ok(vec![
+            (home_dir, "/home"),
+            (project_dir, "/proj"),
```
breathx, line 181: Maybe?
```suggestion
            (project_dir, "/program"),
```

utils/wasm-builder/src/lib.rs
```diff
@@ -159,6 +164,25 @@ impl WasmBuilder {
         }
         Ok(matched_features)
     }
+
+    fn paths_to_remap(&self) -> Result<Vec<(PathBuf, &'static str)>> {
```
breathx, line 168: I'd do this remap for production builds only because with debug features etc it's useful to have ability to tap on real path (e.g. when panic occurred) and your idea goes to the file

breathx (approved):

utils/wasm-builder/src/lib.rs
breathx: Maybe builder flag?

utils/wasm-builder/src/lib.rs
StackOverflowExcept1on: Here we can also consider "code". I tried to give short remaps to reduce the size of the binary file and reduce fee for end-user.

utils/wasm-builder/src/lib.rs
StackOverflowExcept1on: We have `self.wasm_project.profile()`. So, maybe if the smart contract is compiled with the "dev"/"debug" profile, it will have full paths. When you build contract with "release" profile, it will remap the paths.

utils/wasm-builder/src/lib.rs
breathx: > Here we can also consider "code". I tried to give short remaps to reduce the size of the binary file and reduce fee for end-user.

I like "code" one

utils/wasm-builder/src/lib.rs
breathx: > We have `self.wasm_project.profile()`. So, maybe if the smart contract is compiled with the "dev"/"debug" profile, it will have full paths. When you build contract with "release" profile, it will remap the paths.

That is exactly I'm talking about

StackOverflowExcept1on: @breathx should be resolved now

ukint-vs (approved):

utils/wasm-builder/src/lib.rs
```diff
@@ -107,7 +107,9 @@ impl WasmBuilder {
         let profile = if profile == "debug" { "dev" } else { profile };
         self.cargo.set_profile(profile.to_string());
         self.cargo.set_features(&self.enabled_features()?);
-        self.cargo.set_paths_to_remap(&self.paths_to_remap()?);
+        if profile == "release" {
```
breathx, line 110: I'd go with `profile != "dev"` because contracts may introduce custom profiles for production as we did in core

breathx (approved):

PR #2878, opened 2023-06-28
ci(pr): introduce new workflow for PR

### LINKED ISSUE #2875: Trigger CI only when label A0,A2 for label conditions
### Problem to Solve

new problem after #2847, #2851 and #2874 

trigger CI when adding label works fine, but labeling any label ( even not A0, A2 ) can trigger the CI again as well

for example, 

1. when we are ready for review, we label our PR as A0, the CI start working
2. after sometime, somebody add a new label `gotissues`, the CI start rerunning ( this is not expected ) even though there isn't any new commits

### Notes

_No response_

clearloop:
Resolves #2875 

- [x] fix cache: we only have packing cache logic for `push` and unpacking cache logic for `prs`, so no cache are being used. https://github.com/gear-tech/gear/pull/2787#discussion_r1232017502
- [x] introduce `PR.yaml`, which includes logic only works for pull requests
- [x] introduce `labels.yml` ... only works for labels
- [x] introduce `build.yml` shared between `PR.yaml` and `CI.yaml`
- [x] call `build.yml` in `CI.yaml` and remove check steps since they are for prs

**This PR has split the check and build into shared workflows**

```mermaid
---
title: Usages of check and build
---

flowchart LR
PR --> check{check}
CI --> check

CI --> build
PR -->|already-A0-A2| build{build}
Label -->|new-A0| build
```

```mermaid
---
title: Before, CI and PR shared the same workflow with tons of conditions
---
flowchart LR
CI --> Check
CI --> |if-A0-A2-in-pr-or-not-pr| Build
CI --> |if-not-pr| Build-other-targets
```

```mermaid
---
title: After, Both CI and PR have independent workflows
---

flowchart LR
PR --> prc[check]
PR --> status
status --> |already-A0-A2| prb[build]
Label --> |new-A0| prb

CI --> check
CI --> build
CI --> build-other-targets
CI --> ...
```

### Problems described in #2875 with more context

1. once we use `labeled` in `pull_request`, the workflow will be triggered each of time we adding a new label, even though we have have already processed the workflow on the same commit
2. sure we can skip the duplicated or unexpected jobs but the UI will replace the previous result with `skipped` which is not expected

#### Example

1. push commit to the branch without any labels -> CI runs ( expected )
2. add any non-A0,A2 label -> CI re-runs and skip dispatch-ci, UI shows the actions are skipped but not the previous result ( not expected ) 
3. add A0,A2 -> CI re-runs with build jobs ( only the build jobs are expected )
4. remove A0 and add A2 -> CI re-runs with build jobs ( not expected )

@gear-tech/dev


breathx:
>> call build.yml in CI.yaml and remove check steps since they are for prs
that's incorrect
check job should run on master push as well in case of different pr change some code and formatting fails etc

breathx:
Explain current flow.
PR description contains only error explanation, isn't it?

clearloop:
> > call build.yml in CI.yaml and remove check steps since they are for prs
> 
> that's incorrect check job should run on master push as well in case of different pr change some code and formatting fails etc

ahh yeah, messed up by https://github.com/gear-tech/gear/blob/master/.github/workflows/CI.yaml#L50 xd

so mb we can introduce a new `Check.yml` shared between CI and PR


> Explain current flow.
> PR description contains only error explanation, isn't it?

will draw a mermaid flow chart later in the PR desc

breathx:
Rebase master please and don't run check if none appropriate labels checked (or adjust scheme if it's incorrect) Furthermore there are ability to call build for other targets with "force_**platform**" labels that should work for prs as well

clearloop:
> don't run check if none appropriate labels checked (or adjust scheme if it's incorrect)

you mean block the `check` workflow with labels as well? only keep docs? imo clippy and fmt checks are useful for custom commits

```mermaid
flowchart LR
PR --> status
status --> |A0-A2| check
status --> |A0-A2| build
PR --> docs
```

> Furthermore there are ability to call build for other targets with "force_platform" labels that should work for prs as well

yeah, this is easy to apply with the arch in this PR, but mb in another PR

breathx:
> > don't run check if none appropriate labels checked (or adjust scheme if it's incorrect)
> 
> you mean block the `check` workflow with labels as well and only keep docs? imo clippy and fmt checks are useful for custom commits
> 
> ```mermaid
> flowchart LR
> PR --> status
> status --> |A0-A2| check
> status --> |A0-A2| build
> PR --> docs
> ```
> the current flow
> 
> https://github.com/gear-tech/gear/actions/runs/5447177709

yeah, ok, it runs only after status as well
but what for specific platforms with labels? it shouldn't run without check too

clearloop:
> Furthermore, merge master carefully, there are a lot of changes in CI file

I'm about to rewrite this PR with a new PR since this PR is really dirty now

PR #2890, opened 2023-06-29
feat(wasm-builder): Introduce switch for disabling path remapping

DennisInSky:
Path remapping makes compilation under WSL terrible. Cold rebuild, i.e. when there were no changes in any examples, takes 9 min compared to 45 sec without that remapping

@reviewer-or-team


ark0f (approved):

StackOverflowExcept1on:
What version of WSL are you using? You may have an old installation and are not using WSL2.

gshep (approved):
:+1:

shamilsan (approved):

DennisInSky:
> What version of WSL are you using? You may have an old installation and are not using WSL2.

This is for WSL2. As far as I got it from the conversation with you - we have the same issue on the native Linux

PR #2892, opened 2023-07-01
refactor(core-backend): Refactor `MemoryError`

ark0f:


core-backend/common/src/runtime.rs
```diff
@@ -20,9 +20,25 @@
...
 use crate::{
     memory::{MemoryAccessRecorder, MemoryOwner},
-    BackendExternalities, BackendState, TerminationReason,
+    BackendExternalities, BackendExternalitiesError, BackendState, TerminationReason,
 };
 use gear_core::{costs::RuntimeCosts, pages::WasmPage};
+use gear_core_errors::ExtError as FallibleExtError;
+
+#[derive(Debug, Clone, derive_more::From)]
+pub enum RunFallibleError {
```
SabaunT, line 29: I understand why `Run*` is chosen, but the sense is "run fallible syscall". Something related to that must be stated. Maybe it requires changing names in `Runtime` trait

SabaunT: to be considered

core-backend/common/src/memory.rs
```diff
@@ -47,22 +50,52 @@ pub enum ProcessAccessError {
...
+        match self {
+            MemoryAccessError::Memory(MemoryError::AccessOutOfBounds) => {
+                RunFallibleError::FallibleExt(FallibleMemoryError::AccessOutOfBounds.into())
+            }
+            MemoryAccessError::ProcessAccess(ProcessAccessError::OutOfBounds) => {
+                RunFallibleError::FallibleExt(FallibleMemoryError::AccessOutOfBounds.into())
+            }
+            MemoryAccessError::RuntimeBuffer(RuntimeBufferSizeError) => {
+                RunFallibleError::FallibleExt(FallibleMemoryError::RuntimeAllocOutOfBounds.into())
+            }
```
SabaunT, line 97: are they really fallible? I mean these error aren't handled during processing fallible func result. Better map these variants to `unreachable` (if it's unreachable).

SabaunT: @grishasobol 
How shall we treat `RuntimeAllocOutOfBounds` & `AccessOutOfBounds` erros? Are they "fallible"?

SabaunT (changes requested):
I love these changes with `RunFallibleError` :heart: 

Requested changes until clarifications with memory errors will be given

core-backend/common/src/memory.rs
ark0f: > are they really fallible?

Yes. Fallible sys-calls use `read_as` and `read_decoded` methods so they return `RuntimeAllocOutOfBounds & AccessOutOfBounds`.

SabaunT (approved):

core-backend/common/src/memory.rs
grishasobol: All right here for me

core-backend/common/src/lib.rs
```diff
@@ -305,8 +290,12 @@ pub trait BackendExternalities: Externalities + CountersOwner {
 }
 
 /// A trait for conversion of the externalities API error to `TerminationReason`.
-pub trait BackendExternalitiesError: Clone + Sized {
+pub trait BackendExternalitiesError: Sized {
     fn into_termination_reason(self) -> TerminationReason;
+
+    fn into_run_fallible_error(self) -> RunFallibleError {
```
grishasobol, line 296: Remove this default impl please. The code is already very hard to understand

core-backend/common/src/memory.rs
```diff
@@ -47,22 +50,52 @@ pub enum ProcessAccessError {
...
-    GasAllowanceExceeded,
 }
 
-impl From<ProcessAccessError> for MemoryAccessError {
-    fn from(err: ProcessAccessError) -> Self {
-        match err {
-            ProcessAccessError::OutOfBounds => MemoryError::AccessOutOfBounds.into(),
-            ProcessAccessError::GasLimitExceeded => Self::GasLimitExceeded,
-            ProcessAccessError::GasAllowanceExceeded => Self::GasAllowanceExceeded,
+impl BackendExternalitiesError for MemoryAccessError {
```
grishasobol, line 60: If it's implemented for MemoryAccessError then `Externalities` in name is not correct

core-backend/common/src/runtime.rs
```diff
@@ -20,9 +20,25 @@
 
 use crate::{
     memory::{MemoryAccessRecorder, MemoryOwner},
-    BackendExternalities, BackendState, TerminationReason,
+    BackendExternalities, BackendExternalitiesError, BackendState, TerminationReason,
 };
 use gear_core::{costs::RuntimeCosts, pages::WasmPage};
+use gear_core_errors::ExtError as FallibleExtError;
+
+#[derive(Debug, Clone, derive_more::From)]
```
grishasobol, line 28: Also better to remove this derive_mode::From, because this makes code incredibly misunderstandable. Even using IDE does not help to see all deps between errors

grishasobol: Even more may be better to remove `From` for some other errors also, so code with explicit errors names will be bigger but much more human friendly.

ark0f: > Even more may be better to remove From for some other errors also

It's already done in other refactoring PRs

grishasobol (approved):

PR #2897, opened 2023-07-03
feat(gtest): add support of reply code to `Log`

shamilsan:
- Added `Log::auto_reply_builder` to create an empty `Log` entry for auto-reply.
- Added the `Log::reply_code` function to explicitly set the reply code in the `Log` entry.

@gear-tech/dev 

## Release Notes

- **Added a possibility to check auto-replies in `gtest`.**
- Now, it is possible to create `gtest::Log` entries with a custom reply code to check whether we have received an auto-reply or not.


clearloop (approved):

gtest/src/log.rs
```diff
@@ -144,6 +144,14 @@ impl Log {
         log
     }
 
+    /// Create a `Log` with a reply code of `SuccessReplyReason::Auto`.
+    pub fn auto_reply_builder() -> Self {
```
breathx, line 148: auto replies never has payload, better to somehow reflect it

breathx (changes requested):

gtest/src/log.rs
shamilsan: > auto replies never has payload, better to somehow reflect it

How? Add a separate `AutoReplyLog` type? Seems to be overkill.

NikVolf: Why is it overkill?

breathx: At least panic once added payload for such kind of Log

shamilsan: > At least panic once added payload for such kind of Log

Sounds reasonable; I added it.

breathx (changes requested):

gtest/src/log.rs
```diff
@@ -153,6 +161,10 @@ impl Log {
             panic!("Payload was already set for this log");
         }
 
+        if let Some(ReplyCode::Success(SuccessReplyReason::Auto)) = self.reply_code {
```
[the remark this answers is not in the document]
shamilsan: Added `#[track_caller]` to ell functions with `panic!` and `expect`.

gtest/src/log.rs
```diff
@@ -176,6 +188,17 @@ impl Log {
...
         self
     }
+
+    /// Set the reply code for this log.
+    pub fn reply_code(mut self, reply_code: ReplyCode) -> Self {
+        if self.reply_code.is_some() {
+            panic!("Reply code was already set for this log");
+        }
+
+        self.reply_code = Some(reply_code);
```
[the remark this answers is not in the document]
shamilsan: Good catch, thank you! Added a check and panic.

breathx (approved):

PR #2899, opened 2023-07-04
chore(test): Disable path remap in measurements CI

gshep:
This should reduce total execution time

@gear-tech/dev


ukint-vs (approved):

shamilsan (approved):
Great 💯

PR #2900, opened 2023-07-04
fix(wasm-builder): remove limit on generated files

shamilsan:
- Removed a limit to files generated by `gear-wasm-builder` as a 4kb limit for `wasm_binary.rs` is too restrictive.

@gear-tech/dev 

## Release Notes

- **Removed a 4kb limit for files generated by `gear-wasm-builder`**
- This limit caused a failure when building large Gear programs due to big `wasm_binary.rs` auto-generated file. Now there are no limits to making great dApps.

### DIFF
--- a/utils/wasm-builder/src/smart_fs.rs
+++ b/utils/wasm-builder/src/smart_fs.rs
@@ -24,8 +24,6 @@ use anyhow::Result;
 use gmeta::MetadataRepr;
 use std::{fs, io::ErrorKind, path::Path};
 
-const LINEAR_COMPARISON_FILE_SIZE: u64 = 4096;
-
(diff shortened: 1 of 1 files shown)


utils/wasm-builder/src/smart_fs.rs
```diff
@@ -60,9 +58,6 @@ fn check_changed(path: &Path, contents: &[u8]) -> Result<bool> {
         return Ok(true);
     }
 
-    // gear-wasm-builder doesn't write such big files
```
shamilsan, line 63: gear-wasm-builder writes such big files, trust me

ukint-vs (approved):

grishasobol (approved):

PR #2902, opened 2023-07-04
perf(wasm-builder): invert path remapping logic

StackOverflowExcept1on:
The ecosystem team can use the "production" profile to remove the username from the wasm binary.

### DIFF
--- a/utils/wasm-builder/src/lib.rs
+++ b/utils/wasm-builder/src/lib.rs
@@ -107,7 +107,7 @@ impl WasmBuilder {
         let profile = if profile == "debug" { "dev" } else { profile };
         self.cargo.set_profile(profile.to_string());
         self.cargo.set_features(&self.enabled_features()?);
-        if env::var("__GEAR_WASM_BUILDER_NO_PATH_REMAP").is_err() && profile != "dev" {
+        if env::var("GEAR_WASM_BUILDER_PATH_REMAPPING").is_ok() {
             self.cargo.set_paths_to_remap(&self.paths_to_remap()?);
         }
 
--- a/.github/workflows/test-measurements.yaml
+++ b/.github/workflows/test-measurements.yaml
@@ -11,7 +11,6 @@ env:
   TERM: xterm-256color
   COUNT: 100
   BINARYEN_VERSION: version_111
-  __GEAR_WASM_BUILDER_NO_PATH_REMAP: 1
 
 jobs:
   build:


utils/wasm-builder/src/lib.rs
```diff
@@ -167,6 +177,25 @@ impl WasmBuilder {
         Ok(matched_features)
     }
 
+    fn need_remap_path(&self, enabled_features: &[String]) -> bool {
```
DennisInSky, line 180:
```suggestion
    fn need_path_remap(&self, enabled_features: &[String]) -> bool {
```

utils/wasm-builder/src/lib.rs
```diff
@@ -167,6 +177,25 @@ impl WasmBuilder {
...
+            .iter()
+            .filter_map(|feature| crate_info.features.get(feature))
+            .flatten()
+            .any(|feature| feature == "gstd/debug");
+
+        let debug_in_dependencies = crate_info.dependencies.iter().any(|dependency| {
+            dependency.kind == DependencyKind::Normal
+                && dependency.name == "gstd"
+                && dependency.features.contains(&String::from("debug"))
+        });
```
DennisInSky, line 194: Could it be the `debug` feature is enabled via a transitive dependency like
`dapp` crate depends on some `dapp-lib` crate which in turn depends on `gstd` with the `debug` feature on

utils/wasm-builder/src/wasm_project.rs
```diff
@@ -115,6 +115,11 @@ impl WasmProject {
         }
     }
 
+    /// Return the path to the temporary generated `Cargo.toml`.
```
DennisInSky, line 118: The comment has been copy-pasted from the below method

utils/wasm-builder/src/lib.rs
```diff
@@ -167,6 +177,25 @@ impl WasmBuilder {
...
     }
 
+    fn need_path_remap(&self, enabled_features: &[String]) -> bool {
+        let crate_info = CrateInfo::from_manifest(&self.wasm_project.original_manifest_path())
+            .expect("Infallible");
+
+        let debug_in_features = enabled_features
+            .iter()
+            .filter_map(|feature| crate_info.features.get(feature))
+            .flatten()
```
DennisInSky, line 187: Btw, there is the same question about transitive features

breathx (changes requested):
Apply fixes for @DennisInSky comments

StackOverflowExcept1on:
@breathx #2915 will add a panic handler feature to gstd. It makes sense to enable path remapping only when the panic handler is enabled.

If the user uses dependency (such as dapps-gear-lib) that references gstd, the user should disable the panic handler. Otherwise, such dependency will define the panic handler twice, and we will consider it semantically incorrect (also it leads to compile error).

Does it make sense to check transitive dependencies for the `gstd/debug` function in this case?

utils/wasm-builder/src/lib.rs
```diff
@@ -107,7 +107,7 @@ impl WasmBuilder {
         let profile = if profile == "debug" { "dev" } else { profile };
         self.cargo.set_profile(profile.to_string());
         self.cargo.set_features(&self.enabled_features()?);
-        if env::var("__GEAR_WASM_BUILDER_NO_PATH_REMAP").is_err() && profile != "dev" {
+        if env::var("__GEAR_WASM_BUILDER_NO_PATH_REMAP").is_err() && profile == "production" {
```
DennisInSky, line 110: This more resembles exlicit feature and has less flexibility then a dedicated feature. If it is the only option we can suggest to our customers, I would suggest introducing the latter, so they can decide themselves whether and when they want the paths to be remapped.

breathx (approved):

DennisInSky (approved):

PR #2912, opened 2023-07-06
feat(gstd): Introduce max ownership time for Mutex

DennisInSky:
This is an idea of supporting max ownership time for Mutex (the same will be implemented for RwLock if everyone happy with this approach). Basically, before acquiring the mutex ownership, one can declare in the code max number of blocks they are going to own the mutex for. If there happens another attempt to grab the same mutex after the specified number of blocks, but the mutex has not been released by the first owner yet, the first message execution will get terminated via panic indicating the message exceeded lock ownership time.

@reviewer-or-team


gstd/src/lock/mutex.rs
```diff
@@ -193,18 +213,55 @@ impl<'a, T> Future for MutexLockFuture<'a, T> {
...
-        let current_msg_id = crate::msg::id();
-        let lock = unsafe { &mut *self.mutex.locked.get() };
-        if lock.is_none() {
-            *lock = Some(current_msg_id);
+        let current_msg_id = msg::id();
+        let current_block_number = exec::block_height();
+        let locked_by = unsafe { &mut *self.mutex.locked.get() };
+        if let Some((lock_owner_msg_id, deadline_block_number)) = *locked_by {
+            if current_block_number > deadline_block_number {
+                if let Some(msg_future_task) = async_runtime::futures().get_mut(&lock_owner_msg_id)
```
DennisInSky, line 221: I do not really like this dependency on the `async_runtime` here. Probably the better option would be having a separate list of ousted messages inside the `Mutex` struct. Every ousted message gets moved into there and awoken. During the `poll` the future can check its message in that list and panic here instead of the main message loop. Any thoughts?

gstd/src/lock/mutex.rs
```diff
@@ -217,6 +231,46 @@ unsafe impl<T> Sync for Mutex<T> {}
...
+            })
+        }
+    }
+
+    fn acquire_lock_ownership(
+        &mut self,
+        owner_msg_id: MessageId,
+        current_block: BlockNumber,
+    ) -> Poll<MutexGuard<'a, T>> {
+        let locked_by = unsafe { &mut *self.mutex.locked.get() };
```
breathx, line 256: why we aren't satisfied with `get_mut()` fn call?

breathx: and anywhere else

gstd/src/lock/mutex.rs
```diff
@@ -226,22 +280,36 @@ impl<'a, T> Future for MutexLockFuture<'a, T> {
...
-            // and just go back to the waitlist.
-            if !self.mutex.queue.contains(&current_msg_id) {
-                self.mutex.queue.enqueue(current_msg_id);
+        let current_block = exec::block_height();
+        let locked_by = unsafe { &mut *self.mutex.locked.get() };
+        if let Some((lock_owner_msg_id, deadline_block)) = *locked_by {
+            if current_block >= deadline_block {
+                if let Some(msg_future_task) = async_runtime::futures().get_mut(&lock_owner_msg_id)
+                {
+                    msg_future_task.set_lock_exceeded();
```
breathx, line 289: what if I poll my own caught lock?

breathx (changes requested):

pallets/gear/src/tests.rs
```diff
@@ -8669,7 +8669,7 @@ fn waking_message_waiting_for_mx_lock_does_not_lead_to_deadlock() {
         Command as WaiterCommand, LockContinuation, MxLockContinuation, WASM_BINARY as WAITER_WASM,
     };
 
-    let execution = || {
+    fn execution() {
```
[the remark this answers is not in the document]
DennisInSky: May I answer with a question - why is using the closure better than using a local function?

gstd/src/lock/mutex.rs
```diff
@@ -217,6 +231,46 @@ unsafe impl<T> Sync for Mutex<T> {}
...
+}
+
+impl<'a, T> MutexLockFuture<'a, T> {
+    /// Sets the maximum number of blocks the mutex lock can be owned by
+    /// some message before the ownership can be seized by another rival
+    pub fn own_up_for(self, block_count: BlockCount) -> Result<Self> {
+        if block_count == 0 {
+            Err(Error::ZeroMxLockDuration)
+        } else {
+            Ok(MutexLockFuture {
```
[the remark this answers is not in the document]
DennisInSky: To my point of view it is not, as this one more about the builder pattern

gstd/src/lock/mutex.rs
DennisInSky: We can't use `get_mut` as the `mutex` ref field in the `MutexLockFututre` structure is non-mutable. Making it so will break the logic around the `Mutex.lock` method which borrows mutex as non-mutable

gstd/src/lock/mutex.rs
DennisInSky: Can you please provide more details? I do not follow your question

breathx: let lock = Mutex::lock()
lock.poll(); lock.poll() (not with await, but manually by pinning future and providing context)

clearloop (approved):
I'm considering about providing a handbook for program developers since we are introducing more and more features now, the `Learn Gear "std" library` redirects to https://docs.gear.rs/gstd/ (plus https://wiki.gear-tech.io/docs/getting-started-in-5-minutes/) are obvious not friendly enough for new developers

gstd/src/lock/mutex.rs
DennisInSky: Tricky question for me as I am too new to rust. What might be the point of doing polling manually instead of relying on the async runtime? Will all the other stuff (all the futures) work if one decides to go with manul polling?

gstd/src/lock/mutex.rs
```diff
@@ -102,7 +106,10 @@ impl<T> Mutex<T> {
     /// returned to allow scoped unlock of the lock. When the guard goes out
     /// of scope, the mutex will be unlocked.
     pub fn lock(&self) -> MutexLockFuture<'_, T> {
-        MutexLockFuture { mutex: self }
+        MutexLockFuture {
+            mutex: self,
+            own_up_for: None,
```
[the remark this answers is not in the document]
DennisInSky: This is a number of blocks to own the mutex so it is `own up for`

gstd/src/lock/mutex.rs
breathx: Others will work. Furthermore we have test that checks exactly these sync primitives in manual. That's pretty similar to consequent awaits

breathx: anyway it could be checked other time. better to merge current pr already

PR #2916, opened 2023-07-07
fix(gstd): Allow mutex/read/write lock guard access by holder only

DennisInSky:
At the moment it is allowed to obtain mx/read/write lock guard in the context of one message and use it in the context of another which is wrong. This PR fixes this issue.

@reviewer-or-team

### DIFF
--- a/examples/binaries/waiter/Cargo.toml
+++ b/examples/binaries/waiter/Cargo.toml
@@ -15,6 +15,9 @@ gstd.workspace = true
 gear-wasm-builder.workspace = true
 
 [dev-dependencies]
+gtest.workspace = true
+gear-core.workspace = true
+demo-waiter = { path = ".", features = ["debug"] }
 
 [lib]
 
--- a/examples/binaries/waiter/src/lib.rs
+++ b/examples/binaries/waiter/src/lib.rs
@@ -56,14 +56,20 @@ pub enum SleepForWaitType {
 }
 
 #[derive(Debug, Encode, Decode)]
-pub enum MxLockContinuation {
+pub enum LockContinuation {
     Nothing,
     SleepFor(u32),
     MoveToStatic,
 }
 
 #[derive(Debug, Encode, Decode)]
-pub enum MxLockStaticAccessSubcommand {
+pub enum MxLockContinuation {
+    // Here will be Lock
+    General(LockContinuation),
+}
+
+#[derive(Debug, Encode, Decode)]
+pub enum LockStaticAccessSubcommand {
     Drop,
     AsRef,
     AsMut,
@@ -79,8 +85,8 @@ pub enum RwLockType {
 
 #[derive(Debug, Encode, Decode)]
 pub enum RwLockContinuation {
-    Nothing,
-    SleepFor(u32),
+    // Here will be Lock(RwLockType)
+    General(LockContinuation),
 }
 
 #[derive(Debug, Encode, Decode)]
@@ -94,6 +100,7 @@ pub enum Command {
     SleepFor(Vec<u32>, SleepForWaitType),
     WakeUp([u8; 32]),
     MxLock(MxLockContinuation),
-    MxLockStaticAccess(MxLockStaticAccessSubcommand),
+    MxLockStaticAccess(LockStaticAccessSubcommand),
     RwLock(RwLockType, RwLockContinuation),
+    RwLockStaticAccess(RwLockType, LockStaticAccessSubcommand),
 }
--- a/examples/binaries/waiter/src/lib.rs
+++ b/examples/binaries/waiter/src/lib.rs
@@ -77,7 +77,7 @@ pub enum LockStaticAccessSubcommand {
     DerefMut,
 }
 
-#[derive(Debug, Encode, Decode)]
+#[derive(Debug, Encode, Decode, Clone, Copy)]
 pub enum RwLockType {
     Read,
     Write,
--- a/examples/binaries/waiter/src/lib.rs
+++ b/examples/binaries/waiter/src/lib.rs
@@ -59,6 +59,16 @@ pub enum SleepForWaitType {
 pub enum MxLockContinuation {
     Nothing,
     SleepFor(u32),
+    MoveToStatic,
+}
+
+#[derive(Debug, Encode, Decode)]
+pub enum MxLockStaticAccessSubcommand {
+    Drop,
+    AsRef,
+    AsMut,
+    Deref,
+    DerefMut,
 }
 
 #[derive(Debug, Encode, Decode)]
@@ -84,5 +94,6 @@ pub enum Command {
     SleepFor(Vec<u32>, SleepForWaitType),
     WakeUp([u8; 32]),
     MxLock(MxLockContinuation),
+    MxLockStaticAccess(MxLockStaticAccessSubcommand),
     RwLock(RwLockType, RwLockContinuation),
 }
--- a/examples/binaries/waiter/tests/mx_lock_access.rs
+++ b/examples/binaries/waiter/tests/mx_lock_access.rs
@@ -50,7 +50,7 @@ fn access_mx_lock_guard_from_different_msg_fails(
 }
 
 fn init_fixture(system: &System) -> (Program<'_>, MessageId) {
-    system.init_logger_with_fallback_filter("");
+    system.init_logger_with_default_filter("");
     let program = Program::current(system);
     program.send_bytes(USER_ID, []);
     let lock_result = program.send(
--- a/examples/binaries/waiter/tests/mx_lock_access.rs
+++ b/examples/binaries/waiter/tests/mx_lock_access.rs
@@ -1,0 +1,132 @@
+use demo_waiter::{Command, MxLockContinuation, MxLockStaticAccessSubcommand};
+use gear_core::ids::MessageId;
+use gstd::errors::{ErrorReplyReason, ReplyCode, SimpleExecutionError};
+use gtest::{Program, RunResult, System};
+
+const USER_ID: u64 = 10;
+
+#[test]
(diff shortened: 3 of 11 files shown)


gstd/src/lock/mutex.rs
```diff
@@ -117,27 +116,59 @@ impl<T> Mutex<T> {
...
 impl<'a, T> Drop for MutexGuard<'a, T> {
     fn drop(&mut self) {
+        self.ensure_access_by_holder();
         unsafe {
-            *self.mutex.locked.get() = None;
-            if let Some(message_id) = self.mutex.queue.dequeue() {
-                crate::exec::wake(message_id).expect("Failed to wake the message");
+            let locked_by = &mut *self.mutex.locked.get();
+            if let Some(owner_msg_id) = *locked_by {
+                if owner_msg_id != self.holder_msg_id {
```
breathx, line 141: Seems you already have such check 4 lines above

gstd/src/lock/rwlock.rs
```diff
@@ -198,52 +214,87 @@ impl<T> Deref for RwLockReadGuard<'_, T> {
...
     fn drop(&mut self) {
+        self.ensure_access_by_holder();
         unsafe {
-            *self.lock.locked.get() = None;
-            if let Some(message_id) = self.lock.queue.dequeue() {
-                crate::exec::wake(message_id).expect("Failed to wake the message");
+            let locked_by = &mut *self.lock.locked.get();
+            if let Some(owner_msg_id) = *locked_by {
+                if owner_msg_id != self.holder_msg_id {
+                    panic!(
```
breathx, line 252: same

examples/binaries/waiter/tests/utils.rs
```diff
@@ -0,0 +1,22 @@
+use gstd::errors::{ErrorReplyReason, ReplyCode, SimpleExecutionError};
+use gtest::RunResult;
+
+pub const USER_ID: u64 = 10;
+
+#[track_caller]
+pub fn assert_paniced(result: &RunResult, panic_msg: &str) {
```
breathx, line 7:
```suggestion
pub fn assert_panicked(result: &RunResult, panic_msg: &str) {
```

examples/binaries/waiter/tests/utils.rs
```diff
@@ -0,0 +1,22 @@
...
+        result.log()[0].reply_code(),
+        Some(ReplyCode::Error(ErrorReplyReason::Execution(
+            SimpleExecutionError::UserspacePanic
+        )))
+    ));
+    let payload = String::from_utf8(result.log()[0].payload().into())
+        .expect("Unable to decode panic message")
+        .split(',')
+        .map(String::from)
+        .next()
```
breathx, line 19:
```suggestion
        .first()
        .map(String::from)
```

examples/binaries/waiter/tests/rw_lock_access.rs
```diff
@@ -0,0 +1,182 @@
...
+            lock_msg_id,
+            lock_access_result.sent_message_id()
+        ),
+    );
+}
+
+#[test]
+fn deref_mut_w_lock_guard_from_different_msg_fails() {
+    let system = System::new();
+    let (program, lock_msg_id) = init_fixture(&system, RwLockType::Write);
```
breathx, line 153: all the file seems so copy-pasty so I'd introduce common fn for them and reuse for each tests with 2 args: Command and panic message

examples/binaries/waiter/tests/rw_lock_access.rs
```diff
@@ -0,0 +1,182 @@
...
+        &format!(
+            "Write lock guard held by message {} is being accessed by message {}",
+            lock_msg_id,
+            lock_access_result.sent_message_id()
+        ),
+    );
+}
+
+fn init_fixture(system: &System, lock_type: RwLockType) -> (Program<'_>, MessageId) {
+    system.init_logger_with_default_filter("");
```
breathx, line 171: why?

examples/binaries/waiter/tests/mx_lock_access.rs
```diff
@@ -0,0 +1,117 @@
+use demo_waiter::{Command, LockContinuation, LockStaticAccessSubcommand, MxLockContinuation};
```
breathx, line 1: same for the file

examples/binaries/waiter/src/lib.rs
```diff
@@ -69,8 +85,8 @@ pub enum RwLockType {
 
 #[derive(Debug, Encode, Decode)]
 pub enum RwLockContinuation {
```
breathx, line 87: What's the purpose of this to be enum?

examples/binaries/waiter/src/lib.rs
```diff
@@ -56,9 +56,25 @@ pub enum SleepForWaitType {
...
 #[derive(Debug, Encode, Decode)]
-pub enum MxLockContinuation {
+pub enum LockContinuation {
     Nothing,
     SleepFor(u32),
+    MoveToStatic,
+}
+
+#[derive(Debug, Encode, Decode)]
+pub enum MxLockContinuation {
```
breathx, line 66: same here

examples/binaries/waiter/Cargo.toml
```diff
@@ -15,6 +15,9 @@ gstd.workspace = true
 gear-wasm-builder.workspace = true
 
 [dev-dependencies]
+gtest.workspace = true
+gear-core.workspace = true
```
breathx, line 19: add issue to remove gear-core from here. its irrational to import/link/build crate here just for 1 wrapper across [u8; 32]

breathx (changes requested):
Minor changes from comments

gstd/src/lock/mutex.rs
DennisInSky: These are different checkes:
- the first one checks the current message is the guard holder
- the second one checks the current message is the lock owner
in total it checks the guard holder is the lock owner

examples/binaries/waiter/tests/utils.rs
DennisInSky: .first()
   | |         -^^^^^ method not found in `Split<'_, char>`

examples/binaries/waiter/tests/rw_lock_access.rs
DennisInSky: For the testing purposes I build the contract with the `debug` feature on which enables log messages. It has turned out, by default `system.init_logger()` enables all debug messages, but I am not keen on seeing them in tests

examples/binaries/waiter/src/lib.rs
DennisInSky: If you are asking about why I have introduced this wrapping enum, then the answer is that I know that I will need it for the tests I have for the mutex ownership, where for the mutex I will have a continuation like `Lock`, whereas for the RW case there will be RW specific continuation like `Lock(RwLockType)`

examples/binaries/waiter/Cargo.toml
DennisInSky: Actually this is not about a wrapper, it is about the `MessageId` returned by the `RunResult::sent_message_id()` method. Should we instead add an issue that the method should return the `MessageId` type from the `gstd` crate instead?

examples/binaries/waiter/tests/rw_lock_access.rs
DennisInSky: Nice catch. Shame on me I haven't noticed it myself. Addressed

examples/binaries/waiter/Cargo.toml
breathx: > Actually this is not about a wrapper, it is about the `MessageId` returned by the `RunResult::sent_message_id()` method. Should we instead add an issue that the method should return the `MessageId` type from the `gstd` crate instead?

maybe after introducing gstd-primitives crate

breathx (approved):

gtest/src/system.rs
```diff
@@ -42,7 +42,11 @@ impl System {
     }
 
     pub fn init_logger(&self) {
-        let _ = Builder::from_env(Env::default().default_filter_or("gwasm=debug"))
+        self.init_logger_with_default_filter("gwasm=debug")
+    }
+
+    pub fn init_logger_with_default_filter<'a>(&self, default_filter: impl Into<Cow<'a, str>>) {
```
shamilsan, line 48: Isn't it just `filter`, not default?
```suggestion
    pub fn init_logger_with_filter<'a>(&self, filter: impl Into<Cow<'a, str>>) {
```

gstd/src/lock/mutex.rs
```diff
@@ -117,27 +116,59 @@ impl<T> Mutex<T> {
...
 
 impl<'a, T> Drop for MutexGuard<'a, T> {
     fn drop(&mut self) {
+        self.ensure_access_by_holder();
         unsafe {
-            *self.mutex.locked.get() = None;
-            if let Some(message_id) = self.mutex.queue.dequeue() {
-                crate::exec::wake(message_id).expect("Failed to wake the message");
+            let locked_by = &mut *self.mutex.locked.get();
+            if let Some(owner_msg_id) = *locked_by {
```
shamilsan, line 140: As for me `if let Some(a) = b { ... } else { panic!("...") }` can be replaced by `a = b.unwrap_or_else(|| panic!("..."))`

shamilsan (approved):
Awesome 🔥

gtest/src/system.rs
DennisInSky: Not sure this is a proper name for the func. This filter is taken into account only in cases when it is not specified by other means. For the underlaying method this param is called `default_filter`.
What do you think about naming it `init_logger_with_fallback_filter`?

PR #2917, opened 2023-07-09
ci: Speed up tests by giving correct arguments

breathx:


shamilsan (approved):

scripts/src/test.sh
```diff
@@ -41,25 +41,39 @@ workspace_test() {
 }
 
 gsdk_test() {
-  $CARGO test -p gsdk
-  $CARGO test -p gsdk --features vara-testing
+  if [ "$CARGO" = "cargo xwin" ]; then
+    $CARGO test -p gsdk "$@"
```
ark0f, line 45:
```suggestion
    $CARGO test -p gsdk --no-fail-fast "$@"
```

ark0f (approved):

scripts/src/test.sh
breathx: no fail fast is nextest argument

PR #2919, opened 2023-07-10
test(core-processor): Add tests for send message limit

mqxf:
- Added a test for the limit of message `send_init` calls.
- Also checks that the `send_commit`, `send_push` and `send_push_input` functions do not error.
- Added helper methods on `ProcessorContextBuilder` for message config.

### DIFF
--- a/core-processor/src/ext.rs
+++ b/core-processor/src/ext.rs
@@ -1036,7 +1036,7 @@ mod tests {
     use super::*;
     use alloc::vec;
     use gear_core::{
-        message::{ContextSettings, IncomingDispatch, Payload, MAX_PAYLOAD_SIZE},
+        message::{ContextSettings, IncomingDispatch, MAX_PAYLOAD_SIZE},
         pages::PageNumber,
     };
 
@@ -1279,6 +1279,8 @@ mod tests {
                 .build(),
         );
 
+        let data = HandlePacket::default();
+
         let fake_handle = 0;
 
         let msg = ext.send_commit(fake_handle, data, 0);
@@ -1286,8 +1288,6 @@ mod tests {
             msg.unwrap_err(),
             FallibleExtError::Core(FallibleExtErrorCore::Message(MessageError::OutOfBounds))
         );
-
-        let data = HandlePacket::default();
 
         let handle = ext.send_init().unwrap();
 
--- a/core-processor/src/ext.rs
+++ b/core-processor/src/ext.rs
@@ -1379,6 +1379,69 @@ mod tests {
         assert_eq!(dispatch.message().payload_bytes(), &[1, 2, 3, 4, 5, 6]);
     }
 
+    #[test]
+    /// This function tests:
+    ///
+    /// - `send_push_input` on non-existent handle
+    /// - `send_push_input` on valid handle
+    /// - `send_push_input` on used handle
+    /// - `send_push_input` data is added to buffer
+    ///
+    fn test_send_push_input() {
+        let mut ext = Ext::new(
+            ProcessorContextBuilder::new()
+                .with_message_context(
+                    MessageContextBuilder::new()
+                        .with_outgoing_limit(u32::MAX)
+                        .build(),
+                )
+                .build(),
+        );
+
+        let data = HandlePacket::default();
+
+        let fake_handle = 0;
+
+        let res = ext.send_push_input(fake_handle, 0, 1);
+        assert_eq!(
+            res.unwrap_err(),
+            FallibleExtError::Core(FallibleExtErrorCore::Message(MessageError::OutOfBounds))
+        );
+
+        let handle = ext.send_init().unwrap();
+
+        let res = ext
+            .context
+            .message_context
+            .payload_mut()
+            .try_extend_from_slice(&[1, 2, 3, 4, 5, 6]);
+        assert!(res.is_ok());
+
+        let res = ext.send_push_input(handle, 2, 3);
+        assert!(res.is_ok());
+
+        let msg = ext.send_commit(handle, data, 0);
+        assert!(msg.is_ok());
+
+        let res = ext.send_push_input(handle, 0, 1);
+        assert_eq!(
+            res.unwrap_err(),
+            FallibleExtError::Core(FallibleExtErrorCore::Message(MessageError::LateAccess))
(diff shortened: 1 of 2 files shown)


core/src/message/context.rs
```diff
@@ -78,30 +78,61 @@ impl ContextSettings {
         self.sending_fee
     }
 
+    /// Setter for inner sending fee field.
+    pub fn set_sending_fee(&mut self, fee: u64) {
```
SabaunT, line 82: these new `set_*` methods violate aims and expected usage of context objects. Either avoid that by instantiating contexts manually or with builder in the *ext.rs*, or move this new api to some separate `impl` expression and mark it with `#[cfg(test)]` (hope it works as I think).

core-processor/src/ext.rs
```diff
@@ -1211,6 +1248,38 @@ mod tests {
...
+        let handle = ext.send_init();
+        assert!(handle.is_ok());
+
+        let msg = ext.send_push(handle.unwrap(), &[]);
+        assert!(msg.is_ok());
+
+        let handle = ext.send_init();
+        assert!(handle.is_ok());
+
+        let msg = ext.send_push_input(handle.unwrap(), 0, 0);
```
SabaunT, line 1276: assert is ok?

SabaunT (dismissed):
Please add more tests, not just checking limits

breathx (dismissed):
@SabaunT please specify exact logic you want to be tested
And consider here possibility to test not only "low-level" ext, but the `process()` function and its outcome as well

core/src/message/context.rs
mqxf: Removed the setters

core-processor/src/ext.rs
mqxf: Replaced with unwrapping, better aswell since it will show what error happened

breathx: > Replaced with unwrapping, better aswell since it will show what error happened

It's better to avoid unwraps even for tests
Expect does the same but can provide some (even short) explanation for other devs what went wrong within the test.

mqxf:
I have added more tests, separated them into multiple functions and removed the `settings_mut` and other breaking issues.

mqxf:
I pushed the code again after running `cargo fmt` so that it will pass the check test (since that seemed to be what was failing).

SabaunT (commented):
There must be a separate test for each case from cases that we have discussed privately.

core-processor/src/ext.rs
```diff
@@ -1211,6 +1264,184 @@ mod tests {
...
     }
 
+    #[test]
+    /// This function tests:
+    ///
+    /// - `send_commit` on valid handle
+    /// - `send_commit` on invalid handle
+    /// - `send_commit` on used handle
+    /// - `send_init` after limit is exceeded
+    ///
```
SabaunT, line 1274: No need for ending `///`

```suggestion
    /// This function tests:
    ///
    /// - `send_commit` on valid handle
    /// - `send_commit` on invalid handle
    /// - `send_commit` on used handle
    /// - `send_init` after limit is exceeded
```

SabaunT: Same for all the tests below.

core-processor/src/ext.rs
```diff
@@ -1211,6 +1264,184 @@ mod tests {
...
+        let handle = ext.send_init().unwrap();
+
+        let res = ext
+            .context
+            .message_context
+            .payload_mut()
+            .try_extend_from_slice(&[1, 2, 3, 4, 5, 6]);
+        assert!(res.is_ok());
+
+        let res = ext.send_push_input(handle, 2, 3);
```
SabaunT, line 1420: please add test case when `offset` param of `send_push_input` is greater or equal to `len`.

SabaunT (approved):
Pretty important tests :+1:

core-processor/src/ext.rs
```diff
@@ -1211,6 +1264,184 @@ mod tests {
...
+
+        let fake_handle = 0;
+
+        let msg = ext.send_commit(fake_handle, data.clone(), 0);
+        assert_eq!(
+            msg.unwrap_err(),
+            FallibleExtError::Core(FallibleExtErrorCore::Message(MessageError::OutOfBounds))
+        );
+
+        let handle = ext.send_init().unwrap();
```
breathx, line 1292: unwrap -> expect everywhere

core-processor/src/ext.rs
```diff
@@ -1211,6 +1264,184 @@ mod tests {
...
+    #[test]
+    /// This function tests:
+    ///
+    /// - `send_push` on non-existent handle
+    /// - `send_push` on valid handle
+    /// - `send_push` on used handle
+    /// - `send_push` with too large payload
+    /// - `send_push` data is added to buffer
+    ///
+    fn test_send_push() {
```
breathx, line 1321:
```suggestion
    // Test scope:
    //
    // - `send_push` on non-existent handle
    // - `send_push` on valid handle
    // - `send_push` on used handle
    // - `send_push` with too large payload
    // - `send_push` data is added to buffer
    //
    #[test]
    fn test_send_push() {
```

breathx (approved):

core-processor/src/ext.rs
mqxf: Comments instead of documentation?

core-processor/src/ext.rs
mqxf: Looking at the inner code, the `len` param of `send_push_input` being greater than `offset` is a possible outcome, and if the `offset` is greater than the length of the data, it just sets the offset and length of the range to 0, therefore, the call to `send_push_input` that I made should not affect the outcome.

core-processor/src/ext.rs
breathx: > Comments instead of documentation?

yep

mqxf:
Not sure why but it failed on a test I didn't make or change `tests::locking_gas_for_waitlist`

ukint-vs:
> Not sure why but it failed on a test I didn't make or change `tests::locking_gas_for_waitlist`

https://github.com/gear-tech/gear/issues/2967

SabaunT (approved):

PR #2922, opened 2023-07-10
feat(wasm-builder): Don't track features in wasm-builder via env var

ark0f:


DennisInSky (approved):

utils/wasm-builder/src/wasm_project.rs
```diff
@@ -402,7 +402,11 @@ extern "C" fn metahash() {{
             self.postprocess_opt(&original_wasm_path, file_base_name)?;
         }
 
-        self.force_rerun_on_next_run(&original_wasm_path)
+        if env::var("__GEAR_WASM_BUILDER_NO_FEATURES_TRACKING").is_err() {
```
shamilsan, line 405: Why do we use this `__` prefix for Gear env vars?

shamilsan (approved):

PR #2926, opened 2023-07-11
fix(rpc-fns): Read state gas limit restrictions

breathx:


pallets/gear/src/runtime_api.rs
```diff
@@ -302,7 +302,7 @@ where
             None,
             None,
             payload,
-            BlockGasLimitOf::<T>::get() / 4,
+            BlockGasLimitOf::<T>::get() * 6,
```
shamilsan, line 305: Isn't it a hack? Looks like magic.

shamilsan (approved):

ark0f (approved):

pallets/gear/src/runtime_api.rs
breathx: Not exactly. Just previous limitations became under required

core-processor/src/executor.rs
```diff
@@ -602,6 +602,8 @@ where
         .into_ext_info(&memory)
         .map_err(|e| format!("Backend postprocessing error: {e:?}"))?;
 
+    log::debug!("\n\n\tBURNED: {}\n\n", info.gas_amount.burned());
```
grishasobol, line 605: remove this debug or change to trace and remove this garbage  \n

pallets/gear/src/runtime_api.rs
grishasobol, line 305: If your are using * 6 for each, then this must be separate constant with explanation why * 6.

pallets/gear/src/tests.rs
```diff
@@ -77,12 +74,57 @@ use gear_core_errors::*;
...
+                payload.encode(),
+                BlockGasLimitOf::<Test>::get(),
+                0
+            ));
+            let mid = get_last_message_id();
+
+            run_to_next_block(None);
+
+            assert_succeed(mid);
+            let state = Gear::read_state_impl(pid).expect("Failed to read state");
```
grishasobol, line 117: `state` is not using in test. At least the size must be assert_eq!

PR #2930, opened 2023-07-12
chore: Clean up for latest merged PRs

breathx:


pallets/gear/src/runtime_api.rs
```diff
@@ -302,6 +302,8 @@ where
             None,
             None,
             payload,
+            // Multiplier 6 was experimentally found as median value between performance,
+            // security and abilities for calculations.
             BlockGasLimitOf::<T>::get() * 6,
```
ark0f, line 307: I guess we can made it as const

ark0f (approved):

pallets/gear/src/runtime_api.rs
ark0f: Right

examples/binaries/read-big-state/src/lib.rs
```diff
@@ -69,18 +67,13 @@ mod wasm {
     static mut STATE: Option<State> = None;
```
DennisInSky, line 67: Is there any point to have as `Option` at all? Why can't it be initialized statically?

DennisInSky (approved):

examples/binaries/read-big-state/src/lib.rs
breathx: yeah it may have const constructor but see no difference for test

Issue #2934, opened 2023-07-12
Update wasmer to 4.0 in gear_sandbox


grishasobol:
See the title=)

PR #2941, opened 2023-07-13
feat(core-processor): make `dlmalloc` a program library

### LINKED ISSUE #1059: Store smart contracts code as shared libraries
### Problem to Solve

Currently we store contracts code in storage as solid wasm file. This is simple and fast solution, but it has one disadvantage. Each contracts which has been built from rust using gstd has a lot of common parts in code. For example dlmalloc, which has size == 5 kBytes, it is almost the same in each contracts currently. Also there is a lot of parts of gstd, which is equal for many contracts. So, the idea is to store separately some common parts of smart contracts.

This has another advantage: this reduces  size of data which must be sent to relay chain for block finalisation.

### Possible Solution

1) First possible solution is to identify common parts during program submit. Unfortunately this cannot be done in simple way because of call instructions, which contains function number as argument. This number can differ in different wasm files, even in equal code parts.

2) Second solution is to submit program code as already separate libraries. We encourage users to do this reducing cost for program submit if some libraries are already in code storage. And before execution this libraries will be linked.

Unfortunately, currently I cannot find a way, how to compile rust code into several separate libraries:
https://github.com/WebAssembly/tool-conventions/blob/main/DynamicLinking.md

### Notes

_No response_

clearloop:
Resolves #1059 

- [x] introduce `gwasm` for program libraries
- [x] introduce gwasm-dlmalloc which is the wasm library version of dlmalloc
- [ ] provide tests somewhere

### What happens

this PR has introduced `gwasm` feature for `gstd`, if `gwasm` enabled, the program will use dlmalloc as a shared library which requires the backend linking the dlmalloc wasm library when executing the program

for the result of the size of `demo_ping.opt.wasm` after compilation

|allocator | size |
|-|-|
|gwasm|4.0k|
|galloc | 8.0k|

### How to test it? 

TBA

### 1. where to perform the link operation? 

imo the final solution will be in pallet-gear that users can import their uploaded libraries, or some official library like ERC-20, but for the mvp version, introducing it in pallet-gear is risky, so considering core-processor atm

### 2. where to store the program libraries?

hardcode in our node ( dirty but practical ) or upload it as storage ( storage IO requires more weights )

@gear-tech/dev


clearloop:
waiting for #2934

breathx:
Temporary closing as blocked

PR #2943, opened 2023-07-13
fix(rpc-fns): Fix bug with `calculate_gas` returning block gas limit

breathx:
While calculating gas there is getter for current gas limit of the main node, that will be none if some error occurs in gas tree (e.g. node not found). Same should be for query of its balance (even in other functionality).
That affected calculating that way: we check balance of already consumed node that has 0 balance, so calculator means that execution burned all gas from the node, which initially equals block gas limit.

### DIFF
--- a/pallets/gear/src/runtime_api.rs
+++ b/pallets/gear/src/runtime_api.rs
@@ -151,12 +151,22 @@ where
             let journal = step.execute().unwrap_or_else(|e| unreachable!("{e:?}"));
 
             let get_main_limit = || {
+                // For case when node is not consumed and has any (even zero) balance
+                // it means that it burned/sent all the funds and we must return it.
+                //
+                // For case when node is consumed and has zero balance it means that
+                // node moved its funds upstream to its ancestor. So this shouldn't
+                // be returned.
+                //
+                // For case when node is consumed and has non zero balance it means
+                // that it has gasless child that will consume gas further. So we
+                // handle this value as well.
                 GasHandlerOf::<T>::get_limit(main_message_id)
                     .ok()
                     .or_else(|| {
                         GasHandlerOf::<T>::get_limit_consumed(main_message_id)
                             .ok()
-                            .and_then(|limit| (!limit.is_zero()).then_some(limit))
+                            .filter(|limit| !limit.is_zero())
                     })
             };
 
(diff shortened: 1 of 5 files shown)


pallets/gear/src/tests.rs
```diff
@@ -9162,11 +9162,23 @@ fn program_generator_works() {
...
+            USER_1.into_origin(),
+            HandleKind::Handle(generator_id),
+            EMPTY_PAYLOAD.to_vec(),
+            0,
+            true,
+            true,
+        )
+        .expect("calculate_gas_info failed");
+
+        assert_ne!(min_limit, BlockGasLimitOf::<Test>::get());
```
DennisInSky, line 9175: Weird place/test for testing

pallets/gear/src/tests.rs
```diff
@@ -79,6 +79,42 @@ use utils::*;
 
 type Gas = <<Test as Config>::GasProvider as common::GasProvider>::GasTree;
 
+#[test]
+fn calculate_gas_returns_not_block_limit() {
```
DennisInSky, line 83: Does it return a value not equal to the limit always or just in some cases? Would be nice to see in the test name something like `never` or `when_bla_bla_bla`

ark0f (approved):

DennisInSky (approved):

pallets/gear/src/tests.rs
breathx: > Does it return a value not equal to the limit always or just in some cases? Would be nice to see in the test name something like `never` or `when_bla_bla_bla`

Exact cases are not identified. Atm its a lot of them when message is consumed, but for me its like general-purpose test so see no reason to specify something special in test name

DennisInSky (approved):

pallets/gear/src/runtime_api.rs
```diff
@@ -150,7 +150,15 @@ where
...
             let journal = step.execute().unwrap_or_else(|e| unreachable!("{e:?}"));
 
-            let get_main_limit = || GasHandlerOf::<T>::get_limit(main_message_id).ok();
+            let get_main_limit = || {
+                GasHandlerOf::<T>::get_limit(main_message_id)
+                    .ok()
+                    .or_else(|| {
+                        GasHandlerOf::<T>::get_limit_consumed(main_message_id)
+                            .ok()
+                            .and_then(|limit| (!limit.is_zero()).then_some(limit))
```
ark0f, line 159:
```suggestion
                            .filter(|limit| !limit.is_zero())
```

ark0f (approved):

PR #2947, opened 2023-07-14
Add original code storage api to sdk and gclient

grishasobol:


gsdk/tests/rpc.rs
```diff
@@ -264,3 +265,50 @@ async fn test_runtime_wasm_blob_version_history() -> Result<()> {
...
+        )
+        .await?;
+
+    let program = signer.api().gprog(pid).await?;
+    let last_block = signer
+        .api()
+        .rpc()
+        .block(None)
+        .await?
+        .unwrap()
```
clearloop, line 296:
```suggestion
        .await??
```

hmm if this is possible?

clearloop (approved):

gsdk/tests/rpc.rs
clearloop:
```
let rpc = signer.api().rpc();
```

grishasobol: Done

breathx (approved):

PR #2964, opened 2023-07-20
chore(ft-benchmarks): fix wrong path and set PRNG seed

StackOverflowExcept1on:
Because the examples have been moved, this test has stopped working. I also added seed for the number generator so that the test will always show the same result.


breathx:
Similar changes were introduced in #2948, but ok

breathx (approved):

ukint-vs:
Rebase pls to pass ci

PR #2975, opened 2023-07-24
feat: use `alloca` for stack allocations

StackOverflowExcept1on:
Changes:
- use uninitialized memory for stack allocations instead of filling arrays with zeros
- `stack_buffer` has been moved to `gstack-buffer`
- if the size exceeds 64 KiB, then the heap is used

Performance:
```
Gas burned for one transfer operation = 0.9366831083336524 * 10^9 (master)
Gas burned for one transfer operation = 0.9199329316460656 * 10^9 (this branch)
```

Binary size:
| Before (bytes) | After (bytes) | program                         |
|----------------|---------------|---------------------------------|
| 29155          | 26475         | demo_fungible_token.opt.wasm    |
| 12892          | 10955         | demo_stack_allocations.opt.wasm |
| 6873           | 6187          | demo_piggy_bank.opt.wasm        |


gstack-buffer/build.rs
```diff
@@ -0,0 +1,28 @@
+use std::{env, error::Error, path::PathBuf};
+
+fn main() -> Result<(), Box<dyn Error>> {
```
breathx, line 3: for me its better to introduce different main fns under features

gstack-buffer/alloca/alloca.c
```diff
@@ -0,0 +1,16 @@
+#include <stddef.h>
+#include <stdint.h>
+
+#ifdef _MSC_VER
```
breathx, line 4: why is it optional?

gstack-buffer/alloca/README.md
```diff
@@ -0,0 +1,8 @@
+### What is alloca?
```
breathx, line 1: add info how, when and why we should compile

gstack-buffer/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "gstack-buffer"
```
breathx, line 2: I'd really go with gear-stack-buffer, it doesn't look as good as short `g*` stuff

gcore/src/lib.rs
```diff
@@ -67,7 +67,7 @@ pub mod errors;
 pub mod exec;
 pub mod msg;
 pub mod prog;
-pub mod stack_buffer;
+pub use gstack_buffer as stack_buffer;
```
breathx, line 70: pub extern crate

gcore/src/msg.rs
```diff
@@ -190,10 +190,19 @@ pub fn with_read_on_stack<T>(f: impl FnOnce(Result<&mut [u8]>) -> T) -> T {
...
+                    buffer.as_mut_ptr() as *mut u8,
+                    &mut len as *mut u32,
+                )
+            }
         }
 
-        f(SyscallError(len).into_result().map(|_| buffer))
+        f(SyscallError(len)
+            .into_result()
+            .map(|_| unsafe { &mut *(&mut buffer[..size] as *mut _ as *mut [u8]) }))
```
breathx, line 205: what happened here? xd

gstack-buffer/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+//! Stack allocations utils.
+
+#![feature(c_unwind)]
+#![no_std]
+
+extern crate alloc;
+
+use alloc::vec::Vec;
+
+use core::{
```
breathx, line 28:
```suggestion
#![no_std]
#![feature(c_unwind)]

extern crate alloc;

use alloc::vec::Vec;
use core::{
```

gstack-buffer/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+    ffi::c_void,
+    mem::{ManuallyDrop, MaybeUninit},
+    slice,
+};
+
+const MAX_BUFFER_SIZE: usize = 64 * 1024;
```
breathx, line 34:
```suggestion
pub const MAX_BUFFER_SIZE: usize = 64 * 1024;
```

gstack-buffer/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+
+type Callback = unsafe extern "C-unwind" fn(ptr: *mut MaybeUninit<u8>, data: *mut c_void);
```
breathx, line 36:
```suggestion
type Callback = extern "C-unwind" fn(ptr: *mut MaybeUninit<u8>, data: *mut c_void);
```

gstack-buffer/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+#[cfg(any(
+    feature = "compile-alloca",
+    all(not(feature = "compile-alloca"), target_arch = "wasm32")
+))]
+extern "C-unwind" {
+    fn c_with_alloca(size: usize, callback: Callback, data: *mut c_void);
+}
+
+#[cfg(all(not(feature = "compile-alloca"), not(target_arch = "wasm32")))]
+unsafe extern "C-unwind" fn c_with_alloca(_size: usize, callback: Callback, data: *mut c_void) {
```
breathx, line 47: put unsafes in-place pls

gstack-buffer/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+    let mut buffer = MaybeUninit::<[MaybeUninit<u8>; MAX_BUFFER_SIZE]>::uninit().assume_init();
```
breathx, line 48:
```suggestion
    // Create an uninitialized array of `MaybeUninit`. The `assume_init` is
    // safe because the type we are claiming to have initialized here is a
    // bunch of `MaybeUninit`s, which do not require initialization.
    let mut buffer = MaybeUninit::<[MaybeUninit<u8>; MAX_BUFFER_SIZE]>::uninit().assume_init();
```

gstd/src/msg/encoded.rs
```diff
@@ -140,9 +144,10 @@ pub fn reply_on_stack<E: Encode + MaxEncodedLen>(payload: E, value: u128) -> Res
             const ERROR_LOG: &str = "Unexpected encoding behavior: too large input bytes size";
             let end_offset = self.offset.checked_add(bytes.len()).expect(ERROR_LOG);
             if end_offset > self.buffer.len() {
-                panic!("{}", ERROR_LOG);
+                panic!("{ERROR_LOG}");
             }
-            self.buffer[self.offset..end_offset].copy_from_slice(bytes);
+            let this = &mut self.buffer[self.offset..end_offset];
+            this.copy_from_slice(unsafe { transmute(bytes) });
```
breathx, line 150: btw can we even more optimise it by avoiding copying with some other approach or adding additional methods in codec?

gstack-buffer/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+    callback(buffer.as_mut_ptr(), data);
+}
+
+#[inline(always)]
+fn get_trampoline<F: FnOnce(*mut MaybeUninit<u8>)>(_closure: &F) -> Callback {
+    trampoline::<F>
+}
+
+unsafe extern "C-unwind" fn trampoline<F: FnOnce(*mut MaybeUninit<u8>)>(
```
breathx, line 57: why do we need this?

breathx (changes requested):
Comments across code pls

gstack-buffer/alloca/alloca.c
StackOverflowExcept1on: Microsoft compiler does not support [VLA](https://en.wikipedia.org/wiki/Variable-length_array). But `char buffer[size]` is valid syntax for other compilers like clang and gcc

gstack-buffer/src/lib.rs
StackOverflowExcept1on: This hack is used to capture local variables. Also see: https://adventures.michaelfbryan.com/posts/rust-closures-in-ffi/

gstack-buffer/src/lib.rs
StackOverflowExcept1on: This function is kind of a polyfill when we can't use the pre-built library. Since the c_with_alloca function from C is unsafe, this polyfill should also be marked as unsafe.

gstack-buffer/src/lib.rs
StackOverflowExcept1on: I don't agree with this. If we have C code, then its arguments must also be compatible with the C ABI. So the callback should be marked as unsafe
```c
void f(void(*callback)) { }
```

gcore/src/msg.rs
StackOverflowExcept1on: this is https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#method.slice_assume_init_ref

gstd/src/msg/encoded.rs
StackOverflowExcept1on: We have uninitialized memory where we will write data when encoding. It requires copying bytes anyway.

gstack-buffer/alloca/README.md
StackOverflowExcept1on: We don't have to compile it. On wasm32-unknown-unknown we use pre-built library so that the end user of gstd does not need installed C compiler when building contracts. However, the end user will benefit from our optimizations due to the pre-built library.

gcore/src/msg.rs
StackOverflowExcept1on: I also checked it with https://github.com/rust-lang/rust-bindgen
```
$ bindgen alloca_.h

...

extern "C" {
    pub fn c_with_alloca(
        size: usize,
        callback: ::std::option::Option<
            unsafe extern "C" fn(arg1: *mut u8, arg2: *mut ::std::os::raw::c_void),
        >,
        data: *mut ::std::os::raw::c_void,
    );
}
```

gstack-buffer/Cargo.toml
StackOverflowExcept1on: I can do similar to gear-core-errors:
- folder: /stack-buffer/
- name in Cargo.toml: gear-stack-buffer

gstack-buffer/src/lib.rs
StackOverflowExcept1on: I also checked it with https://github.com/rust-lang/rust-bindgen
This CLI tool also shows the `unsafe` modifier in the callback.
```
$ bindgen alloca_.h

...

extern "C" {
    pub fn c_with_alloca(
        size: usize,
        callback: ::std::option::Option<
            unsafe extern "C" fn(arg1: *mut u8, arg2: *mut ::std::os::raw::c_void),
        >,
        data: *mut ::std::os::raw::c_void,
    );
}
```

gcore/src/lib.rs
StackOverflowExcept1on: This is bad for documentation.

`pub use gstack_buffer as stack_buffer;`
image
`pub extern crate gstack_buffer as stack_buffer;`
image

gstd/src/msg/encoded.rs
breathx: but atm we copy them twice: once for writable buffer, secondly for uninitialised memory

gstack-buffer/alloca/README.md
breathx: So that's exactly I was expecting in Readme, now see it

gstack-buffer/Cargo.toml
breathx: that's ok

breathx (approved):

breathx:
@grishasobol @ukint-vs @ark0f

ark0f (changes requested):

gcore/src/lib.rs
```diff
@@ -67,7 +67,7 @@ pub mod errors;
 pub mod exec;
 pub mod msg;
 pub mod prog;
-pub mod stack_buffer;
+pub extern crate gear_stack_buffer as stack_buffer;
```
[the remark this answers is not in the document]
StackOverflowExcept1on: @breathx asked me to use `pub use gear_stack_buffer` here

ukint-vs (approved):

gcore/src/lib.rs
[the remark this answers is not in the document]
breathx: @ark0f whats the matter?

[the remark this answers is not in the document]
StackOverflowExcept1on: Since Rust 2018 edition:
- extern crate is no longer needed in 99% of circumstances.
- An exception: "sysroot" crates, `#[macro_use] extern crate bar;`

Source: https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html

ark0f (approved):

PR #2976, opened 2023-07-24
feat(wasm-builder,wasm-opt): use only mvp features, enable sign-ext

ukint-vs:
By the way using wasm-opt from lib gives wasm an even smaller size and speeds it up a bit.

@gear-tech/dev


shamilsan (approved):

clearloop (approved):

clearloop:
> By the way using wasm-opt from lib gives wasm an even smaller size and speeds it up a bit.
> 
> @gear-tech/dev

feeling it could be caused by the default settings

Issue #2970, opened 2023-07-21
`test_runtime_wasm_blob_version` failed in `build-win-native` on master


clearloop:
### Problem

from bc5442828048b05933205d18a63ccf21cc6fbe3a

### Steps

https://github.com/gear-tech/gear/actions/runs/5556675247/job/15055261882#step:14:300

### Relevant Log Output

<details><summary>Click to expand/collapse</summary>
<p>

```
<Paste your log here>
```

</p>
</details>

PR #2978, opened 2023-07-25
Add extra diagnostic to wasm blob version test

DennisInSky:
To shed light on the #2970


breathx:
@DennisInSky stale pr?

NikVolf:
Looks like stale

PR #2979, opened 2023-07-25
feat(gsdk): introduce events API

### LINKED ISSUE #2896: Introduce new method events for gsdk API
btw, can't we cast events as root events just like here?

_Originally posted by @breathx in https://github.com/gear-tech/gear/pull/2744#discussion_r1249321776_

clearloop:
Resolves #2896 

- [x] introduce events API
- [x] rename module types to subscription
- [x] introduce module api

@gear-tech/dev


gsdk/src/lib.rs
```diff
@@ -51,84 +68,38 @@ pub mod gp {
...
 
-/// Gear api wrapper.
-#[derive(Clone)]
-pub struct Api {
-    /// How many times we'll retry when rpc requests failed.
-    pub retry: u16,
-    client: OnlineClient<GearConfig>,
+/// Information of gas
+#[derive(Clone, Debug, Decode, Encode, PartialEq, Eq, Serialize, Deserialize)]
+pub struct GasInfo {
```
breathx, line 76: Original gas info from pallet is differ. Why?
```rust
/// The struct contains results of gas calculation required to process
/// a message.
#[derive(Clone, Debug, Decode, Encode, PartialEq, Eq, TypeInfo)]
pub struct GasInfo {
    pub min_limit: u64,
    /// Gas amount that we reserve for some other on-chain interactions.
    pub reserved: u64,
    /// Contains number of gas burned during message processing.
    pub burned: u64,
    /// The value may be returned if a program happens to be executed
    /// the second or next time in a block.
    pub may_be_returned: u64,
    /// Was the message placed into waitlist at the end of calculating.
    ///
    /// This flag shows, that `min_limit` makes sense and have some guarantees
    /// only before insertion into waitlist.
    pub waited: bool,
}
```

gsdk/src/lib.rs
```diff
@@ -51,84 +68,38 @@ pub mod gp {
...
-
-impl Deref for Api {
-    type Target = OnlineClient<GearConfig>;
-
-    fn deref(&self) -> &Self::Target {
-        &self.client
-    }
-}
+/// Transaction status.
+pub type TxStatus = tx::TxStatus<GearConfig, OnlineClient<GearConfig>>;
```
breathx, line 98:
```suggestion
pub type TxStatus = TxStatus<GearConfig, OnlineClient<GearConfig>>;
```

gsdk/src/lib.rs
```diff
@@ -51,84 +68,38 @@ pub mod gp {
...
+pub type GearPages = HashMap<u32, Vec<u8>>;
 
-    /// Subscribe finalized blocks
-    pub async fn finalized_blocks(&self) -> Result<types::Blocks> {
-        Ok(types::Blocks(
-            self.client.blocks().subscribe_finalized().await?,
-        ))
-    }
+/// Transaction in block.
+pub type InBlock = Result<TxInBlock<GearConfig, OnlineClient<GearConfig>>>;
```
breathx, line 95: Introduce TxInBlock wraper and InBlockResult or something like this

gsdk/src/rpc.rs
```diff
@@ -18,7 +18,7 @@
 
 //! gear api rpc methods
 #![allow(clippy::too_many_arguments)]
```
breathx, line 20: Move it to the only places requires this
```suggestion
```

gsdk/src/signer/rpc.rs
```diff
@@ -18,7 +18,7 @@
 
 //! RPC calls with signer
 #![allow(clippy::too_many_arguments)]
```
breathx, line 20: same

gsdk/src/subscription.rs
```diff
@@ -0,0 +1,83 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//! Shared types
```
breathx, line 19:
```suggestion
//! Shared types

```

breathx (approved):
Please adjust some examples and write comment how it works now

gsdk/src/lib.rs
clearloop: https://github.com/gear-tech/gear/issues/3021

need to be updated, still works caused by the deserializer has skipped the non-exist fields

gsdk/src/lib.rs
clearloop: the name of `TxStatus` is duplicated

gsdk/src/subscription.rs
```diff
@@ -0,0 +1,84 @@
...
+            })
+        })
+    }
+}
+
+impl From<BlockSubscription> for Events {
+    fn from(sub: BlockSubscription) -> Self {
+        Self(sub.into())
+    }
+}
```
clearloop, line 84: This is the wrapper for the new introduced `Events` api

gsdk/src/api.rs
```diff
@@ -0,0 +1,92 @@
...
+
+    /// Subscribe finalized blocks
+    pub async fn finalized_blocks(&self) -> Result<Blocks> {
+        Ok(self.client.blocks().subscribe_finalized().await?.into())
+    }
+
+    /// Subscribe all events
+    pub async fn events(&self) -> Result<Events> {
+        Ok(self.client.blocks().subscribe_all().await?.into())
+    }
```
clearloop, line 67: This is actually the only updates in this PR

clearloop (commented):
This PR mainly added `api.events()` which returns an iterator `gsdk::Events`, which works like `api.blocks()`

plus, including some cleaning stuffs only

gsdk/src/lib.rs
breathx: `GearTxStatus maybe`

breathx: to specify underlying generic type

breathx:
@ark0f @StackOverflowExcept1on @tltsutltsu

gsdk/src/lib.rs
clearloop: makes sense, but since our library is just for our API, I think we can take the name of `TxStatus`, instead, `subxt::TxStatus` -> `SubxtTxStatus`

gsdk/src/lib.rs
```diff
@@ -51,84 +68,41 @@ pub mod gp {
...
+pub type TxInBlock = SubxtTxInBlock<GearConfig, OnlineClient<GearConfig>>;
 
-    /// Subscribe finalized blocks
-    pub async fn finalized_blocks(&self) -> Result<types::Blocks> {
-        Ok(types::Blocks(
-            self.client.blocks().subscribe_finalized().await?,
-        ))
-    }
+/// Transaction in block with result wrapper.
+pub type TxInBlockResult = Result<TxInBlock>;
```
ark0f, line 98: I think this alias is not actually necessary

ark0f (approved):

tltsutltsu (approved):
:thumbsup:

PR #2981, opened 2023-07-26
refactor(pallet-airdrop): Add tests and benchmarks

### LINKED ISSUE #2949: Pallets refactoring after M1 review
### File Location(s)

_No response_

### Proposal

- [ ] Some pallet refactoring can be done (e.g. too long functions in airdrop pallet).
- [ ] Add benchmarking for `transfer_vested`.
- [ ] Cover more tests for negative scenarios (e.g. `NotVesting`, `ScheduleIndexOutOfBounds`).

ukint-vs:
Resolves #2949  .

- Add tests for `Notvesting` and `ScheduleIndexOutOfBounds` errors
- Fix `transfer` benchmark
- Add `transfer_vested` benchmark
- Add benchmarks to runtime
- Update pallet weights

But no fee will be paid anyway as these calls are only allowed for sudo

@gear-tech/dev

### DIFF
--- a/pallets/airdrop/src/benchmarking.rs
+++ b/pallets/airdrop/src/benchmarking.rs
@@ -37,7 +37,7 @@ benchmarks! {
         <T as pallet_gear::Config>::Currency::deposit_creating(&source, (1u128 << 60).unique_saturated_into());
         let recipient: T::AccountId = benchmarking::account("recipient", 0, 0);
         // Keeping in mind the existential deposit
-        let amount = 100_000_u128.saturating_add(10_u128.saturating_mul(q.into()));
+        let amount = 10_000_000_000_000_u128.saturating_add(10_u128.saturating_mul(q.into()));
 
     }: _(RawOrigin::Root, source, recipient.clone(), amount.unique_saturated_into())
     verify {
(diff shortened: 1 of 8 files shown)


shamilsan (approved):

pallets/airdrop/src/benchmarking.rs
```diff
@@ -37,12 +38,33 @@ benchmarks! {
...
-        let amount = 100_000_u128.saturating_add(10_u128.saturating_mul(q.into()));
+        let amount = 10_000_000_000_000_u128.saturating_add(10_u128.saturating_mul(q.into()));
 
     }: _(RawOrigin::Root, source, recipient.clone(), amount.unique_saturated_into())
     verify {
         assert_eq!(pallet_balances::Pallet::<T>::total_balance(&recipient), amount.unique_saturated_into());
     }
+
+    transfer_vested {
+        let q in 1 .. 256;
```
ekovalev, line 49: I'm not quite getting the meaning of the complexity parameter `q` here: seems it only affects the amount being transferred. Off the top of my head the resulting weight shouldn't depend on it so we could have a constant weight function for this call. Or am I mistaken?

ukint-vs: Yeah you're right. But after I implemented benchmarks I remembered about sudo call😅

ekovalev (approved):

PR #3004, opened 2023-07-30
feat(ci): run time consuming tests using github action

### LINKED ISSUE #2811: Run ignored tests on CI
### Problem to Solve

Stress test with fungible token is really time consuming so it's decided to avoid on `make pre-commit` but it should be ruined somehow on CI

### Possible Solution

/ .. /

### Notes

_No response_

StackOverflowExcept1on:
Resolves #2811
Successful build: https://github.com/gear-tech/gear/actions/runs/5706348278

### DIFF
--- a/scripts/src/test.sh
+++ b/scripts/src/test.sh
@@ -24,8 +24,7 @@ test_usage() {
     syscalls       run syscalls integrity test in benchmarking module of pallet-gear
     docs           run doc tests
     validators     run validator checks
-    ignored        run ignored tests
-
+    time-consuming run time consuming tests
 EOF
 }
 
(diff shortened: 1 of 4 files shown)


scripts/src/test.sh
```diff
@@ -100,3 +101,7 @@ doc_test() {
 
   __GEAR_WASM_BUILDER_NO_BUILD=1 SKIP_WASM_BUILD=1 SKIP_GEAR_RUNTIME_WASM_BUILD=1 SKIP_VARA_RUNTIME_WASM_BUILD=1 $CARGO test --doc --workspace --exclude runtime-fuzzer --exclude runtime-fuzzer-fuzz --manifest-path="$MANIFEST" --no-fail-fast "$@"
 }
+
+ignored_tests() {
+  $CARGO test -p demo-fungible-token --no-fail-fast "$@" -- --nocapture --ignored
+}
```
clearloop, line 107: mb using ft tests instead?

clearloop (commented):
Feeling just naming/running `ignored tests` is confused and unpredictable, bcz sometimes we giving ignore tags to the tests which will never work / blocking by some issues

scripts/src/test.sh
StackOverflowExcept1on: I think what we could do `$CARGO test --workspace --manifest-path="$MANIFEST" --no-fail-fast "$@" -- --nocapture --ignored`, but it will probably take a long time to build.

clearloop: running all ignored tests is arbitrary, not all of the ignored tests are caused by taking too much building time

StackOverflowExcept1on: I renamed it "time-consuming tests". In the future, we may have such tests not only in fungible-token.

breathx (approved):

clearloop (approved):

PR #3005, opened 2023-07-31
fix: increase gas limit for state reading

shamilsan:
@gear-tech/dev


shamilsan:
We need to find a more robust solution.

PR #3007, opened 2023-07-31
feat(gstd): propagate stack allocations

### LINKED ISSUE #2880: use encoded_size in gstd::msg::reply<D: Encode>
### Problem to Solve

In many simple cases when user does not use custom types with heap allocation (this should be checked for complex structs, because derive Encode can be simple in that cases and just use default impl which is very ineffective).

So, we can use encoded_size  in reply in order to use stack allocation, which requires the known size before encoding.

### Possible Solution

-

### Notes

_No response_

StackOverflowExcept1on:
Resolves #2880

Changes:
- `msg::send*(...)`, `msg::reply*(...)`, `{prog, ProgramGenerator}::create_program<E>*(...)` __might use__ stack allocation if `payload.encoded_size()` is less than 64 KiB
- removed `load_on_stack` and `reply_on_stack` from `gstd`
- `gstd::prelude` re-exports `MaxEncodedLen`
- if the user uses `#[derive(MaxEncodedLen)]`, then due to the specialization, encoding on stack is faster (no need to calculate the size)


gstd/src/msg/encoded.rs
```diff
@@ -25,14 +25,41 @@ use crate::{
...
+        buffer: &'a mut [MaybeUninit<u8>],
+        offset: usize,
+    }
+
+    impl<'a> Output for ExternalBufferOutput<'a> {
+        fn write(&mut self, bytes: &[u8]) {
+            const ERROR_LOG: &str = "Unexpected encoding behavior: too large input bytes size";
+            let end_offset = self.offset.checked_add(bytes.len()).expect(ERROR_LOG);
+            if end_offset > self.buffer.len() {
+                panic!("{ERROR_LOG}");
```
breathx, line 47: Add str right here, what's the point to format anything?

StackOverflowExcept1on: It was written by Gregory, I just made change:
```diff
- panic!("{}", ERROR_LOG);
+ panic!("{ERROR_LOG}");
```
We can't use just string literal here:
```
error: format argument must be a string literal
 --> src/main.rs:3:12
  |
3 |     panic!(ERROR_LOG);
  |            ^^^^^^^^^
```

breathx: So why do you need this const? Add some simple expl and copy-paste

StackOverflowExcept1on: Maybe move this conversation to #2975?

breathx: kk

breathx: > Maybe move this conversation to #2975?

will we merge 2975 first?

breathx (approved):

gstd/src/msg/encoded.rs
StackOverflowExcept1on: Yes, this is an additional PR to 2975, because we solved the problem with increasing the size of the binary file.

ukint-vs (approved):

ukint-vs (approved):

Issue #2700, opened 2023-05-30
Consider another error for `Ext::check_forbidden_destination`


techraed:
### File Location(s)

https://github.com/gear-tech/gear/blob/8c82ff9554fd5c4db7574b80852ba83dc6636c0e/core-processor/src/ext.rs#L366

### Proposal

It must be something like `SyscallUsageError`, which we already have in errors crate. Better to Maybe there's a need to make it more complex error variant.

SabaunT:
Need first to solve the problem with backwards compatibility.

PR #3010, opened 2023-08-02
refactor(core): Add nonce to program id generation

### LINKED ISSUE #2821: Handle case when `gr_create_program` results in already existing program
### Problem to Solve

/ .. /

### Possible Solution

Consider adding extra data at backend side except code_id and salt (maybe message id and counter etc)

### Notes

_No response_

mqxf:
Resolves #2821

@gear-tech/dev


core/src/message/init.rs
```diff
@@ -133,7 +133,7 @@ impl InitPacket {
     /// Create new InitPacket without gas.
     pub fn new(code_id: CodeId, salt: Salt, payload: Payload, value: Value) -> Self {
         Self {
-            program_id: ProgramId::generate(code_id, salt.inner()),
+            program_id: ProgramId::generate(0u32, code_id, salt.inner()),
```
SabaunT, line 136: Nope.

`InitPacket::new` must accet nonce, that will be used for program id generation

SabaunT (changes requested):
What about changes in core-processor when `create_program` is called? or `upload_program`? There are lots of dependant changes like if you change program id generation, then you must change `InitPacket` instantiation and all places where it's used.

core/src/message/init.rs
mqxf: The issue is that when we take in a InitPacket, like in the core, it doesn't have a message id yet. What it does, is it gives over the InitPacket to the create_program function, which means the ProgramId would already be created, the ext create_program actually makes the messageid, which means that I'd need to change around the way InitPacket works and is used. Should I do that?

mqxf: So `InitPacket::new` should not take in the nonce, instead, it will keep the CodeId and salt, and not generate a new ProgramId. Then, in the Ext `create_program` function, the data from the packet is used with the newly made message id to make the ProgramId.

SabaunT: > The issue is that when we take in a InitPacket, like in the core, it doesn't have a message id yet. What it does, is it gives over the InitPacket to the create_program function, which means the ProgramId would already be created, the ext create_program actually makes the messageid, which means that I'd need to change around the way InitPacket works and is used. Should I do that?

I guess you should create a new program id for `gr_create_program` from currently executing message id.
I think there's a way to obtain it from ext in `core-backend` sys-call implementation.

SabaunT: Or you can play around `InitPacket` type somehow. For example, `Externalities::create_program` can accept not the `InitPacket`, but something like `SemiInitPacket` (choose name by yourself), which later must be converted to init packet using the nonce/message id to generate program id.

You aren't required to implement exactly that, but use similar approach with types.

SabaunT: > So `InitPacket::new` should not take in the nonce, instead, it will keep the CodeId and salt, and not generate a new ProgramId. Then, in the Ext `create_program` function, the data from the packet is used with the newly made message id to make the ProgramId.

Yeah, nice idea. I described something similar in the previous comment.

core/src/ids.rs
```diff
@@ -224,41 +225,63 @@ impl ReservationId {
     }
 }
 
+pub trait ProgramNonce {
```
breathx, line 228: Can't see any reason for the trait and impls for mid, pid and u32/64

core/src/ids.rs
```diff
@@ -204,10 +205,10 @@ impl ProgramId {
     pub const SYSTEM: Self = Self(*b"geargeargeargeargeargeargeargear");
 
     /// Generate ProgramId from given CodeId and salt
-    pub fn generate(code_id: CodeId, salt: &[u8]) -> Self {
+    pub fn generate<T: ProgramNonce>(nonce: T, code_id: CodeId, salt: &[u8]) -> Self {
         const SALT: &[u8] = b"program";
 
-        let argument = [SALT, code_id.as_ref(), salt].concat();
+        let argument = [nonce.as_nonce().as_ref(), SALT, code_id.as_ref(), salt].concat();
```
breathx, line 211: overhead out of nowhere, because your as_nonce allocates vec just to use it as slice :/

gcli/tests/common/mod.rs
```diff
@@ -102,7 +102,7 @@ pub fn login_as_alice() -> Result<()> {
 
 /// Generate program id from code id and salt
 pub fn program_id(bin: &[u8], salt: &[u8]) -> ProgramId {
```
breathx, line 104: add comment that it may differ for tests where multiple programs created

breathx (changes requested):

breathx:
Once already addressed comments solved, do not forget to replace no-op block with panic in journal handler's program creation on gear pallet

gtest/src/program.rs
```diff
@@ -520,7 +520,7 @@ fn read_file<P: AsRef<Path>>(path: P, extension: &str) -> Vec<u8> {
 }
 
 pub fn calculate_program_id(code_id: CodeId, salt: &[u8]) -> ProgramId {
-    ProgramId::generate(code_id, salt)
+    ProgramId::generate(0u32, code_id, salt)
```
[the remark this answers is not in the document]
mqxf: Like have a function which doesn't take in a nonce to generate a program from the user? And then a separate function `generate_with_nonce` or something similar for when a program generates it?

core/src/ids.rs
mqxf: The only way I found of doing this was by using 

```unsafe { slice::from_raw_parts(self as *const u32 as *const u8, 4) }```

That allows me to return a &[u8]. This is still safe since we know that the u32 should still live after the bytes are returned thus it will be a valid pointer for when it is added to the Hash. Please confirm that this is okay, or some other alternative solution to removing the Vec.

File: core/src/ids.rs (hunk 2 above)
Comment (mqxf) [reply core/src/ids.rs:228]: The idea was to limit the nonce variable to only be one of the three types, but this could be removed and a couple extra functions could be added for the u32 and u64 case. This is only an issue since u32 and u64 don't implement something like AsRef<[u8]>

File: core/src/ids.rs [hunk 6]
```diff
@@ -224,41 +225,65 @@ impl ReservationId {
...
+
+impl ProgramNonce for MessageId {
+    fn as_nonce(&self) -> &[u8] {
+        self.as_ref()
+    }
+}
+
+impl ProgramNonce for u32 {
+    fn as_nonce(&self) -> &[u8] {
+        unsafe { slice::from_raw_parts(self as *const u32 as *const u8, 4) }
```
SabaunT, line 242: Is it possible to avoid `unsafe` anyhow?

gtest/src/program.rs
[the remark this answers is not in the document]
breathx: > Like have a function which doesn't take in a nonce to generate a program from the user? And then a separate function `generate_with_nonce` or something similar for when a program generates it?

yep

core/src/ids.rs
breathx: That's mostly about my misunderstanding why do we need trait here at all

core/src/ids.rs
SabaunT: maybe with introducing `NonceBytes` carrier which owns `[u8; T]`.

core-errors/src/lib.rs
```diff
@@ -159,6 +159,10 @@ pub enum MessageError {
     /// An error occurs in attempt to charge gas for dispatch stash hold.
     #[display(fmt = "Not enough gas to hold dispatch message")]
     InsufficientGasForDelayedSending = 399,
+
+    /// An error when the program id is marked as invalid
+    #[display(fmt = "Program id is forbidden")]
+    InvalidProgramId = 400,
```
SabaunT, line 165: Better stick one wording: either forbidden or invalid.

But I'd omit these changes. Better introduce these changes in a separate PR (this will resolve #2700 )

core/src/message/init.rs
```diff
@@ -49,7 +49,7 @@ impl InitMessage {
     pub fn from_packet(id: MessageId, packet: InitPacket) -> Self {
         Self {
             id,
-            destination: packet.program_id,
+            destination: packet.destination(id),
```
SabaunT, line 52: That's wrong. It definetely creates untraceable id duplicates. Program can use same salt and code id, as was used to create itself. The message id will remain same if the call described previously happens in `init`.

The idea was that from pallet-level you use a global program ids nonce, but from `Externalities` level you use `u32`/`u64` nonce.

SabaunT (changes requested):

core/src/ids.rs
SabaunT: This carrier can be instatiated from `u32`/`u64` bytes or message id bytes.

core-errors/src/lib.rs
mqxf: In that issue it mentions a `SyscallUsageError`, should that be returned instead?

core/src/ids.rs
```diff
@@ -224,41 +225,65 @@ impl ReservationId {
...
+    }
+}
+
+impl ProgramNonce for u64 {
+    fn as_nonce(&self) -> &[u8] {
+        unsafe { slice::from_raw_parts(self as *const u64 as *const u8, 8) }
```
NikVolf, line 248: please no

not only this is platform dependant, it is also unsafe

see
https://doc.rust-lang.org/nightly/std/primitive.u64.html#method.to_le_bytes

core/src/message/init.rs
NikVolf: nonces should be local to the program that creates other programs

There should not be global id, as it will lead to parallelization problem (since everybody will be able to access global counter)

core/src/ids.rs
mqxf: That was the initial thing but I couldn't return a reference so I had to allocate a Vec, which wasnt nice, though I could revert it back to that.

NikVolf: You can't have platform-independent slice in memory for the integer

core/src/message/init.rs
breathx: > nonces should be local to the program that creates other programs
> 
> There should not be global nonce, as it will lead to parallelization problem (since everybody will be able to access global counter)

yeah, optimal solution is to follow message id generation except in-block counter for creating from user

breathx:
Resolve finished conversations, fix CI and re-request reviews please @mqxf

mqxf:
Will do closer to evening today. I am on multiple flights today.

core/src/message/context.rs
```diff
@@ -231,6 +231,7 @@ pub struct MessageContext {
     outcome: ContextOutcome,
     store: ContextStore,
     settings: ContextSettings,
+    nonce: u32,
```
SabaunT, line 234: That's wrong nonce. Use the `self.store.initialized.len()`

SabaunT (changes requested):
Please use the next approach:
- program ids from pallet - `hash(concat(message_id, code_id, salt))`
- program ids from wasm program - `hash(concat(message_id, code_id, salt, store.initialized.len()))`

core/src/ids.rs
```diff
@@ -224,41 +230,66 @@ impl ReservationId {
     }
 }
 
+/// Trait representing the extra info needed for program id generation.
+/// This is usually a MessageId, but it can be a handle in the form of a u32 or u64.
+pub trait ProgramNonce {
+    /// Turn the nonce into a vector of bytes to be used in the program id generation.
+    fn as_nonce(&self) -> Vec<u8>;
```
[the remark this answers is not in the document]
mqxf: Not sure how anything like that would help, since its always a dynamic amount of bytes, so it cannot be a statically allocated anything. u32 gives 4 bytes, u64 gives 8 and MessageId gives way more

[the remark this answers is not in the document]
SabaunT: Actually the array len can be generic.
Anyway, implement this and you can get rid of this trait https://github.com/gear-tech/gear/pull/3010#pullrequestreview-1584247512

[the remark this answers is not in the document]
mqxf: I'm wondering if I even need the trait, instead just opting to always using a u64, since I don't seem to use the messageid, instead I use the outgoing length.

core-errors/src/lib.rs
```diff
@@ -159,6 +159,10 @@ pub enum MessageError {
     /// An error occurs in attempt to charge gas for dispatch stash hold.
     #[display(fmt = "Not enough gas to hold dispatch message")]
     InsufficientGasForDelayedSending = 399,
+
+    /// An error when a syscall is used incorrectly
+    #[display(fmt = "Syscall was used incorrectly")]
+    SyscallUsageError = 400,
```
ark0f, line 165:
```suggestion
    SyscallUsageError = 311,
```

`InsufficientGasForDelayedSending` is assigned to 399 as to be removed

Moreover, such generic error as `sys-call usage` must be discussed. We have forbidden code 0xffff for such error

ark0f (changes requested):

core/src/message/init.rs
```diff
@@ -46,10 +46,10 @@ pub struct InitMessage {
 
 impl InitMessage {
     /// Create InitMessage from InitPacket.
-    pub fn from_packet(id: MessageId, packet: InitPacket) -> Self {
+    pub fn from_packet(id: MessageId, packet: InitPacket, nonce: usize) -> Self {
```
SabaunT, line 49: `nonce` should be `Option`, right?

pallets/gear/src/lib.rs
```diff
@@ -1254,7 +1255,7 @@ pub mod pallet {
                 false,
             );
 
-            let message = InitMessage::from_packet(message_id, packet);
+            let message = InitMessage::from_packet(message_id, packet, 0);
```
SabaunT, line 1258: idk how it works, but it's seems to be incorrect. You create program id without nonce, then you create message with destination that uses nonce equal to zero.

pallets/gear/src/lib.rs
```diff
@@ -1193,6 +1194,8 @@ pub mod pallet {
             gas_limit: u64,
             value: BalanceOf<T>,
         ) -> Result<InitPacket, DispatchError> {
+            let program_id = ProgramId::generate(code_id, &salt);
```
SabaunT, line 1197: what's this? it's a wrong program id to be checked.

SabaunT: there are no program ids that are generated from salt and code id only.

SabaunT (changes requested):
I don't know why CI is green when there are wrong program ids generation.

core/src/message/init.rs
mqxf: yes

pallets/gear/src/lib.rs
mqxf: Not sure how to replace this, might need to remove the check from here and put it elsewhere

pallets/gear/src/lib.rs
```diff
@@ -606,6 +606,8 @@ pub mod pallet {
             let code_and_id = CodeAndId::new(code);
             let code_info = CodeInfo::from_code_and_id(&code_and_id);
 
+            let program_id = ProgramId::generate(code_and_id.code_id(), &salt);
```
SabaunT, line 609: its' wrong. You create a message with another program id below (you call `from_packet` which uses not only salt and code_id, but message id)

pallets/gear/src/lib.rs
```diff
@@ -1233,13 +1227,19 @@ pub mod pallet {
...
                 expiration_block,
             );
 
-            let program_id = packet.destination();
+            let program_id = packet.destination(message_id, None);
+
+            ensure!(
+                !Self::program_exists(program_id),
+                Error::<T>::ProgramAlreadyExists
+            );
```
SabaunT, line 1241:
```suggestion
            if Self::program_exists(program_id) {
                unreachable!("Program id uniqueness rule violated: generated existing program id");
             }
```

SabaunT (commented):
please think of more representative names instead of `generate_for_pallet`/`generate_for_wasm`

core/src/ids.rs
```diff
@@ -210,6 +210,34 @@ impl ProgramId {
         let argument = [SALT, code_id.as_ref(), salt].concat();
         hash(&argument).into()
     }
+
+    /// Generate ProgramId from given CodeId, MessageId and salt
+    pub fn generate_for_pallet(code_id: CodeId, salt: &[u8], message_id: MessageId) -> Self {
```
breathx, line 215: we don't need that fn at all.
just rename already existing `generate` into `generate_from_user` and use it only on pallet

core/src/ids.rs
```diff
@@ -210,6 +210,34 @@ impl ProgramId {
...
+        hash(&argument).into()
+    }
+
+    /// Generate ProgramId from given CodeId, MessageId, salt and nonce
+    pub fn generate_for_wasm(
+        code_id: CodeId,
+        salt: &[u8],
+        message_id: MessageId,
+        nonce: usize,
+    ) -> Self {
```
breathx, line 228:
```suggestion
    pub fn generate_from_program(
        code_id: CodeId,
        salt: &[u8],
        message_id: MessageId,
    ) -> Self {
```
And keep duplicate init error here if salt matches

core/src/message/context.rs
```diff
@@ -278,20 +278,26 @@ impl MessageContext {
...
 
         let message_id = MessageId::generate_outgoing(self.current.id(), last);
-        let message = InitMessage::from_packet(message_id, packet);
+
+        let message =
+            InitMessage::from_packet(message_id, packet, Some(self.store.initialized.len()));
+
+        let program_id = message.destination();
+
+        if program_id == ProgramId::SYSTEM {
```
breathx, line 294: you don't need that check
and especially - here

core/src/message/init.rs
```diff
@@ -115,8 +115,6 @@ impl InitMessage {
 /// This structure is preparation for future init message sending. Has no message id.
 #[derive(Clone, Default, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Decode, Encode, TypeInfo)]
 pub struct InitPacket {
-    /// Newly created program id.
-    program_id: ProgramId,
```
breathx, line 119: why did you remove the field?

core/src/message/init.rs
```diff
@@ -161,8 +157,12 @@ impl InitPacket {
     }
 
     /// Packet destination (newly created program id).
-    pub fn destination(&self) -> ProgramId {
-        self.program_id
+    pub fn destination(&self, message_id: MessageId, nonce: Option<usize>) -> ProgramId {
```
breathx, line 160: 1) you need to have predefined id
2) you need to declare different constructors for in-program and in-extrinsics generations

core-errors/src/lib.rs
```diff
@@ -159,6 +159,10 @@ pub enum MessageError {
     /// An error occurs in attempt to charge gas for dispatch stash hold.
     #[display(fmt = "Not enough gas to hold dispatch message")]
     InsufficientGasForDelayedSending = 399,
+
+    /// An error when a syscall is used incorrectly
+    #[display(fmt = "Syscall was used incorrectly")]
```
breathx, line 164: Why do you need this error?

breathx (changes requested):

core-errors/src/lib.rs
mqxf: It was an error that already existed that was being used if the programid made was the system id, and I just renamed it.

core/src/message/init.rs
mqxf: programid needs messageid, so i cannot generate it until a messageid is also generated, this was the whole thing I was doing in this PR

core/src/message/init.rs
mqxf: I cannot have a predetermined id if it requires a messageId and nonce, this can only be called once those are already made, such as in the case of the InitMessage

core/src/message/context.rs
mqxf: It was a check that existed previously, I can remove it if you believe it is not needed

core-processor/src/ext.rs
```diff
@@ -308,6 +308,14 @@ impl BackendExternalities for Ext {
 }
 
 impl Ext {
+    fn check_forbidden_destination(&mut self, id: ProgramId) -> Result<(), FallibleExtError> {
```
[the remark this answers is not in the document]
mqxf: that was multiple commits, one removed and another readded, just in a different location by accident

core/src/message/init.rs
breathx: you need to use message id of the message during which processing create_program was called

core/src/message/context.rs
breathx: it's no longer needed so unreachable may be put on journal notes processing level but I don't see the real profit in it

core-errors/src/lib.rs
breathx: can be removed now, see our conversation above

core/src/message/init.rs
SabaunT: > I cannot have a predetermined id if it requires a messageId and nonce, this can only be called once those are already made, such as in the case of the InitMessage

You will have pre-defined id if you follow the logic described here https://github.com/gear-tech/gear/pull/3010#discussion_r1301975721. Not sure whether that will be possible for the pallet level. I guess 2 constructors serve that purpose.

NikVolf:
@breathx @SabaunT re-review please

core/src/ids.rs
```diff
@@ -204,38 +204,18 @@ impl ProgramId {
...
-
-    /// Generate ProgramId from given CodeId, MessageId, salt and nonce
-    pub fn generate_with_nonce(
-        code_id: CodeId,
-        salt: &[u8],
-        message_id: MessageId,
-        nonce: usize,
-    ) -> Self {
+    pub fn generate_with_id(code_id: CodeId, salt: &[u8], message_id: MessageId) -> Self {
         const SALT: &[u8] = b"program_for_wasm";
```
SabaunT, line 216: `b"program_from_wasm"`

core/src/ids.rs
```diff
@@ -204,38 +204,18 @@ impl ProgramId {
     pub const SYSTEM: Self = Self(*b"geargeargeargeargeargeargeargear");
 
     /// Generate ProgramId from given CodeId and salt
-    pub fn generate_raw(code_id: CodeId, salt: &[u8]) -> Self {
-        const SALT: &[u8] = b"program";
+    pub fn generate(code_id: CodeId, salt: &[u8]) -> Self {
+        const SALT: &[u8] = b"program_for_pallet";
```
SabaunT, line 208: `b"program_from_user"`

core/src/message/init.rs
```diff
@@ -46,10 +46,10 @@ pub struct InitMessage {
 
 impl InitMessage {
     /// Create InitMessage from InitPacket.
-    pub fn from_packet(id: MessageId, packet: InitPacket, nonce: Option<usize>) -> Self {
+    pub fn from_packet(id: MessageId, packet: InitPacket) -> Self {
         Self {
             id,
-            destination: packet.destination(id, nonce),
+            destination: packet.destination(Some(id)),
```
SabaunT, line 52: wrong id

core/src/message/context.rs
```diff
@@ -278,21 +278,22 @@ impl MessageContext {
...
-
         let last = self.store.outgoing.len() as u32;
 
         if last >= self.settings.outgoing_limit {
             return Err(Error::OutgoingMessagesAmountLimitExceeded);
         }
 
         let message_id = MessageId::generate_outgoing(self.current.id(), last);
+
         let message = InitMessage::from_packet(message_id, packet);
```
SabaunT, line 289: wrong id is used to create a program id

mqxf: What should be used then? self.current.id()? that seems like it would generate duplicates.

core/src/ids.rs
```diff
@@ -205,11 +205,19 @@ impl ProgramId {
...
 
         let argument = [SALT, code_id.as_ref(), salt].concat();
         hash(&argument).into()
     }
+
+    /// Generate ProgramId from given CodeId, MessageId and salt
+    pub fn generate_with_id(code_id: CodeId, salt: &[u8], message_id: MessageId) -> Self {
+        const SALT: &[u8] = b"program_from_wasm";
+
+        let argument = [message_id.as_ref(), SALT, code_id.as_ref(), salt].concat();
```
SabaunT, line 218:
```suggestion
        let argument = [SALT, message_id.as_ref(), code_id.as_ref(), salt].concat();
```

just for consistency

core/src/message/context.rs
```diff
@@ -278,20 +278,23 @@ impl MessageContext {
...
 
         if last >= self.settings.outgoing_limit {
             return Err(Error::OutgoingMessagesAmountLimitExceeded);
         }
 
         let message_id = MessageId::generate_outgoing(self.current.id(), last);
-        let message = InitMessage::from_packet(message_id, packet);
+
+        // Not sure if this is correct, it seems much more likely to generate duplicates
+        // if the same program is uploaded twice from the same parent program.
```
SabaunT, line 290: don't write such comments please
it can be a comment on PR, but not the part of code.

core/src/message/context.rs
```diff
@@ -278,20 +278,23 @@ impl MessageContext {
...
+        let message = InitMessage::from_packet(self.current.id(), packet);
```
SabaunT, line 291: wrong `message_id` is used to create `InitMessage`
```suggestion
        let message = InitMessage::from_packet(message_id, packet);
```

pallets/gear/src/lib.rs
```diff
@@ -638,6 +634,12 @@ pub mod pallet {
             let message_id = Self::next_message_id(origin);
             let block_number = Self::block_number();
 
+            let program_id = packet.destination(None);
```
SabaunT, line 637: move it up where it was.

pallets/gear/src/lib.rs
```diff
@@ -1227,14 +1225,19 @@ pub mod pallet {
...
             ExtManager::<T>::default().set_program(
-                packet.destination(),
+                program_id,
                 &code_info,
                 message_id,
                 expiration_block,
             );
 
-            let program_id = packet.destination();
+            if Self::program_exists(program_id) {
```
SabaunT, line 1237: same, move this check

core/src/message/init.rs
```diff
@@ -160,7 +177,7 @@ impl InitPacket {
         }
     }
 
-    /// Packet destination (newly created program id).
+    /// Program id.
```
SabaunT, line 180: The comment on master is more clarifying.

pallets/gear/src/lib.rs
```diff
@@ -1227,14 +1222,20 @@ pub mod pallet {
             let block_number = Self::block_number();
             let expiration_block = block_number.saturating_add(RentFreePeriodOf::<T>::get());
 
+            let program_id = packet.destination();
+
+            ensure!(
+                !Self::program_exists(program_id),
+                Error::<T>::ProgramAlreadyExists
+            );
+
```
SabaunT, line 1231: what about reverting that to be the same as on master?

SabaunT (approved):
Please apply suggestions from the review.

core/src/ids.rs
```diff
@@ -205,11 +205,19 @@ impl ProgramId {
 
     /// Generate ProgramId from given CodeId and salt
     pub fn generate(code_id: CodeId, salt: &[u8]) -> Self {
```
breathx, line 207:
```suggestion
    pub fn generate_from_user(code_id: CodeId, salt: &[u8]) -> Self {
```

core/src/ids.rs
```diff
@@ -205,11 +205,19 @@ impl ProgramId {
...
-        const SALT: &[u8] = b"program";
+        const SALT: &[u8] = b"program_from_user";
 
         let argument = [SALT, code_id.as_ref(), salt].concat();
         hash(&argument).into()
     }
+
+    /// Generate ProgramId from given CodeId, MessageId and salt
+    pub fn generate_with_id(code_id: CodeId, salt: &[u8], message_id: MessageId) -> Self {
```
breathx, line 215:
```suggestion
    pub fn generate_from_program(code_id: CodeId, salt: &[u8], message_id: MessageId) -> Self {
```

core/src/message/init.rs
```diff
@@ -131,9 +131,20 @@ pub struct InitPacket {
 
 impl InitPacket {
     /// Create new InitPacket without gas.
-    pub fn new(code_id: CodeId, salt: Salt, payload: Payload, value: Value) -> Self {
+    pub fn new(
+        code_id: CodeId,
+        salt: Salt,
+        payload: Payload,
+        message_id: Option<MessageId>,
```
breathx, line 138: please split it and `new_with_gas` into two different functions as it did for ProgramId generation

core/src/message/init.rs
```diff
@@ -49,7 +49,7 @@ impl InitMessage {
     pub fn from_packet(id: MessageId, packet: InitPacket) -> Self {
         Self {
             id,
-            destination: packet.program_id,
+            destination: packet.destination(),
```
breathx, line 52:
```suggestion
            destination: packet.program_id,
```

breathx: Just revert it back

pallets/gear/src/tests.rs
```diff
@@ -7114,107 +7120,6 @@ fn pay_program_rent_extrinsic_works() {
     });
 }
 
-#[test]
-fn test_create_program_duplicate() {
```
breathx, line 7118: we still can try to create duplicate program, revert test removal

breathx (changes requested):

core/src/message/init.rs
mqxf: Cannot, this will cause the `host` macro to break if I try to change these definitions.

breathx: As we discussed, don't use host macro for the case

mqxf: Should I remove the #[host] from the init packet function in funcs.rs and paste in what it would expand into? This would be inconsistent with all the other functions in funcs.rs. I think removing the host macro should be its own issue, or changing it such that it works with this, since that is also possible.

breathx: For me: Temporary copy-paste > bad design and usage of the structs
@SabaunT

SabaunT: `host` is to be dead, so let's postpone renaming until we remove `host`.

btw, @mqxf you can remove `host` in a separate PR.

SabaunT (approved):

breathx (approved):

PR #3025, opened 2023-08-03
refactor(pallet-gear,core-processor): Lazy-pages agnostic design

ark0f:


ark0f:
Non lazy-pages code must be removed in favor of future 4GB allocations

PR #3033, opened 2023-08-05
feat(tests, cli): Add subcommand to `gear-replay-cli` tool

ekovalev:
Existing CLI replays blocks that have already been finalized. However, if an extrinsic panicked at the time of block creation, it would have been dropped and not included in a block, thereby destroying the whole purpose of attempting to get some meaningful information about the root cause of the panic.

To circumvent that, a subcommand is being introduced that would mimic the steps of a block producer (not validator): initialize a block and apply extrinsic through calling the `BlockBuilder_apply_extrinsic` runtime api. Current implementation only supports the `Gear::run()` extrinsic, mainly because it is where the message queue is processed, but also because it doesn't have any arguments, therefore the known encoded version of it can be supplied to the runtime api call (without a need of creating an `UncheckedExtrinsic` by hand and then encoding it - that would require having the `Runtime` as an explicit dependency, which we want to avoid).

### DIFF
--- a/utils/gear-replay-cli/README.md
+++ b/utils/gear-replay-cli/README.md
@@ -108,7 +108,7 @@ General command format and available subcommand are:
 
 ```bash
     gear-replay-cli -h 
-    Commands of `gear-reply` CLI
+    Commands of `gear-replay` CLI
 
     Usage: gear-replay-cli [OPTIONS] <COMMAND>
 
--- a/utils/gear-replay-cli/README.md
+++ b/utils/gear-replay-cli/README.md
@@ -104,12 +104,33 @@ Another difference from the `try-runtime` API is that in `gear-replay-cli` the b
 
 In order to use the native runtime build make sure the node is built with the Runtime spec version that matches the one currently uploaded on chain.
 
-- Replay a block on Vara live chain
+General command format and available subcommand are:
+
+```bash
+    gear-replay-cli -h 
+    Commands of `gear-reply` CLI
+
+    Usage: gear-replay-cli [OPTIONS] <COMMAND>
+
+    Commands:
+    replay-block  Replay block subcommand
+    gear-run      GearRun subcommand
+    help          Print this message or the help of the given subcommand(s)
+
+    Options:
+    -u, --uri <URI>            The RPC url [default: wss://archive-rpc.vara-network.io:443]
+    -l, --log [<NODE_LOG>...]  Sets a custom logging filter. Syntax is `<target>=<level>`, e.g. -lsync=debug
+    -h, --help                 Print help (see more with '--help')
+```
+
+Currently supported cases include:
+
+- Replaying a block on a live chain
 
   - current latest finalized block on Vara chain
 
     ```bash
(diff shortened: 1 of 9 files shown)


ukint-vs (approved):
Great. Works as expected.

utils/gear-replay-cli/README.md
```diff
@@ -108,12 +108,33 @@ Another difference from the `try-runtime` API is that in `gear-replay-cli` the b
 
 In order to use the native runtime build make sure the node is built with the Runtime spec version that matches the one currently uploaded on chain.
 
-* Replay a block on Vara live chain
+General command format and available subcommand are:
+
+```bash
+    gear-replay-cli -h 
+    Commands of `gear-reply` CLI
```
shamilsan, line 115: Typo?
```suggestion
    Commands of `gear-replay` CLI
```

ekovalev: I think the name of the tool should be changed again, now that it not only replays blocks but applies individual extrinsics and soon will be useful also for more precise gas calculation. Good catch anyway.

utils/gear-replay-cli/src/lib.rs
```diff
@@ -0,0 +1,116 @@
+// This file is part of Gear.
+
+// Copyright (C) 2021-2022 Gear Technologies Inc.
```
breathx, line 3: copyrights

utils/gear-replay-cli/src/cmd/replay_block.rs
```diff
@@ -0,0 +1,179 @@
...
+    let block = fetch_block::<Block>(&rpc, current_hash).await?;
+
+    // A digest item gets added when the runtime is processing the block, so we need to pop
+    // the last one to be consistent with what a gossiped block would contain.
+    let (mut header, extrinsics) = block.deconstruct();
+    header.digest_mut().pop();
+    let block = Block::new(header, extrinsics);
+
+    // Create executor, suitable for usage in conjunction with the preferred execution strategy.
+    #[cfg(all(not(feature = "always-wasm"), feature = "vara-native"))]
```
breathx, line 134: do we panic in cases of vara and gear native at the same time

breathx (approved):
Remove cache step from win CI

utils/gear-replay-cli/src/cmd/replay_block.rs
ekovalev: Well, eventually it will panic if the executor doesn't match the actual runtime being used.
I don't think we should be over-protective here as this is not a production code and we rely on the caller to know what they are doing.

breathx: I'd add comment about this for such places only then

NikVolf:
@ekovalev rebase to master and rerun CI please

utils/gear-replay-cli/src/cmd/replay_block.rs
ekovalev: On second thoughts, we shouldn't even have a panic in such case: if the native runtime version (expected by the actual native executor) doesn't match the onchain version from the downloaded runtime, the "fallback" WASM executor will be used.

PR #3035, opened 2023-08-06
refactoring(gsdk): structure gsdk API

### LINKED ISSUE #2895: Structure gsdk APIs
I'm not pretty sure that it affects api positively. I'd go with
`signer.calls.transfer`
and
`signer.storage.number`
to split functionality and make it more readable + intuitive

_Originally posted by @breathx in https://github.com/gear-tech/gear/pull/2744#discussion_r1249324057_

there could be another solution, wrapping storage and tx into traits

```rust
{ // use storage query
  use gsdk::Storage;

  api.number().await?;
}

{ // run transaction
  use gsdk::Transaction;

  api.transfer(..).await?;
}
```

mertwole:
Resolves #2895


gcli/src/cmd/send.rs
```diff
@@ -56,6 +56,7 @@ pub struct Send {
 impl Send {
     pub async fn exec(&self, signer: Signer) -> Result<()> {
         signer
+            .calls
```
breathx, line 59: why send_message is from "calls", but `transfer` is from `.balance`?

gsdk/src/signer/mod.rs
```diff
@@ -17,22 +17,55 @@
...
+/// [`Signer::balance`], [`Signer::calls`], [`Signer::rpc`].
 #[derive(Clone)]
 pub struct Signer {
+    signer: Arc<SignerInner>,
+    /// Calls that require sudo.
+    pub sudo: SignerSudo,
+    /// Calls to interact with account balance.
+    pub balance: SignerBalance,
+    /// Calls for interaction with on-chain programs.
+    pub calls: SignerCalls,
```
clearloop, line 50: sudo and balances could be merged in to calls

gsdk/src/signer/mod.rs
```diff
@@ -17,22 +17,55 @@
...
+    /// Calls to fetch data from node.
+    pub rpc: SignerRpc,
+}
+
+#[derive(Clone)]
+pub struct SignerSudo(Arc<SignerInner>);
```
clearloop, line 56: docs plz

gcli/src/cmd/send.rs
mertwole: I think all the calls intended to communicate with programs should be placed in `calls` and all the calls working with balance in `balance`

gsdk/src/signer/mod.rs
```diff
@@ -17,22 +17,55 @@
...
 mod utils;
 
+/// Signer representation that provides access to gear API.
+/// Implements low-level methods such as [`run_tx`](`SignerInner::run_tx`)
+/// and [`force_batch`](`SignerInner::force_batch`).
+/// Other higher-level calls are provided by [`Signer::sudo`],
+/// [`Signer::balance`], [`Signer::calls`], [`Signer::rpc`].
 #[derive(Clone)]
 pub struct Signer {
+    signer: Arc<SignerInner>,
```
clearloop, line 44:
```rust
pub struct Signer {
  inner: Arc<RefCell<Inner>>
  // ...
}
```

gsdk/src/signer/mod.rs
```diff
@@ -42,32 +75,72 @@ pub struct Signer {
 impl Signer {
     /// New signer api.
     pub fn new(api: Api, suri: &str, passwd: Option<&str>) -> Result<Self> {
-        Ok(Self {
+        let signer = SignerInner {
```
clearloop, line 78:
```rust
let signer = PairSigner::new(
    Pair::from_string(suri, passwd).map_err(|_| Error::InvalidSecret)?,
);

self.inner.get_mut() = *signer;
```

gsdk/src/signer/mod.rs
```diff
@@ -42,32 +75,72 @@ pub struct Signer {
...
+        let signer = Arc::new(signer);
+
+        Self {
+            sudo: SignerSudo(signer.clone()),
+            balance: SignerBalance(signer.clone()),
+            calls: SignerCalls(signer.clone()),
+            rpc: SignerRpc(signer.clone()),
+            signer,
+        }
+    }
```
clearloop, line 99: since we are using `Arc` now, don't have to clone everything again, better find a way applying the changes via just updating the inner

clearloop (commented):
The idea of using smart pointer looks great! but however the usage seems not that correct lol, besides, @breathx mentioned, balances and sudo belongs to calls as well

gsdk/src/signer/mod.rs
mertwole: Even if I move `transfer` to the calls then there still remains `get()` and `log_spent()` fns. And why `sudo` should be merged into calls? These calls are belong to different use-cases and the border between them is obvious

gsdk/src/signer/mod.rs
mertwole: But how do I implement `Deref<SignerInner>` for Signer?

gcli/src/cmd/send.rs
breathx: Isn't it better to split them by pallet names?

gsdk/src/signer/mod.rs
clearloop:
```rust
impl Deref for Signer {
  target = Inner;

  fn deref(&self) -> &Inner {
    self.inner.borrow()
  }
}
```

gsdk/src/signer/mod.rs
breathx: how do you feel about
* ...pallets
    * calls
    * storage
* rpc

gcli/src/cmd/send.rs
mertwole: Hmm, But will it make sense for API user that's not familiar with gear pallet organisation? There will be 7 fields in `Signer`, 1 per pallet, + `utils` + `rpc`

gsdk/src/signer/mod.rs
clearloop: > how do you feel about

hmmm, feeling this discussion is endless, but sure we can write some macro to implement it, I'm more closed to 

storage.pallet_name.query because it is synced with the polkadot.js API

mertwole: > currently, imo just introducing calls/storage/rpc is enough ))

where do I put `log_balance_spent()` then?

gsdk/src/signer/mod.rs
clearloop: try `self.inner.clone().borrow()`, I remember there is a solution for it, it doesn't matter we cloning it since we are using smart pointer

gsdk/src/signer/mod.rs
mertwole: > storage.pallet_name.query

But why do we need to let user know about pallet existence?

clearloop: > where do I put log_balance_spent() then?

I got you now, mb we can move this method to `Singer` or just give it to `gclient`

```rust
struct Signer {
  inner
}
```

because Signer has the access to the pallet balances and all of other calls

gsdk/src/signer/mod.rs
mertwole: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=0185a54060d277dda11ff8d85ee5e5d3

clearloop: we can re-export these APIs by introducing new methods momentary, keep all of the current tests working, for the solution of `Deref`, we create a new issue for it

mertwole: I think this approach is good enough because 5x Arc clones is a little overhead (and occurs not so often) and reference integrity is provided by this https://github.com/gear-tech/gear/blob/5b63e54c421cd71725a818b3e95c4616df3c4d22/gsdk/src/signer/mod.rs#L99

clearloop: hmmm, makes sense tbh, however we can leave some docs on it for the future refactoring is necessary

mertwole: You mean docs on line 99?

gsdk/src/signer/calls.rs
```diff
@@ -235,7 +236,7 @@ impl SignerInner {
 }
 
 // pallet-system
-impl SignerSudo {
+impl SignerStorage {
```
clearloop, line 239: mb move this kind of implementations to a new file like `storage.rs`? and please create issues for the problems we discussed in the comments

clearloop (approved):

gsdk/src/signer/mod.rs
clearloop: yeah 

- `L42`: sth like `NOTE: figure a better layout for this struct in the future`
- `L99`: sth like `TODO: split calls into pallets`

ukint-vs (approved):

gsdk/src/signer/rpc.rs
```diff
@@ -82,7 +98,8 @@ impl Signer {
         allow_other_panics: bool,
         at: Option<H256>,
     ) -> Result<GasInfo> {
-        self.api
+        self.0
+            .api
             .calculate_handle_gas(
```
breathx, line 103: mb go with .api.rpc.calculate_gas... ?

breathx: for code consistency with calls and storage

breathx (approved):

PR #3043, opened 2023-08-08
fix calc gas: enable process queue before calculation

grishasobol:


pallets/gear/src/lib.rs
```diff
@@ -770,20 +770,28 @@ pub mod pallet {
...
                     kind.clone(),
                     initial_gas,
                     payload.clone(),
                     value,
                     allow_other_panics,
                     allow_skip_zero_replies,
-                )
-                .map_err(|e| {
+                );
+                GasAllowanceOf::<T>::put(gas_allowance);
```
breathx, line 787: same for QueueProcessing?

breathx (approved):

pallets/gear/src/lib.rs
ukint-vs: yeah

ukint-vs (approved):

clearloop:
workflow build not triggered is caused by https://github.com/gear-tech/gear/issues/3048, need 1 more empty commit or we can dispatch it in actions after #3049

PR #3045, opened 2023-08-09
fix(gstd): log the panic message for the debugging convenience

fluiderson:
#2200 broke the the printing of panic messages in `gtest` tests due to removing a log function to do so. This PR adds it back for debug builds just before calling `gr_panic`.


clearloop (approved):

shamilsan (approved):
Great catch 🔥

PR #3047, opened 2023-08-09
refactor(gstd): align gstd reexports & prelude items with `std`

fluiderson:
In `gstd`'s `prelude` module, the following is stated:
https://github.com/gear-tech/gear/blob/fdb21befac6b7dcd1b07f1cc035058c0cdfe870d/gstd/src/prelude.rs#L19-L20
That's not entirely true. Several modules and types, which can be found in `std`'s `prelude` module, are missing in `gstd`'s one. So, in this PR, I refactored the `prelude` module to make `gstd` align to `std` as much as possible.

The only problem here is that I'd like to leave reexports in this format for easier alignment maintaining with `std`, but `rustfmt` is configured to merge `use` statements. For now, I removed `imports_granularity = "Crate"` from `rustfmt.toml`, but the final solution needs to be discussed.


rustfmt.toml
```diff
@@ -1,2 +1 @@
 edition = "2021"
-imports_granularity = "Crate"
```
shamilsan, line 2: Why has this option been removed?

fluiderson: > The only problem here is that I'd like to leave reexports in this format for easier alignment maintaining with `std`, but `rustfmt` is configured to merge `use` statements. For now, I removed `imports_granularity = "Crate"` from `rustfmt.toml`, but the final solution needs to be discussed.

gstd/src/prelude.rs
```diff
@@ -19,25 +19,81 @@
...
+}
+pub use ::alloc::fmt;
+pub use ::alloc::rc;
+pub use ::alloc::str;
+pub use ::alloc::string;
+pub use ::alloc::vec;
+pub use core::future;
+pub use core::hash;
+pub use core::hint;
+pub use core::iter;
```
grishasobol, line 69: What the reason to look like std prelude? How this can help smart contracts developers? Also you need in this case to sort it by hands

grishasobol: Adding additional re-exporting is good. My question is only about fmt

fluiderson:
At first, I thought it'd be more convenient to maintain it this way if new modules would be added in `core`/`alloc`, but now I think it's easier to merge reimports than to try to exclude them from the common formatting rules.

@ark0f, glob imports create ambiguity because there are a lot of modules with the same name from `core` and `alloc`. Also I recreated some modules (e.g. `collections`), and they conflict with glob imports too.

grishasobol (approved):

ukint-vs (approved):

examples/fungible-token/io/src/lib.rs
```diff
@@ -20,7 +20,8 @@
 
 use core::ops::Range;
 use gmeta::{In, InOut, Metadata, Out};
-use gstd::{codec::MaxEncodedLen, prelude::*, ActorId};
+use gstd::{prelude::*, ActorId};
+use parity_scale_codec::MaxEncodedLen;
```
breathx, line 24: why not reimporting from codec?

examples/waiter/src/code.rs
```diff
@@ -4,13 +4,13 @@ use crate::{
...
 use futures::future;
-use gstd::{errors::Error, exec, format, lock, msg, MessageId};
+use gstd::{ exec, format, sync, msg};
 
-static mut MUTEX: lock::Mutex<()> = lock::Mutex::new(());
-static mut MUTEX_LOCK_GUARD: Option<lock::MutexGuard<()>> = None;
-static mut RW_LOCK: lock::RwLock<()> = lock::RwLock::new(());
-static mut R_LOCK_GUARD: Option<lock::RwLockReadGuard<()>> = None;
-static mut W_LOCK_GUARD: Option<lock::RwLockWriteGuard<()>> = None;
+static mut MUTEX: sync::Mutex<()> = sync::Mutex::new(());
```
breathx, line 9: Best practice to import directly the type from mod.
```suggestion
static mut MUTEX: Mutex<()> = Mutex::new(());
```

gstd/src/prelude.rs
```diff
@@ -19,25 +19,50 @@
...
+///
+/// [`alloc::ffi`]: ::alloc::ffi
+pub mod ffi {
+    pub use ::alloc::ffi::*;
+    pub use core::ffi::*;
+}
+
+// Reexports from third-party libraries
+
+pub use parity_scale_codec::{self as codec, Decode, Encode};
```
breathx, line 67: MaxEncodedLen, Encode(/Decode)Like

breathx (approved):

PR #3050, opened 2023-08-09
refactor(gstd): split `prog::create_program` into two functions

### LINKED ISSUE #3026: Proposal: `create_program` should have `<E: Encode>`
### File Location(s)

gstd/src/prog/mod.rs

### Proposal

We have `msg::send_bytes()` function that accepts bytes to send to some address. However, for some reason `create_program` accepts bytes, but doesn't have a proper name.

We could do the following:
1. Rename `create_program` to `create_program_bytes`
2. Add `create_program<E: Encode>()`
3. Apply stack optimizations for `prog::create_program*` functions

StackOverflowExcept1on:
Resolves #3026

### DIFF
--- a/examples/constructor/src/call.rs
+++ b/examples/constructor/src/call.rs
@@ -68,7 +68,7 @@ mod wasm {
             let delay = delay.value();
 
             let res = if let Some(gas_limit) = gas_limit {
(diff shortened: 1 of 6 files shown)


shamilsan (approved):
Now it looks more consistent, great!

PR #3055, opened 2023-08-10
feat(manifest): trim manifest with wildcard

clearloop:
Resolves # .

merged the field `members` in the root manifest

- [x] core-backend
- [x] node
- [x] runtime
- [x] utils

@gear-tech/dev


tltsutltsu (approved):

Cargo.toml
```diff
@@ -13,63 +13,10 @@ members = [
...
-    "examples/sys-calls",
-    "examples/syscall-error",
-    "examples/vec",
-    "examples/wait",
-    "examples/wait-timeout",
-    "examples/wait_wake",
-    "examples/waiter",
-    "examples/waiting-proxy",
-    "examples/wat",
+    "examples/[!big-wasm]*",
```
shamilsan, line 19: We had trouble with wildcards in `examples` some time ago. When you have some new example in a branch, you get a compilation error when returning to the master branch as you have a new example directory (with `.binpath`), but no manifest inside it.

Cargo.lock
```diff
@@ -2053,15 +2053,6 @@ dependencies = [
...
 ]
 
-[[package]]
-name = "demo-ncompose"
-version = "0.1.0"
-dependencies = [
- "gear-wasm-builder",
- "gstd",
- "hex",
-]
```
shamilsan, line 2063: Some examples seem to be accidentally removed.

Cargo.toml
clearloop: mb we can  `examples/[!big-wasm|!.binpath]` when meet that problem?

Cargo.lock
clearloop: yeah just fount it is pretty buggy, considering creating another issue for the examples

clearloop: https://github.com/gear-tech/gear/issues/3069

Cargo.toml
```diff
@@ -83,29 +80,11 @@ members = [
     "gstd",
     "gsys",
     "gtest",
-    "node/cli",
-    "node/service",
-    "node/authorship",
-    "node/testing",
+    "node/[!res]*[!md]",
```
shamilsan, line 83: I'd leave the old list for `node` as this `[!res]*[!md]` looks not too cool to me.

shamilsan (approved):

Cargo.toml
clearloop: but this is a one-line solution )) mb create a poll : D

clearloop: leave it to #3069

PR #3056, opened 2023-08-10
chore(common): rename common to pallets

### LINKED ISSUE #2869: Proposal: Consider re-organizing core packages
### File Location(s)

_No response_

### Proposal

now

```
/common
/core
/core-backend
/core-errors
/core-processor
```

proposal

```
pallets/common
core                # re-export everything
core/common         # the previous `/core`
core/backend
core/errors
core/processor
```

clearloop:
Resolves #2869 

since the package common is only used by pallets and runtime related stuffs, we can just move it to pallets, since this naming could be misguided for other packages under `/`

since `core-*` belong to `core`, just move them into it

- [x] `common` -> `pallets/common`
- [x] `core-*` -> `core/*`

@gear-tech/dev


breathx (changes requested):
`gear-common` isn't aligned to only pallets logic. This just a temporary coincidence. That is common parts and traits of gear internal primitives that will be used for a lot of other test toolings as well

breathx:
To be re-opened once decided to implement

Issue #3058, opened 2023-08-11
move all simple demos in demo-collection


grishasobol:
### File Location(s)

examples/collection/src/lib.rs

### Proposal

Reduce amount of demos in examples using demo collection.

PR #3059, opened 2023-08-11
feat!: append payload in `read_state` rpc call

### LINKED ISSUE #3006: Proposal: expand `extern "C" fn state()` and `read_state` functionality
### Problem to Solve

User has no ability to get some specific part of state just using `read_state`. User need to use `read_state_using_wasm` which is much more difficult.
Other programs may be also wanna to get some information about state of another program. You may ask me: why we cannot just use `send_for_reply()`?  The problem is that executing export `handle()` may cause changes in program state. Even with using stack allocations, program can make hidden allocations in external crates. So, it can be really difficult task to remove all heap allocations, which always cause read/writes in storage.

### Possible Solution

1) For user we just need to add payload param into read_state api:
```rust
fn read_state(program_id: H256, payload: Vec<u8>) -> Result<Vec<u8>, Vec<u8>>;
```
And use this payload as msg payload into ext during `state()` execution . With that user possible to use `msg::load()` to identify what exactly from state need to be replied:
```rust

enum StateAction {
    All,
    OneUser(AccountId),
    …
}

extern "C" fn state(){
   let state_action = msg::load().unwrap();
   match state_action {
      StateAction::All => { ... },
      StateAction::OneUser(user_id) => { ... },
      ...
   }
}
```

2) After export state() execution, program state is not changed in storage, so, this can be much more cheaper than using the same logic thru `handle()`. We can make one new sys-call:
```rust
    pub fn gr_request_state(
        program_id: *const Hash,
        payload: *const BufferStart,
        len: Length,
        delay: BlockNumber,
        err_mid: *mut ErrorWithHash,
    );
```
This will cause dispatch `State` creation in message queue. Which will execute `state` export with all weights like in handle execution except costs for pages writes in storage.

### Notes

_No response_

grishasobol:
Part of #3006

1) Append payload argument in read_state rpc call and all places where it uses including gclient and gsdk
2) In order to test this patch, but do not increase demos amount I have made new demo : demo-collection. This demo can be configured  to act like any demo inside it's collection. Currently collection is btree, capacitor and backend-error. It can be simply extended. (see also #3058)


Cargo.toml
```diff
@@ -364,12 +362,10 @@ demo-async-init = { path = "examples/async-init" }
 demo-async-recursion = { path = "examples/async-recursion" }
 demo-async-signal-entry = { path = "examples/async-signal-entry" }
 demo-async-tester = { path = "examples/async-tester" }
-demo-backend-error = { path = "examples/backend-error" }
-demo-btree = { path = "examples/btree" }
 demo-calc-hash = { path = "examples/calc-hash" }
 demo-calc-hash-in-one-block = { path = "examples/calc-hash/in-one-block" }
 demo-calc-hash-over-blocks = { path = "examples/calc-hash/over-blocks" }
-demo-capacitor = { path = "examples/capacitor" }
+demo-collection = { path = "examples/collection" }
```
grishasobol, line 368: May be better to name it configurable.

shamilsan (approved):
Awesome 🔥 

As a suggestion, I'd make a separate PR for such things as `capacitor` -> `collection` in the future.

examples/collection/src/lib.rs
```diff
@@ -0,0 +1,87 @@
...
+    unsafe { STATE = Some(state) };
+}
+
+#[no_mangle]
+extern "C" fn handle() {
+    let state = unsafe { STATE.as_mut().expect("State must be set in handle") };
+    match state {
+        State::Capacitor(state) => capacitor::handle(state),
+        State::BTree(state) => btree::handle(state),
+        _ => {}
```
breathx, line 77: better to panic like `unimplemented!()`

breathx: for all match cases within the main demo

breathx (approved):
@osipov-mit please implement such logic for JS side, it break compatibility!

Cargo.toml
shamilsan: Or `demo-custom`

examples/collection/src/lib.rs
grishasobol: No, because if we will take contract which does not have handle then no panic will happen. If some contract have panic in handle then this is separate and special logic

grishasobol: By default we suppose logic like without handle export

ukint-vs (approved):

shamilsan:
@grishasobol Is there something blocking it? We need some time to update `gear-js` etc., to reflect the new RPC signature and feature.

shamilsan:
Should be changed to `Types`: https://github.com/gear-tech/gear/blob/master/gmeta/src/lib.rs#L463

shamilsan (approved):
Ideally

shamilsan:
> Ideally

I mean "bezuprechno"

ukint-vs (approved):

shamilsan:
Please don't forget to increment the metadata version: https://github.com/gear-tech/gear/blob/master/gmeta/src/lib.rs#L228

PR #3068, opened 2023-08-15
refactor(sandbox): Extract dispatch_common function in wasmer backend

### LINKED ISSUE #3067: Unify dispatch_function{, _v2} in wasmer_backend
### File Location(s)

_No response_

### Proposal

extract common generic function and introduce a trait for (Wasm)ReturnValue

gshep:
Resolves #3067 

@gear-tech/dev

### DIFF
--- a/sandbox/host/src/sandbox/wasmer_backend.rs
+++ b/sandbox/host/src/sandbox/wasmer_backend.rs
@@ -114,15 +114,7 @@ pub fn invoke(
         .get_function(export_name)
         .map_err(|error| Error::Sandbox(error.to_string()))?;
 
-    let args: Vec<wasmer::Val> = args
-        .iter()
-        .map(|v| match *v {
-            Value::I32(val) => wasmer::Val::I32(val),
-            Value::I64(val) => wasmer::Val::I64(val),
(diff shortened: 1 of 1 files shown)


ukint-vs (approved):

sandbox/host/src/sandbox/wasmer_backend.rs
```diff
@@ -133,22 +125,13 @@ pub fn invoke(
...
-                    return Err(Error::Sandbox(format!(
-                        "Unsupported return value: {:?}",
-                        wasm_value,
-                    )))
-                }
-            };
-
-            Ok(Some(wasmer_value))
-        }
+        [wasm_value] => match into_value(wasm_value) {
```
shamilsan, line 128: Don't you like [`ok_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else)?

shamilsan (approved):
Good job 💯

sandbox/host/src/sandbox/wasmer_backend.rs
gshep: yep, I do. However it requires to return Ok(Some(_)) if it is some not just Ok(_).

PR #3070, opened 2023-08-15
ci(label): trigger workflows on labelling 

### LINKED ISSUE #2888: More conditions for the build job of workflow Labels
### Problem to Solve

once #2878 merged, there will be a problem, each time we add `A0` will trigger the build workflow whatever if it has already been triggered in PR

https://github.com/gear-tech/gear/pull/2959#discussion_r1268936757

### Possible Solution

use Github API to trigger the workflow

```js
const {
  data: { total_count: count, workflow_runs: runs },
} = await github.rest.actions.listWorkflowRuns({
  owner: context.repo.owner,
  repo: context.repo.repo,
  workflow_id: "PR.yaml",
});

// Get the latest workflow run.
const cis = runs
  .filter((r) => r.name == "Build")
  .sort((a, b) => b.number - a.number);
if (cis.length == 0) return;
const {
  data: { jobs },
} = await github.rest.actions.listJobsForWorkflowRun({
  owner: context.repo.owner,
  repo: context.repo.repo,
  run_id: cis[0].id,
});

// Get job ids and trigger build jobs
const builds = jobs
  .filter((j) =>j.name == "build" && j.conclusion == "skipped")
  .sort((a, b) => b.run_attempt - a.run_attempt);

if (builds.length != 1) return;
// dispatch build
```

### Notes

_No response_

clearloop:
Resolves #2888

1. list all [check runs][0] for the PR
2. find if the latest check has processed workflow `build`
3. dispatch workflow with github API
4. create checks from the dispatched workflow
5. update the status of the checks till they are completed

- [x] trigger build on `A0-pleasereview`
- [x] dispatch workflow and create specified build checks
- [x] maintain the status of the new created checks
- [x] not trigger workflow build when it has already been triggered

---

Why create checks instead of just triggering the workflows with labels? 

Triggering the workflows directly will cause the workflows be triggered on each labelling and even if we can skip triggering on `non-A0/E2` labels, the state of the check suite will be reset which means our build checks will be disappeared as well

@gear-tech/dev 

[0]: https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#list-check-runs-for-a-git-reference


clearloop:
`A0-pleasereview` triggered build successfully, now, re-labeling `A0-pleasereview` to see if it re-trigger building

clearloop:
details_url doesn't work: https://github.com/orgs/community/discussions/26757

clearloop:
test label

clearloop:
re-labelling

clearloop:
great, the build processes are not re-triggered after re-labelling `A0-pleasereview`

clearloop:
show case here:

push without `A0-pleasereview` in PR, now labelling `A0-pleasereview`

shamilsan:
Did you check how it works with `[skip ci]` prefix?

clearloop:
> Did you check how it works with `[skip ci]` prefix?

It follows the rules of the provided skip CI implementation https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs since the label workflow is `on: pull_request`

.github/actions/label/build.js
```diff
@@ -0,0 +1,179 @@
...
+ */
+const dispatchWorkflow = async ({ core, github }) => {
+  await github.rest.actions.createWorkflowDispatch({
+    owner,
+    repo,
+    workflow_id,
+    ref: REF,
+  });
+
+  await sleep(10000);
```
shamilsan, line 77: Why sleep? And how this value had been chosen?

shamilsan (approved):

.github/actions/label/build.js
clearloop: There will be short delays inside the API service of gtihub after dispatching the workflow, 10s is insanely safe, in the next step `listWorkflowRuns` requires workflow info from the API service of github

PR #3071, opened 2023-08-15
increase max size of program memory from 32 MB to 4 GB

grishasobol:


common/src/paused_program_storage.rs
```diff
@@ -40,7 +35,8 @@ impl From<(BTreeSet<WasmPage>, H256, MemoryMap)> for Item {
     fn from(
         (allocations, code_hash, mut memory_pages): (BTreeSet<WasmPage>, H256, MemoryMap),
     ) -> Self {
-        let remaining_pages = memory_pages.split_off(&GearPage::from(SPLIT_COUNT));
+        let split_page: GearPage = WasmPage::from(MAX_WASM_PAGE_COUNT / 2).to_page();
```
grishasobol, line 38: @gshep please see this place. I cannot understand why did we need MAX_WASM_PAGE_COUNT here. May be we can just use some in-depended constant WasmPage(256) ?

core-processor/src/executor.rs
```diff
@@ -215,9 +215,6 @@ fn get_pages_to_be_updated<ProcessorExt: ProcessorExternalities>(
     if ProcessorExt::LAZY_PAGES_ENABLED {
         // In lazy pages mode we update some page data in storage,
         // when it has been write accessed, so no need to compare old and new page data.
-        new_pages_data.keys().for_each(|page| {
```
grishasobol, line 218: Just remove this annoying trace. Better to make it may be in journal notes

common/src/paused_program_storage.rs
gshep: That is actually why we cannot increase memory limit. Will discuss it

grishasobol:
Close until problem with program pausing is not solved

PR #3102, opened 2023-08-18
feat(runtime-fuzzer): Introduce "precise" `runtime-fuzzer`

### LINKED ISSUE #2759: Precise fuzz-testing
### Problem to Solve

So I have learnt that our fuzzer has quite good coverage, but still has some problems which make fuzz testing ineffective.

The first and the most major problem is that generated data is just a `u64` seed. The seed is used in rng later to create random data inputs for the fuzzer. Of course, mutating this 8 bytes and then shrinking them is just a waste of CPU and fuzzer's internal memory (as libFuzzer is in-memory fuzzer). So the generated by fuzzer data must be actually a random input.
Solving this problem by implementing `Arbitrary` on `GearCall` (https://github.com/gear-tech/gear/issues/2314) will not give the expected result. That's because our fuzzer depends on persistent data. Each generated `GearCall` is executed in the context of the previous `GearCall`, which is stored in the static variable. So to make shrinking and reproduction successful we have to store the context on disk.
But there's problem with storing data too. Fuzzer won't create a set of calls to reproduce an error, as input data is just a `GearCall`. We won't have minimal test case to lead persistent context to such state that the next `GearCall` will make runtime panic. It's a problem when you run hunderds of thousands of `GearCall`s to lead state to panic and you don't know which subset of these calls is really needed to reproduce failure.

### Possible Solution

So instead of trying to save the context and state between newly generated calls we can use another approach, which is used in fuzzing by parity. We can create a temporary state with our state transition functions. All state transitions within this temporary state are checked for unwanted errors and panics.
More precisely, a vector of `GearCall`s with a valid structure will be generated and processed by `pallet_gear`. By valid structure I mean most of `GearCall`s will address existing programs and mailbox messages and etc.

### Notes
After researching usage practices of fuzzers, I have found out that having a focused on "some part of the code\logic" fuzzer is a widely spread solution. It gives us the simplicity of input data and less areas to be covered. As a result, we can get a higher quality of shrinking and a better fuzzing speed.

From my perspective, there are only 2 main areas to be checked by precise fuzzers: memory safety (safety against segmentation faults, overflows and underflows, invalid reads/writes and such) and business logic (storage invariants, gas and value economics).
*Should be stated, that the latter solution also gives an ability to test these 2 areas.

Alternatively we can have 2 fuzzers focused on these 2 areas:
##### Executor level fuzzer
This fuzzer mainly generates data for the `core_processor::execute` function applying some rules for generated inputs (for example, if you have generated `IncomingDispatch` with some `Context`, then `GasReserver` can also be generated non-empty). 

##### Processor level fuzzer
This fuzzer generates "semantically" valid vector of `JournalNote`s that will reproduce the execution of the program. Later this vector is processed, which will cause changes in `pallet-gear` storage. State transition must not panic and meet "cross-storage" invariants.

SabaunT:
Resolves #2759  .

- [x] Adjust fuzz-target so fuzzing becomes more precise
- [x] Adjust `check-fuzzer` task (corresponding changes in scripts)

@gear-tech/dev


utils/runtime-fuzzer/fuzz/fuzz_targets/main.rs
```diff
@@ -19,53 +19,13 @@
...
+fuzz_target!(|gear_calls: GearCalls| {
     gear_utils::init_default_logger();
 
-    dump_seed(seed).unwrap_or_else(|e| unreachable!("internal error: failed dumping seed: {e}"));
-
-    log::info!("Running the seed {seed}");
-    runtime_fuzzer::run(seed);
+    // Newline to easily browse logs.
+    println!();
+    log::info!("Executing generated gear calls");
```
shamilsan, line 29: You mix `println` (stdout) and `log::info` (may be filtered and redirected). Is it good? Maybe just add `\n` before the text in `log::info`?

utils/runtime-fuzzer/src/arbitrary_call.rs
```diff
@@ -42,9 +39,23 @@ static_assertions::const_assert!(MAX_PAYLOAD_SIZE <= gear_core::message::MAX_PAY
...
+/// That's done because when fuzzer finds a crash it prints a [`Debug`] string of the [`GearCalls`].
+/// Fuzzer executes [`GearCalls`] with pretty large codes and payloads, therefore to avoid printing huge
+/// amount of data we do a mock implementation of [`Debug`].
+///
+/// If one wants to see a real debug string of the type, a separate wrapper over [`GearCall`]s array must be
+/// implemented. This wrapper must implement [`Debug`] then.
+impl Debug for GearCalls {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("GearCalls")
+            .field(&"Mock `Debug` impl")
```
shamilsan, line 54: Should the `field` be `&&str`? Doesn't `&str` work?

shamilsan (approved):

utils/runtime-fuzzer/fuzz/fuzz_targets/main.rs
mertwole: this code snipped can also be found at the other places in this PR

scripts/check-fuzzer.sh
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env sh
 
 main() {
+    echo " >> Getting random bytes from /dev/urandom"
+    mkdir -p utils/runtime-fuzzer/fuzz/corpus/main
+    dd if=/dev/urandom of=utils/runtime-fuzzer/fuzz/corpus/main/check-fuzzer-bytes bs=1 count=27000000
```
mertwole, line 6: Please describe here why we should fill this file with random bytes

mertwole (approved):

utils/runtime-fuzzer/src/arbitrary_call.rs
SabaunT: It expects `&dyn Debug`, so `&str` doesn't compile

StackOverflowExcept1on (approved):

utils/runtime-fuzzer/fuzz/fuzz_targets/main.rs
SabaunT: discussed

scripts/src/test.sh
```diff
@@ -85,8 +85,8 @@ run_fuzzer() {
   cd $ROOT_DIR/utils/runtime-fuzzer
 
   # Run fuzzer
-  RUST_LOG="debug,runtime_fuzzer_fuzz=debug,wasmi,libfuzzer_sys,node_fuzzer=debug,gear,pallet_gear,gear-core-processor,gear-backend-wasmi,gwasm'" \
-  cargo fuzz run --release --sanitizer=none main -- -rss_limit_mb=8192
+  RUST_LOG=debug,syscalls,gear_wasm_gen,runtime_fuzzer_fuzz=debug,wasmi,libfuzzer_sys,runtime_fuzzer=trace,gear,pallet_gear,gear-core-processor,gear-backend-wasmi,gwasm \
```
breathx, line 88: seems you set debug for all targets so no need in their specification

utils/runtime-fuzzer/fuzz/fuzz_targets/main.rs
```diff
@@ -19,53 +19,13 @@
...
-fuzz_target!(|seed: u64| {
+fuzz_target!(|gear_calls: GearCalls| {
     gear_utils::init_default_logger();
 
-    dump_seed(seed).unwrap_or_else(|e| unreachable!("internal error: failed dumping seed: {e}"));
-
-    log::info!("Running the seed {seed}");
-    runtime_fuzzer::run(seed);
+    // Newline to easily browse logs.
+    println!();
```
breathx, line 28: using "log" mb?

breathx (approved):

scripts/check-fuzzer.sh
SabaunT: done

utils/runtime-fuzzer/fuzz/fuzz_targets/main.rs
SabaunT: discussed

scripts/src/test.sh
SabaunT: done

PR #3110, opened 2023-08-19
feat(cli): Add node CLI arg to set max block gas value

ekovalev:
Enables `--max-gas` command line argument to set a hard limit for the amount of gas a validator can spend in one block for the message queue processing.

Useful to reduce pressure on validators in case of the network congestion (DDoS attack or whatnot) to keep producing blocks at constant rate and not disable the queue processing completely.

Implementation wise, this value, if set, is propagated to the block authorship module in the client and then passed as a parameter to the `Gear::run()` extrinsic through a `runtime-api`, which will from now on take an `Option<u64>` as an argument.

### DIFF
--- a/node/cli/src/cli.rs
+++ b/node/cli/src/cli.rs
@@ -29,6 +29,10 @@ pub struct RunCmd {
     /// Force using Vara native runtime.
     #[arg(long = "force-vara")]
     pub force_vara: bool,
+
+    /// The upper limit for the amount of gas a validator can burn in one block.
+    #[arg(long, short = 'g')]
+    pub max_gas: Option<u64>,
 }
 
 #[derive(Debug, Parser)]
--- a/node/cli/src/cli.rs
+++ b/node/cli/src/cli.rs
@@ -31,7 +31,7 @@ pub struct RunCmd {
     pub force_vara: bool,
 
     /// The upper limit for the amount of gas a validator can burn in one block.
-    #[arg(long, short = 'g')]
+    #[arg(long)]
     pub max_gas: Option<u64>,
 }
 
--- a/node/service/src/lib.rs
+++ b/node/service/src/lib.rs
@@ -368,6 +368,7 @@ pub fn new_full_base<RuntimeApi, ExecutorDispatch>(
         >,
         &sc_consensus_babe::BabeLink<Block>,
     ),
+    max_gas: Option<u64>,
 ) -> Result<NewFullBase<RuntimeApi, ExecutorDispatch>, ServiceError>
 where
     RuntimeApi: ConstructRuntimeApi<Block, FullClient<RuntimeApi, ExecutorDispatch>>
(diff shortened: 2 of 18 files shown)


ukint-vs (approved):

pallets/gear/src/lib.rs
```diff
@@ -1716,8 +1719,12 @@ pub mod pallet {
             let max_weight = <T as frame_system::Config>::BlockWeights::get().max_block;
             let remaining_weight = max_weight.saturating_sub(weight_used.total());
 
-            // Remaining weight may exceed the minimum block gas limit determined by the Limiter trait
-            let adjusted_gas = GasAllowanceOf::<T>::get().max(remaining_weight.ref_time());
+            // Remaining weight may exceed the minimum block gas limit set by the Limiter trait.
+            let mut adjusted_gas = GasAllowanceOf::<T>::get().max(remaining_weight.ref_time());
+            // Gas for queue processing can never exceed the hard limit, if the latter is provided.
+            if let Some(max_gas) = max_gas {
+                adjusted_gas = adjusted_gas.min(max_gas);
```
shamilsan, line 1726: Here `max_gas` can be only less than the gas allowance. Are there use cases when we want to set `max_gas` greater than the current gas allowance?

ekovalev: The target use of this is to limit the queue processing time to some reasonably small value (supposedly smaller than the default block gas limit). If the provided value turns out to be greater than the actual allowance it will simply have no effect on the usual queue processing flow.

shamilsan: In my opinion, it would be great to use max gas, e.g., in RPC calls to (say here: https://github.com/gear-tech/gear/blob/master/pallets/gear/src/runtime_api.rs#L110). But if its purpose is only to limit block gas during production, there is no sense in making it bigger indeed.

node/service/src/lib.rs
```diff
@@ -484,13 +485,16 @@ where
...
     if let sc_service::config::Role::Authority { .. } = &role {
-        let proposer = authorship::ProposerFactory::new(
+        let mut proposer = authorship::ProposerFactory::new(
             task_manager.spawn_handle(),
             client.clone(),
             transaction_pool.clone(),
             prometheus_registry.as_ref(),
             telemetry.as_ref().map(|x| x.handle()),
         );
+        if let Some(max_gas) = max_gas {
```
NikVolf, line 495: why mutating it and not just setting field above during the creation?

NikVolf (approved):

node/cli/src/cli.rs
```diff
@@ -29,6 +29,10 @@ pub struct RunCmd {
     /// Force using Vara native runtime.
     #[arg(long = "force-vara")]
     pub force_vara: bool,
+
+    /// The upper limit for the amount of gas a validator can burn in one block.
+    #[arg(long, short = 'g')]
```
breathx, line 34: such short variant isn't intuitive for me

breathx (approved):
Add todo about considering previously discussed case with rerunning the same first from the queue message

node/cli/src/cli.rs
ekovalev: `-g` for "gas". Also unlikely to conflict with any other args should there be any in the future.

Other suggestions?

ekovalev: or we can disable the short variant completely: `--max-gas` is short enough

PR #3114, opened 2023-08-19
feat(wasm-gen): generate instructions handling fallible sys-calls results

### LINKED ISSUE #3099: Make instructions to handle fallible sys-calls results
### Problem to Solve

-

### Possible Solution

Would be also nice to use results of one sys-call in other sys-calls (at least consider a separate issue for that)

### Notes

_No response_

mertwole:
Resolves #3099


utils/wasm-gen/src/generator/syscalls.rs
```diff
@@ -82,9 +87,85 @@ impl InvocableSysCall {
...
+        SysCallInfo {
+            signature: self.into_signature(),
+            fallible,
+        }
+    }
+}
+
+// If syscall changes from fallible into infallible or vice versa in future,
+// we'll see it by analyzing code coverage stats produced by fuzzer.
+fn is_syscall_fallible(syscall: SysCallName) -> bool {
```
StackOverflowExcept1on, line 113: Maybe we can introduce a new `SysCallName` method?

mertwole: We don't check anywhere in this `fn` or in tests that these syscalls are divided into fallible/infallible correctly. On the level of `runtime-fuzzer` we will see by analyzing coverage if there's errors, but if I place it on the level of `SysCallName` (inside `wasm-instrument`), then anyone can use this unchecked `fn` freely

utils/wasm-gen/src/generator/syscalls/imports.rs
```diff
@@ -280,22 +280,21 @@ impl<'a, 'b> SysCallsImportsGenerator<'a, 'b> {
...
+        let rid_pid_value_ptr = pid_value_ptr + mem::size_of::<HashWithValue>() as i32;
 
         let func_instructions = Instructions::new(vec![
             // Amount of gas to reserve
             Instruction::GetLocal(4),
             // Duration of the reservation
             Instruction::GetLocal(5),
             // Pointer to the LengthWithHash struct
-            Instruction::I32Const(reserve_gas_result_ptr),
+            Instruction::GetLocal(6),
```
StackOverflowExcept1on, line 294: Is local variable with index 6 defined? I see indexes from 0 to 5 in signature
https://github.com/gear-tech/gear/blob/8b54acaf652807d4e93cffc9b2b719751a939483/utils/wasm-gen/src/generator/syscalls.rs#L78-L85

mertwole: https://github.com/gear-tech/gear/blob/md-fuzzer-syscall-results/utils/wasm-gen/src/generator/syscalls.rs#L79C6-L95C6

utils/wasm-gen/src/generator/syscalls.rs
StackOverflowExcept1on: We can do it in another PR and write tests for `SysCallName::is_fallible()`.

mertwole: But for now fallible/infallible syscall division makes sense only for `wasm-gen`, and asserting this is not a trivial task. So I think we can have this solution

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -266,19 +295,66 @@ impl<'a, 'b> SysCallsInvocator<'a, 'b> {
...
+    fn build_result_processing(
         &mut self,
-        params: &[ParamType],
-        results: &[ValueType],
-        call_indexes_handle: CallIndexesHandle,
-    ) -> Result<Vec<Instruction>> {
+        syscall: SysCallInfo,
+        param_setters: &[ParamSetter],
+    ) -> Vec<Instruction> {
+        if syscall.fallible {
```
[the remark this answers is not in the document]
mertwole: But it also have logics for processing infallible syscalls returning error

utils/wasm-gen/src/config/module.rs
```diff
@@ -24,7 +24,10 @@
 //! and [`ConstantParams`].
 
 use arbitrary::{Arbitrary, Result, Unstructured};
-use wasm_smith::{InstructionKind::*, InstructionKinds, SwarmConfig};
+use wasm_smith::{
+    InstructionKind::{self, *},
```
SabaunT, line 28: `InstructionKind` or kinds must be `pub use` as you obligate users to set config using these types.

utils/wasm-gen/src/config/module.rs
```diff
@@ -115,6 +119,12 @@ impl From<(SelectableParams, ArbitraryParams)> for WasmModuleConfig {
             max_memory_pages,
         } = arbitrary_params;
 
+        let allowed_instructions: Vec<_> = allowed_instructions
+            .into_iter()
+            .filter(|ix| !forbidden_instructions.contains(ix))
+            .collect();
+        let allowed_instructions = InstructionKinds::new(&allowed_instructions);
```
SabaunT, line 126: I think `allowed_instructions` must go to `SelectableParams` too.

mertwole: Hm, but will we use it when setting up fuzzer? For now It's used only for tests

SabaunT: You can have default implementation of it. I suggest moving it, because it seems wrong to define instructions in one config and then invalidate them in the other.

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -454,12 +454,27 @@ impl<'a, 'b> SysCallsInvocator<'a, 'b> {
     }
 
     fn build_result_processing_infallible(signature: SysCallSignature) -> Vec<Instruction> {
+        // TODO: #3129
```
SabaunT, line 457: please a small comment on reasons of #3129

utils/wasm-gen/src/generator/syscalls/imports.rs
```diff
@@ -280,22 +280,21 @@ impl<'a, 'b> SysCallsImportsGenerator<'a, 'b> {
...
-        let reservation_send_result_ptr = pid_value_ptr + mem::size_of::<HashWithValue>() as i32;
+        let pid_value_ptr = memory_size_in_bytes.saturating_sub(100) as i32;
+        let rid_pid_value_ptr = pid_value_ptr + mem::size_of::<HashWithValue>() as i32;
 
         let func_instructions = Instructions::new(vec![
             // Amount of gas to reserve
             Instruction::GetLocal(4),
             // Duration of the reservation
             Instruction::GetLocal(5),
             // Pointer to the LengthWithHash struct
```
StackOverflowExcept1on, line 293:
```suggestion
            // Pointer to the ErrorWithHash struct
```
since ba15f037f7f5d8dc4cca1f8e3ee65154d1c2c3e0 `LengthWithHash` has been renamed to `ErrorWithHash`

utils/wasm-gen/src/generator/syscalls/imports.rs
```diff
@@ -280,22 +280,21 @@ impl<'a, 'b> SysCallsImportsGenerator<'a, 'b> {
...
             initial_mem_size.memory_size()
         };
+
         // subtract to be sure we are in memory boundaries.
-        let reserve_gas_result_ptr = memory_size_in_bytes.saturating_sub(100) as i32;
-        let rid_pid_value_ptr = reserve_gas_result_ptr + mem::size_of::<Length>() as i32;
-        let pid_value_ptr = reserve_gas_result_ptr + mem::size_of::<ErrorWithHash>() as i32;
-        let reservation_send_result_ptr = pid_value_ptr + mem::size_of::<HashWithValue>() as i32;
+        let pid_value_ptr = memory_size_in_bytes.saturating_sub(100) as i32;
+        let rid_pid_value_ptr = pid_value_ptr + mem::size_of::<HashWithValue>() as i32;
```
StackOverflowExcept1on, line 286: `HashWithValue` is allocated here, but `gr_reservation_send` takes `rid_pid_value: *const TwoHashesWithValue`
looks like need to fix the instructions since `*mut ErrorWithHash` comes from arguments and we can no longer construct `TwoHashesWithValue` in easy way

utils/wasm-gen/src/config/syscalls/param.rs
```diff
@@ -33,6 +33,27 @@ use std::{collections::HashMap, ops::RangeInclusive};
 pub struct SysCallsParamsConfig(HashMap<ParamType, SysCallParamAllowedValues>);
 
 impl SysCallsParamsConfig {
+    /// New [`SysCallsParamsConfig`] with all rules set to produce one constant value.
+    pub fn constant_value(value: i64) -> Self {
```
SabaunT, line 37:
```suggestion
    pub fn all_constant_value(value: i64) -> Self {
```

utils/wasm-gen/src/tests.rs
```diff
@@ -157,6 +167,245 @@ fn injecting_addresses_works() {
...
+        message::{ContextSettings, DispatchKind, IncomingDispatch, MessageContext},
+        reservation::GasReserver,
+    };
+    use gear_core_processor::{configs::PageCosts, ProcessorContext, ProcessorExternalities};
+
+    const INITIAL_PAGES: u16 = 1;
+    const INJECTED_SYSCALLS: u32 = 8;
+
+    let buf = vec![0; UNSTRUCTURED_SIZE];
+    let mut unstructured = Unstructured::new(&buf);
```
SabaunT, line 293: you create program from zero bytes?

utils/wasm-gen/src/tests.rs
```diff
@@ -157,6 +167,245 @@ fn injecting_addresses_works() {
...
+
+    let mut injection_amounts = SysCallsInjectionAmounts::all_never();
+    injection_amounts.set(syscall, INJECTED_SYSCALLS, INJECTED_SYSCALLS);
```
SabaunT, line 296: these lines of code does the same, doesn't it?

utils/wasm-gen/src/tests.rs
```diff
@@ -157,6 +167,245 @@ fn injecting_addresses_works() {
...
+                ..MemoryPagesConfig::default()
+            })
+            .with_sys_calls_config(
+                SysCallsConfigBuilder::new(injection_amounts)
+                    .with_params_config(params_config)
+                    .set_ignore_fallible_syscall_errors(ignore_fallible_errors)
+                    .build(),
+            )
+            .with_entry_points_config(EntryPointsSet::Init)
+            .with_recursions_removed(true)
```
SabaunT, line 311: seems odd when only one function is generated

utils/wasm-gen/src/tests.rs
```diff
@@ -157,6 +167,245 @@ fn injecting_addresses_works() {
...
+
+    let module = generate_gear_program_module(&mut unstructured, gear_config)
+        .expect("failed wasm generation");
+
+    let module = gear_wasm_instrument::inject(
+        module,
+        &gear_wasm_instrument::rules::CustomConstantCostRules::new(0, 0, 0),
+        "env",
+    )
+    .unwrap();
```
SabaunT, line 330: please use `expect` with a some clarifying text

SabaunT: Same for all `unwraps`.

utils/wasm-gen/src/tests.rs
```diff
@@ -157,6 +167,245 @@ fn injecting_addresses_works() {
...
+    let init_msg = IncomingMessage::new(
+        MessageId::default(),
+        ProgramId::default(),
+        payload,
+        0,
+        0,
+        None,
+    );
+
+    let default_pc = ProcessorContext {
```
SabaunT, line 342: what about having a separate `Default` trait impl only for tests?

utils/wasm-gen/src/tests.rs
```diff
@@ -157,6 +167,245 @@ fn injecting_addresses_works() {
...
+            termination_reason,
+            TerminationReason::Actor(ActorTerminationReason::Success),
+            "syscall: {}",
+            syscall.to_str()
+        );
+    }
+}
+
+#[derive(Clone)]
+struct MemoryWrite {
```
SabaunT, line 217: Discussed a way to remove it, please try to execute.

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -40,8 +40,13 @@ pub(crate) enum ProcessedSysCallParams {
         value_type: ValueType,
         allowed_values: Option<SysCallParamAllowedValues>,
     },
-    MemoryArray,
-    MemoryPtrValue,
+    MemoryArray {
+        pointer_allowed_values: Option<SysCallParamAllowedValues>,
+        length_allowed_values: Option<SysCallParamAllowedValues>,
```
SabaunT, line 45: That creates an issue with memory arrays. Rules for `ProcessedSysCallParams` must be enabled by desire in config for each sys-call separately. What I mean is that when you define a rule for `ParamType::Ptr` and `ParamType::Size`, it affects not only `Value`, `MemoryPtrValue`, but `MemoryArray`. Default implementation of instructions generation for `MemoryArray` is good enough - it chooses from the existing memory range. Your changes override that with constant values that aren't guaranteed to be in an existing wams memory range. Maybe, it will work for default memory config.

SabaunT (commented):
Awesome work has been done so far. Keep it going!

utils/wasm-gen/src/tests.rs
mertwole: nope, from array filled with zeroes

SabaunT: ha, i see, the test instantiates wasm with only one function that has only one sys-call. It's ok in this case

utils/wasm-gen/src/tests.rs
mertwole: what do you mean? we generate config with all injection amounts = 0 then set our syscall to `(INJECTED_SYSCALLS..=INJECTED_SYSCALLS)`

utils/wasm-gen/src/tests.rs
mertwole: but it's enough for us, as we testing the logics independent of generated function count

utils/wasm-gen/src/tests.rs
mertwole: Renamed variable, `Default` impl not fits because `ProcessorContext` instantiated from params local to this `fn` and make sense only in context of this `fn`

utils/wasm-gen/src/tests.rs
SabaunT: sorry, thought it was 0, not 8

utils/wasm-gen/src/tests.rs
SabaunT: discussed, seems no way to remove that to implement test for pay_program_rent

utils/wasm-gen/src/generator/syscalls/invocator.rs
mertwole: What if we change `SysCallsParamsConfig` to work not with `ParamType` but with higher-level entities like memory arrays?

SabaunT: `MemoryArray` is not a user-space type.

core-processor/src/ext.rs
```diff
@@ -106,6 +106,48 @@ pub struct ProcessorContext {
     pub rent_cost: u128,
 }
 
+impl Default for ProcessorContext {
```
SabaunT, line 109: I believe we must move this impl under `test` cfg

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -310,57 +377,152 @@ impl<'a, 'b> SysCallsInvocator<'a, 'b> {
...
         );
 
-        Ok(instructions)
+        assert_eq!(setters.len(), params.len());
+
+        Ok(setters)
+    }
+
+    fn build_result_processing_ignored(signature: SysCallSignature) -> Vec<Instruction> {
+        std::iter::repeat(Instruction::Drop)
```
StackOverflowExcept1on, line 447: We already has `iter` import in `use std::{collections::BTreeMap, iter}`
```suggestion
        iter::repeat(Instruction::Drop)
```

core-processor/src/ext.rs
mertwole: This `Default` may be useful not only for tests, and actually it's a little bit different with one used in test below

utils/wasm-gen/src/tests.rs
```diff
@@ -157,6 +163,169 @@ fn injecting_addresses_works() {
...
+            max_funcs: 1,
+        },
+    );
+
+    let module = generate_gear_program_module(&mut unstructured, gear_config)
+        .expect("failed wasm generation");
+
+    let module = gear_wasm_instrument::inject(
+        module,
+        &gear_wasm_instrument::rules::CustomConstantCostRules::new(0, 0, 0),
```
StackOverflowExcept1on, line 277:
```suggestion
        &CustomConstantCostRules::new(0, 0, 0),
```

full diff: https://gist.github.com/StackOverflowExcept1on/bd009da4be071e8d64d7064cde5fe5a0

SabaunT (approved):

StackOverflowExcept1on (approved):

core-processor/src/ext.rs
breathx: > @breathx were there any specific intentions not to implement `Default`/`Debug`/`Clone` for `ProcessorContext`?

In general, it doesnt matter. But we should never use any of such implicit defaults for runtime, so I'd avoid it

Issue #3091, opened 2023-08-18
Introduce smart sys-calls invocations


techraed:
### Problem to Solve

Better coverage with closer to reality sys-calls invocations in wasm.

### Possible Solution

This is actually a huge task with possibly an infinite number of subtasks. We can implement various kinds of rules/patterns of how sys- calls must be injected/called in the wasm. Looks pretty like making precise sys-calls invocations like it’s done for precise reservation send (`SysCallsImportsGenerator::generate_send_from_reservation`).

### Notes

Possible patterns:
- [x] `reply_from_reservation`
- [x] filling the message with payload: `gr_send_init`, `gr_send_push` & `gr_send_push_input`, `gr_send_commit`
- [ ] consider incorrect patterns to improve coverage

techraed:
Introduced lots of patterns. Considered to be resolved.

PR #3116, opened 2023-08-19
refactor(wasm-gen): add methods to generate precise sys-calls

StackOverflowExcept1on:
Related to #3091 (Currently does not solve the problem, but makes it convenient to write patterns)


utils/wasm-gen/src/generator/syscalls/imports.rs
```diff
@@ -242,36 +242,83 @@ impl<'a, 'b> SysCallsImportsGenerator<'a, 'b> {
...
-            match maybe_reserve_gas.zip(maybe_reservation_send) {
-                Some(indexes) => indexes,
-                None => {
-                    log::trace!("Wasm hasn't got either gr_reserve_gas, or gr_reservation_send, or both of them");
-                    return;
-                }
-            }
-        };
+    /// Returns the indexes of invocable sys-calls.
+    fn invocable_sys_calls_indexes(&self, sys_calls: &[SysCallName]) -> Option<Vec<usize>> {
```
SabaunT, line 246: Basically, I guess the `invocable_sys_calls_indexes` will be called from a method, which builds precise sys-calls invocation. That means such method will "know" sys-calls it needs. So `invocable_sys_calls_indexes` can accept not the slice, but the `[SysCalls; N]` and give as a result array of the same size. This allows us to avoid additional vector allocation

utils/wasm-gen/src/generator/syscalls/imports.rs
```diff
@@ -242,36 +242,83 @@ impl<'a, 'b> SysCallsImportsGenerator<'a, 'b> {
...
+            })
+            .collect();
+
+        if ret.is_none() {
+            log::trace!(
+                "Wasm should import the following syscalls: {missing_sys_calls}",
+                missing_sys_calls = sys_calls
+                    .iter()
+                    .map(|sys_call| sys_call.to_str())
+                    .collect::<Vec<_>>()
```
SabaunT, line 262: collect to an array of known size?

utils/wasm-gen/src/generator/syscalls/imports.rs
```diff
@@ -242,36 +242,83 @@ impl<'a, 'b> SysCallsImportsGenerator<'a, 'b> {
...
+            .map(|&sys_call| {
+                self.sys_calls_imports
+                    .get(&InvocableSysCall::Loose(sys_call))
+                    .map(|&(_, call_indexes_handle)| call_indexes_handle)
+            })
+            .collect();
+
+        if ret.is_none() {
+            log::trace!(
+                "Wasm should import the following syscalls: {missing_sys_calls}",
```
SabaunT, line 258: Also log which "precise" sys-call invocation required that. Currently there's only `gr_reservation_send`.

utils/wasm-gen/src/generator/syscalls/imports.rs
```diff
@@ -242,36 +242,83 @@ impl<'a, 'b> SysCallsImportsGenerator<'a, 'b> {
...
-        let send_from_reservation_signature = invocable_sys_call.into_signature();
+        ret
+    }
 
-        let send_from_reservation_func_ty = send_from_reservation_signature.func_type();
+    /// Generates a function which calls "properly" the given sys-call.
+    fn generate_sys_call_pattern(
+        &mut self,
+        sys_call: SysCallName,
+        sys_call_amount: u32,
```
SabaunT, line 274: this is a constant, I guess. If you want to insert "precise" invocations several times, that must be defined from the config. Unless we have that, I'd remove this argument and would just use `1`.

utils/wasm-gen/src/generator/syscalls/imports.rs
```diff
@@ -242,36 +242,83 @@ impl<'a, 'b> SysCallsImportsGenerator<'a, 'b> {
...
+        }
 
-        let invocable_sys_call = InvocableSysCall::Precise(SysCallName::ReservationSend);
-        let send_from_reservation_signature = invocable_sys_call.into_signature();
+        ret
+    }
 
-        let send_from_reservation_func_ty = send_from_reservation_signature.func_type();
+    /// Generates a function which calls "properly" the given sys-call.
+    fn generate_sys_call_pattern(
```
SabaunT, line 271: Maybe `generate_proper_sys_call_invocation` ? I mean, at least avoid "pattern"

utils/wasm-gen/src/generator/syscalls/imports.rs
StackOverflowExcept1on: rust has no way to collect into an array in stdlib

utils/wasm-gen/src/generator/syscalls/imports.rs
StackOverflowExcept1on: I can set this to 1, but in general it should probably have separate config in the future

utils/wasm-gen/src/generator/syscalls/imports.rs
StackOverflowExcept1on: done in 41651b831611ce73584f0f580835a7db43159fde

utils/wasm-gen/src/generator/syscalls/imports.rs
StackOverflowExcept1on: done in 41651b831611ce73584f0f580835a7db43159fde

utils/wasm-gen/src/generator/syscalls/imports.rs
SabaunT: Yeah, but, I mean we should avoid allocating. Maybe introduce some external converter function or something similar

StackOverflowExcept1on: done in 6aa698f1f6f75fffd984f8773d5acc4345da14e1

utils/wasm-gen/src/generator/syscalls/imports.rs
StackOverflowExcept1on: This would require defining the precise sys-call at the top level and storing it in a local variable and then passing it to this function. It doesn't look good, but I can do it

StackOverflowExcept1on: done in faf88bd231a295964785a02d4d68a1a894e98104

utils/wasm-gen/src/generator/syscalls/imports.rs
```diff
@@ -242,36 +242,92 @@ impl<'a, 'b> SysCallsImportsGenerator<'a, 'b> {
...
+
+        log::trace!(
+            "Built proper call to {precise_sys_call_name}",
+            precise_sys_call_name = InvocableSysCall::Precise(sys_call).to_str()
+        );
+
+        let call_indexes_handle = self.call_indexes.len();
+        self.call_indexes.add_func(func_idx.signature as usize);
+
+        // TODO: make separate config for precise sys-calls
```
SabaunT, line 318: Please create a separate issue for this and mark it here

utils/wasm-gen/src/generator/syscalls/imports.rs
```diff
@@ -242,36 +242,92 @@ impl<'a, 'b> SysCallsImportsGenerator<'a, 'b> {
...
-                .sys_calls_imports
-                .get(&InvocableSysCall::Loose(SysCallName::ReservationSend))
-                .map(|&(_, call_indexes_handle)| call_indexes_handle);
-
-            match maybe_reserve_gas.zip(maybe_reservation_send) {
-                Some(indexes) => indexes,
+    /// Returns the indexes of invocable sys-calls.
+    fn invocable_sys_calls_indexes<const N: usize>(
+        &self,
+        sys_call: SysCallName,
```
SabaunT, line 248: btw, if you still don't want `sys_call` be in the signature of this method, you can remove that, but need to log name of the precise sys-call invocation being built (in generate_send_from_reservation, for example)

utils/wasm-gen/src/generator/syscalls/imports.rs
```diff
@@ -242,36 +242,92 @@ impl<'a, 'b> SysCallsImportsGenerator<'a, 'b> {
...
                 None => {
-                    log::trace!("Wasm hasn't got either gr_reserve_gas, or gr_reservation_send, or both of them");
-                    return;
+                    log::trace!(
+                        "To build the {precise_sys_call_name} sys-call, wasm must import the following sys-calls: {missing_sys_calls}",
+                        precise_sys_call_name = InvocableSysCall::Precise(sys_call).to_str(),
+                        missing_sys_calls = sys_calls
+                            .iter()
+                            .map(|sys_call| sys_call.to_str())
+                            .collect::<Vec<_>>()
```
SabaunT, line 268: This allocation is really odd. Just create a separate converter function in `utlis.rs`

SabaunT (approved):

utils/wasm-gen/src/generator/syscalls/imports.rs
StackOverflowExcept1on: On the other hand, it's easier to understand what we want to register as precise sys-call if we write it at the beginning of the function.

mertwole: But the only thing this `fn` does is mapping sys_calls[] <-> import_idx[] so ideally it shouldn't contain any other logics.

utils/wasm-gen/src/generator/syscalls/imports.rs
```diff
@@ -242,36 +242,88 @@ impl<'a, 'b> SysCallsImportsGenerator<'a, 'b> {
...
+        });
+
+        for (index, maybe_index) in indexes.iter_mut().zip(iter) {
+            match maybe_index {
+                Some(idx) => *index = idx,
                 None => {
-                    log::trace!("Wasm hasn't got either gr_reserve_gas, or gr_reservation_send, or both of them");
-                    return;
+                    log::trace!(
+                        "To build the {precise_sys_call_name} sys-call, wasm must import the following sys-calls: {missing_sys_calls:?}",
```
mertwole, line 263:
```suggestion
                        "The following sys-calls must be imported: {missing_sys_calls:?}",
```

mertwole: And then the higher-level logs will explain why they must be imported

StackOverflowExcept1on: In that case you will have to write this for each new sys-call pattern. I think `let ... else { return; }` is better anyway.

mertwole:
Btw, can we introduce parsing WAT instead of writing instructions like this? In this PR or in separate
https://github.com/gear-tech/gear/blob/master/utils/wasm-gen/src/generator/syscalls/imports.rs#L289-L344
@StackOverflowExcept1on  @SabaunT

StackOverflowExcept1on:
@mertwole converting WAT to bytes will take time. We already have problems with the performance of wasm-gen.

utils/wasm-gen/src/generator/syscalls/imports.rs
mertwole: But the only log you need in precise syscalls construction is something like `"constructing precise syscall {name}..."` and you anyway need it

SabaunT:
> Btw, can we introduce parsing WAT instead of writing instructions like this? In this PR or in separate https://github.com/gear-tech/gear/blob/master/utils/wasm-gen/src/generator/syscalls/imports.rs#L289-L344 @StackOverflowExcept1on @SabaunT

interesting idea, but currently it seems that we don't need it. Using `Instruction` instead of WAT hasn't created any inconvenience yet.

utils/wasm-gen/src/generator/syscalls/imports.rs
StackOverflowExcept1on: done in 0c6cfb7aa445d5b5517fc7e02ff4a551498848fb

StackOverflowExcept1on:
@CodiumAI-Agent /review

StackOverflowExcept1on:
@CodiumAI-Agent /improve

utils/wasm-gen/src/generator/syscalls/imports.rs
```diff
@@ -242,36 +242,91 @@ impl<'a, 'b> SysCallsImportsGenerator<'a, 'b> {
...
+                    .with_signature(func_signature)
+                    .body()
+                    .with_instructions(func_instructions)
+                    .with_locals(func_locals.unwrap_or_default())
+                    .build()
+                    .build(),
+            );
+
+            (module_builder.build(), idx)
+        });
```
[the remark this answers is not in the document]
StackOverflowExcept1on: useless change

utils/wasm-gen/src/generator/syscalls/imports.rs
```diff
@@ -242,36 +242,91 @@
...
+                        missing_sys_calls = sys_calls.map(|sys_call| sys_call.to_str()),
+                    );
+                    return None;
                 }
             }
-        };
+        }
+
+        Some(indexes)
+    }
```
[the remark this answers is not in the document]
StackOverflowExcept1on: cool, but it is easier to use `return;` instead of logging string/missing_syscall every time

mertwole (approved):

PR #3117, opened 2023-08-19
test(runtime-fuzzer): Test that `GearCalls` generation is deterministic

### LINKED ISSUE #3095: Test that GearCalls generation is deterministic
### File Location(s)

_No response_

### Proposal

Also test that fuzz target itself (targets entry point) is deterministic.

SabaunT:
Resolves #3095  .

Also tests that fuzzer is deterministic.

@gear-tech/dev

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -80,6 +80,9 @@ jobs:
       - name: "Install: Rust toolchain"
         uses: dsherret/rust-toolchain-file@v1
 
+      - name: "Install: cargo-nextest"
+        run: curl -LsSf https://get.nexte.st/latest/linux | tar zxf - -C ${CARGO_HOME:-~/.cargo}/bin
+
       - name: Install cargo-fuzz
         run: cargo install cargo-fuzz
 
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -80,9 +80,6 @@ jobs:
       - name: "Install: Rust toolchain"
         uses: dsherret/rust-toolchain-file@v1
 
-      - name: "Install: cargo-nextest"
-        run: curl -LsSf https://get.nexte.st/latest/mac | tar zxf - -C ${CARGO_HOME:-~/.cargo}/bin
-
       - name: Install cargo-fuzz
         run: cargo install cargo-fuzz
 
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -80,10 +80,16 @@ jobs:
       - name: "Install: Rust toolchain"
         uses: dsherret/rust-toolchain-file@v1
 
+      - name: "Install: cargo-nextest"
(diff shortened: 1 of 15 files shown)


grishasobol (approved):

.github/workflows/check.yml
```diff
@@ -86,4 +86,7 @@ jobs:
       - name: "Check fuzzer with mutation test"
         run: ./scripts/check-fuzzer.sh
 
+      - name: "Check fuzzer reproduction"
+        run: ./scripts/gear.sh test fuzz-repr
```
StackOverflowExcept1on, line 90: probably need to add to install `cargo nextest`
```yml
- name: "Buid: Init"
  run: ./scripts/gear.sh init cargo
```

StackOverflowExcept1on (approved):

utils/call-gen/src/upload_program.rs
```diff
@@ -29,7 +29,7 @@ type UploadProgramArgsInner = (Vec<u8>, Vec<u8>, Vec<u8>, u64, u128);
 /// Upload program args
 ///
 /// Main type used to generate arguments for the `pallet_gear::Pallet::<T>::upload_program` call.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
```
breathx, line 32: mb add for future other default traits like Default and Ords?

utils/runtime-fuzzer/src/lib.rs
```diff
@@ -16,8 +16,12 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+#![allow(clippy::items_after_test_module)]
```
breathx, line 19: ?

breathx (approved):

utils/runtime-fuzzer/src/lib.rs
SabaunT: clippy will warn because of items (functions, in this case) being defined after `mod tests`
https://rust-lang.github.io/rust-clippy/master/index.html#/items_after_test_module

.github/workflows/check.yml
SabaunT: Right. done here https://github.com/gear-tech/gear/pull/3117/commits/eec074ec7e94b8d2dbdf88d3830c03ece934b35d

I don't run the `gear.sh` script for that, as I don't need the `cargo-hack`.

utils/call-gen/src/upload_program.rs
SabaunT: Will add Ords, but not `Default`, as it looks wrong. How can we decide which GearCall is Default? I mean, the impl of the trait won't be meaningful, it will just help bypass type-level boundaries.

SabaunT: [a3b1603](https://github.com/gear-tech/gear/pull/3117/commits/a3b160371c299e258eb8eb2b98e3eda5ea3eac1d)

PR #3118, opened 2023-08-19
fix(authorship): Forward-fix a test that will require the gear-bank account to exist

ekovalev:
The `pallet-bank` that is about to be merged introduces the bank account and requires the latter exists at the time an `init_package` is created. This would make one of the tests fail.
The current PR forward-fixes that.


breathx (approved):
Awesome! Thanks

ukint-vs (approved):

PR #3130, opened 2023-08-21
chore(core): move GasInfo to core

### LINKED ISSUE #3021: Use GasInfo from the generated metadata in rpc calls of gsdk
Original gas info from pallet is differ. Why?
```rust
/// The struct contains results of gas calculation required to process
/// a message.
#[derive(Clone, Debug, Decode, Encode, PartialEq, Eq, TypeInfo)]
pub struct GasInfo {
    pub min_limit: u64,
    /// Gas amount that we reserve for some other on-chain interactions.
    pub reserved: u64,
    /// Contains number of gas burned during message processing.
    pub burned: u64,
    /// The value may be returned if a program happens to be executed
    /// the second or next time in a block.
    pub may_be_returned: u64,
    /// Was the message placed into waitlist at the end of calculating.
    ///
    /// This flag shows, that `min_limit` makes sense and have some guarantees
    /// only before insertion into waitlist.
    pub waited: bool,
}
```

_Originally posted by @breathx in https://github.com/gear-tech/gear/pull/2979#discussion_r1282250943_

clearloop:
Resolves #3021 

`GasInfo` currently only used by `pallet-gear` and the related runtime APIs ( plus rpc APIs ), the term of moving it to `gear-core` is for importing it in `gsdk` and make the definition sync since it is really not necessary and even impossible to import `pallet-gear` in `gsdk`

- [x] move `GasInfo` to `core`
- [x] use GasInfo from `core`

@gear-tech/dev


breathx (approved):
Merge master pls

PR #3135, opened 2023-08-22
feat(node): Make sure `pallet_gear::run()` pseudo-inherent only runs once per block

### LINKED ISSUE #3111: Ensure there cannot be more than one gear_run() extrinsic in a block
Put necessary guards in the code and write tests to ensure the `pallet_gear::run()` extrinsic is never run more than once in a block. Tests scenarios may include (but are not necessarily limited to):
- check that an attempt of a user to submit it as un unsigned extrinsic deliberately via some client interface results in an error;
- check the `pallet_gear::run()` extrinsic as having the `DispatchClass:Mandatory` is always invalidated even if somehow it makes it in the txpool;
- check that a repeated execution of the `pallet_gear::run()` within the same block yields an error.

ekovalev:
Resolves #3111.

While resolving the above issue, some refactoring has been done pertaining  the way we determine whether the `gear::run()` extrinsic has been executed in a block or not:
- instead of doing some not too straightforward math with the gear internal block number and making the conclusion from that, a simple `Option` storage value has been introduced that is set to `Some()` in the extrinsic and `taken` (set to `None`) in the `on_finalize()` hook thereby never being committed to the storage. This gives an extra guarantee the gear internal block numbering is consistent no matter what.


ukint-vs (approved):

pallets/gear/src/tests.rs
```diff
@@ -13696,6 +13696,57 @@ fn wasm_ref_types_doesnt_work() {
     });
 }
 
+/// Test that the `Gear::run()` extrinsic can only run once per block,
+/// even if somehow included in a block multiple times.
+#[test]
```
breathx, line 13701: If gear::run won't be included to the block than QueueReverted event will be deposit when shouldn't

breathx (changes requested):

pallets/gear/src/tests.rs
ekovalev: It is always included (unless panics during a block creation - in that case `QueueProcessingReverted` event is exactly what is expected), but not more than once in a block. All subsequent attempts (if, let's say, a malicious validator for whatever reason decides to inject another `gear::run()` directly into the block) will result in an error and thereby not be executed.
It's the same type of guards all inherents in substrate have.

ukint-vs:
@CodiumAI-Agent /review

breathx (approved):
No need in checks if block author didn't include gear_run tx, but seems its relevant to rename event to smth like QueueWasNotProcessed or GearRunDidNotTakePlace

Issue #3069, opened 2023-08-15
Find a way to merge the examples' definition in the root manifest


clearloop:
### Problem to Solve

https://github.com/gear-tech/gear/blob/master/Cargo.toml#L24-L74

simply use `examples/[!big-wasm]*` doesn't work, see https://github.com/gear-tech/gear/pull/3055#discussion_r1292418166

### Possible Solution

try to modify the `workspace = "../../"` in each of the examples as well

PR #3138, opened 2023-08-22
refactor(examples): make demos be more consistent

shamilsan:
- Removed unused things.
- Made manifest style united.
- Split contract source files into `lib.rs` + `wasm.rs`.

@gear-tech/dev


StackOverflowExcept1on:
Can we consider removing `parity-scale-codec` and replacing with `gstd` re-export?

shamilsan:
> Can we consider removing `parity-scale-codec` and replacing with `gstd` re-export?

I don't find importing `parity-scale-codec` too bad.

clearloop (approved):
finally can remove `workspace = "../../"` lol

examples/async-init/Cargo.toml
```diff
@@ -4,7 +4,6 @@ version = "0.1.0"
 authors.workspace = true
 edition.workspace = true
 license.workspace = true
-workspace = "../../"
```
breathx, line 7: homepage.workspace = true
repository.workspace = true

examples/fungible-token/io/Cargo.toml
```diff
@@ -4,7 +4,6 @@ version = "0.1.4"
 authors.workspace = true
 edition.workspace = true
 license.workspace = true
-workspace = "../../../"
 
 [dependencies]
 gstd = { workspace = true }
```
breathx, line 9: Seems you didn't totally unify the toml style 
```suggestion
gstd.workspace = true
```

breathx: And introduce similar feature gate for all programs
Furthermore you can't split them all into lib.rs and wasm.rs, cause some of our demo doesnt compile correct but don't show up on CI and tests due to cargo workspace issues

breathx (changes requested):

examples/fungible-token/io/Cargo.toml
clearloop: ref #3069

shamilsan:
@breathx Please re-review

examples/compose/src/wasm.rs
```diff
@@ -0,0 +1,131 @@
...
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+// This contract recursively composes itself with another contract (the other contract
```
breathx, line 19: doc comment

examples/gas-burned/src/wasm.rs
```diff
@@ -0,0 +1,21 @@
+use gstd::{msg, prelude::*};
```
breathx, line 1: maybe keep mod wasm inside lib rs for small programs?

breathx (approved):
Btw you can significantly reduce diff by initially just raneming previous lib.rs into wasm.rs and moving crate-related stuff into lib.rs (it will be the only diff just to keep history of demo authors etc)

examples/gas-burned/src/wasm.rs
shamilsan: I prefer a common approach for all.

PR #3145, opened 2023-08-22
chore: Version 0.3.0 / 300 in master

NikVolf:


breathx (approved):

shamilsan:
If we mark PR's title with `[skip ci]`, it will block running CI by adding a tag. Changed the title.

By the way `[skip ci]` seems still to not work for commits.

PR #3154, opened 2023-08-22
Replace log::error! with log::trace!

NikVolf:


ukint-vs:
https://github.com/gear-tech/gear/blob/5d8fb07221d187b573b1d68fd7f2445682465ef8/sandbox/sandbox/src/embedded_executor.rs#L253

Here also

sandbox/sandbox/src/embedded_executor.rs
```diff
@@ -348,11 +348,11 @@ impl<State: 'static> super::SandboxInstance<State> for Instance<State> {
         }
 
         let instance_pre = linker.instantiate(&mut store, &module).map_err(|e| {
-            log::error!(target: TARGET, "Error instantiating module: {:?}", e);
+            log::trace!(target: TARGET, "Error instantiating module: {:?}", e);
```
shamilsan, line 351:
```suggestion
            log::trace!(target: TARGET, "Error instantiating module: {e:?}");
```

PR #3162, opened 2023-08-23
fix(node-loader): Fix anoying error with too large RPC data

SabaunT:
Pretty frequently we receive an RPC error, which tell us that RPC payload is too big. This fixes the client side of that error.

Server side is "fixed" with running the node (loaded by `gear-node-loader`) with  `--rpc-max-response-size 100` cli param.

@gear-tech/dev

### DIFF
--- a/gsdk/src/client.rs
+++ b/gsdk/src/client.rs
@@ -37,6 +37,7 @@ use subxt::{
 
 const DEFAULT_GEAR_ENDPOINT: &str = "wss://rpc-node.gear-tech.io:443";
 const DEFAULT_TIMEOUT: u64 = 60_000;
+const ONE_HUNDRED_MEGA_BYTES: u32 = 100 * 1024 * 1024;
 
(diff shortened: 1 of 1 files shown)


ukint-vs (approved):
lgtm

breathx (approved):

gsdk/src/client.rs
```diff
@@ -37,6 +37,7 @@ use subxt::{
 
 const DEFAULT_GEAR_ENDPOINT: &str = "wss://rpc-node.gear-tech.io:443";
 const DEFAULT_TIMEOUT: u64 = 60_000;
+const ONE_HUNDRED_MEGA_BYTES: u32 = 100 * 1024 * 1024;
```
shamilsan, line 40: I prefer more sensible names à la `MAX_BODY_SIZE_MB` but not critical.

shamilsan (approved):

clearloop (approved):

PR #3173, opened 2023-08-24
feat(rpc)!: Introduce payload parameter for read_state_using_wasm rpc call

DennisInSky:
Mimics #3059 

@reviewer-or-team


shamilsan (approved):

gcli/src/cmd/program.rs
```diff
@@ -93,7 +93,7 @@ impl Program {
         at: Option<H256>,
```
breathx, line 93: append field to that method as well?

breathx (approved):

PR #3176, opened 2023-08-25
chore(runtime): Adjust target inflation value

ekovalev:
The target inflation parameter has to be changed to 6% to reflect the latest changes in token economics.


pallets/staking-rewards/src/migration.rs
```diff
@@ -16,14 +16,87 @@
...
+            // Adjusted target inflation parameter: 6.00%
+            let adjusted_inflation: Perquintill = Perquintill::from_percent(6);
+            pallet::TargetInflation::<T>::put(adjusted_inflation);
+
+            current.put::<Pallet<T>>();
+
+            log::info!("Successfully migrated storage from v1 to v2");
+
+            // 1 write for `TargetInflation`
+            weight += T::DbWeight::get().writes(1)
```
breathx, line 49: you also put new storage version

pallets/staking-rewards/src/migration.rs
```diff
@@ -16,14 +16,87 @@
...
+        let inflation = pallet::TargetInflation::<T>::get();
+        assert_eq!(inflation, Perquintill::from_rational(578_u64, 10_000_u64));
+        Ok(inflation.encode())
+    }
+
+    #[cfg(feature = "try-runtime")]
+    fn post_upgrade(state: Vec<u8>) -> Result<(), &'static str> {
+        use parity_scale_codec::Decode;
+
+        let inflation: Perquintill = Decode::decode(&mut &state[..]).unwrap();
```
breathx, line 70: `state` here is an outcome of `pre_upgrade` so it's previous Inflation
here you need to assert_ne current storage value and previous one + assert_eq current and expected 6%

pallets/staking-rewards/src/migration.rs
```diff
@@ -16,14 +16,87 @@
...
+            .initial_authorities(vec![(VAL_1_STASH, VAL_1_CONTROLLER, VAL_1_AUTH_ID)])
+            .stash(VALIDATOR_STAKE)
+            .endowment(ENDOWMENT)
+            .target_inflation(Perquintill::from_rational(578_u64, 10_000_u64))
+            .build()
+            .execute_with(|| {
+                StorageVersion::new(1).put::<Pallet<Test>>();
+
+                let weight = MigrateToV2::<Test>::on_runtime_upgrade();
+                assert_ne!(weight.ref_time(), 0);
```
breathx, line 94: you can exactly assert it

pallets/staking-rewards/src/tests.rs
```diff
@@ -1183,6 +1183,57 @@ fn unclaimed_rewards_burn() {
...
+        // Getting up-to-date data on era duration (they may differ from runtime constants)
+        let sessions_per_era = <Test as pallet_staking::Config>::SessionsPerEra::get() as u64;
+        let epoch_duration = SESSION_DURATION;
+        let era_duration = sessions_per_era * epoch_duration;
+
+        let (initial_total_issuance, _, _, initial_rewards_pool_balance) = chain_state();
+        assert_eq!(initial_rewards_pool_balance, 0); // ED is auto-deducted by the getter function
+
+        // Running chain until era rollover
+        run_to_block(era_duration + 1);
```
breathx, line 1201: you can just skip blocks by setting new bn
please do the same for already existing tests

pallets/staking-rewards/src/lib.rs
```diff
@@ -367,19 +367,22 @@ impl<T: Config> OnUnbalanced<PositiveImbalanceOf<T>> for Pallet<T> {
     fn on_nonzero_unbalanced(minted: PositiveImbalanceOf<T>) {
         let amount = minted.peek();
 
-        let burned = T::Currency::withdraw(
+        if let Ok(burned) = T::Currency::withdraw(
             &Self::account_id(),
             amount,
             WithdrawReasons::TRANSFER,
             ExistenceRequirement::KeepAlive,
```
breathx, line 374: are you sure we need to keep alive?
for cases of big withdrawals this may fail each time because resulting balance will go lower than ED but with keep alive we cannot kill it, so error will be returned each time, burn won't take place while total supply increases what leads to invalidation of tokenomic model - better to have a test for it

breathx (changes requested):

pallets/staking-rewards/src/lib.rs
```diff
@@ -367,19 +367,22 @@ impl<T: Config> OnUnbalanced<PositiveImbalanceOf<T>> for Pallet<T> {
...
-
-        Self::deposit_event(Event::Burned { amount });
+        ) {
+            // Offsetting rewards against rewards pool until the latter is not depleted.
+            // After that the positive imbalance is dropped adding up to the total supply.
+            let _ = minted.offset(burned);
+
+            Self::deposit_event(Event::Burned { amount });
+        } else {
+            log::warn!(
```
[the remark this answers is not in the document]
ekovalev: Not really. It just means the inflation offset pool doesn't have enough funds.

pallets/staking-rewards/src/migration.rs
ekovalev: Right. But does it provide any extra guarantees? Being different from zero would do, would it?

pallets/staking-rewards/src/lib.rs
ekovalev: > for cases of big withdrawals

This should never be the case. The purpose of this account is to store the inflation offset pool. It is never supposed to be completely depleted: one-time withdrawal can only be as big as a single era reward for one validator, whereas the pool in the beginning is supposed to store 10% of the total token supply, that is a value of totally different order of magnitude which never gets even close to the ED. According to the tokenomics model the pool goes out of the picture in a year or two anyway.

pallets/staking-rewards/src/migration.rs
ekovalev: Of course - my bad. Corrected.

pallets/staking-rewards/src/tests.rs
ekovalev: It won't work: we need authorship and session information to be updated block by block. Not in this particular test maybe but in other tests in this pallet. I'd keep it this way for the sake of uniformity.

Anyway, removed irrelevant hooks invocations.

ukint-vs (approved):

pallets/staking-rewards/src/lib.rs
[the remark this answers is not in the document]
breathx: So wouldn't its emptiness invalidating tokenomic at all? Since it should never happen, the error seems ok for me, but ok

pallets/staking-rewards/src/migration.rs
breathx: Since we can't write proper tests such dumb check look ok for me)

breathx (approved):

pallets/staking-rewards/src/migration.rs
```diff
@@ -67,8 +69,10 @@ impl<T: Config> OnRuntimeUpgrade for MigrateToV2<T> {
     fn post_upgrade(state: Vec<u8>) -> Result<(), &'static str> {
         use parity_scale_codec::Decode;
 
-        let inflation: Perquintill = Decode::decode(&mut &state[..]).unwrap();
-        assert_eq!(inflation, Perquintill::from_percent(6),);
+        let old_inflation: Perquintill = Decode::decode(&mut &state[..]).unwrap();
```
NikVolf, line 72: no unwraps ever

pallets/staking-rewards/src/migration.rs
```diff
@@ -67,8 +69,10 @@ impl<T: Config> OnRuntimeUpgrade for MigrateToV2<T> {
...
+        let new_inflation = pallet::TargetInflation::<T>::get();
+        assert_ne!(old_inflation, new_inflation);
```
NikVolf, line 74: no asserts

breathx: It's a tests that runs locally even without any chain state

NikVolf (approved):

pallets/staking-rewards/src/lib.rs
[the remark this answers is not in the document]
ekovalev: Maybe, but it's not the concern of this pallet logic: it provides an instrument to adhere to the token economic model, but the correct use of this instrument must be ensured at a higher level.

PR #3209, opened 2023-09-02
feat(pallet-gear): enable stack limit instrumentation

grishasobol:
Enable stack limit instrumentation, because `wasmer` does not support deterministic execution in case of stack overflow (see https://github.com/wasmerio/wasmer/issues/4181#issuecomment-1697025557). Also this can cause problems for example if some nodes will use `wasmi` as host sandbox executors.

Stack limits check injection is made before gas injection, because gas injection inserts gas_charge call which cannot consumes stack significantly and does not contains recursion inside. On other hand stack limit check injection  can be expensive in some cases and need to be taken in account in gas charging. For example:
```
for i in 1..1000000 {
     call $func
}
```
In this example instrumentation injects stack counter increase and check before function call and stack counter decrease after call
```
for i in 1..1000000 {
     stack_counter += func_stack_height
     if stack_counter >= stack_limit {
           unrechable
     }
     call $func
     stack_counter -= func_stack_height
}
```


pallets/gear/src/tests.rs
```diff
@@ -7799,14 +7799,24 @@ fn gas_spent_precalculated() {
         )
     )"#;
 
-    let wat_no_counter = r#"
+    let wat_empty_init = r#"
```
grishasobol, line 7802: This f* test is ate too much of my time.

examples/fungible-token/tests/benchmarks.rs
```diff
@@ -208,15 +208,15 @@ async fn stress_test() -> Result<()> {
 async fn stress_transfer() -> Result<()> {
     let mut rng = StdRng::seed_from_u64(42);
 
-    let api = GearApi::dev_from_path(GEAR_PATH).await?;
+    let api = GearApi::dev_from_path(GEAR_PATH).await.unwrap();
```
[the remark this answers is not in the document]
breathx: No reason to have this test `-> Result<_>` tho
And replace unwraps with expects please

pallets/gear/src/benchmarking/tests/mod.rs
```diff
@@ -23,8 +23,73 @@
...
+    .into_iter()
+    .find_map(|note| match note {
+        JournalNote::MessageDispatched { outcome, .. } => Some(outcome),
+        _ => None,
+    })
+    .map(|outcome| match outcome {
+        DispatchOutcome::InitFailure { reason, .. } => {
+            assert_eq!(
+                reason,
+                "Reason is unknown. Possibly `unreachable` instruction is occurred"
```
breathx, line 89:
```suggestion
TrapExplanation::Unknown.to_string()
```

pallets/gear/src/tests.rs
```diff
@@ -7819,129 +7829,132 @@ fn gas_spent_precalculated() {
...
-            0,
-            true, true,
-        )
-        .expect("calculate_gas_info failed");
+            assert_eq!(gas_spent_calculated, gas_spent_expected);
+        };
 
-        assert_eq!(gas_spent_1, gas_spent_2);
+        // Check also, that gas spent is the same if we calculate it twice.
+        make_check(gas_spent_expected);
```
breathx, line 7956: It runs in different copies of externalities, but ok

sandbox/host/src/sandbox/wasmer_backend.rs
```diff
@@ -116,7 +117,17 @@ pub fn invoke(
...
             .call(&args)
-            .map_err(|error| Error::Sandbox(error.to_string()))
+            .map_err(|error| {
+                if error.clone().to_trap() == Some(TrapCode::StackOverflow) {
+                    // Panic stops process queue execution in that case.
+                    // This allows to avoid error lead to consensus failures, that must be handled
+                    // in node binaries forever. If this panic occur, then we must increase stack memory size,
+                    // or tune stack limit injection.
+                    // see also https://github.com/wasmerio/wasmer/issues/4181
+                    unreachable!("Suppose that this can not happen, because we have a stack limit instrumentation in programs");
```
breathx, line 127: won't it be the problem of node sync? :/

core/src/code.rs
```diff
@@ -287,33 +287,88 @@ fn check_mut_global_exports(module: &Module) -> Result<(), CodeError> {
...
+        get_gas_rules: Option<GetRulesFn>,
+        config: TryNewCodeConfig,
     ) -> Result<Self, CodeError>
     where
         R: Rules,
         GetRulesFn: FnMut(&Module) -> R,
     {
-        wasmparser::validate(&raw_code).map_err(|_| CodeError::Decode)?;
+        if config.make_validation {
+            wasmparser::validate(&original_code).map_err(|_| CodeError::Decode)?;
```
breathx, line 354: How do you feel about introducing CodeError::Validation?
To concrete each error case.

breathx (approved):
Really appreciate `core::Code` refactoring!
Awesome

pallets/gear/src/tests.rs
grishasobol: Cannot see where different copy is

sandbox/host/src/sandbox/wasmer_backend.rs
grishasobol: Yes, we should restart webs with this errors

examples/fungible-token/tests/benchmarks.rs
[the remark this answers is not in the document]
grishasobol: What the reason to do this in tests? unwrap shows error debug string, file name and line number - this is more then enough for developer to localise problem.

core/src/code.rs
grishasobol: Dood idea. Unfortunately this would not make upload_program errors more user friendly because we need to refactor pallet errors, which has one error ProgramConstrunctionFailed for all cases

pallets/gear/src/benchmarking/tests/mod.rs
grishasobol: Done

core/src/code.rs
grishasobol: Done

examples/fungible-token/tests/benchmarks.rs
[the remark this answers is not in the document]
breathx: Anyway expects with a small explanation in more intuitive

sandbox/host/src/sandbox/wasmer_backend.rs
breathx: Better to know concrete networks that should be restarted

examples/fungible-token/tests/benchmarks.rs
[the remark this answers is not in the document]
breathx: Not the deal at all, ok

breathx (approved):

sandbox/host/src/sandbox/wasmer_backend.rs
grishasobol: Vara Production - OK
VST - OK
VIT - OK
V7 - need to be re started
PST - does not check,  suppose that same as V7

sandbox/host/src/sandbox/wasmer_backend.rs
```diff
@@ -21,6 +21,7 @@
 use std::{cell::RefCell, collections::HashMap, rc::Rc};
 
 use wasmer::{Exportable, RuntimeError};
+use wasmer_types::TrapCode;
```
gshep, line 24: doesn't wasmer reexport that?

core/src/code.rs
```diff
@@ -242,16 +240,17 @@ pub enum CodeError {
...
     /// Code instrumented with the latest schedule.
     code: Vec<u8>,
     /// The uninstrumented, original version of the code.
-    raw_code: Vec<u8>,
+    original_code: Vec<u8>,
     /// Exports of the wasm module.
     exports: BTreeSet<DispatchKind>,
+    /// Static pages count from memory import.
     static_pages: WasmPage,
-    #[codec(compact)]
```
gshep, line 254: I'm curious doesn't it break backward compatibility? Namely db storage

core/src/code.rs
```diff
@@ -322,172 +382,178 @@ impl Code {
...
+    ///    (func $init
+    ///       call $f1
+    ///       call $f2
+    ///       call $f3
+    ///       <-- some code -->
+    ///    )
+    /// )
+    /// ```
+    ///
+    /// After instrumentation code will be:
```
gshep, line 456:
```suggestion
    /// After instrumentation code will look like:
```

core/src/code.rs
```diff
@@ -322,172 +382,178 @@ impl Code {
...
+    ///    (func $f3
+    ///       <-- f3 code -->
+    ///    )
+    ///    (func $init
+    ///       call $f1
+    ///       call $f2
+    ///       call $f3
+    ///       <-- some code -->
+    ///    )
+    /// )
```
gshep, line 453: It would be nice to make docs more shorter and better, for example use only two funcs.
```
/// (module
    ///    (import "env" "memory" (memory 1))
    ///    (export "init" (func $init))
    ///
    ///    (func $f1
    ///       <-- f1 code -->
    ///    )
    ///
    ///    (func $f2
    ///       if (i32.eqz (i32.const 0))
    ///          <-- some code -->
    ///       else
    ///          <-- some code -->
    ///       end
    ///    )
    ///
    ///    (func $init
    ///       call $f1
    ///       call $f2
    ///       <-- some code -->
    ///    )
    /// )
```

core/src/code.rs
```diff
@@ -322,172 +382,178 @@ impl Code {
...
+    ///      <-- some code -->
+    ///   )
+    ///   (func $init_export
+    ///      i32.const 123
+    ///      call $gas_charge
+    ///      <-- stack limit check and increase -->
+    ///      call $init
+    ///      <-- stack limit decrease -->
+    ///   )
+    /// )
```
gshep, line 509: how does `init_export` appear? It doesn't present in the original wat ).

Also why funcs bodies (f1/2/3) doesn't contain logic of stack limiting?

pallets/gear/src/lib.rs
```diff
@@ -674,24 +675,16 @@ pub mod pallet {
...
-                false,
-                true,
-            )
-            .map_err(|e| {
-                log::debug!("Code failed to load: {:?}", e);
-                Error::<T>::ProgramConstructionFailed
-            })?;
+            let code = Code::try_new_mock_const_or_no_rules(code, false, Default::default())
+                .map_err(|e| {
+                    log::debug!("Code failed to load: {:?}", e);
```
gshep, line 685: just for the sake of consistency
```suggestion
                    log::debug!("Code failed to load: {e:?}");
```

pallets/gear/src/schedule.rs
```diff
@@ -726,7 +726,15 @@ impl<T: Config> Default for Schedule<T> {
...
-            stack_height: None,
+            // Constant for `stack_height` is chosen to be small enough to avoid stack overflow in
+            // wasmer and wasmi executors. Currently it's just heuristic value.
+            // Unfortunately it's very hard to calculate this value precisely,
+            // because of difference of how stack height is calculated in injection and
+            // how wasmer and wasmi actually uses stack.
+            // To avoid potential stack overflow problems we have a panic in sandbox in case,
+            // execution is ended with stack overflow error. So, process queue execution will be
+            // stopped and we will be able to investigate the problem and decrease this constant if needed.
+            stack_height: Some(20000),
```
gshep, line 737:
```suggestion
            stack_height: Some(20_000),
```

gshep (approved):
Good job! :100:

core/src/code.rs
ukint-vs: why you removed?

ukint-vs (changes requested):

ukint-vs (approved):

sandbox/host/src/sandbox/wasmer_backend.rs
grishasobol: Cannot find this re-export. May be in future versions

core/src/code.rs
grishasobol: resolved

core/src/code.rs
grishasobol: done

core/src/code.rs
grishasobol: done

pallets/gear/src/lib.rs
grishasobol: done

pallets/gear/src/schedule.rs
grishasobol: done

core/src/code.rs
grishasobol: It's logic of stack injection. Each call instruction is surrounded by stack check. ini_export is in order to surround export function

sandbox/host/src/sandbox/wasmer_backend.rs
grishasobol: PST need to be restarted

PR #3239, opened 2023-09-10
feat(runtime)!: Add customisation for gas price

breathx:


ukint-vs (approved):

common/src/gas_provider/mod.rs
```diff
@@ -235,6 +258,9 @@ pub trait Provider {
     /// Type representing a quantity of value.
     type Balance;
 
+    /// Type representing a quantity of token balance.
+    type Funds;
```
gshep, line 262: For the sake of readability I would suggest to rename this field to `ValueBalance` and the former one to `GasBalance`.

gcli/tests/cmd/claim.rs
```diff
@@ -23,7 +23,7 @@ use crate::common::{
 };
 use gsdk::Api;
 
-const REWARD_PER_BLOCK: u128 = 3_000_000; // 3_000 gas * 1_000 value per gas
+const REWARD_PER_BLOCK: u128 = 75_000; // 3_000 gas * 25 value per gas
```
gshep, line 26: Can we use the corresponding constant from runtime(s) there?

gshep (approved):
:+1:

PR #3242, opened 2023-09-11
feat(wasm-gen): insert syscall invocations into wasm wisely

### LINKED ISSUE #3093: Insert sys-calls invocations into functions wisely
### Proposal

Positions to insert the sys call must be chosen wisely, so there won't be a situation when you insert sys-call in the middle of valid instructions which call another sys-calll.

mertwole:
Resolves #3093


utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -169,21 +171,23 @@ impl<'a, 'b> SysCallsInvocator<'a, 'b> {
             self.unstructured.len()
         );
 
+        let mut invoke_instructions = vec![];
```
SabaunT, line 174: `with_capacity`

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -169,21 +171,23 @@ impl<'a, 'b> SysCallsInvocator<'a, 'b> {
...
                 invocable.to_str(),
                 amount
             );
 
-            for instructions in iter::repeat(&instructions).take(amount as usize) {
-                self.insert_sys_call_instructions(instructions)?;
+            for _ in 0..amount {
+                let instructions =
+                    self.build_sys_call_invoke_instructions(invocable, call_indexes_handle)?;
+                invoke_instructions.push(instructions);
```
SabaunT, line 185: what were reasons for that? we will increase amount of "build sys-call" logs by "amount"

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -513,51 +517,73 @@ impl<'a, 'b> SysCallsInvocator<'a, 'b> {
...
+                insertion_mapping.partition_point(|&(func_id, _)| func_id < insert_into);
+
+            if !insertion_mapping.is_empty()
+                && partition != insertion_mapping.len()
+                && insertion_mapping[partition].0 == insert_into
+            {
+                insertion_mapping[partition].1.push(ixs);
+            } else {
+                insertion_mapping.insert(partition, (insert_into, vec![ixs]));
+            }
```
SabaunT, line 544: low level details must be moved  into other fn

SabaunT (commented):
why did you decide to insert only into custom funcs. when it was previously denied?

utils/wasm-gen/src/generator/syscalls/invocator.rs
mertwole: I'll need to iterate over `sys_calls_imports` to achieve this. And also this `Vec` contains `Vec`s, which have the size of around 16 bytes

utils/wasm-gen/src/generator/syscalls/invocator.rs
mertwole: I think it's more desirable to have different args to the syscalls, as they all have been inserted with the same args in the previous implementation

SabaunT: wise, but make sure the logs will be well-organised and easy to investigate

utils/wasm-gen/src/generator/syscalls/invocator.rs
SabaunT: what about calling inserting sys-calls invocation instructions only for one sys-call per iteration, but not to collect all instructions?

utils/wasm-gen/src/generator/syscalls/invocator.rs
mertwole: resolved?

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -529,52 +515,139 @@ impl<'a, 'b> SysCallsInvocator<'a, 'b> {
...
+        let syscalls_to_insert = self
+            .sys_call_imports
+            .clone()
+            .into_iter()
+            .map(|(syscall, (amount, _))| {
+                iter::repeat(syscall)
+                    .take(amount as usize)
+                    .collect::<Vec<_>>()
+            })
+            .flatten();
```
StackOverflowExcept1on, line 538: use `.flat_map()`, clippy should show how
```
$ cargo clippy -p gear-wasm-gen
warning: called `map(..).flatten()` on `Iterator`
```

StackOverflowExcept1on: maybe also enable CI in .github/workflows/PR.yml?
like here: https://github.com/gear-tech/gear/pull/3262/commits/f4f9965ac12c74e6bf93e1eb1ff13158491fded5

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -529,52 +515,139 @@ impl<'a, 'b> SysCallsInvocator<'a, 'b> {
...
+
+        // Sort in descending order.
+        insertion_positions.sort_by(|a, b| b.0.cmp(&a.0));
+
+        for (pos, syscall) in insertion_positions {
+            let call_indexes_handle = self
+                .sys_call_imports
+                .get(&syscall)
+                .expect("Syscall presented in sys_call_imports")
+                .1;
```
SabaunT, line 621: use `.map(|(_, call_indexes_handle)| call_indexes_handle)`
avoid ambiguous `.0` and `.1`

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -529,52 +515,139 @@ impl<'a, 'b> SysCallsInvocator<'a, 'b> {
...
+        // index is used as an insertion point.
+        let last = if fn_code_len > 1 { fn_code_len - 2 } else { 0 };
+
+        let mut insertion_positions = syscalls
+            .into_iter()
+            .map(|syscall| Ok((self.unstructured.int_in_range(0..=last)?, syscall)))
+            .collect::<Result<Vec<_>>>()?;
+
+        // Sort in descending order.
+        insertion_positions.sort_by(|a, b| b.0.cmp(&a.0));
```
SabaunT, line 614: please, give explanation above this that it gives guarantee that no inserting invocations instructions will intersect each other

SabaunT (commented):
Please follow the common line for giving `sys_call` name to variables and functions and using `sys-call` in logs. I understand, that `syscall` can be used too, but we should follow one path to make it clear and easy to read the code.

utils/wasm-gen/src/generator/syscalls/invocator.rs
SabaunT: Are logs well organised after that and not confusing?

utils/wasm-gen/src/wasm.rs
```diff
@@ -167,6 +167,14 @@ impl WasmModule {
     }
 }
 
+impl Deref for WasmModule {
```
SabaunT, line 170: It's against design. Please remove the `Deref`.

SabaunT (changes requested):

utils/wasm-gen/src/wasm.rs
mertwole: It provides readonly access to the underlying struct so why not to give it to the user? Anyway it already have method into_inner that can be used to expose underlying type

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -529,52 +515,139 @@ impl<'a, 'b> SysCallsInvocator<'a, 'b> {
...
-            let res = self.unstructured.int_in_range(0..=last).map(|pos| {
-                log::trace!(" -- Inserting into position {pos}");
-                code.splice(pos..pos, instructions.iter().cloned());
-            });
+        Ok(())
+    }
+
+    fn get_syscall_insertion_mapping<I>(
+        &mut self,
+        syscalls: I,
```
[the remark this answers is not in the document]
mertwole: `syscalls_to_insert` aren't collected from iterator, so this helps evaluate mapping lazily

utils/wasm-gen/src/wasm.rs
SabaunT: calling `into_inner` is a signal that wasm generation work-flow is done. Check generatorts transitions.

SabaunT: User must not do anything with `Module` when he gives it to generator.

StackOverflowExcept1on: It seems to look legit because `Deref` provides immutable reference to `Module`. So, user will not be able to do anything with it.

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -173,43 +178,161 @@ impl<'a, 'b> SysCallsInvocator<'a, 'b> {
...
+                }
+                Entry::Vacant(entry) => {
+                    entry.insert(vec![syscall]);
+                }
+            }
+        }
+
+        let mut insertion_mapping: Vec<_> = insertion_mapping.into_iter().collect();
+        // Sort it to avoid indeterminance appeared because of HashMap usage.
+        insertion_mapping.sort_by(|a, b| a.0.cmp(&b.0));
```
SabaunT, line 260: if you sort by func ids, maybe just use `BTreeMap` instead of `HashMap`?

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -173,43 +178,161 @@ impl<'a, 'b> SysCallsInvocator<'a, 'b> {
...
+        // The end of insertion range is second-to-last index, as the last
+        // index is defined for `Instruction::End` of the function body.
+        // But if there's only one instruction in the function, then `0`
+        // index is used as an insertion point.
+        let last = if fn_code_len > 1 { fn_code_len - 2 } else { 0 };
+
+        let mut insertion_positions = syscalls
+            .into_iter()
+            .map(|syscall| Ok((self.unstructured.int_in_range(0..=last)?, syscall)))
+            .collect::<Result<Vec<_>>>()?;
```
SabaunT, line 286: not sure whether more optimised, but you can collect it to min binary heap.

SabaunT: do not implement if it doesn't require only changing the `collect` stuff with removing the `insertion_positions.sort_by(|a, b| b.0.cmp(&a.0));`

SabaunT (approved):

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -173,43 +178,156 @@ impl<'a, 'b> SysCallsInvocator<'a, 'b> {
...
+    /// Returns mapping `func_id` <-> `syscalls which should be inserted into func_id`.
+    fn build_syscalls_insertion_mapping<I>(
+        &mut self,
+        syscalls: I,
+        insert_into_funcs: &[usize],
+    ) -> Result<BTreeMap<usize, Vec<InvocableSysCall>>>
+    where
+        I: Iterator<Item = InvocableSysCall>,
+    {
+        let mut insertion_mapping: BTreeMap<usize, Vec<InvocableSysCall>> = BTreeMap::new();
```
StackOverflowExcept1on, line 243: type can be auto inferred
```suggestion
        let mut insertion_mapping = BTreeMap::new();
```

StackOverflowExcept1on (approved):

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -162,6 +165,8 @@ impl ParamSetter {
     }
 }
 
+pub type SysCallInvokeInstructions = Vec<Instruction>;
```
breathx, line 168: Strange alias

breathx (approved):

utils/wasm-gen/src/generator/syscalls/invocator.rs
mertwole: It's returned from fn as `Vec<SysCallInvokeInstructions>`, so without it return type will be like `Vec<Vec<_>>`

PR #3250, opened 2023-09-11
fix(node-loader): Handle tx not found errors

SabaunT:
When Tx not found error occurres, all following batches end up with an error `Rpc error: RPC error: The background task been terminated because: Invalid request ID; restart required`. That's very hard to reproduce and pretty obscure situation, but from error string it seems that we need not to re-instantiate client (`GearApi` instance).

This PR is a try to solve the issue with a little blood.

@gear-tech/dev


mertwole (approved):

breathx (approved):

ukint-vs (approved):

SabaunT:
Closing as our `try-runtime` tests had the same error, which was mainly caused by the size of the request/response. Closing in favor of adding this `--rpc-max-request-size 100000 --rpc-max-response-size 100000` when starting node.

PR #3256, opened 2023-09-12
feat: add support for Linux on ARM

shamilsan:
- Added support for Linux on ARM.

@gear-tech/dev


breathx (changes requested):
We should never allow compilation for targets we don't support (mainten/test etc)

PR #3315, opened 2023-09-20
fix(gstd-async): Handle locks for default wait sendings

breathx:


gstd/src/async_runtime/signals.rs
```diff
@@ -51,14 +51,18 @@ impl WakeSignals {
...
             waiting_reply_to,
             WakeSignal {
-                message_id: crate::msg::id(),
+                message_id,
                 payload: None,
                 waker: None,
             },
         );
+
+        crate::async_runtime::locks().lock(message_id, waiting_reply_to, Default::default());
```
DennisInSky, line 65: Would be great to have some comment here explaining why the lock is required

DennisInSky (approved):

gstd/src/async_runtime/signals.rs
breathx: To be added later in covering with logs

PR #3335, opened 2023-09-22
feat(utils): add util to retrieve default peer ID secret keys

### LINKED ISSUE #3311: Create a Script to Retrieve the Default Node Secret Key
### Problem to Solve

We need script for gear noders.
This script is designed to locate the gear-node path and extract the secret key for exporting into your wallet.

### Possible Solution

Navigate to the gear-node directory: cd gear-node/chains/vara/keystore
keep in minde secret key in ED25519 format, which is not compatible with most wallets.
For greater compatibility with wallets, consider providing the secret key as a seed phrase."

### Notes

_No response_

mqxf:
Resolves #3311

@gear-tech/dev


utils/key-finder/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+
+[[bin]]
+name = "gear-key-finder"
+path = "src/main.rs"
+
+[dependencies]
+hex = {workspace = true, features = ["alloc"] }
```
shamilsan, line 12:
```suggestion
hex = { workspace = true, features = ["alloc"] }

```

utils/key-finder/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+const APPDATA: &'static str = r"/.local/share/";
+
+#[cfg(target_os = "macos")]
+const APPDATA: &'static str = r"/Library/Application Support/";
+
+#[cfg(target_os = "windows")]
+const APPDATA: &'static str = r"\AppData\Local\";
+
+#[cfg(not(any(unix, windows, target_os = "macos")))]
+compile_error!("Unsupported OS");
```
shamilsan, line 15: You may just use [`sc_service::config::BasePath`](https://docs.subsocial.network/rust-docs/latest/sc_service/config/enum.BasePath.html) as we do it here for example:

https://github.com/gear-tech/gear/blob/master/node/cli/src/command.rs#L150

utils/key-finder/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+    let keys = PATHS.iter().flat_map(|p| {
+        let mut path = home.clone() + APPDATA + p;
+        if Path::new(&path).is_dir() {
+            path.push(MAIN_SEPARATOR);
+            path.push_str("chains");
+            path.push(MAIN_SEPARATOR);
+            VERSIONS.iter().filter_map(|v| {
+                let mut path = path.clone() + v;
+                if Path::new(&path).is_dir() {
+                    path.push(MAIN_SEPARATOR);
```
shamilsan, line 47: You may use [`Path::join`](https://doc.rust-lang.org/std/path/struct.Path.html#method.join) to not mess around path separator.

utils/key-finder/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+        }
+        else {
+            eprintln!("Failed to read key from {}", path);
+            None
+        }
+    }).collect::<Vec<_>>();
+
+    for key in keys {
+        println!("Key found in \"{}\": 0x{}", key.0, key.1);
+    }
```
shamilsan, line 73: You can just find and print in one loop. No need to use iterators here.

utils/key-finder/src/main.rs
shamilsan: Or even [`directories::ProjectDirs`](https://docs.rs/directories/latest/directories/struct.ProjectDirs.html) as `BasePath` does: https://docs.subsocial.network/rust-docs/latest/src/sc_service/config.rs.html#242-248

utils/key-finder/src/main.rs
```diff
@@ -0,0 +1,54 @@
...
+        })
+        .filter_map(|(path, p, v)| {
+            if let Ok(key) = std::fs::read(&path) {
+                Some((p.to_string() + "/" + v, encode(key)))
+            } else {
+                eprintln!("Failed to read key from {:?}", path);
+                None
+            }
+        })
+        .for_each(|(path, key)| {
```
shamilsan, line 51: Isn't nested loop just shorter and clearer?

```suggestion
    for path in PATHS {
        let chains_path = ProjectDirs::from("", "", path)
            .unwrap()
            .data_local_dir()
            .join("chains");
        if chains_path.is_dir() {
            for version in VERSIONS {
                let mut key_path = chains_path.join(version);
                key_path.extend(&["network", "secret_ed25519"]);
                if key_path.is_file() {
                    let key = std::fs::read(&key_path);
                    if let Ok(key) = key {
                        println!("Key found in \"{path}/{version}\": 0x{}", hex::encode(key));
                    } else {
                        eprintln!("Failed to read key from {path:?}",);
                    }
                }
            }
        }
    }
```

shamilsan (approved):

StackOverflowExcept1on (approved):

StackOverflowExcept1on:
I'm not sure, but maybe we can move this code to `./gear` subcommand?

shamilsan:
> I'm not sure, but maybe we can move this code to `./gear key` subcommand?

This will be a one-time utility for specific needs.

PR #3341, opened 2023-09-22
feat(gstd): Mutex lock monitor

DennisInSky:
This PR implements functionality with regards to mutex locks when one lock is still in bound of its duration, i.e. it can't be ousted, whereas another lock comes, but when the first lock excceeds its time limit it has to be ousted, and the second one should get into possesion


gstd/src/sync/mutex.rs
```diff
@@ -30,6 +30,22 @@ use core::{
...
+pub(crate) struct MutexId(u32);
+
+impl MutexId {
+    pub const fn new() -> Self {
+        MutexId(1)
+    }
+
+    pub fn next(self) -> Self {
+        let id = self.0.wrapping_add(1);
+        MutexId(if id == 0 { 1 } else { id })
```
breathx, line 45: ?

gstd/src/sync/mutex.rs
```diff
@@ -121,6 +140,15 @@ impl<T> Mutex<T> {
     fn locked_by_mut(&self) -> &mut Option<(MessageId, BlockNumber)> {
         unsafe { &mut *self.locked.get() }
     }
+
+    fn get_or_set_id(&self) -> MutexId {
```
breathx, line 144: so is it getter or setter tho?

gstd/src/async_runtime/locks.rs
```diff
@@ -173,35 +179,58 @@ impl LocksMap {
...
+
+    pub(crate) fn insert_mx_lock_monitor(
+        &mut self,
+        message_id: MessageId,
+        mutex_id: MutexId,
+        wake_up_at: BlockNumber,
+    ) {
+        let locks = self.message_locks(message_id);
+        locks.insert(
+            LockContext::MxLockMonitor(mutex_id),
```
breathx, line 221: and where is it processed?

breathx (changes requested):

gstd/src/sync/mutex.rs
DennisInSky: Re-written to make the intent more clear

gstd/src/sync/mutex.rs
DennisInSky: It is sort of lazy assignment. Renamed to make it more clear

gstd/src/async_runtime/locks.rs
DennisInSky: It is processed same way as all the other locks when their `wait` method is called. The purpose of this lock is waking up a message owning an mx lock after its ownership period elapses.

breathx (approved):

NikVolf (approved):

PR #3345, opened 2023-09-23
[dep] ci(dep): introduce label `[dep]`

clearloop:
Introduce `[dep]` prefix for appending passed `build / linux` for dependabot's PRs, we should leave the `check / linux` with dependabot's PRs since we need to forbid merging them separately when there is sth wrong

@gear-tech/dev

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,7 +7,7 @@ updates:
       interval: "weekly"
       day: "friday"
     commit-message:
-      prefix: "[skip-ci] "
+      prefix: "[dep] "
 
   - package-ecosystem: "cargo"
     directory: "/examples/"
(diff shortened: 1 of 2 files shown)


tltsutltsu (approved):

.github/dependabot.yml
```diff
@@ -2,27 +2,27 @@ version: 2
...
-    labels: ["A0-pleasereview", "A4-insubstantial", "E2-forcemacos"]
+    labels: ["A4-insubstantial", "E2-forcemacos"]
     schedule:
       interval: "weekly"
       day: "friday"
     commit-message:
-      prefix: "[depbot] "
+      prefix: "[dep] "
 
   - package-ecosystem: "cargo"
```
breathx, line 12: Seems this step isn't required because we have unified workspace for main crates and demos

breathx (approved):

clearloop:
https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/customizing-dependency-updates#grouping-dependabot-version-updates-into-one-pull-request

mb this will be the better solution

PR #3347, opened 2023-09-24
ci(dep): introduce groups feature in dependabot

clearloop:
Ref #3345 

dependabot has introduced grouped feature in 2023-06-30 the https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/

## Open Question

How to test it...

@gear-tech/dev


.github/dependabot.yml
```diff
@@ -2,27 +2,39 @@ version: 2
 updates:
   - package-ecosystem: "cargo"
     directory: "/"
-    labels: ["A0-pleasereview", "A4-insubstantial", "E2-forcemacos"]
+    labels: ["A4-insubstantial", "E2-forcemacos"]
     schedule:
       interval: "weekly"
       day: "friday"
```
breathx, line 8: and consider moving it on sat-sun mb to avoid hanging ci on Friday

breathx (approved):

breathx:
Can we provide on workflow dispatch as well to test it?

shamilsan (approved):
Agree with @breathx about the `/examples/` entry. We don't need it anymore.

.github/dependabot.yml
clearloop: moved to `Sunday` so that we can have a chill Saturday and handle it on Monday

PR #2516, opened 2023-04-06
feat(gcli): support new meta

clearloop:
Resolves #2367 

- [x] displaying new metadata
- [x] remove wasm executor in `gcli`
- [x] read `MetawasmData` from `core-processor`
- [x] displaying MetawasmData ( function exports )
- [x] displaying derived type
- [x] merge command `upload-program` into `upload`

<details>
  <summary><kbd>gcli</kbd> <kbd>program</kbd> <kbd>meta</kbd> <kbd>*.txt</kbd></summary>
  <pre><code>
2:08AM ./target/release/gcli program meta examples/binaries/new-meta/demo_new_meta.meta.txt
Metadata {
    init:  {
        input: MessageInitIn {
            amount: u8,
            currency: String,
        },
        output: MessageInitOut {
            exchange_rate: Result&ltu8, u8&gt,
            sum: u8,
        },
    },
    handle:  {
        input: MessageIn {
            id: Id,
        },
        output: MessageOut {
            res: Option&ltWallet&gt,
        },
    },
    reply:  {
        input: str,
        output: [u16],
    },
    others:  {
        input: MessageAsyncIn {
            empty: (),
        },
        output: Option&ltu8&gt,
    },
    signal: (),
    state: [Wallet { id: Id, person: Person }],
}
  </code></pre>
</details>

<details>
  <summary><kbd>gcli</kbd> <kbd>program</kbd> <kbd>meta</kbd> <kbd>*.txt</kbd> <kbd>--derive</kbd> <kbd>Type</kbd></summary>
  <pre><code>
2:08AM ./target/release/gcli program meta examples/binaries/new-meta/demo_new_meta.meta.txt --derive Wallet
Wallet { id: Id, person: Person }
  </code></pre>
</details>

<details>
  <summary><kbd>gcli</kbd> <kbd>program</kbd> <kbd>meta</kbd> <kbd>*.wasm</kbd></summary>
  <pre><code>
2:08AM ./target/release/gcli program meta target/wasm32-unknown-unknown/release/demo_meta_state_v1.meta.wasm
Exports {
    first_and_last_wallets:  {
        input: (),
        output: (
            Option&ltWallet&gt,
            Option&ltWallet&gt,
        ),
    },
    first_wallet:  {
        input: (),
        output: Option&ltWallet&gt,
    },
    last_wallet:  {
        input: (),
        output: Option&ltWallet&gt,
    },
}
  </code></pre>
</details>

<details>
[...]

PR #3366, opened 2023-09-27
feat(gcli): introduce metadata executor

### LINKED ISSUE #3222: Refactor metadata parser in gcli
### Problem to Solve

`gcli` requires `gear-sandbox-host` After removing `gear-backend-wasmi`, which has `sp-allocator` as dependency, since substrate has archived their repo, this dependency is really hard to maintain

### Possible Solution

implement the metadata parser on our own like before if there is not a better solution

### Notes

ref #3103

clearloop:
Ref: #2516 
Resolves #3222

- Introduce metadata executor

@gear-tech/dev


gcli/executor.rs
```diff
@@ -0,0 +1 @@
+
```
breathx, line 1: ?

gcli/src/meta/executor.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    let instance = linker
+        .instantiate(&mut store, &module)
+        .unwrap()
+        .start(&mut store)?;
+
+    let metadata = instance
+        .get_export(&store, method)
+        .and_then(Extern::into_func)
+        .ok_or_else(|| anyhow!("could not find function \"metadata\""))?
```
breathx, line 73:
```suggestion
        .ok_or_else(|| anyhow!("could not find function \"$method\""))?
```

gcli/src/meta/funcs.rs
```diff
@@ -0,0 +1,156 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use super::executor::HostState;
+use wasmi::{
+    core::{memory_units::Pages, Trap, TrapCode},
+    AsContext, AsContextMut, Caller, Extern, Func, Memory, Store,
+};
+
```
breathx, line 24: I don't actually see the reason of separation these fns from executor mod

breathx (approved):

breathx:
@ark0f @shamilsan please take a look

gcli/src/meta/executor.rs
```diff
@@ -0,0 +1,218 @@
...
+    pub fn gr_read(ctx: &mut Store<HostState>, memory: Memory) -> Extern {
+        Extern::Func(Func::wrap(
+            ctx,
+            move |mut caller: Caller<'_, HostState>, at: u32, len: i32, buff: i32, err: i32| {
+                let (at, len, buff, err) = (at as _, len as _, buff as _, err as _);
+
+                let msg = &caller.host_data().msg;
+                let mut payload = vec![0; len];
+                if at + len <= msg.len() {
+                    payload.copy_from_slice(&msg[at..(at + len)]);
```
shamilsan, line 131: Could we alloc it inside the `if` branch? Otherwise, `payload` allocates and deallocates memory in case of an error.
```suggestion
                let payload = if at + len <= msg.len() {
                    msg[at..(at + len)].to_vec();
```

shamilsan (approved):

gcli/src/meta/executor.rs
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    pub fn free(ctx: impl AsContextMut) -> Extern {
+        Extern::Func(Func::wrap(ctx, |_: i32| 0))
+    }
+
+    pub fn gr_panic(ctx: &mut Store<HostState>, _memory: Memory) -> Extern {
+        Extern::Func(Func::wrap(
+            ctx,
+            move |mut _caller: Caller<'_, HostState>, _ptr: u32, _len: i32| {},
```
ark0f, line 111: What about log here too?

gcli/src/meta/executor.rs
```diff
@@ -0,0 +1,218 @@
...
+use anyhow::{anyhow, Result};
+use wasmi::{AsContextMut, Engine, Extern, Linker, Memory, MemoryType, Module, Store};
+
+const PAGE_STORAGE_PREFIX: [u8; 32] = *b"gcligcligcligcligcligcligcligcli";
+
+/// HostState for the WASM executor
+#[derive(Default)]
+pub struct HostState {
+    pub msg: Vec<u8>,
+    pub timestamp: u64,
```
ark0f, line 28: Field is not used

ark0f (changes requested):

gcli/src/meta/executor.rs
```diff
@@ -0,0 +1,224 @@
...
+use anyhow::{anyhow, Result};
+use wasmi::{AsContextMut, Engine, Extern, Linker, Memory, MemoryType, Module, Store};
+
+const PAGE_STORAGE_PREFIX: [u8; 32] = *b"gcligcligcligcligcligcligcligcli";
+
+/// HostState for the WASM executor
+#[derive(Default)]
+pub struct HostState {
+    pub msg: Vec<u8>,
+    pub height: u64,
```
ark0f, line 28: The field is not used too. I guess they should be used in `gr_block_height` and `gr_block_timestamp`?

clearloop: actually I'm not pretty sure about these two functions since I don't see gmeta using them but they are required in the generated metadata WASM, so I just gave them empty implementation

clearloop: oh I got it 

https://github.com/gear-tech/gear/blob/master/examples/new-meta/state-v3/src/wasm.rs

breathx: > oh I got it
> 
> https://github.com/gear-tech/gear/blob/master/examples/new-meta/state-v3/src/wasm.rs
> 
> `block_number` and `timestamp` require real execution env, however, the meta command in gcli only parse the data types of the state of programs, so we actually don't have to implement these two functions in the metadata executor for now

so maybe there is no need in calling it like "executor", introduce just "fn call_metadata()" and that is?

gcli/src/meta/executor.rs
```diff
@@ -0,0 +1,228 @@
...
+
+    // Execution environment
+    {
+        let memory = Memory::new(store.as_context_mut(), MemoryType::new(256, None)).unwrap();
+        linker.define("env", "memory", Extern::Memory(memory))?;
+        linker.define("env", "gr_read", funcs::gr_read(&mut store, memory))?;
+        linker.define("env", "alloc", funcs::alloc(&mut store, memory))?;
+        linker.define("env", "free", funcs::free(&mut store))?;
+        linker.define("env", "gr_size", funcs::gr_size(&mut store, memory))?;
+        linker.define("env", "gr_reply", funcs::gr_reply(&mut store, memory))?;
```
breathx, line 54: gr_debug may appear as well

breathx: and gr_reply_push + gr_reply_commit

breathx (changes requested):

ark0f (approved):

gcli/src/meta/executor.rs
clearloop: https://github.com/gear-tech/gear/pull/3366/commits/ad94391ea95f1886ef5a0e563afb0e38455c6f62

gcli/src/meta/executor.rs
clearloop: https://github.com/gear-tech/gear/pull/3366/commits/18cdf6bb414348bf9cdf3708b16567e1e8597c19

gcli/src/meta/executor.rs
```diff
@@ -0,0 +1,318 @@
...
+
+    let mut store = Store::new(&engine, HostState::default());
+    let mut linker = <Linker<HostState>>::new();
+
+    // Execution environment
+    //
+    // TODO: refactor this after #3416.
+    {
+        let memory = Memory::new(store.as_context_mut(), MemoryType::new(256, None)).unwrap();
+        linker.define("env", "memory", Extern::Memory(memory))?;
```
breathx, line 77: why dont you want to iter through SysCall enum as its done in prod backend? except calls you really need

breathx (changes requested):

gcli/src/meta/executor.rs
clearloop: When the executor loads a WASM module, it validates all of the imports of the module.

For example, if we don't have imports of `block_height` and `block_timestamp`, we can parse metadata for `new-meta-v1` and `new-meta-v2` but not `new-meta-v3`, because it will be failed on the module validation ( missing imports `block_height` and `block_timestamp` )

breathx: I mean add bindings to linker from iter of all syscalls instead of their manual declaration here

clearloop: oo got it

breathx (approved):

PR #3364, opened 2023-09-26
feat(vara): Store dust in treasury

ukint-vs:
Dust will move into treasury pot.

@reviewer-or-team

PR #3381, opened 2023-09-30
feat(vara,migrations): Revive burnt dust into the treasury

ukint-vs:
In addition to #3364 .

@gear-tech/dev


ekovalev (approved):

breathx:
This migrations are outdated and no more needed, cause supply is bigger than expected and we need to apply any of solutions only after #3405

Issue #3382, opened 2023-10-01
Lower amount of traps when `Control` instructions are off for wasms from `wasm-gen`


SabaunT:
### Proposal

If you turn off `Control` instructions from `wasm-gen`, the execution will become really fast, but the rate of `Traps` (`Possibly 'unreachable' instruction occurred`) becomes too high. There are basically 2 main reasons (wasmi): `out of bounds memory access` and `integer divide by zero`. The task is to lower the rate of these erros.

StackOverflowExcept1on:
~1/2 of unreachables comes from `free` syscall: 
https://github.com/gear-tech/gear/blob/5b931e25403d62a1231c117c63baeed533972147/core-backend/src/funcs.rs#L330-L340

Issue #3313, opened 2023-09-20
Instrument loops in wasms generated with `wasm-gen` with breaks


SabaunT:
Heavy loops are the main reason why `Not enough gas` error occurs during fuzz run. To avoid having this error frequently, an instrumentation for `loop` should be done.

PR #3383, opened 2023-10-01
refactor(wasm-gen, runtime-fuzzer): Increase successful messages execution rate in `runtime-fuzzer`.

SabaunT:
This task was a hot one from the beginning of `runtime-fuzzer` launch when first production coverages were available. The issue was that there were too many traps, which caused a very weak coverage for messaging logic in the `pallet_gear`.

After researching it thoroughly, I have found out that there were 2 many trap reasons: `Possibly 'unreachable' instruction reached` and `Not enough gas to continue execution`. 
The former trap reason was caused not only by `unreachable` instruction occurrences, but there were some other internal wasmi errors causing that termination reason. I have turned off `unreachable` instructions here https://github.com/gear-tech/gear/pull/3307.
The latter one was mainly due to large loops in wasm - each function body consisted of about 100_000 instructions and these instructions could call other functions with the same size in a loop. Turning off `Control` instructions can be a solution to that, but it brings some other problems described here https://github.com/gear-tech/gear/issues/3382. This seems to be the best solution to the problem - https://github.com/gear-tech/gear/issues/3313.

Basically, there are 3 available params that have a significant effect on execution:
1. Turning off `Control` (`if`, `loop`, `try`, `catch` and etc.)  instructions. This makes the execution of the ~8gb corpus directory really fast (3 minutes), there are no `Not enough gas to continue execution` errors, but it there is a still a huge number of traps.
2. Instructions number in the function body.
3. Amount of internal functions generated by `wasm-smith`.

After running 2 corpuses sets each about 8 gb with different values for these params, I have found out that amount of instructions doesn't have a huge effect on execution time or execution results. In theory, less instructions make less gas errors and more frequent syscalls invocations. In practice, this has a very weak effect when amount of functions is large. The third  param caused the most effect on increasing the corpus execution speed and amount of successfull executions. Even if there are 100_000 instructions, but less functions, it empirically shows great results.

### DIFF
--- a/utils/runtime-fuzzer/src/gear_calls.rs
+++ b/utils/runtime-fuzzer/src/gear_calls.rs
@@ -43,6 +43,14 @@ use gear_wasm_gen::{
 const MAX_PAYLOAD_SIZE: usize = 512 * 1024;
 static_assertions::const_assert!(MAX_PAYLOAD_SIZE <= gear_core::message::MAX_PAYLOAD_SIZE);
 
+/// Maximum salt size for the fuzzer - 512 bytes.
+///
+/// There's no need in large salts as we have only 35 extrinsics
+/// for one run. Also small salt will make overall size of the
+/// corpus smaller.
+const MAX_SALT_SIZE: usize = 512;
+static_assertions::const_assert!(MAX_SALT_SIZE <= gear_core::message::MAX_PAYLOAD_SIZE);
+
 /// This trait provides ability for [`ExtrinsicGenerator`]s to fetch messages
 /// from mailbox, for example [`UploadProgramGenerator`] and
 /// [`ClaimValueGenerator`] use it.
@@ -238,8 +246,8 @@ impl UploadProgramGenerator {
     }
 
     const fn unstructured_size_hint(&self) -> usize {
-        // 1024 KiB for payload and salt and 50 KiB for code.
-        1080 * 1024
+        // 512 KiB for payload, 50 KiB for code and 512 bytes for salt.
+        572 * 1024 + 512
     }
 }
 
@@ -377,7 +385,7 @@ fn arbitrary_message_id_from_mailbox(
 }
 
 fn arbitrary_salt(u: &mut Unstructured) -> Result<Vec<u8>> {
-    arbitrary_limited_bytes(u, MAX_PAYLOAD_SIZE)
+    arbitrary_limited_bytes(u, MAX_SALT_SIZE)
 }
 
 fn arbitrary_payload(u: &mut Unstructured) -> Result<Vec<u8>> {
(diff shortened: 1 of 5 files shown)


gshep (approved):

utils/runtime-fuzzer/src/gear_calls.rs
```diff
@@ -238,8 +246,8 @@ impl UploadProgramGenerator {
     }
 
     const fn unstructured_size_hint(&self) -> usize {
-        // 1024 KiB for payload and salt and 50 KiB for code.
-        1080 * 1024
+        // 512 KiB for payload, 50 KiB for code and 512 bytes for salt.
+        572 * 1024 + 512
```
mertwole, line 250:
```suggestion
        const AUXILIARY_SIZE: usize = 512;
        MAX_PAYLOAD_SIZE + 50 * 1024 + MAX_SALT_SIZE + AUXILIARY_SIZE
```

mertwole (approved):

utils/runtime-fuzzer/src/gear_calls.rs
SabaunT: done

SabaunT:
DON'T MERGE, some performance checks are running

SabaunT:
> This task was a hot one from the beginning of `runtime-fuzzer` launch when first production coverages were available. The issue was that there were too many traps, which caused a very weak coverage for messaging logic in the `pallet_gear`.
> 
> After researching it thoroughly, I have found out that there were 2 many trap reasons: `Possibly 'unreachable' instruction reached` and `Not enough gas to continue execution`. The former trap reason was caused not only by `unreachable` instruction occurrences, but there were some other internal wasmi errors causing that termination reason. I have turned off `unreachable` instructions here #3307. The latter one was mainly due to large loops in wasm - each function body consisted of about 100_000 instructions and these instructions could call other functions with the same size in a loop. Turning off `Control` instructions can be a solution to that, but it brings some other problems described here #3382. This seems to be the best solution to the problem - #3313.
> 
> Basically, there are 3 available params that have a significant effect on execution:
> 
> 1. Turning off `Control` (`if`, `loop`, `try`, `catch` and etc.)  instructions. This makes the execution of the ~8gb corpus directory really fast (3 minutes), there are no `Not enough gas to continue execution` errors, but it there is a still a huge number of traps.
> 2. Instructions number in the function body.
> 3. Amount of internal functions generated by `wasm-smith`.
> 
> After running 2 corpuses sets each about 8 gb with different values for these params, I have found out that amount of instructions doesn't have a huge effect on execution time or execution results. In theory, less instructions make less gas errors and more frequent syscalls invocations. In practice, this has a very weak effect when amount of functions is large. The third param caused the most effect on increasing the corpus execution speed and amount of successfull executions. Even if there are 100_000 instructions, but less functions, it empirically shows great results.

After additional performance checks it occurred that changing amount of functions and lowering instructions doesn't actually give much improvement to successful execution rate. Anyway the PR will be merged as it brings some execution improvements (speed).

PR #3403, opened 2023-10-11
feat(gstd, backend): Introduce gr_exec_settings sys-call

DennisInSky:
Introduce `gr_exec_settings` sys-call providing user space with various system settings. The settings are supposed to be extandable via versioning so old clients can still obtain the settings w/o any issues when backend introduces new versions


core/src/exec_settings.rs
```diff
@@ -0,0 +1,53 @@
...
+#[derive(Debug, Clone, Copy)]
+pub struct ExecSettingsV1 {
+    /// Performance multiplier percentage
+    pub performance_multiplier_percent: u32,
+    /// Existential deposit
+    pub existential_deposit: u128,
+    /// Mailbox threshold
+    pub mailbox_threshold: u64,
+    /// Multiplier for converting gas into value
+    pub gas_to_value_multiplier: u128,
```
DennisInSky, line 52: @breathx would it be okay to make the `core` crate dependent on the `gsys` one, so at least I can use the `Value` and `Gas` type aliases here instead of uXX

pallets/gear/src/benchmarking/syscalls.rs
```diff
@@ -363,6 +363,28 @@ where
...
+
+        let module = ModuleDefinition {
+            memory: Some(ImportedMemory::new(SMALL_MEM_SIZE)),
+            imported_functions: vec![SysCallName::ExecSettings],
+            handle_body: Some(body::syscall(
+                repetitions,
+                &[
+                    // offset where to write settings
+                    InstrI32Const(settings_offset),
+                    // version. TODO: Should it be benched based on version?
```
DennisInSky, line 378: Should this be benched based on version? I.e. should we have different benchmarks for different version?

core-backend/src/mock.rs
```diff
@@ -115,15 +115,23 @@ impl Externalities for MockExt {
     fn free(&mut self, _page: WasmPage) -> Result<(), Self::AllocError> {
         Err(Error)
     }
+    fn exec_settings(&self, version: u32) -> Result<ExecSettings, Self::FallibleError> {
```
ark0f, line 118:
```suggestion
    fn exec_settings(&self, version: u32) -> Result<ExecSettings, Self::UnrecoverableError> {
```

gsys/src/lib.rs
```diff
@@ -331,6 +331,13 @@ impl TwoHashesWithValue {
 
 #[allow(improper_ctypes)]
 extern "C" {
+    /// Infallible `gr_exec_settings` get syscall.
+    ///
+    /// Arguments type:
+    /// - `settings`: `mut ptr` for buffer to store requested version of settings.
+    /// - `version`: `u32` defining version of settings to get.
+    pub fn gr_exec_settings(settings: *mut BufferStart, version: u32);
```
ark0f, line 339:
```suggestion
    pub fn gr_exec_settings(version: u32, settings: *mut BufferStart);
```

utils/wasm-instrument/src/syscalls.rs
```diff
@@ -285,6 +287,12 @@ impl SysCallName {
                 SysCallSignature::gr([Ptr(PtrInfo::new_mutable(PtrType::ErrorWithSignalCode))])
             }
             Self::MessageId => SysCallSignature::gr([Ptr(PtrInfo::new_mutable(PtrType::Hash))]),
+            Self::ExecSettings => SysCallSignature::gr([
+                // The PtrType::BufferStart doesn't work here as it requires length_param_idx
+                // Should we introduce something like PtrType::Buffer
+                Ptr(PtrInfo::new_mutable(PtrType::BlockNumber)),
+                Size,
```
ark0f, line 294:
```suggestion
                I32,
```

ark0f (changes requested):

gsys/src/lib.rs
DennisInSky: Any explanation for the requested change?

utils/wasm-instrument/src/syscalls.rs
DennisInSky: There is no such variant in the `ParamType` enum

NikVolf:
@ark0f re-review?

NikVolf (approved):

pallets/gear/src/benchmarking/syscalls.rs
NikVolf: I'd do different version, since this call will be used by pretty much every single program

core-backend/src/error.rs
```diff
@@ -127,6 +127,8 @@ pub enum UnrecoverableExecutionError {
     TooBigReadLen,
     #[display(fmt = "Cannot take data in payload range from message with size")]
     ReadWrongRange,
+    #[display(fmt = "Unexpected version of execution settings encountered ")]
+    UnexpectedExecSettingsVersion,
```
ark0f, line 131: Unused error

gsys/src/lib.rs
ark0f: Consistency with other sys-calls where input arguments are first, output ones are second

gcore/src/exec.rs
```diff
@@ -26,9 +26,35 @@ use crate::{
...
+    HashWithValue, Value,
 };
 
+/// Current version of execution settings.
+///
+/// Backend maintains backward compatibility with previous versions of execution
+/// settings. This structure matches to the most recent version of execution
+/// settings supported by backend.
+#[repr(C, packed)]
+#[derive(Debug, Default, Clone, Copy)]
```
ark0f, line 39:
```suggestion
#[derive(Debug, Clone, Copy)]
```

gcore/src/exec.rs
```diff
@@ -26,9 +26,35 @@ use crate::{
...
+    pub existential_deposit: Value,
+    /// Current value of mailbox threshold.
+    pub mailbox_threshold: Gas,
+    /// Current value of value per gas multiplier.
+    pub gas_to_value_multiplier: Value,
+}
+
+/// Get current version of execution settings.
+pub fn settings() -> Settings {
+    let mut settings = Settings::default();
```
ark0f, line 53:
```suggestion
    let mut settings = MaybeUninit::new();
```

utils/wasm-instrument/src/syscalls.rs
ark0f: @SabaunT, I guess we need to introduce new parameter type?

core/src/exec_settings.rs
ark0f: `core` contain too many dependencies when `gsys` is intended to be for definitions only

gcore/src/exec.rs
```diff
@@ -26,9 +26,35 @@ use crate::{
...
 };
 
+/// Current version of execution settings.
+///
+/// Backend maintains backward compatibility with previous versions of execution
+/// settings. This structure matches to the most recent version of execution
+/// settings supported by backend.
+#[repr(C, packed)]
+#[derive(Debug, Default, Clone, Copy)]
+pub struct Settings {
```
ark0f, line 40: I'd move structure into `gsys` and use it in `gear-core` and `gcore`

ark0f (changes requested):

core-backend/src/error.rs
DennisInSky: I think this is counted as usage, isn't it?
https://github.com/gear-tech/gear/blob/3cdd83aedafed5473d15d9b5c4b1a521dedff95d/core-processor/src/ext.rs#L736

gcore/src/exec.rs
DennisInSky: I haven't done it on purpose as we are going to have multiple versions of this struct on the backend side whereas there should be a single one on the userspace side. I don't think we need any confusing on the latter.

core/src/exec_settings.rs
ark0f: Oh, I'm sorry. I confused what depends on what. Seems like a good idea to me

gcore/src/exec.rs
DennisInSky: That's been done

utils/wasm-instrument/src/syscalls.rs
breathx: > @SabaunT, I guess we need to introduce new parameter type?

I'd add something like `Version`

breathx: It's never a problem to code something new xd

gsys/src/lib.rs
```diff
@@ -329,8 +329,33 @@ impl TwoHashesWithValue {
...
 }
 
+/// Current version of execution settings.
+///
+/// Backend maintains backward compatibility with previous versions of execution
+/// settings. This structure matches to the most recent version of execution
+/// settings supported by backend.
+#[repr(C, packed)]
+#[derive(Debug, Clone, Copy)]
+pub struct ExecSettings {
```
ark0f, line 339: I would make name like this to clearly indicate version
```suggestion
pub struct ExecSettingsV1 {
```

utils/wasm-instrument/src/syscalls.rs
```diff
@@ -293,15 +295,13 @@ impl SysCallName {
...
+                SysCallSignature::gr([Version, Ptr(PtrInfo::new_mutable(PtrType::Buffer))])
+            }
             Self::Read => SysCallSignature::gr([
                 MessagePosition,
                 Size,
-                // TODO #3375, the PtrType::BlockNumber is incorrect here.
-                // Should be:
-                // Ptr(PtrInfo::new_mutable(PtrType::BufferStart {
-                //     length_param_idx: 1,
-                // }))
```
ark0f, line 303: Looks like exactly `BufferStart` is required, not just `Buffer`

utils/wasm-instrument/src/syscalls.rs
```diff
@@ -534,6 +526,7 @@ pub enum ParamType {
     Handler,         // i32 handler number
     Alloc,           // i32 alloc pages
     Free,            // i32 free page
+    Version,         // i32 version number of exec settings
```
ark0f, line 609: You need to add config in `utils/wasm-gen/src/config/syscalls/param.rs`:
```rust
impl Default for SysCallsParamsConfig {
```

ark0f (approved):

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -204,6 +204,7 @@ fn default_processor_context<T: Config>() -> ProcessorContext {
         reservation: 0,
         random_data: ([0u8; 32].to_vec(), 0),
         rent_cost: 0,
+        gas_to_value_multiplier: Default::default(),
```
breathx, line 207: pls replace all related data to non-zero values because it's cheaper to write zeroes into memory so bench isn't that accurate

utils/wasm-instrument/src/syscalls.rs
```diff
@@ -293,15 +295,13 @@ impl SysCallName {
             Self::MessageId => SysCallSignature::gr([Ptr(PtrInfo::new_mutable(PtrType::Hash(
                 HashType::MessageId,
             )))]),
+            Self::ExecSettings => {
+                SysCallSignature::gr([Version, Ptr(PtrInfo::new_mutable(PtrType::Buffer))])
+            }
             Self::Read => SysCallSignature::gr([
                 MessagePosition,
                 Size,
-                // TODO #3375, the PtrType::BlockNumber is incorrect here.
```
breathx, line 299: where did todo disappeared?

breathx: as we discussed, I still think that it should be buffer start
if wasmgen buffer start requires len that it should be edited into BufferStart and SizedBufferStart

breathx: > as we discussed, I still think that it should be buffer start
if wasmgen buffer start requires len that it should be edited into BufferStart and SizedBufferStart

just make some renaming there

breathx (changes requested):
Requesting changes due to gas_to_value converter
to avoid using enums you can hack the system by assuming multiplier as u128 where last 64bits are u64, first - just defining byte :/

And one more thing about naming: does "settings" fit well? Maybe "params"? like env_params... to be considered

Otherwise LveryGTM

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -188,7 +187,7 @@ fn default_processor_context<T: Config>() -> ProcessorContext {
             ContextSettings::new(0, 0, 0, 0, 0, 0),
         ),
         block_info: Default::default(),
-        performance_multiplier: Percent::new(100),
+        performance_multiplier: gsys::Percent::new(100),
         max_pages: TESTS_MAX_PAGES_NUMBER.into(),
         page_costs: PageCosts::new_for_tests(),
         existential_deposit: 0,
```
breathx, line 193: non zero ed and mb_threshold as well please, same reason

breathx (approved):
Awesome.

So how do you feel about renaming: gr_exec_settings into gr_env_parameters, or even maybe gr_env_vars ? Cause it most likely variable env params rather than settings of some standalone execution.

utils/wasm-instrument/src/syscalls.rs
```diff
@@ -293,19 +295,18 @@ impl SysCallName {
...
-                // Should be:
-                // Ptr(PtrInfo::new_mutable(PtrType::BufferStart {
-                //     length_param_idx: 1,
-                // }))
-                Ptr(PtrInfo::new_mutable(PtrType::BlockNumber)),
+                // TODO #3375, in fact it has to be SizedBufferStart
+                Ptr(PtrInfo::new_mutable(PtrType::BufferStart)),
```
SabaunT, line 305: Fixed here https://github.com/gear-tech/gear/pull/3422/

Please remove `BufferStart`

SabaunT (commented):
I'm not sure that we need to generate `gr_exec_settings` from wasm-gen for loader and fuzzer. What about adjusting wasm-gen configs for those test tools accordingly?

PR #3405, opened 2023-10-12
feat(runtime): Adding dust and slash handlers

ekovalev:
- Dust and slashes handlers to pallets configs to keep the total issuance consistent
- Tests

### DIFF
--- a/runtime/vara/src/integration_tests.rs
+++ b/runtime/vara/src/integration_tests.rs
@@ -247,6 +247,23 @@ pub(crate) fn get_last_program_id() -> [u8; 32] {
     }
 }
 
+pub(crate) fn get_treasury_events() -> (Balance, Balance, Balance) {
+    System::events()
+        .into_iter()
+        .fold((0, 0, 0), |r, e| match e.event {
+            RuntimeEvent::Treasury(pallet_treasury::Event::Spending { budget_remaining }) => {
+                (budget_remaining, r.1, r.2)
+            }
+            RuntimeEvent::Treasury(pallet_treasury::Event::Burnt { burnt_funds }) => {
+                (r.0, burnt_funds, r.2)
+            }
+            RuntimeEvent::Treasury(pallet_treasury::Event::Rollover { rollover_balance }) => {
+                (r.0, r.1, rollover_balance)
+            }
+            _ => r,
+        })
+}
+
 #[test]
 fn tokens_locking_works() {
     init_logger();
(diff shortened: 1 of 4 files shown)


ukint-vs (approved):

runtime/vara/src/integration_tests.rs
```diff
@@ -345,10 +348,239 @@ fn tokens_locking_works() {
...
+            // Run chain for a day so that `Treasury::spend_funds()` is triggered
+            run_to_block(DAYS);
+
+            // Treasury had a surplus, but none of it was burned
+            assert_eq!(Treasury::pot(), 1_000 * UNITS);
+
+            // The total issuance persisted
+            assert_eq!(Balances::total_issuance(), initial_total_issuance);
+
+            // Run chain until another `Treasury::spend_funds()` invocation
```
breathx, line 419: do we have some event on spend_funds call to assert that it was triggered?

runtime/vara/src/lib.rs
```diff
@@ -603,7 +601,7 @@ impl pallet_staking::Config for Runtime {
     type ElectionProvider = ElectionProviderMultiPhase;
     type GenesisElectionProvider = onchain::OnChainExecution<OnChainSeqPhragmen>;
     // Burning the reward remainder for now.
-    // TODO: set remainder back to `RewardsStash<Self, Treasury>` to stop burning `Treasury` part.
+    // TODO: set remainder back to `RewardProxy<Self, Treasury>` to stop burning `Treasury` part.
```
breathx, line 604: what this todo is about and when should it be solved?

breathx (approved):

breathx:
Seems all holes for leak of total issuance solved so we are able to introduce migrations with burning token from offset pool to stabilise issuance?

runtime/vara/src/lib.rs
ekovalev: This essentially means that instead of burning tokens minted in excess during the current era (that is, those that have been minted but not supposed to be added to the validator's rewards in order to respect the `max_roi` threshold) we want to move this amount from the offset pool to the `Treasury`. This can only be happening if the current `staking_ratio` is very low (~10% of stakeable tokens and lower) - this will likely never be the case for Vara.

We can switch it back on any time, I guess, but from practical perspective there is no urgency anymore.

runtime/vara/src/integration_tests.rs
ekovalev: The `Treasury` pallet emits the events every `T::SpendPeriod` blocks (once a day in out runtime)

PR #3418, opened 2023-10-17
feat!(runtime): Add ability to align total supply

breathx:
This PR brings new extrinsic supposed to be called only once by governance voting on minting/burning some tokens from offset pool in order to align total issuance to 10kkk tokens, because network ran into imbalance here due to previous runtime bugs (see PRs related to dust removal and staking rewards)

### DIFF
--- a/pallets/staking-rewards/src/lib.rs
+++ b/pallets/staking-rewards/src/lib.rs
@@ -94,6 +94,7 @@ pub struct InflationInfo {
 #[frame_support::pallet]
 pub mod pallet {
     use super::*;
+    use core::cmp::Ordering;
     use frame_support::pallet_prelude::*;
     use frame_system::pallet_prelude::*;
 
@@ -237,6 +238,8 @@ pub mod pallet {
         Withdrawn { amount: BalanceOf<T> },
         /// Burned from the pool.
         Burned { amount: BalanceOf<T> },
+        /// Minted to the pool.
+        Minted { amount: BalanceOf<T> },
     }
 
     /// Error for the staking rewards pallet.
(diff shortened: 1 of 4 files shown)


ukint-vs (approved):

gshep (approved):
:100:

pallets/staking-rewards/src/lib.rs
```diff
@@ -326,6 +329,26 @@ pub mod pallet {
 
             Ok(())
         }
+
+        #[pallet::call_index(3)]
+        #[pallet::weight(<T as Config>::WeightInfo::align_supply())]
+        pub fn align_supply(origin: OriginFor<T>, target: BalanceOf<T>) -> DispatchResult {
+            ensure_root(origin)?;
```
ekovalev, line 336: In this pallet we have `RefillOrigin` (which is better be renamed as `DepositOrigin`) and `WithdrawOrigin` for respective operations. Maybe we can introduce another generalized origin one for this type of operation, too, to have more freedom in setting permissions here, as well?

pallets/staking-rewards/src/lib.rs
```diff
@@ -438,7 +461,7 @@ impl<T: Config> OnUnbalanced<NegativeImbalanceOf<T>> for OffsetPool<T> {
         // Should resolve into existing but resolving with creation is a safer bet anyway
         T::Currency::resolve_creating(&Pallet::<T>::account_id(), amount);
 
-        Pallet::deposit_event(Event::<T>::Deposited {
+        Pallet::deposit_event(Event::<T>::Minted {
```
ekovalev, line 464: If I understood the rationale behind having another event (`Minted`), the latter should indicate the imbalance has been created inside this pallet. But in this case the imbalance is being passed in from outside. In my opinion `Deposited` is a closer match.

ark0f (approved):

ekovalev (approved):

PR #3422, opened 2023-10-18
refactor(wasm-gen): add new approach of processing syscall parameters

### LINKED ISSUE #3375: Refactor `gear-wasm-gen` approach of processing syscall parameters
### File Location(s)

`utils/wasm-gen/src/generator/syscalls/invocator.rs::process_sys_call_params`
`utils/wasm-instrument/src/syscalls.rs`

### Proposal

Rewrite `process_sys_call_params` such a way that it'll not depend on the assumption that `Length` argument will follow corresponding pointer in all syscall signatures.

StackOverflowExcept1on:
Resolves #3375

Instead of skipping the parameter, we will first push the parameter with `value=0` and then mutate it by `length_param_idx`.

### DIFF
--- a/utils/wasm-gen/src/generator/syscalls/invocator.rs
+++ b/utils/wasm-gen/src/generator/syscalls/invocator.rs
@@ -459,7 +459,13 @@ impl<'a, 'b> SysCallsInvocator<'a, 'b> {
         let mem_size = Into::<WasmPageCount>::into(mem_size_pages).memory_size();
 
         let mut setters = Vec::with_capacity(params.len());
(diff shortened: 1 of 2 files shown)


breathx (approved):

SabaunT (commented):
There are some implicit properties that I would move to type level as the code is under development:
- syscalls params slice, processed syscalls slice and setters vector has the same size
- pointers to length and offset params are valid (point to the desired data)
- only post processed params are used to build a call

SabaunT (approved):

breathx (approved):

PR #3423, opened 2023-10-19
feat(gsdk): Expose `GearConfig` and `PairSigner`

ukint-vs:
If we want to use subxt in external crates with gsdk we need access to subxt Signer and Config

@gear-tech/dev

### DIFF
--- a/gsdk/src/lib.rs
+++ b/gsdk/src/lib.rs
@@ -17,10 +17,6 @@
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
 //! Gear api
-use crate::metadata::runtime_types::gear_common::{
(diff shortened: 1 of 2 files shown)


gsdk/src/lib.rs
```diff
@@ -17,20 +17,18 @@
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
 //! Gear api
+use crate::metadata::runtime_types::gear_common::{
```
breathx, line 20: please don't mix up public and private exports

breathx (approved):

gsdk/src/lib.rs
ukint-vs: сargo fmt

Issue #3442, opened 2023-10-23
Make usage of a message payload related syscalls more complex


techraed:
### Proposal

Currently syscalls like `gr_read` or any of `gr_*input*` doesn't do crucial for runtime fuzzer checks. I suggest introducing some "precise syscalls" type of functions that will use such syscalls in a way that really tests runtime for panics. Otherwise we are just consuming fuzzer resources ineffectively.

PR #3443, opened 2023-10-23
refactor(runtime-fuzzer): Decrease payload length in `runtime-fuzzer`

SabaunT:
This decreases corpus length as the simplicity of wasms invoking syscalls related to the message payload makes it inefficient to generate large message payloads. For more info read #3442


StackOverflowExcept1on (approved):

utils/runtime-fuzzer/src/gear_calls.rs
```diff
@@ -254,8 +256,8 @@ impl UploadProgramGenerator {
     }
 
     const fn unstructured_size_hint(&self) -> usize {
-        // Max code size - 50 KiB.
-        const MAX_CODE_SIZE: usize = 50 * 1024;
+        // Max code size - 25 KiB.
```
mertwole, line 259: Why have you changed this?

mertwole (approved):

breathx (approved):

utils/runtime-fuzzer/src/gear_calls.rs
SabaunT: As I got from logs, fuzzer doesn't create wasm (in current configuration) with size larger than 10 kb. So 25 kb is absolutely enough.

PR #3459, opened 2023-10-30
fix(wasm-builder): allow building with stable toolchain

StackOverflowExcept1on:
This PR fixes build with `+stable`

Tests:
```
cd utils/wasm-builder/test-program
cargo +stable build --release
```

@gear-tech/dev


clearloop:
plz provide a test of stable build with test-program btw

https://github.com/gear-tech/gear/tree/master/utils/wasm-builder/test-program

mb override the rust-toolchain of this crate or using environment to do the test

StackOverflowExcept1on:
We have smoke tests in https://github.com/gear-tech/gear/blob/master/utils/wasm-builder/tests/smoke.rs but it's marked with ignored to do not waste time on `make pre-commit`. It's tested every Monday here: https://github.com/gear-tech/gear/actions/workflows/time-consuming-tests.yml

I think we can test this program with both toolchains in `time-consuming-tests` environment.

For the `check` action, we can also add `cargo +stable build`

https://github.com/gear-tech/gear/blob/a34fcfc353c522669016f8d4433a56433e327fc3/.github/workflows/check.yml#L61-L62

clearloop (approved):

NikVolf:
> We have smoke tests in https://github.com/gear-tech/gear/blob/master/utils/wasm-builder/tests/smoke.rs but it's marked with ignored to do not waste time on `make pre-commit`. It's tested every Monday here: https://github.com/gear-tech/gear/actions/workflows/time-consuming-tests.yml
> 
> I think we can test this program with both toolchains in `time-consuming-tests` environment.
> 
> For the `check` action, we can also add `cargo +stable build`
> 
> https://github.com/gear-tech/gear/blob/a34fcfc353c522669016f8d4433a56433e327fc3/.github/workflows/check.yml#L61-L62

yeah, add this pls

utils/wasm-builder/src/cargo_command.rs
```diff
@@ -135,10 +135,37 @@ impl CargoCommand {
...
+            "TARGET",
+            "HOST",
+            "NUM_JOBS",
+            "OPT_LEVEL",
+            "PROFILE",
+            "RUSTC",
+            "RUSTDOC",
+            "RUSTC_WRAPPER",
+            "RUSTC_WORKSPACE_WRAPPER",
+            "RUSTC_LINKER",
```
clearloop, line 157: mb reconsider:

- `TARGET`,
- `OUT_DIR`
- `RUSTC_WRAPPER`

for cache managing,

- `OPT_LEVEL`,
- `PROFILE`

for syncing the original package?

clearloop:
> time

wait, I believe this test should be

```
cargo +stable check -p gstd --target wasm32-unknown-unkown
```

the current

```
cargo +stable check -p gstd 
```

missed the wasm target

StackOverflowExcept1on:
@clearloop I think we shouldn't use `--target` because the `wasm-builder` will switch to nightly toolchain.

utils/wasm-builder/src/cargo_command.rs
StackOverflowExcept1on: @clearloop `TARGET` will be passed using `.arg("--target=wasm32-unknown-unknown")`, it's not host target

This will be ignored by cargo and will not change behavior ([Environment variables Cargo reads](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-reads)):
- `OUT_DIR`
- `OPT_LEVEL`
- `PROFILE`

`RUSTC_WRAPPER` maybe make sence

clearloop:
> @clearloop I think we shouldn't use `--target` because the `wasm-builder` will switch to nightly toolchain.

there's no `wasm-builder` involved just checking `gstd`, as result, whatever we run

```
cargo +stable check -p gstd
```

or 

```
cargo +stable build -p gsdk
```

we can not reproduce this problem on master, 

```
cargo +stable check -p gstd --target wasm32-unknown-unkonwn
```

will reproduce it, bcz the problem is actually occurred while compiling `gsdk` to target wasm32-unknown-unknown on stable channel

utils/wasm-builder/src/cargo_command.rs
StackOverflowExcept1on: done in 89fc8998b77929c2f3431bca18031cfeedc4817e

breathx (approved):

breathx:
@ark0f please take a look

utils/wasm-builder/src/cargo_toolchain.rs
```diff
@@ -35,6 +35,9 @@ static TOOLCHAIN_CHANNELS: &[&str] = &[
 pub(crate) struct Toolchain(String);
 
 impl Toolchain {
+    /// This is a hardcoded version of nightly toolchain, tested on our CI.
+    const PINNED_NIGHTLY_TOOLCHAIN: &'static str = "nightly-2023-09-18";
```
NikVolf, line 39: can you take it from `rust-toolchain.toml` ?

StackOverflowExcept1on: I doubt it, especially if we publish wasm builder to crates.io

ukint-vs (approved):

utils/wasm-builder/src/cargo_toolchain.rs
breathx: and add comment in out `rust-toolchain.toml` that if changed need to change in wasm builder as well

StackOverflowExcept1on: > what about try to find file in the root and if not found go with const? or its too unintuitive?

I was thinking about generating code in `src/` using `build.rs`. We can try this if it doesn't lead to https://github.com/rust-lang/cargo/issues/3076

StackOverflowExcept1on: done in last 2 commit

breathx (approved):

utils/wasm-builder/build.rs
```diff
@@ -1,4 +1,25 @@
...
+        .map(|workspace_dir| workspace_dir.join("rust-toolchain.toml"))
+        .filter(|path| path.exists())
+    {
+        let content = fs::read_to_string(rust_toolchain).unwrap();
+        let line = content
+            .lines()
+            .find(|line| line.starts_with("channel"))
+            .unwrap();
+        let (start, end) = line.find('"').map(|i| i + 1).zip(line.rfind('"')).unwrap();
+        let toolchain = &line[start..end];
```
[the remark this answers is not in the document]
StackOverflowExcept1on: does it make sense to add `toml` and `serde` to `dev-dependencies` to read toolchain? we assume that `rust-toolchain.toml` is always valid

breathx (changes requested):
Decided to implement different way

StackOverflowExcept1on:
Will be continued after #3527

breathx (changes requested):
@StackOverflowExcept1on @ark0f @breathx 
Relates https://github.com/gear-tech/gear/pull/3527#pullrequestreview-1816091458

```toml
# Wasm builder changes
# Use the same toolchain: `gear_wasm_builder::build::<Metadata>();`
# Use recommended nightly: `gear_wasm_builder::recommended_nightly::<Metadata>();` // ERROR IF VERSION IS DIFFERENT
```

utils/wasm-builder/src/cargo_command.rs
```diff
@@ -45,6 +46,7 @@ impl CargoCommand {
             target_dir: "target".into(),
             features: vec![],
             toolchain: Toolchain::nightly(),
```
breathx, line 48: we still compile nightly

utils/wasm-builder/src/cargo_toolchain.rs
```diff
@@ -35,6 +35,9 @@ static TOOLCHAIN_CHANNELS: &[&str] = &[
 pub(crate) struct Toolchain(String);
 
 impl Toolchain {
```
breathx, line 37: I'd remove all of these redundant methods for `Toolchain`

breathx: like `nightly` etc

.github/workflows/check.yml
```diff
@@ -56,10 +56,18 @@ jobs:
           fi
 
       - name: "Install: Rust stable toolchain"
-        uses: dtolnay/rust-toolchain@stable
+        run: rustup toolchain install stable --component llvm-tools-preview --target wasm32-unknown-unknown
```
breathx, line 59: https://github.com/dtolnay/rust-toolchain#inputs

.github/workflows/check.yml
```diff
@@ -56,10 +56,18 @@ jobs:
...
 
       - name: "Check: Compiling gstd on stable"
-        run: cargo +stable check -p gstd
+        run: |
+          TOOLCHAIN=$(grep 'channel' rust-toolchain.toml | cut -d '"' -f 2)
+          CARGO_TOOLCHAIN="utils/wasm-builder/src/cargo_toolchain.rs"
```
breathx, line 64: split into different check step pls

.github/workflows/check.yml
```diff
@@ -56,10 +56,18 @@ jobs:
...
+          if ! grep -q "$TOOLCHAIN" $CARGO_TOOLCHAIN; then
+            echo "Please update PINNED_NIGHTLY_TOOLCHAIN constant in $CARGO_TOOLCHAIN to match rust-toolchain.toml."
+            exit 1
+          fi
+          cargo +stable check -p gstd -F debug --target wasm32-unknown-unknown
```
breathx, line 69: remove debug or extend with all stable features set

breathx (changes requested):

utils/wasm-builder/src/cargo_command.rs
StackOverflowExcept1on: Well I can rename it stable but later wasm-builder just uses `Toolchain::try_from_rustup`

.github/workflows/check.yml
StackOverflowExcept1on: Hmm, I'm not sure if I should use actions for this because they override the toolchain

utils/wasm-builder/src/lib.rs
```diff
@@ -82,6 +82,20 @@ impl WasmBuilder {
...
+    /// Add check of recommended toolchain.
+    pub fn with_recommended_toolchain(mut self) -> Self {
+        self.cargo.set_check_recommended_toolchain(true);
+        self
+    }
+
+    /// Force the recommended toolchain to be used, but do not check whether the
+    /// current toolchain is recommended.
+    #[doc(hidden)]
+    pub unsafe fn with_forced_recommended_toolchain(mut self) -> Self {
```
ark0f, line 94: I don't see actual usage. Can be removed because check throws error and thus forces to use recommended nightly

ark0f (changes requested):

utils/wasm-builder/src/lib.rs
breathx: @StackOverflowExcept1on isn't it a time to already use the method for `out-of-memory` demo, for example?

breathx (approved):

breathx (approved):

examples/out-of-memory/build.rs
```diff
@@ -16,6 +16,11 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+use gear_wasm_builder::WasmBuilder;
+
 fn main() {
-    gear_wasm_builder::build();
+    WasmBuilder::new()
+        .exclude_features(vec!["std"])
+        .with_forced_recommended_toolchain()
```
ark0f, line 24: I'd add comment about internal usage because this demo is in `examples`

ark0f (approved):

examples/out-of-memory/build.rs
StackOverflowExcept1on: beb4d72dacad8c348c42d5ea1ea7a1a1540efea9

PR #3462, opened 2023-10-30
fix(windows): Precise stack height limit

ark0f:
Also: 
- Respect `CARGO_TARGET_DIR` in `gear-wasm-builder`


core/src/code.rs
```diff
@@ -398,10 +418,20 @@ impl Code {
...
                 log::trace!("Failed to inject stack height limits: {err}");
                 CodeError::StackLimitInjection
             })?;
+
+            if let Some(globals_before) = globals {
+                // ensure stack limiter injector has created global
+                let globals_after = module.globals_space();
+                assert_eq!(globals_after, globals_before + 1);
+
+                module = export_stack_height(module);
```
StackOverflowExcept1on, line 433: maybe we can move this code to `gwasm-instrument` crate?

core/src/code.rs
```diff
@@ -515,6 +545,7 @@ impl Code {
             TryNewCodeConfig {
                 version,
                 stack_height,
+                export_stack_height: true,
```
grishasobol, line 548: What the reason to set it as true here. Also cannot see reasons to make logic of this export creating in core. You can use wasm-instruments in your crate `calc-stack-height` in order to insert this export after creating new code:
```rust
let code = Code::try_new_mock_with_rules(
        inf_recursion.as_ref().to_vec(),
        |module| schedule.rules(module),
        TryNewCodeConfig {
            version: schedule.instruction_weights.version,
            stack_height: Some(u32::MAX),
            export_stack_height: true,
            ..Default::default()
        },
    )
    .map_err(|e| anyhow::anyhow!("{e}"))?;
    *code.code_mut() = insert_stack_height_export(code.code())
```

grishasobol (changes requested):

core/src/code.rs
ark0f: > What the reason to set it as true here? 

Artifact

> Also cannot see reasons to make logic of this export creating in gear-core

We cannot guarantee global index is persistent

grishasobol: > We cannot guarantee global index is persistent

ok

core/src/code.rs
ark0f: I'd leave it here to avoid many patches in our fork

grishasobol (approved):

node/cli/src/command.rs
```diff
@@ -149,6 +149,8 @@ pub fn run() -> sc_cli::Result<()> {
 
     let base = &mut cli.run.base;
 
+    let is_dev = base.is_dev().unwrap_or(false);
```
breathx, line 152: It's already below on line 163 `base.shared_params.dev`
```suggestion
```

pallets/gear/src/benchmarking/tests/mod.rs
```diff
@@ -70,18 +60,24 @@ where
...
+                JournalNote::SendDispatch { dispatch, .. } => Some(dispatch),
+                _ => None,
+            })
+            .unwrap();
+
+    let code = dispatch
+        .reply_details()
+        .expect("reply details")
+        .to_reply_code();
+    assert_eq!(
```
breathx, line 77:
```suggestion
        .to_reply_code();

    assert_eq!(
```

pallets/gear/src/schedule.rs
```diff
@@ -732,16 +732,13 @@ impl<T: Config> Default for Schedule<T> {
...
-            // because of difference of how stack height is calculated in injection and
-            // how wasmer and wasmi actually uses stack.
+            // Constant for `stack_height` is calculated to be small enough to avoid stack overflow in
+            // wasmer and wasmi executors.
             // To avoid potential stack overflow problems we have a panic in sandbox in case,
             // execution is ended with stack overflow error. So, process queue execution will be
             // stopped and we will be able to investigate the problem and decrease this constant if needed.
             // TODO #3435. Disabled stack height is a temp solution.
-            stack_height: cfg!(not(feature = "fuzz")).then_some(20_000),
+            stack_height: cfg!(not(feature = "fuzz")).then_some(18_369),
```
breathx, line 741: separate as const maybe for quick find?

breathx: Or even make it configurable param to be able rule that with governance?

utils/calc-stack-height/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "calc-stack-height"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+homepage.workspace = true
+repository.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
breathx, line 10:
```suggestion
```

utils/calc-stack-height/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+
+[dependencies]
+gear-core.workspace = true
+vara-runtime = { workspace = true, features = ["std"] }
```
breathx, line 14:
```suggestion
vara-runtime = { workspace = true, features = ["std", "dev"] }
```

utils/wasm-builder/src/wasm_project.rs
```diff
@@ -91,9 +91,12 @@ impl WasmProject {
             .to_string_lossy()
             .into();
 
+        let target_dir_name = env::var("CARGO_TARGET_DIR")
```
breathx, line 94: I'd split it into different pr but ok

breathx: @StackOverflowExcept1on ptal

utils/wasm-instrument/src/lib.rs
```diff
@@ -99,7 +99,10 @@ pub fn inject<R: Rules>(
     let out_of_gas_index = import_count as u32 - 1;
 
     let gas_charge_index = module.functions_space();
-    let gas_index = module.globals_space() as u32;
+    let gas_index = module
+        .global_section()
+        .map(|section| section.entries().len())
```
breathx, line 104: Seems its not necessary should be the last for future, isn't it?

core/src/code.rs
```diff
@@ -398,10 +418,20 @@ impl Code {
         }
 
         if let Some(stack_limit) = config.stack_height {
+            let globals = config.export_stack_height.then(|| module.globals_space());
```
breathx, line 421: we must take care that this is never used in prod because of the fact that such change requires re-intrumentation (!)

breathx (changes requested):

utils/wasm-builder/src/wasm_project.rs
StackOverflowExcept1on: lgtm, this code is trying to not hardcode the target directory

pallets/gear/src/benchmarking/tests/mod.rs
ark0f: Why?

utils/calc-stack-height/Cargo.toml
ark0f: Why do we remove these lines?

pallets/gear/src/schedule.rs
ark0f: > Or even make it configurable param to be able rule that with governance?

I'd leave const because stack height is expected to be changed only during executor update

utils/calc-stack-height/Cargo.toml
breathx: Why no? It's template-informational comments

breathx (approved):

StackOverflowExcept1on:
what about the code above?
https://github.com/gear-tech/gear/blob/ed9a5b226c85a68a238f792cea92f98a448a705a/utils/wasm-builder/src/wasm_project.rs#L80-L84

shouldn't we use `target_dir_name` instead of `"target"`?

```
# ok
CARGO_TARGET_DIR=target-custom cargo build --package demo-custom --release

# fails
cargo build --package demo-custom --release --target-dir target-custom

error: failed to run custom build command for `demo-custom v0.1.0 (/home/alex/work/gear/examples/custom)`

Caused by:
  process didn't exit successfully: `/home/alex/work/gear/target-custom/release/build/demo-custom-00c0ba042ca70753/build-script-build` (exit status: 1)
  --- stderr
  error: Failed to create WASM target directory: /wasm32-unknown-unknown/release
  |      Permission denied (os error 13)
```

utils/wasm-builder/src/wasm_project.rs
```diff
@@ -91,9 +91,12 @@ impl WasmProject {
...
+            .map(PathBuf::from)
+            .unwrap_or_else(|_err| "target".into());
         let mut target_dir = out_dir.clone();
         while target_dir.pop() {
-            if target_dir.ends_with("target") {
+            if target_dir.ends_with(&target_dir_name) {
                 break;
             }
         }
```
StackOverflowExcept1on, line 102: should also work with `--target-dir`: https://github.com/rust-lang/cargo/issues/9661#issuecomment-1722358176
```suggestion
        let mut target_dir = out_dir
            .ancestors()
            .find(|path| path.ends_with(&profile))
            .and_then(|path| path.parent().map(|p| p.to_owned()))
            .expect("Could not find target directory");
```

StackOverflowExcept1on (approved):

utils/wasm-builder/src/wasm_project.rs
ark0f: Nice find

PR #3468, opened 2023-11-01
refactor(scripts): Clean up shell scripts

### LINKED ISSUE #781: Use `cargo-make` instead of scripts
### Problem to Solve

Bash is harder to be supported as cross-platform script and may have some unexpected troubles as language.
Also, `cargo-make` will introduce "native" cargo integration

### Possible Solution

Introduction of `cargo-make`

### Notes

Scripts must be left yet.

`gear.sh` must be left as wrapper

thecaralice:
Resolves #781.

## Things to consider

- [ ] Do we actually need `cargo-make`?
  Please discuss in #781. 
- [ ] Refactor shell scripts as they have a lot of problems right now (according to [shellcheck](https://shellcheck.net/))
  - [ ] Maybe add shellcheck to CI/CD config


shamilsan:
Stale PR? Need to decide what to do.
@thecaralice

PR #3470, opened 2023-11-01
fix(backport)!: Move nightly features behind feature gates

thecaralice:
- `default` now compiles on stable
- `alloc_error_handler` is behind `oom-handler`
- `panic_info_message` is behind `panic-messages`
- Both can be activated with `nightly`

## TODO
- [ ] Cover stable toolchain compiling with tests

### DIFF
--- a/gstd/Cargo.toml
+++ b/gstd/Cargo.toml
@@ -24,9 +24,7 @@ static_assertions.workspace = true
 [features]
 #! ## Default features
 
-default = ["default-stable", "nightly"]
-## Stable subset of default features.
-default-stable = ["panic-handler"]
+default = ["panic-handler"]
 ## When enabled, a panic handler is provided by this crate.
 panic-handler = []
 
--- a/gstd/Cargo.toml
+++ b/gstd/Cargo.toml
@@ -6,6 +6,7 @@ edition.workspace = true
 license.workspace = true
 
 [dependencies]
+document-features = { version = "0.2.7", optional = true }
 galloc.workspace = true
 gcore = { workspace = true, features = ["codec"] }
 gstd-codegen = { path = "codegen" }
@@ -13,7 +14,7 @@ gear-core-errors.workspace = true
 hashbrown.workspace = true
 bs58 = { workspace = true, features = ["alloc"] }
 hex = { workspace = true, features = ["alloc"] }
-parity-scale-codec = { workspace = true, features = [ "derive" ] }
+parity-scale-codec = { workspace = true, features = ["derive"] }
 primitive-types = { workspace = true, features = ["scale-info"] }
 scale-info = { workspace = true, features = ["derive"] }
 futures = { workspace = true, features = ["alloc"] }
@@ -21,10 +22,34 @@ futures = { workspace = true, features = ["alloc"] }
 static_assertions.workspace = true
 
 [features]
+#! ## Default features
+
 default = ["default-stable", "nightly"]
+## Stable subset of default features.
 default-stable = ["panic-handler"]
-nightly = ["panic-messages", "oom-handler"]
+## When enabled, a panic handler is provided by this crate.
 panic-handler = []
-debug = ["galloc/debug", "gcore/debug"]
+
+#! ## Nightly features
+
+## Enables all features below.
+## These features depend on unstable Rust API and require nightly toolchain.
+nightly = ["panic-messages", "oom-handler"]
+## When enabled, additional context information is available from
+## panic messages in debug mode. Relies on [`panic_info_message`][rust-66745].
 panic-messages = []
+## When enabled, an OOM error handler is provided.
+## Relies on [`alloc_error_handler`][rust-51540],
 oom-handler = []
+
+#! ## Additional features
+
+## Enables debug logging; this heavily impacts gas cost 
+## and is therefore disabled by default.
+debug = ["galloc/debug", "gcore/debug"]
+
+#! [rust-66745]: https://github.com/rust-lang/rust/issues/66745
+#! [rust-51540]: https://github.com/rust-lang/rust/issues/51540
+
+[package.metadata.docs.rs]
+all-features = true
--- a/gstd/Cargo.toml
+++ b/gstd/Cargo.toml
@@ -21,6 +21,10 @@ futures = { workspace = true, features = ["alloc"] }
 static_assertions.workspace = true
 
 [features]
-default = ["panic-handler"]
+default = ["default-stable", "nightly"]
+default-stable = ["panic-handler"]
+nightly = ["panic-messages", "oom-handler"]
 panic-handler = []
 debug = ["galloc/debug", "gcore/debug"]
+panic-messages = []
+oom-handler = []
--- a/gstd/src/common/handlers.rs
+++ b/gstd/src/common/handlers.rs
@@ -25,26 +25,29 @@
 //! debug and non-debug mode, for programs built in `wasm32` architecture.
 //! For `debug` mode it provides more extensive logging.
 
+#[cfg(target_arch = "wasm32")]
+use crate::ext;
+#[cfg(target_arch = "wasm32")]
+use core::panic::PanicInfo;
+
 #[cfg(feature = "oom-handler")]
 #[cfg(target_arch = "wasm32")]
 #[alloc_error_handler]
 pub fn oom(_: core::alloc::Layout) -> ! {
-    crate::ext::oom_panic()
+    ext::oom_panic()
 }
 
 #[cfg(not(all(feature = "panic-message", feature = "debug")))]
 #[cfg(target_arch = "wasm32")]
 #[panic_handler]
 pub fn panic(_: &PanicInfo) -> ! {
-    crate::ext::panic("no info")
+    ext::panic("no info")
 }
 
 #[cfg(all(feature = "panic-message", feature = "debug"))]
 #[cfg(target_arch = "wasm32")]
 #[panic_handler]
-pub fn panic(panic_info: &core::panic::PanicInfo) -> ! {
-    use crate::{ext, prelude::format};
-
+pub fn panic(panic_info: &PanicInfo) -> ! {
     let msg = match (panic_info.message(), panic_info.location()) {
         (Some(msg), Some(loc)) => format!(
             "'{:?}', {}:{}:{}",
--- a/gstd/src/common/handlers.rs
+++ b/gstd/src/common/handlers.rs
@@ -32,15 +32,14 @@ pub fn oom(_: core::alloc::Layout) -> ! {
     crate::ext::oom_panic()
 }
 
-#[cfg(not(any(feature = "debug", debug_assertions)))]
+#[cfg(not(all(feature = "panic-message", feature = "debug")))]
 #[cfg(target_arch = "wasm32")]
 #[panic_handler]
 pub fn panic(_: &PanicInfo) -> ! {
     crate::ext::panic("no info")
 }
 
-#[cfg(feature = "panic-message")]
-#[cfg(any(feature = "debug", debug_assertions))]
+#[cfg(all(feature = "panic-message", feature = "debug"))]
 #[cfg(target_arch = "wasm32")]
 #[panic_handler]
 pub fn panic(panic_info: &core::panic::PanicInfo) -> ! {
--- a/gstd/src/common/handlers.rs
+++ b/gstd/src/common/handlers.rs
@@ -24,10 +24,9 @@
 //! Panic handlers are available in two implementations -
 //! debug and non-debug mode, for programs built in `wasm32` architecture.
 //! For `debug` mode it provides more extensive logging.
-#![cfg(target_arch = "wasm32")]
 
+#[cfg(feature = "oom-handler")]
 use crate::ext;
-use core::panic::PanicInfo;
 
 #[cfg(feature = "oom-handler")]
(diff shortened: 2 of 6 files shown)


gstd/src/common/handlers.rs
```diff
@@ -25,27 +25,26 @@
...
 #[cfg(not(any(feature = "debug", debug_assertions)))]
 #[cfg(target_arch = "wasm32")]
 #[panic_handler]
 pub fn panic(_: &PanicInfo) -> ! {
-    ext::panic("no info")
+    crate::ext::panic("no info")
 }
 
+#[cfg(feature = "panic-message")]
 #[cfg(any(feature = "debug", debug_assertions))]
```
breathx, line 43: It was a long time I was thinking of removal here cfg(debug_assertions) and the time has come, isn't it?

gstd/Cargo.toml
```diff
@@ -21,6 +21,10 @@ futures = { workspace = true, features = ["alloc"] }
 static_assertions.workspace = true
 
 [features]
-default = ["panic-handler"]
+default = ["default-stable", "nightly"]
+default-stable = ["panic-handler"]
```
breathx, line 25: I don't think we need such precise customization so to make it as easytouse as it could be, I'd remove default-stable and separate features of oom handling and panic info in order to keep only:
- panic-handler (includes panic handler)
- nightly (includes panic info and oom handler in case of panic-handler used)

And add docs for each feature of gstd pls

gstd/src/common/handlers.rs
```diff
@@ -25,27 +25,26 @@
...
-pub fn oom(_: Layout) -> ! {
-    ext::oom_panic()
+pub fn oom(_: core::alloc::Layout) -> ! {
+    crate::ext::oom_panic()
 }
 
 #[cfg(not(any(feature = "debug", debug_assertions)))]
 #[cfg(target_arch = "wasm32")]
 #[panic_handler]
 pub fn panic(_: &PanicInfo) -> ! {
```
breathx, line 38: Path to panic info?

breathx (changes requested):
Btw, we need somehow deal with running clippy in stable bcs by default we run them with --all-features and --all-targets

gstd/src/common/handlers.rs
thecaralice: Done in 24351be

gstd/src/common/handlers.rs
thecaralice: Reverted in cf7b030

thecaralice:
> Btw, we need somehow deal with running clippy in stable bcs by default we run them with --all-features and --all-targets

I believe that'll need a separate PR.

gstd/Cargo.toml
thecaralice: Docs added by 7048dd8.

gstd/src/common/handlers.rs
```diff
@@ -24,29 +24,26 @@
 //! Panic handlers are available in two implementations -
 //! debug and non-debug mode, for programs built in `wasm32` architecture.
 //! For `debug` mode it provides more extensive logging.
+#![cfg(target_arch = "wasm32")]
```
StackOverflowExcept1on, line 27: I'd move `target_arch="wasm32"` to `mod.rs`
https://github.com/gear-tech/gear/blob/3d0d23f85ec3528015e8ae2a0e534b0283ea6a36/gstd/src/common/mod.rs#L22-L23

thecaralice: Fixed in 13e8c07

gstd/Cargo.toml
```diff
@@ -21,6 +21,10 @@ futures = { workspace = true, features = ["alloc"] }
...
+nightly = ["panic-messages", "oom-handler"]
```
breathx, line 26: What about this?
```suggestion
default = ["stable", "nightly"]
stable = ["panic-handler"]
nightly = ["panic-messages", "oom-handler"]
```

gstd/src/common/handlers.rs
```diff
@@ -25,43 +25,47 @@
 //! debug and non-debug mode, for programs built in `wasm32` architecture.
 //! For `debug` mode it provides more extensive logging.
 
-#[cfg(target_arch = "wasm32")]
-use {crate::ext, core::alloc::Layout, core::panic::PanicInfo};
+#[cfg(feature = "oom-handler")]
+use crate::ext;
```
breathx, line 29: For the case you can move it inside function to make file more readable

gstd/src/common/handlers.rs
```diff
@@ -25,43 +25,47 @@
...
+                msg,
+                loc.file(),
+                loc.line(),
+                loc.column()
+            ),
+            (Some(msg), None) => format!("'{msg:?}'"),
+            (None, Some(loc)) => {
+                format!("{}:{}:{}", loc.file(), loc.line(), loc.column())
+            }
+            _ => ext::panic("no info"),
```
breathx, line 63:
```rust
pub const NO_INFO_PANIC: &str = "no info";
```

gstd/src/common/handlers.rs
```diff
@@ -25,43 +25,47 @@
 //! debug and non-debug mode, for programs built in `wasm32` architecture.
 //! For `debug` mode it provides more extensive logging.
 
-#[cfg(target_arch = "wasm32")]
-use {crate::ext, core::alloc::Layout, core::panic::PanicInfo};
+#[cfg(feature = "oom-handler")]
+use crate::ext;
 
-#[cfg(target_arch = "wasm32")]
+#[cfg(feature = "oom-handler")]
```
breathx, line 31: here and below cfg attr seems better

gstd/src/lib.rs
```diff
@@ -28,6 +28,25 @@
 //! asynchronous programming primitives, arbitrary types encoding/decoding,
 //! providing convenient instruments for creating programs from programs, etc.
 //!
+//! # Crate features
+//!
+//! - **debug** - When enabled, logging becomes more verbose; this heavily
```
breathx, line 33: Not the verbose, but in fact only then logging appears

gstd/src/lib.rs
```diff
@@ -28,6 +28,25 @@
 //! asynchronous programming primitives, arbitrary types encoding/decoding,
 //! providing convenient instruments for creating programs from programs, etc.
 //!
+//! # Crate features
```
breathx, line 31: For me it's a good practice to doc crate features be placed/duplicated in toml

breathx (approved):
sp-io compilation will be solved into separate pr-s, so this one is ready to be merged

breathx:
@ark0f please take a look

gstd/src/common/handlers.rs
ark0f: Why?

gstd/Cargo.toml
ark0f, line 25: I don't understand why this feature is needed

thecaralice: That would be a breaking change, bwcause users of this crate will no longer get OOM handler and verbose panic messages by default. Prior to this change, crate wouldn't compile on stable, so changing defaults for stable users isn't a braking change.

thecaralice: On the other hand, any codebase using nightly features is prone to breaking due to Rust upgrades, but that should be considered separately IMHO.

gstd/src/common/handlers.rs
thecaralice: OOS

gstd/src/common/handlers.rs
thecaralice: Guess it's better to remove the `use` altogether then

gstd/Cargo.toml
thecaralice: The current name highlights the fact that `default-stable` is actually a default set of features that compiles on stable; `stable` wouldn't be as apparent

gstd/src/common/handlers.rs
thecaralice: That would make `oom` exist without any features enabled; is that intentional?

gstd/src/lib.rs
thecaralice: Fixed in 0defe87.

gstd/src/lib.rs
thecaralice: Fixed in 5e08210.

gstd/src/common/handlers.rs
thecaralice: Fixed in 2489cd2.

gstd/Cargo.toml
breathx: > The current name highlights the fact that `default-stable` is actually a default set of features that compiles on stable; `stable` wouldn't be as apparent

why "nightly" is intuitive than?

gstd/src/common/handlers.rs
breathx: Why?

`#[cfg_attr(feature = "oom-handler", alloc_error_handler)]`

breathx (changes requested):
Please fix CI
For example, by passing the required nightly feature to all of the programs used in pallet-gear tests

gstd/Cargo.toml
ark0f: > so changing defaults for stable users isn't a breaking change.

> any codebase using nightly features is prone to breaking due to Rust upgrades

Agree

> but that should be considered separately IMHO.

I think this is implied by default like it is in nightly toolchain and crates with features like  `nightly` or `unstable`

gstd/src/common/handlers.rs
```diff
@@ -25,43 +25,45 @@
...
-        (None, Some(loc)) => {
-            format!("{}:{}:{}", loc.file(), loc.line(), loc.column())
-        }
-        _ => ext::panic("no info"),
-    };
+    #[cfg(all(feature = "panic-messages", feature = "debug"))]
+    #[panic_handler]
+    pub fn panic(panic_info: &PanicInfo) -> ! {
+        use crate::prelude::format;
+        let msg = match (panic_info.message(), panic_info.location()) {
```
ark0f, line 49:
```suggestion
        #[cfg(all(feature = "panic-messages", feature = "debug"))]
        let message = panic_info.message();
        #[cfg(not(all(feature = "panic-messages", feature = "debug")))]
        let message = None;

        let msg = match (message, panic_info.location()) {
```

gstd/src/common/handlers.rs
```diff
@@ -25,43 +25,45 @@
...
-        (Some(msg), None) => format!("'{msg:?}'"),
-        (None, Some(loc)) => {
-            format!("{}:{}:{}", loc.file(), loc.line(), loc.column())
-        }
-        _ => ext::panic("no info"),
-    };
+    #[cfg(all(feature = "panic-messages", feature = "debug"))]
+    #[panic_handler]
+    pub fn panic(panic_info: &PanicInfo) -> ! {
+        use crate::prelude::format;
```
ark0f, line 48:
```suggestion
        use alloc::format;
```

And move to other imports

ark0f (changes requested):

gstd/Cargo.toml
thecaralice: Well, I see two options:

- Break everything with this PR right now (breakage has a big chance to appear unnoticed until the user opens the logs and finds out they are almost empty; expect issues on "Why are my logs empty?"), immediately affects everyone
- Do not break anything (if the nightly features used become stabilized, compiler will emit a warning, but nothing more than that)

I do not think "nightly can break with updates sometimes" is enough for making actual breaking changes with no motivation behind.

gstd/src/common/handlers.rs
thecaralice: Pretty sure it was `crate::prelude::format` prior to this PR; why should that be changed?

> And move to other imports

By "other imports" you mean https://github.com/gear-tech/gear/blob/8dc668a38d52843aaea01b5a8158d4bff613998c/gstd/src/common/handlers.rs#L36-L37?
It is only used in this implementation, the other one does not need it

gstd/src/common/handlers.rs
thecaralice: Makes sense

thecaralice: Done in 5d9baa86f.

gstd/src/common/handlers.rs
thecaralice: Became irrelevant with 5d9baa86f

gstd/src/common/handlers.rs
thecaralice: Had to be partly undone in f12f67055 since the `debug` and non-`debug` implementations still need to be separated.

gstd/Cargo.toml
thecaralice: `nightly` is a feature that enables all nightly features, `default-nightly` would be "default features for nightly toolchain" - that's covered by `default`. In some sense, `nightly` is a "modifier" unlike `default-stable` which is a basic feature set.

gstd/src/common/handlers.rs
thecaralice: Then `oom` function will still exist, although not annotated with `#[alloc_error_handler]`; but there is no need in keeping it there when compiling without OOM handling enabled

ark0f (approved):

breathx (approved):

breathx:
Solve todo by adding test for stable compilation pls

PR #3475, opened 2023-11-07
fix(core): Allow waking the same message between waits

### LINKED ISSUE #2763: Allow waking the same message between waits
### Problem

Case of waking the same message in different executions should be allowed.

### Steps

See pallet-gear test referencing the issue

### Possible Solution

- Remove `awaken` field from `gear_core::message::context::ContextStore`
- Write migrations for `StoredDispatch` in `MessageQueue`, `DispatchStash` and `Waitlist` storages of `pallet-gear-messenger`
- Remove `#[should_panic]` attribute from test
// extra due to required migrations
- Remove `reply_sent` field from `gear_core::message::context::ContextStore` (Calling waits after replies is forbidden since auto-replies, so the field useless)

### Notes

_No response_

### Relevant Log Output

_No response_

mqxf:
Resolves #2763 

@reviewer-or-team


SabaunT (commented):
Up to now seems ok. Need to implement migrations.

core/src/message/context.rs
```diff
@@ -484,13 +483,9 @@ impl MessageContext {
 
     /// Wake message by it's message id.
     pub fn wake(&mut self, waker_id: MessageId, delay: u32) -> Result<(), Error> {
-        if self.store.awaken.insert(waker_id) {
-            self.outcome.awakening.push((waker_id, delay));
+        self.outcome.awakening.push((waker_id, delay));
```
breathx, line 486: duplicate waking should still exist, discussed in dm

core/src/message/context.rs
```diff
@@ -182,7 +182,6 @@ pub struct ContextStore {
     outgoing: BTreeMap<u32, Option<Payload>>,
     reply: Option<Payload>,
     initialized: BTreeSet<ProgramId>,
-    awaken: BTreeSet<MessageId>,
```
breathx, line 185: not the context store points to storing in db

breathx (changes requested):

breathx (changes requested):
1) You don't need to store awakening at all
2) You need runtime migrations

NikVolf:
Rebase this properly to master to remove all the commits above

pallets/gear-messenger/src/migrations.rs
```diff
@@ -0,0 +1,327 @@
...
+use sp_std::marker::PhantomData;
+#[cfg(feature = "try-runtime")]
+use {
+    frame_support::codec::{Decode, Encode},
+    sp_std::vec::Vec,
+};
+
+pub struct MigrateToV3<T: Config>(PhantomData<T>);
+
+impl<T: Config> MigrateToV3<T> {
```
breathx, line 38: It's not necessary to be impl of MigrateToV3. Btw just Into within the module seems ok

pallets/gear-messenger/src/migrations.rs
```diff
@@ -0,0 +1,327 @@
...
+                    next: store.next,
+                    value: StoredDispatch {
+                        kind: store.value.kind,
+                        message: store.value.message,
+                        context: store.value.context.map(Self::migrate_context_store),
+                    },
+                })
+            });
+
+            DispatchStash::<T>::translate(
```
breathx, line 89: btw you don't need to actually migrate dispatch stash because context there is always None. In this migrations you can change its data structure into sth like StoredDelayedDispatch { kind, msg }
In that case assert in tests that they're always empty and propagate here log::error in decode err case

pallets/gear-messenger/src/migrations.rs
```diff
@@ -0,0 +1,327 @@
...
+                    reply: None,
+                    initialized: Default::default(),
+                    awaken: Default::default(),
+                    reply_sent: false,
+                    reservation_nonce: Default::default(),
+                    system_reservation: None,
+                }),
+            };
+
+            v2::Waitlist::<Test>::insert(
```
breathx, line 289: generate different dispatches pls in cycle

breathx (changes requested):

core-processor/src/common.rs
```diff
@@ -138,6 +140,7 @@ impl DispatchResult {
             system_reservation_context,
             page_update: Default::default(),
             allocations: Default::default(),
+            reply_sent: false,
```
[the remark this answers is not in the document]
mqxf: We need it in one place after program execution is over. It has to be saved since this is the only struct which is passed onto the processing.rs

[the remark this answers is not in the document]
mqxf: The value is also always reset to false, after every execution. Just the way we talked about. I cant just not initialize it or remove it from the struct or that would break the processing.rs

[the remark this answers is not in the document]
breathx: It's not always reset to false. But this fn is used only for case of non-executing with success. Specify it in a comment or add `reply_sent: bool` argument to the fn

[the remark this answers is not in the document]
mqxf: This function is only used to create a dispatchresult when there is no code execution whatsoever, ie, the export we wanted to call was not found (so we send a message but handle method doesnt exist), therefore a reply can never be sent. I can add a comment if you feel that is necessary, but due to this usecase being the way it is, i dont think it is necessary.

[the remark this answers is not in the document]
breathx: Nobody will remember this small thing if any new use-case of the fn will appear some time soon, so small comment is a must-have

pallets/gear-messenger/src/migrations.rs
```diff
@@ -0,0 +1,347 @@
...
+    message::{ContextStore, StoredDispatch},
+};
+use sp_std::marker::PhantomData;
+#[cfg(feature = "try-runtime")]
+use {
+    frame_support::codec::{Decode, Encode},
+    sp_std::vec::Vec,
+};
+
+pub struct MigrateToV3<T: Config>(PhantomData<T>);
```
breathx, line 36: It's required to add this migrations to runtime otherwise it won't be executed! See runtime/vara/src/migrations.rs and add this to tuple.

breathx (changes requested):
Never resolve conversations if they're not solved in any manner
@SabaunT please review

mqxf:
Any conversations I resolve without comment means I believe that I have resolved them in the commit I pushed.

breathx:
Still waiting of @SabaunT

breathx:
@techraed please review

pallets/gear-messenger/src/migrations.rs
```diff
@@ -0,0 +1,363 @@
...
+    };
+    use common::storage::{Interval, LinkedNode};
+    use frame_support::{pallet_prelude::StorageVersion, traits::OnRuntimeUpgrade};
+    use gear_core::{
+        ids::{MessageId, ProgramId},
+        message::{StoredDelayedDispatch, StoredMessage},
+    };
+
+    #[test]
+    fn migration_to_v3_works() {
```
techraed, line 255: can't see @breathx suggestions on tests to be implemented

techraed (changes requested):
Everything seems fine except for `StoredDelayedDispatch`.
Waiting for additional review of @breathx

pallets/gear-messenger/src/migrations.rs
mqxf: Except they are? Compared to the old code, it generates three different StorageDispatches, some with Some and some with None, adds them to the old storage, migrates to the new one, then asserts that they are migrated correctly. I'm decently sure that is what he meant when he first mentioned my tests which just ran the migrations without any assertions. Or am I missing something?

pallets/gear-messenger/src/migrations.rs
```diff
@@ -0,0 +1,363 @@
...
+    }
+
+    pub struct DispatchesPrefix<T: Config>(PhantomData<T>);
+
+    impl<T: Config> StorageInstance for DispatchesPrefix<T> {
+        fn pallet_prefix() -> &'static str {
+            <<T as frame_system::Config>::PalletInfo as PalletInfo>::name::<Pallet<T>>()
+                .expect("No name found for the pallet in the runtime!")
+        }
+        const STORAGE_PREFIX: &'static str = "Dispatches";
```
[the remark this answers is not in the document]
mqxf: Pretty sure clippy would complain, since thats how the trait was defined, and it doesn't like it when I change the order of trait objects.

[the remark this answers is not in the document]
techraed: resolved then

core/src/message/stored.rs
```diff
@@ -216,3 +216,59 @@ impl Deref for StoredDispatch {
...
+
+impl From<StoredDelayedDispatch> for StoredDispatch {
+    fn from(dispatch: StoredDelayedDispatch) -> Self {
+        StoredDispatch::new(dispatch.kind, dispatch.message, None)
+    }
+}
+
+/// Stored message with entry point.
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Decode, Encode, TypeInfo)]
+pub struct StoredDelayedDispatch {
```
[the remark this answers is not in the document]
techraed: Having a separate type for stored delayed dispatch is a good practice, but introducing it only for the migration seems odd.

Except for tests, we can check the `None` with `post_upgrade`, isn't it?
Instead of introducing a temporary type, I would recommend even make fields of `StoredDispatch` public until the migration is applied to the real chain and then revert this (also with migrations deletion).

@breathx what do you think?

pallets/gear-messenger/src/migrations.rs
mqxf: hmm ok, I will do that

core/src/message/stored.rs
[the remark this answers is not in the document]
breathx: @techraed,

The `StoredDelayedDispatch` is not temporary solution and even more is not for this migrations.

Dispatches that are just sent with delay are never have `ContextStore`. It's much better to see type-level guarantees for storing data instead of just every time check that it's none.

Implementing it within the PR is preferable way because we can fit two different independent (!) changes on storage in one migration cycle.

Speaking of migrations itself: we will perform transformation with checks: if some of the message context is not None, which is impossible in case of our runtime has no significant bugs, we will log::error it

core/src/message/stored.rs
```diff
@@ -216,3 +216,59 @@ impl Deref for StoredDispatch {
...
+impl StoredDelayedDispatch {
+    /// Create new StoredDelayedDispatch.
+    pub fn new(kind: DispatchKind, message: StoredMessage) -> Self {
+        Self { kind, message }
+    }
+
+    /// Convert StoredDispatch into IncomingDispatch for program processing.
+    pub fn into_incoming(self, gas_limit: GasLimit) -> IncomingDispatch {
+        IncomingDispatch::new(self.kind, self.message.into_incoming(gas_limit), None)
+    }
```
breathx, line 250:
```suggestion
```

pallets/gear-messenger/src/migrations.rs
```diff
@@ -0,0 +1,435 @@
...
+                weight = weight.saturating_add(T::DbWeight::get().reads_writes(1, 1));
+                Some(LinkedNode {
+                    next: store.next,
+                    value: store.value.into(),
+                })
+            });
+
+            DispatchStash::<T>::translate(
+                |_, store: (v2::StoredDispatch, Interval<BlockNumberFor<T>>)| {
+                    if store.0.context.is_some() {
```
breathx, line 67: could you add similar check in pre_uprade, that none of delayed dispatches has context?

pallets/gear-messenger/src/migrations.rs
```diff
@@ -0,0 +1,435 @@
...
+                        log::error!("Previous context on StoredDispatch in DispatchStash should always be None, but was {:?}", store.0.context);
```
breathx, line 68: print message id instead of context

breathx (changes requested):
Minor comments below. Otherwise lgtm.

@techraed

breathx:
Kindly asking @ark0f to take a look as well

core/src/message/stored.rs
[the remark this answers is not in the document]
techraed: > @techraed,
> 
> The `StoredDelayedDispatch` is not temporary solution and even more is not for this migrations.
> 
> Dispatches that are just sent with delay are never have `ContextStore`. It's much better to see type-level guarantees for storing data instead of just every time check that it's none.
> 
> Implementing it within the PR is preferable way because we can fit two different independent (!) changes on storage in one migration cycle.
> 
> Speaking of migrations itself: we will perform transformation with checks: if some of the message context is not None, which is impossible in case of our runtime has no significant bugs, we will log::error it

discussed, thank you.

techraed (approved):

pallets/gear-messenger/src/migrations.rs
```diff
@@ -0,0 +1,435 @@
...
+        );
+
+        // 1 read for on chain storage version
+        let mut weight = T::DbWeight::get().reads(1);
+
+        if current == 3 && onchain == 2 {
+            Waitlist::<T>::translate(
+                |_, _, store: (v2::StoredDispatch, Interval<BlockNumberFor<T>>)| {
+                    weight = weight.saturating_add(T::DbWeight::get().reads_writes(1, 1));
+                    Some((store.0.into(), store.1))
```
ark0f, line 53:
```suggestion
                |_, _, (dispatch, interval): (v2::StoredDispatch, Interval<BlockNumberFor<T>>)| {
                    weight = weight.saturating_add(T::DbWeight::get().reads_writes(1, 1));
                    Some((dispatch.into(), interval))
```

core/src/message/stored.rs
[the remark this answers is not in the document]
ark0f: This definitely should be described in comments

pallets/gear-messenger/src/lib.rs
```diff
@@ -591,6 +593,7 @@ pub mod pallet {
         type MailboxSecondKey = MessageId;
         type MailboxedMessage = UserStoredMessage;
         type QueuedDispatch = StoredDispatch;
```
ark0f, line 595: It looks like we need to migrate queue, too

pallets/gear-messenger/src/migrations.rs
```diff
@@ -0,0 +1,435 @@
...
+        ProgramId,
+        Identity,
+        MessageId,
+        (StoredDispatch, Interval<BlockNumberFor<T>>),
+    >;
+}
+
+#[cfg(test)]
+#[cfg(feature = "try-runtime")]
+mod test {
```
ark0f, line 244:
```suggestion
mod tests {
```

pallets/gear-messenger/src/migrations.rs
```diff
@@ -0,0 +1,435 @@
...
+                    random_dispatch(true),
+                    Interval {
+                        start: random(),
+                        finish: random(),
+                    },
+                )
+            })
+            .collect::<Vec<_>>();
+
+            for dispatch in dispatch_stash.iter() {
```
ark0f, line 398:
```suggestion
            for (msg_id, dispatch, interval) in dispatch_stash.clone() {
```

pallets/gear-messenger/src/migrations.rs
```diff
@@ -0,0 +1,435 @@
...
+                v2::DispatchStash::<Test>::insert(
+                    dispatch.0,
+                    (dispatch.1.clone(), dispatch.2.clone()),
+                );
+            }
+
+            let state = MigrateToV3::<Test>::pre_upgrade().unwrap();
+            let _ = MigrateToV3::<Test>::on_runtime_upgrade();
```
ark0f, line 406: Assert weight is not zero

ark0f (dismissed):

core/src/message/context.rs
```diff
@@ -484,9 +498,8 @@ impl MessageContext {
 
     /// Wake message by it's message id.
     pub fn wake(&mut self, waker_id: MessageId, delay: u32) -> Result<(), Error> {
-        if self.store.awaken.insert(waker_id) {
+        if !self.outcome.awakening.iter().any(|v| v.0 == waker_id) {
```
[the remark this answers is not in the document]
mqxf: This is a Vec<(MessageId, u64)>, contains does not take a function, but a value to compare with Eq, but we do not care what the u64 is since only if the MessageId is the same, it is a duplicate, therefore, I cannot use contains.

pallets/gear-messenger/src/migrations.rs
mqxf: I did, a while ago, but it fails, because, Test weights for everything is zero, so it always adds zero for every read and write.

ark0f: https://github.com/gear-tech/gear/blob/ed3290bd7440103deb39dbfc715c3bc6794e90be/pallets/gear-messenger/src/mock.rs#L50

Replace to`DbWeight = RocksDbWeight`

pallets/gear-messenger/src/lib.rs
breathx: Migrations contain translation step for message queue (storage `Dispatches`)

mqxf: We are migrating the inner storage of the queue, which is the StoredDispatch if i remember correctly

pallets/gear-messenger/src/migrations.rs
```diff
@@ -87,7 +87,13 @@ impl<T: Config> OnRuntimeUpgrade for MigrateToV3<T> {
     fn pre_upgrade() -> Result<Vec<u8>, DispatchError> {
         let mut count = v2::Waitlist::<T>::iter().count();
         count += v2::Dispatches::<T>::iter().count();
-        count += v2::DispatchStash::<T>::iter().count();
+        count += v2::DispatchStash::<T>::iter().inspect(
+            |store| {
+                if store.1.0.context.is_some() {
+                    log::error!("Previous context on StoredDispatch in DispatchStash should always be None, but was Some for message id {:?}", store.1.0.message.id());
```
breathx, line 93: panic in pre-upgrade, it's only for testing purpose

breathx (approved):

PR #3474, opened 2023-11-07
fix(readme): Update binary usage examples

NikVolf:

PR #3479, opened 2023-11-07
ci(label): provide precise logs of check runs on skipping build

### LINKED ISSUE #3316: More logs in the label checks
### Problem to Solve

https://github.com/gear-tech/gear/actions/runs/6252495002/job/16975888383#step:3:19

the checks are not fully printed 

### Possible Solution

plus logs for the result of appending checks to the matched PRs

### Notes

ref https://github.com/gear-tech/gear/pull/3314

### LINKED ISSUE #3318: Generate & use `.binpath` in the `target` directory or reject it
### Problem to Solve

See the 6 point:
gear-foundation/cargo-program#103

### Possible Solution

This is only point related not only to `gear-wasm-builder` but also to `gtest`. So, we need to store `.binpath` somewhere in the `target` directory or reject it completely.

### Notes

_No response_

clearloop:
Resolves #3316

@gear-tech/dev

### DIFF
--- a/.github/actions/label/build.js
+++ b/.github/actions/label/build.js
@@ -30,29 +30,24 @@ const skip = async ({ core, github }) => {
     ref: REF,
   });
 
-  core.info(`check runs: ${check_runs}`);
-
   const runs = linux
     ? check_runs.filter(
       (run) => run.name === "build" || run.name === "build / linux"
     )
     : check_runs.filter((run) => run.name === "build / macos-x86");
 
-  // Skip this action by default.
-  let skipped = false;
+  let skipAction = false;
   for (run of runs) {
-    // Process this action only if the previous build has been skipped.
+    // If there is already a build, skip this action.
     if (
-      (run.name === "build" && run.conclusion === "skipped")
-    )
-      skipped = true;
-
-    // If there is already a build, skip this action without more conditions.
-    if (run.name === "build / linux" || run.name === "build / macos-x86")
+      run.name === "build / linux"
+      || run.name === "build / macos-x86"
+      || (run.name === "build" && run.conclusion !== "skipped")) {
       return true;
+    }
   }
 
-  return !skipped;
+  return [skipAction, JSON.stringify(check_runs, null, 2)];
 };
 
 /**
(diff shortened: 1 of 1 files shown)


.github/actions/label/build.js
```diff
@@ -30,29 +30,24 @@ const skip = async ({ core, github }) => {
...
-    // If there is already a build, skip this action without more conditions.
-    if (run.name === "build / linux" || run.name === "build / macos-x86")
+      run.name === "build / linux"
+      || run.name === "build / macos-x86"
+      || (run.name === "build" && run.conclusion !== "skipped")) {
       return true;
+    }
   }
 
-  return !skipped;
```
clearloop, line 55: #3474 not triggering build again, could be caused by we skipping this action by default, and only runs it on finding `build (status: skipped)`, which missed a situation that skipping `build` from github CI processes after this label CI.

instead, we skip build only on build has been processed without any other conditions, plus removing the confused variable `skipped`.

.github/actions/label/build.js
```diff
@@ -30,29 +30,24 @@ const skip = async ({ core, github }) => {
...
-      (run.name === "build" && run.conclusion === "skipped")
-    )
-      skipped = true;
-
-    // If there is already a build, skip this action without more conditions.
-    if (run.name === "build / linux" || run.name === "build / macos-x86")
+      run.name === "build / linux"
+      || run.name === "build / macos-x86"
+      || (run.name === "build" && run.conclusion !== "skipped")) {
       return true;
```
breathx, line 46: Is it legit to return here just bool and below array with bool on first, further accepting only array for js?

breathx: @osipov-mit if it won't bother you, could you take a look pls?

breathx (approved):

.github/actions/label/build.js
osipov-mit:
```suggestion
      return [true];
```

breathx (changes requested):
Address comments above pls

.github/actions/label/build.js
clearloop: it works in js ))

breathx (approved):

shamilsan (approved):

PR #3484, opened 2023-11-08
fix(wasm-gen): Include value into argument of invocable syscalls with destination param

SabaunT:
Loose and precise syscalls with destination param received only destination (`gsys::Hash`) instead of destination with value (`gsys::HashWithValue`) when their invocation instructions were generated in `wasm-gen`.

### DIFF
--- a/utils/wasm-gen/src/generator/syscalls/invocator.rs
+++ b/utils/wasm-gen/src/generator/syscalls/invocator.rs
@@ -31,9 +31,10 @@ use gear_wasm_instrument::{
     parity_wasm::elements::{BlockType, Instruction, Internal, ValueType},
     syscalls::{ParamType, PtrInfo, PtrType, SysCallName, SysCallSignature},
 };
+use gsys::Hash;
 use std::{
     collections::{btree_map::Entry, BTreeMap, BinaryHeap, HashSet},
-    iter,
+    iter, mem,
     num::NonZeroU32,
 };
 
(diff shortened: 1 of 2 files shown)


utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -376,13 +377,36 @@ impl<'a, 'b> SysCallsInvocator<'a, 'b> {
...
-            vec![
-                // call `gsys::gr_source` with a memory offset
+            let mut ret = {
+                // 3 instructions for invoking `gsys::gr_source` and possibly 3 more
+                // for defining value param so HashWithValue will be constructed.
+                let capacity = 3 + invocable
+                    .has_destination_param_with_value()
+                    .then_some(3)
+                    .unwrap_or_default();
+                Vec::with_capacity(capacity)
```
DennisInSky, line 387: Is it crucial to be that precise here? I.e. does it matter much whether there are not less than 3 or 6 items? If not, then I think this code can be re-written in a more readable way when you put the first 3 items all the time and then if append yet another 3 based on the condition (now you compute that condition twice)

DennisInSky (approved):

gshep (approved):

StackOverflowExcept1on (approved):

utils/wasm-gen/src/generator/syscalls/invocator.rs
SabaunT: done

DennisInSky (approved):

PR #3486, opened 2023-11-09
perf(core-backend): Register write access before execution

ukint-vs:
This will avoid unnecessary interruption of program execution.

@reviewer-or-team

### DIFF
--- a/core-backend/src/runtime.rs
+++ b/core-backend/src/runtime.rs
@@ -111,11 +111,10 @@ impl<'a, 'b, Ext: BackendExternalities + 'static> CallerWrap<'a, 'b, Ext> {
             gas,
             cost,
             |ctx: &mut Self| -> Result<_, UndefinedTerminationReason> {
+                let write_res = ctx.manager.register_write_as::<R>(res_ptr);
+
                 let res = f(ctx);
                 let res = ctx.process_fallible_func_result(res)?;
-
-                // TODO: move above or make normal process memory access.
-                let write_res = ctx.manager.register_write_as::<R>(res_ptr);
 
                 ctx.write_as(write_res, R::from(res)).map_err(Into::into)
             },


breathx (approved):

core-backend/src/runtime.rs
```diff
@@ -111,12 +111,11 @@ impl<'a, 'b, Ext: BackendExternalities + 'static> CallerWrap<'a, 'b, Ext> {
             gas,
             cost,
             |ctx: &mut Self| -> Result<_, UndefinedTerminationReason> {
+                let write_res = ctx.manager.register_write_as::<R>(res_ptr);
```
grishasobol, line 114: This can cause a problems with fallible syscall result. For example

- during `let res = f(ctx);` execution `write_res`  considered to be out of memory.
- then you will receive error here: `let res = ctx.process_fallible_func_result(res)?;`,
but does not exit function cause error is fallible and must be returned to user.
-  `ctx.write_as(write_res, R::from(res)).map_err(Into::into)` `write_res` was not processed, because it was out of memory. Cause registred accesses lists are already clear - nothing will be pre-processed, and access to `write_res` memory will cause signal which will be considered as out of memory and execution will finished with out of memory access, as expected.
- Although execution result is as expected: out of memory access, !!!! charged gas will be different cause globals was not updated and we will loose this gas. It's very dangerous and can be possible way to hack as.

grishasobol (changes requested):
This can cause a problems with fallible sys-calls

core-backend/src/runtime.rs
gshep: can it be covered with a test case?

grishasobol: Even worse situation can happen if `write_res` is ok, but some other accesses are not ok, so nothing will be pre-processed, so we will also go to signal handler where globals will be updated, but that global not yet synced with globals in `ext`, that can also cause potential risks.

grishasobol: > can it be covered with a test case?

yes can be )

breathx:
Buggy due to show-case #3490

PR #3487, opened 2023-11-09
feat(demo-constructor): Support multiple calls in `if_else`

mqxf:
@reviewer-or-team


examples/constructor/src/call.rs
```diff
@@ -246,18 +246,20 @@ mod wasm {
             Some((left == right).encode())
         }
 
-        fn if_else(self, previous: Option<CallResult>) -> Option<Vec<u8>> {
-            let Self::IfElse(flag, true_call, false_call) = self else {
+        fn if_else(self, mut previous: Option<CallResult>) -> Option<Vec<u8>> {
```
clearloop, line 249:
```suggestion
        fn if_else(self, mut previous: Option<CallResult>) -> Option<Vec<Vec<u8>>> {
```

shouldn't this result be `Vec<Vec<u8>>`, or we will only have the result of the last call

clearloop (changes requested):

examples/constructor/src/call.rs
breathx: nope, we don't need all of the return values

examples/constructor/src/builder.rs
```diff
@@ -297,20 +297,13 @@ impl Calls {
     pub fn if_else(
```
breathx, line 297: some tests, especially returning some value from if_else would be great

examples/constructor/src/call.rs
```diff
@@ -246,18 +246,20 @@ mod wasm {
...
-            let call = if flag { true_call } else { false_call };
+            let calls = if flag { true_calls } else { false_calls };
 
-            let (_call, value) = call.process(previous);
+            for call in calls {
+                previous = Some(call.process(previous));
+            }
 
-            value
+            previous.map(|res| res.1).unwrap_or(None)
```
breathx, line 262:
```suggestion
            previous.map(|res| res.1).flatten()
```

examples/constructor/src/call.rs
```diff
@@ -246,18 +246,20 @@ mod wasm {
...
             };
 
             let flag = flag.value();
 
-            let call = if flag { true_call } else { false_call };
+            let calls = if flag { true_calls } else { false_calls };
 
-            let (_call, value) = call.process(previous);
+            for call in calls {
+                previous = Some(call.process(previous));
```
breathx, line 259: Whats the point?
```suggestion
                previous = call.process(previous);
```

breathx (changes requested):

breathx:
Once more, to run CI set the label `A0-pleasereview` and only then push commits

examples/constructor/src/call.rs
mqxf: Because the type of `previous` is Option<...>, so we just reuse it instead of making a whole new variable.

clearloop (approved):

examples/constructor/src/builder.rs
breathx: Test still actual. Impl it with return of some value from block. Use unit tests with non-related to syscalls commands

pallets/gear/src/tests.rs
```diff
@@ -15086,6 +15086,80 @@ fn test_handle_signal_wait() {
...
+        let handle = Calls::builder()
+            .if_else(
+                Arg::get("switch"),
+                Calls::builder().add_call(Call::Bool(false)),
+                Calls::builder().add_call(Call::Bool(true)),
+            )
+            .store("switch")
+            .if_else(
+                Arg::get("switch"),
+                Calls::builder().wait_for(1),
```
SabaunT, line 15105: You can simplify that by invoking `gr_exit`. Try that, it will reduce test.

pallets/gear/src/tests.rs
```diff
@@ -15086,6 +15086,80 @@ fn test_handle_signal_wait() {
...
+fn test_constructor_if_else() {
+    use demo_constructor::{Arg, Call, Calls, Scheme, WASM_BINARY};
+
+    init_logger();
+    new_test_ext().execute_with(|| {
+        let init = Calls::builder().bool("switch", false);
+        let handle = Calls::builder()
+            .if_else(
+                Arg::get("switch"),
+                Calls::builder().add_call(Call::Bool(false)),
```
breathx, line 15099: the pr introduces vec of calls, so add here vec of something.
for example, if true than panic(expected message), else wait .

breathx: furthermore, as I said you can test in in examples/constructor crate exactly, you don't need wasm execution here

examples/constructor/src/call.rs
```diff
@@ -246,18 +246,20 @@ mod wasm {
...
+            }
 
-            value
+            previous.and_then(|res| res.1)
```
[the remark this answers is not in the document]
breathx: Because that's correct, previous: `Option<(Call, Option<Vec<u8>>)>`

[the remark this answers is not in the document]
breathx: and_then here concats logic of map + flatten

breathx:
@SabaunT please review once more

examples/constructor/src/call.rs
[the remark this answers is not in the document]
SabaunT: Oops, it's an Option for the second param on tuple, right.

SabaunT:
> @SabaunT please review once more

If tests continue to be in pallet, I suggest calling gr_exit instead of checking the result of "if-else" with the wait.

SabaunT:
@mqxf it's your choice actually whether to leave the test in pallet, or to move to constructor crate.

pallets/gear/src/tests.rs
mqxf: The multiple calls will end up being tested anyways when I rewrite the signal test.

PR #3490, opened 2023-11-10
test(pallet-gear): add test for gas limit exceeded when out of memory bounds case

grishasobol:
check that execution ends with GasLimitExceeded error even if some out of bounds accesses occurs in sys-call

minor changes: fix \t in pallet tests


pallets/gear/src/tests.rs
```diff
@@ -3331,6 +3331,79 @@ fn memory_access_cases() {
...
+        let message_id = get_last_message_id();
+
+        run_to_block(2, None);
+        assert_last_dequeued(1);
+
+        // We have sent message with `gas_limit`, but it must not be enough,
+        // because one write access to memory costs 100_000_000 gas (storage write cost).
+        // Fallible syscall error is written in each iteration to new gear page,
+        // so to successfully finish execution must be at least 100_000_000 * 512 * 4 = 204_800_000_000 gas,
+        // which is bigger than than provided `gas_limit`.
```
clearloop, line 3399:
```suggestion
        // which is bigger than provided `gas_limit`.
```

gshep (approved):

ukint-vs (approved):

breathx (approved):

PR #2053, opened 2023-01-01
Add gas lock in send_delayed_dispatch

MikitaS:
Resolves #2002

PR #3496, opened 2023-11-16
bugfix(runtime): Correct charging on delayed sendings from reservations

breathx:
Resolves artefacts after #2053

### DIFF
--- a/pallets/gear/src/internal.rs
+++ b/pallets/gear/src/internal.rs
@@ -535,8 +535,7 @@ where
         let hold_builder = HoldBoundBuilder::<T>::new(StorageType::DispatchStash);
 
         // Calculating correct gas amount for delay.
-        let bn_delay = delay.saturated_into::<BlockNumberFor<T>>();
-        let delay_hold = hold_builder.clone().duration(bn_delay);
+        let delay_hold = hold_builder.duration(delay.saturated_into());
         let gas_for_delay = delay_hold.lock_amount();
 
         let interval_finish = if to_user {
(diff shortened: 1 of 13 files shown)


ukint-vs (approved):

pallets/gear/src/internal.rs
```diff
@@ -591,22 +590,18 @@ where
                 Self::remove_gas_reservation_with_task(dispatch.source(), reservation_id);
             }
 
-            // Calculating correct hold bound to lock gas.
-            let maximal_hold = hold_builder.maximum_for_message(dispatch.id());
```
SabaunT, line 595: did I get it right, that it was a copy paste from `wait_dispatch` and it's incorrect calculating `maximal_hold` here?

pallets/gear/src/tests.rs
```diff
@@ -83,6 +83,218 @@ use utils::*;
...
+
+        assert_failed(
+            mid,
+            ActorExecutionErrorReplyReason::Trap(TrapExplanation::Panic(error_text.into())),
+        );
+    });
+}
+
+#[test]
+fn delayed_reservations_to_mailbox() {
```
SabaunT, line 241: fyi, this test works fine on 1d4d95d6fea1881ab8c6c35261ef20b8a7b4f3f7
was that expected?

SabaunT (approved):

SabaunT:
Seems that charging for message delay differs for regularly sent messages and messages sent from reservations. When it's a regular message, than gas is charged from the original message, otherwise - from reservation. Shall we do something with that?

ark0f (changes requested):

pallets/gear/src/internal.rs
breathx: Yeah, no idea why here we should take maximum for message if we have pre-defined amount of blocks for storing in dispatch stash. Probably copy-pasted from wait_dispatch where wait(None) does it for maximum amount of blocks

pallets/gear/src/tests.rs
breathx: Yeah, just making sure that nothing fails with my change in logic of removal re-calculating of hold bound

ark0f (approved):

PR #3503, opened 2023-11-18
feat(gstd): Introduce critical hook

ark0f:
**Release notes:**
Introduce critical hook API. Similar to [`std::panic::set_hook`](https://doc.rust-lang.org/std/panic/fn.set_hook.html) but sets on per-message basis and takes into account specific program halts like gas outage, memory overflow, backend error and so on


gstd/src/critical.rs
```diff
@@ -0,0 +1,130 @@
...
+        self.fns.remove(&TypeId::of::<F>());
+    }
+
+    /// Executes every saved critical section once.
+    ///
+    /// Must be called in `handle_signal` entry point
+    /// if you don't use async runtime.
+    pub fn execute_all(&mut self) {
+        for (_, mut f) in mem::take(&mut self.fns) {
+            (f)();
```
StackOverflowExcept1on, line 98: What if we register multiple critical sections and then they are executed in random order due to `HashMap<K, V, S = RandomState>`?

breathx: So there is a question should we care about the order?

DennisInSky:
I am a bit confused by the behaviour of the `HandleAction::Panic` and the `HandleAction::WaitAndPanic` - in both cases we create a section with no the `execute` method being called, but in the second case it gets executed anyway. I can imagine this is a result of having the wait op before the panic, but it is absolutely not intuitive when you read the code w/o impl behind it

DennisInSky:
> I am a bit confused by the behaviour of the `HandleAction::Panic` and the `HandleAction::WaitAndPanic` - in both cases we create a section with no the `execute` method being called, but in the second case it gets executed anyway. I can imagine this is a result of having the wait op before the panic, but it is absolutely not intuitive when you read the code w/o impl behind it

To me it would be more comprehended if I had an option to register my compensation logic in relation to a future returned by some async call. This would clearly say that this functionality is about executing compensation with regards to this async/await. It doesn't seem the implemented functionality is relevant to any other cases.

gstd/src/critical.rs
ark0f: We use `hashbrown` crate which uses [`BuildDefaultHasher`](https://doc.rust-lang.org/nightly/core/hash/struct.BuildHasherDefault.html) by default so there are no random keys. 
But order is not still sequential because usually hash maps don't guarantee it.

Conceptual question is still theme to discuss

DennisInSky:
> > but it is absolutely not intuitive when you read the code w/o impl behind it
> 
> What about module docs?

This is not quite about docs - I read some code which uses some libraries and from reading it, it is vague what happens and how it works

NikVolf:
User-facing API should be _per-call_

i.e. critical section is created, fired, and disposed once corresponding call handling is out of scope

gstd/src/critical.rs
```diff
@@ -0,0 +1,130 @@
...
+//! ```
+
+use alloc::boxed::Box;
+use core::{any::TypeId, mem};
+use hashbrown::HashMap;
+
+static mut SECTIONS: Option<Sections> = None;
+
+pub(crate) struct Sections {
+    fns: HashMap<TypeId, Box<dyn FnMut()>>,
```
grishasobol, line 63: If we will have two sections to the same function, whether this will not break logic in case one is dropped:
```rust
fn func() {}

let s1 = Section::new(func);

{
     let _ = Section::new(func);
     // in drop remove section from table by `func` TypeId
}

something.await;

panic!(); // whether this panic cause `s1` execution in signal handler?

```

gstd/src/critical.rs
```diff
@@ -0,0 +1,116 @@
...
+//! // get source outside of critical section
+//! // because `gr_source` sys-call is forbidden inside `handle_signal` entry point
+//! let source = msg::source();
+//!
+//! msg::send_for_reply(msg::source(), "for_reply", 0, 0)
+//!     .expect("Failed to send message")
+//!     // register section
+//!     .critical(move || {
+//!         msg::send(source, "example", 0).expect("Failed to send message");
+//!     })
```
grishasobol, line 38: also it's not clear in doc, that two futures with critical sections cannot exists simultaneously .

gstd/src/critical.rs
```diff
@@ -0,0 +1,116 @@
...
+#[pinned_drop]
+impl<Fut> PinnedDrop for SectionFuture<Fut> {
+    fn drop(self: Pin<&mut Self>) {
+        let _func = sections().remove(&msg::id());
+        // future drops after `.await` is complete
+        // so `_func == Some(_)`
+        //
+        // and also drops in `handle_signal` during futures cleanup
+        // so `_func == None`
+        // because failing message ID must be obtained via `msg::signal_from()`
```
grishasobol, line 92: I completely cannot understand this comment. Could you make it more clear please ?

examples/async-critical/src/wasm.rs
```diff
@@ -0,0 +1,92 @@
...
+//! If an approval is obtained the method replies with "PONG".
+
+use crate::HandleAction;
+use gstd::{critical::SectionFutureExt, exec, msg, prelude::*};
+
+#[gstd::async_init]
+async fn init() {}
+
+#[gstd::async_main]
+async fn main() {
```
NikVolf, line 36: i'd added some clear cases where memory state of the example is modified/unmodified/rolled back following critical section execution (or lack of)

examples/async-critical/src/wasm.rs
```diff
@@ -0,0 +1,96 @@
...
+//! The program demonstrates asynchronous execution and
+//! how to use macros `gstd::async_init`/`gstd::async_main`.
+//!
+//!  `Init` method gets three addresses, sends "PING" messages
+//! to them and waits for at least two replies with any payload ("approvals").
+//!
+//! `Handle` processes only "PING" messages. When `handle` gets such message
+//! it sends empty requests to the three addresses and waits for just one approval.
+//! If an approval is obtained the method replies with "PONG".
+
```
mqxf, line 28: Add docs/description please, as per #2845

ark0f: It is test demo. Such demos will be moved their own directory in the future

examples/async-critical/src/wasm.rs
```diff
@@ -0,0 +1,86 @@
...
+            // call `gr_source` outside because it is forbidden in `handle_signal`
+            let source = msg::source();
+
+            // should send message because panic occurs below
+            critical::set_hook(move || {
+                msg::send_bytes(source, b"critical", 0).unwrap();
+            });
+
+            // wait occurs inside so hook is saved
+            gstd::msg::send_bytes_for_reply(msg::source(), b"for_reply", 0, 0)
```
NikVolf, line 55: what if wait does not occurs (i.e., reply has already arrived during previous break) ?

gstd/src/critical.rs
```diff
@@ -0,0 +1,116 @@
...
+//! only across `.await` calls because section has to be saved.
+//!
+//! ```rust,no_run
+//! use gstd::{critical::SectionFutureExt, msg};
+//!
+//! # async fn _dummy() {
+//!
+//! // get source outside of critical section
+//! // because `gr_source` sys-call is forbidden inside `handle_signal` entry point
+//! let source = msg::source();
```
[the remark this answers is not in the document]
ark0f: I think it's better to do in another PR with automation and entrypoints description in Rust docs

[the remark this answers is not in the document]
grishasobol: so, please add TODO. We always forget docs issues

examples/async-critical/src/wasm.rs
```diff
@@ -0,0 +1,86 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use crate::HandleAction;
+use gstd::{critical, exec, msg, prelude::*};
+
+#[gstd::async_init]
+async fn init() {}
```
breathx, line 23: ?
```suggestion
```

pallets/gear/src/tests.rs
```diff
@@ -14976,6 +14976,179 @@ fn test_handle_signal_wait() {
...
+            msg.id(),
+            EMPTY_PAYLOAD.to_vec(),
+            10_000_000_000,
+            0,
+            false,
+        ));
+
+        run_to_block(4, None);
+
+        assert_eq!(MailboxOf::<Test>::iter_key(USER_1).count(), 0);
```
breathx, line 15026: what if message didn't enter MB due to lack of gas? check events by `last_mail(user)` or sth like that fn exists in utils mod

pallets/gear/src/tests.rs
```diff
@@ -14976,6 +14976,179 @@ fn test_handle_signal_wait() {
...
+        assert_ok!(Gear::send_reply(
+            RuntimeOrigin::signed(USER_1),
+            msg.id(),
+            EMPTY_PAYLOAD.to_vec(),
+            10_000_000_000,
+            0,
+            false,
+        ));
+
+        run_to_block(5, None);
```
breathx, line 15148: and whats the point of the test? you don't trigger critical to assert if it was replaced

examples/async-critical/src/wasm.rs
breathx: If so, add at least 1 test case for init function that it works similarly

breathx (changes requested):

breathx:
Otherwise LGTM

breathx:
@ark0f kind reminder to finish this awesome work

NikVolf:
why is it stalling? @DennisInSky @breathx ?

DennisInSky (approved):
You know the difficulties arose during taking this design decision better, but I am still not quite happy with the solution as it still doesn't hihjlight that functionality is applicable to async code executed after await completion only. Probably, at least renaming the `set_hook` method into something like `set_hook_for_async` or moving it into the `async_runtime` module would give some hint to users about the context of this method.
I would also showcase that
```
set_hook(...);
panic!(...);
```
doesn't cause the hook to get executed. Whereas
```
some_async(...).await;
take_hook();
panic!(...);
```
will do

examples/async-critical/src/wasm.rs
```diff
@@ -0,0 +1,112 @@
...
+                msg::send_bytes(source, b"from_handle_reply", 0).unwrap();
+            });
+        }
+    }
+}
+
+fn my_handle_signal() {
+    unsafe {
+        if SIGNAL_SET_HOOK {
+            let source = msg::source();
```
breathx, line 104: aren't we forbid taking source for handle signal?

ark0f: Nice catch

PR #3504, opened 2023-11-18
feat(core, runtime): Runtime built-in actor

ekovalev:
Adds a so-called "built-in" actor implemented as a Runtime pallet that can handle messages from normal actors (contracts) in applications built on top of some blockchain logic (staking, governance etc.). See detailed description in the crate's readme file.

- [x] Framework implementation as a pallet
- [x] Staking operations
- [x] Common protocol the contracts should adhere to when talking to the built-in actor
- [x] Example contract (`staking broker`)
- [x] Benchmarking to calibrate the correct cost in terms of gas
- [x] Unit tests for liquid staking use case
<br/>

Good to have, but can be done later:
- Complex tests involving real applications


pallets/gear/src/lib.rs
```diff
@@ -456,6 +485,10 @@ pub mod pallet {
         GearRunAlreadyInBlock,
         /// The program rent logic is disabled.
         ProgramRentDisabled,
+        /// User messages to this destination are not allowed.
+        ///
+        /// Occurs if a user attempts to send a message to a built-in actor.
+        IllegalDestination,
```
NikVolf, line 491: why is it illegal to user to send messages to built-in actor?

ekovalev: Maybe they could, but what is the use case? Users can directly call extrinsics from other runtime pallets.
Allowing users to send messages to built-in actor would complicate implementation - we'd need to be able to prepare a user reply message on the built-in actor side.
Let's think about it

ekovalev: Ok, re-enabled user messages to the built-in actor.
Indeed, it is a legit flow. However, sending messages from a user to the built-in actor doesn't make sense economically, therefore we shouldn't expect anyone wanting to do this... often.

pallets/gear-built-in-actors/src/lib.rs
```diff
@@ -0,0 +1,443 @@
...
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//! # Gear BuiltIn Actors Pallet
+//!
+//! The BuiltIn Actors pallet provides a a set of unique accounts that are treated as built-in
```
gshep, line 21:
```suggestion
//! The BuiltIn Actors pallet provides a set of unique accounts that are treated as built-in
```

pallets/gear-built-in-actors/src/lib.rs
```diff
@@ -0,0 +1,443 @@
...
+//!
+//! - [`Config`]
+//! - [`Call`]
+//!
+//! ## Overview
+//!
+//! The pallet implements a set of actors to handle messages from the message queue that
+//! require some sort of runtime-related logic to shortcut the usual messages processing flow.
+//! This provides an easy way for programs to interact with "pure runtime" actors thereby
+//! allowing contracts to build on top of functionality provided by the Runtime.
```
gshep, line 32: It is better to use more common term `program`

```suggestion
//! allowing programs to build on top of functionality provided by the Runtime.
```

pallets/gear-built-in-actors/src/lib.rs
```diff
@@ -0,0 +1,443 @@
...
+        /// The overarching event type.
+        type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+
+        /// The overarching call type.
+        type RuntimeCall: Parameter
+            + Dispatchable<RuntimeOrigin = Self::RuntimeOrigin, PostInfo = PostDispatchInfo>
+            + GetDispatchInfo
+            + From<pallet_staking::Call<Self>>;
+
+        /// The staking proxy actor pallet id, used for deriving its sovereign account ID.
```
gshep, line 167: Perhaps comment should be edited since the id is used to derive not only staking actor ID but for all future actors too

pallets/gear-built-in-actors/src/lib.rs
```diff
@@ -0,0 +1,443 @@
...
+    #[pallet::storage]
+    #[pallet::getter(fn actors)]
+    pub type Actors<T> = StorageMap<_, Identity, ActorType, ProgramId>;
+
+    #[pallet::hooks]
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+        fn on_initialize(_n: T::BlockNumber) -> Weight {
+            Weight::zero()
+        }
+    }
```
gshep, line 189: maybe that is not necessary?

pallets/gear-built-in-actors/src/lib.rs
```diff
@@ -0,0 +1,443 @@
...
+    fn dispatch_call(
+        origin: T::AccountId,
+        call: <T as Config>::RuntimeCall,
+        gas_limit: u64,
+    ) -> Result<(u64, Result<(), StakingErrorReason>), BuiltInActorReason> {
+        let call_info = call.get_dispatch_info();
+        Self::check_gas_limit(gas_limit, &call_info)?;
+        // Execute call
+        let res = call.dispatch(frame_system::RawOrigin::Signed(origin).into());
+        let actual_gas = extract_actual_weight(&res, &call_info).ref_time();
```
gshep, line 360: There is a suggestion to charge for 5-15% more gas in these cases. It would allow to incentivize validators to execute built-in actors.  The second argument is that messages to built-in actor don't pass the usual execution flow so consume less gas.

gshep (approved):
Incredible work! 🔥 

There is a couple of remarks however most of them are insubstantial. Please also address all todos or create issues and make corresponding references in the code.

examples/staking-broker/src/wasm.rs
```diff
@@ -0,0 +1,247 @@
...
+const BUILT_IN: ActorId = ActorId::new(hex!(
+    "9d765baea1938d17096421e4f881af7dc4ce5c15bb5022f409fc0d6265d97c3a"
+));
+
+pub type AccountId = [u8; 32];
+
+#[derive(Debug, Default)]
+struct StakingBroker {
+    /// Has bonded any amount yet
+    has_bonded_any: bool,
```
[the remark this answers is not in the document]
ekovalev: Not quite. It's a bit more complex than that: calling the "last" `unbond` on the `Staking` pallet doesn't immediately delete the so-called `ledger` - there is the unbonding period during which the ledger still lingers around until an explicit call to `do_withdraw_unbonded` is made (and there are no other unbonding chunks anymore). Only then the information about bonding is cleared.
Calling `bond` again while the controller's ledger still exists in storage will yield an error.

So this program is not what a real liquid staking provider should look like but merely a simplified demo program.

examples/staking-broker/src/wasm.rs
```diff
@@ -0,0 +1,247 @@
...
+}
+
+#[gstd::async_main]
+async fn main() {
+    let broker = unsafe { STATE.get_or_insert(Default::default()) };
+
+    let payload = msg::load().unwrap();
+    match payload {
+        StakingMessage::Bond { value } | StakingMessage::BondExtra { value } => {
+            broker.bond(msg::value().min(value)).await;
```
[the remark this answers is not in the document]
ekovalev: Yes and no ).
This was the first that comes across one's mind. The problem is that when a value is sent to a program, it is immediately deposited to the `gear-bank` and disappears from the program balance temporarily (until the `handle` call has completed successfully). But this amount has to be locked on the program's balance during bonding, and if the current program free balance is lower than the requested `value`, the smallest of the two will be bonded.
Hence we have to resort to encoding the bonded value in the payload.

pallets/gear-built-in-actors/common/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+extern crate alloc;
+
+use alloc::vec::Vec;
+use scale_info::scale::{self, Decode, Encode};
+
+// /// Actor id as it is seen by a contract.
+// #[derive(Encode, Decode, Clone, Copy, Default, PartialEq, Eq, Debug)]
+// #[codec(crate = scale)]
+// pub struct ActorId(pub [u8; 32]);
+pub type AccountId = [u8; 32];
```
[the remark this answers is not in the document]
ekovalev: We don't want the `gstd` as a dependency to avoid cyclic dependencies issue.
It's a type alias only after all - not even a new type. Will remove the dead code not to cause confusion.

pallets/gear/src/lib.rs
```diff
@@ -248,6 +274,9 @@ pub mod pallet {
             VoucherId = Self::AccountId,
         >;
 
+        /// A built-in actor enabling staking features for programs.
```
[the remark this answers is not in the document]
ekovalev: You're right. This is a leftover from the very first implementation version.

pallets/gear-built-in-actors/common/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+
+pub mod staking {
+    use super::*;
+
+    /// Type that should be used to create a message to the staking built-in actor.
+    /// It is a [partial] mirror of the staking pallet interface. Not all extrinsics
+    /// are supported, more can be added as needed for real-world use cases.
+    #[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+    #[codec(crate = scale)]
+    pub enum StakingMessage {
```
NikVolf, line 70: First, about naming -- better call it `Staking`, Message suffix is not needed

NikVolf, line 70: Second, better design call versionning right here 

So that we could change message format by incrementing version

pallets/gear/src/lib.rs
NikVolf: Those users can be bridges or some other external actors, not actual users, so it should be fine

examples/staking-broker/src/wasm.rs
```diff
@@ -0,0 +1,187 @@
...
+    prelude::*,
+    ActorId,
+};
+use hashbrown::HashMap;
+use hex_literal::hex;
+use parity_scale_codec::Encode;
+
+// Staking proxy built-in actor id (hardcoded for all runtimes)
+const BUILT_IN: ActorId = ActorId::new(hex!(
+    "9d765baea1938d17096421e4f881af7dc4ce5c15bb5022f409fc0d6265d97c3a"
```
shamilsan, line 35: Why this address is not declared in the `built_in` module?

pallets/gear-built-in-actors/Cargo.toml
```diff
@@ -0,0 +1,83 @@
+[package]
+name = "pallet-gear-built-in-actor"
```
shamilsan, line 2: Also,
```suggestion
name = "pallet-gear-builtin-actor"
```

examples/staking-broker/src/wasm.rs
ekovalev: There can potentially be more than one built-in actor who are essentially differentiated by their addresses (actor ids). The idea was that these addresses could be published off chain and kept up-to-date in case new actors are added or the algorithm how we derive an address changes.
But seems that declaring them as `built_in` module constants is a good idea.

pallets/gear-built-in-actors/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+        Self::check_gas_limit(gas_limit, &call_info)?;
+        // Skipping the actual call dispatch
+        let _ = extract_actual_weight(&Ok(Default::default()), &call_info).ref_time();
+        Self::deposit_event(Event::MessageExecuted { result: Ok(()) });
+
+        Ok((0, Ok(())))
+    }
+}
+
+pub mod staking_proxy {
```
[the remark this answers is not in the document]
ekovalev: It's the matter of preference, I guess, is it not? In my opinion keeping everything in one file doesn't affect readability at the moment while there are no other built-in actors.

pallets/gear-built-in-actors/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+        let mut output = vec![];
+        let actor_id = dispatch.destination();
+        if actor_id == Self::staking_proxy_actor_id() {
+            output = staking_proxy::handle::<T>(&dispatch, gas_limit).map_or_else(
+                |e| Self::process_error(&dispatch, e),
+                |(gas_spent, dispatch_result)| {
+                    Self::process_success(&dispatch, gas_spent, dispatch_result.err())
+                },
+            );
+        }
```
gshep, line 240: Maybe add note to log in the remaining 'else' case for diagnostic

pallets/gear-built-in-actors/common/src/lib.rs
ekovalev: This is kind of "inspired" by something like grpc protocol where they have a `request` and `response`. Here we are also defining a two-way protocol.
If we drop the suffix, how to differentiate between the message and response? Maybe better drop the prefix, instead?

pallets/gear-built-in-actors/src/lib.rs
```diff
@@ -0,0 +1,469 @@
...
+        gas_spent: u64,
+        reason: Option<StakingErrorReason>,
+    ) -> Vec<JournalNote> {
+        let message_id = dispatch.id();
+        let origin = dispatch.source();
+        let actor_id = dispatch.destination();
+
+        let mut journal = vec![];
+
+        // No call dispatced, so no gas burned except for the base
```
gshep, line 262: seems to be unnecessary comment

gshep (approved):

shamilsan (approved):
Many TODOs in the code. Also, maybe we are to create a dedicated crate for builtins instead of including them in `gstd`.  We could rename `gear-builtin-actor-common` to `gbuiltins` or `gear-builtin-actors` (the good name is the subject to discuss) and import it to Gear programs (alongside `gstd`) as well as to the pallet.

The remaining looks great to me.

examples/staking-broker/src/wasm.rs
```diff
@@ -0,0 +1,275 @@
...
+        let payload = Request::WithdrawUnbonded {
+            num_slashing_spans: 0,
+        };
+        debug!(
+            "[StakingBroker] Sending `withdraw_unbonded` message {:?} at broker's state {:?}",
+            payload, self
+        );
+        do_send_message(payload, || {
+            // TODO: send a part of withdrawn amount to the sender and/or
+            // some other users who requested unbonding earlier
```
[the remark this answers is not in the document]
ekovalev: It'd complicate the contract itself - we'd need to maintain more data structures in the contract to keep track of who else nominates which validator, and what are the proportions the reward should be split into so every nominator gets their share. This `TODO` is just a hint to a real contract developer that this should be taken care of.
Since it's rather urgent to get the builtin actors basics merged to not hold back other actors development, I've decided to leave this for the future.

gbuiltin/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    /// Type that should be used to create a message to the staking built-in actor.
+    ///
+    /// A [partial] mirror of the staking pallet interface. Not all extrinsics
+    /// are supported, more can be added as needed for real-world use cases.
+    #[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+    #[codec(crate = scale)]
+    pub enum Request {
+        /// Version 1 of the staking built-in actor protocol.
+        V1(RequestV1),
```
NikVolf, line 131: can we add a fixed tag (=0 ?) to that ?

breathx (changes requested):
Required changes were discussed. PR must be splitted in advance.

breathx:
Will be implemented with different PRs, including #3624

PR #3508, opened 2023-11-20
Draft: merge example programs

### LINKED ISSUE #3058: move all simple demos in demo-collection

holykol:
Resolves #3058

This is an example approach of how we can reduce the number of crates in `examples/` 

**todo:**
- [ ] way to reduce boilerplate code in `init()` even further? Suggestions are welcome

**moved examples:**
- [x] Ping
- [x] Vec
- [x] PiggyBank
- [x] Distributor

**downsides to this approach:**
* The examples are becoming less copy-pasteable, because they are built around `Program` trait, which hides `extern "C"` callback logic

**upsides:**
* faster compilation time (hopefully)
* less boilerplate for writing 3 lines of code (as seen in ping)


examples/custom/src/lib.rs
```diff
@@ -38,55 +39,68 @@ mod code {
...
+}
+
+pub trait Program {
+    fn init(args: gstd::Box<dyn gstd::any::Any>) -> Self
+    where
+        Self: Sized;
+
+    fn handle(&'static mut self) {}
+    fn state(&self) {}
 }
```
grishasobol, line 60: May be I don't understand something but can we simplify this code like this for example? If init does not take something then we can just put `P = ()` there. 
```suggestion
pub trait Program<P>: Sized {
    fn init(init_payload: P) -> Self;
    fn handle(&mut self) {}
    fn state(&self) {}
}
```

breathx:
Please, re-open once work continued

Issue #3216, opened 2023-09-05
Proposal: introduce gr_break


grishasobol:
### Problem to Solve

Currently we have 8 non-return syscalls:
```
gr_panic(payload: *const BufferStart, len: Length)
gr_oom_panic()
gr_exit(inheritor_id: *const Hash)
gr_leave()
gr_wait_for(duration: BlockNumber)
gr_wait_up_to(duration: BlockNumber)
gr_wait()
gr_out_of_gas()
```

Aslo we need one more sys-call `gr_stack_overflow` for stack limit exceeded (currently simple unreachable is using). Unfortunately instruction `unreachable` has no literal argument, so we cannot use it to identify the reason of unreachable execution. 

So, the problem is that for each one new special exit from execution reason we need to make separate sys-call, which cause problems in support, code size, code complexity and etc. 

### Possible Solution

Make two sys-calls instead:
```
gr_user_break(data: u64) -> !;
gr_system_break(data: u64) -> !;
```

`gr_user_break` is instead of all user available sys-calls. `data` contains next information:
1) `data.to_le_bytes[0]` - contains code of sys-call functionality. Should it act like gr_panic, gr_leave, gr_exit and etc.
2) `data.to_le_bytes[1]` - contains custom user information code. For some sys-calls it can be useful. For example: instead of panic with some panic string program can panic with some code and send this code in system reply.
3) in bytes `data.to_le_bytes[2..8]` special sys-call data is encoded (not scale codec). For example for `gr_wait_for` is `BlockNumber`. For `gr_panic` `data.to_le_bytes[2..6]` is  `*const BufferStart` and `data.to_le_bytes[6..8]` is `Length` (65536 bytes for panic payload should be enough)

`gr_system_break` is instead of `gr_out_of_gas` and `gr_stack_overflow`. Argument `data` just identifies whether it's `out_of_gas` or  `stack_overflow` occur.

### Notes

The signatures of this sys-calls can be expanded for more flexibility. For example:
```
gr_user_break(code: u64, data: u64) -> !;
gr_system_break(code: u64, data: u64) -> !;
```

gr_system_break will be restricted at instrumentation in original user code

breathx:
Solved in #3512

PR #3512, opened 2023-11-21
feat(core, wasm-instrument): add `gr_system_break` syscall

StackOverflowExcept1on:
Related to #3216

Instead of `gr_out_of_gas() -> !` we will have `gr_system_break(code: u32) -> !`

| code | meaning              |
|------|----------------------|
| 0    | out of gas           |
| 1    | stack limit exceeded |

`gr_out_of_gas()` is replaced with `gr_system_break(code=0)`
`unreachable` instructions are replaced with `gr_system_break(code=1)`

TODO:
- [x] create test that can catch `SimpleExecutionError::StackLimitExceeded` in `handle_signal` (in wasmer environment)
- [x] correctly calculate new stack limit
- [x] review gear-tech/wasm-instrument#7
- [x] publish `gwasm-instrument` under `v0.3.0` to avoid errors in dapps


grishasobol (changes requested):
unfortunately it's not possible now to add this changes because we already have working programs on Vara, so we must support now all legacy

StackOverflowExcept1on:
Yes, but isn't updating the instrumented version enough to solve this problem? `gr_out_of_gas` is a thing that exists only at the instrumentation level.

grishasobol:
> Yes, but isn't updating the instrumented version enough to solve this problem? `gr_out_of_gas` is a thing that exists only at the instrumentation level.

Ah yes. Forgot about this.

breathx:
> unfortunately it's not possible now to add this changes because we already have working programs on Vara, so we must support now all legacy

Yeah it was discussed before. This fn should have docs on syscall name enum maybe that it's synthetic syscall that we can modify anytime

Cargo.toml
```diff
@@ -189,8 +189,8 @@ tempfile = "3.8.1"
 #
 # fork of `wasm-instrument`
 #
-# https://github.com/gear-tech/wasm-instrument/tree/v0.2.3-sign-ext
-gwasm-instrument = { version = "0.2.3", default-features = false }
+# https://github.com/gear-tech/wasm-instrument/tree/v0.3.0-sign-ext-utils
```
breathx, line 192: btw can't we rename this branch to just 0.3.0 or gear-stable-0.3.0 etc?

core-backend/src/funcs.rs
```diff
@@ -1297,21 +1298,36 @@ where
...
+    }
+
+    pub fn system_break(_gas: Gas, code: u64) -> impl Syscall<Ext> {
+        RawSyscall::new(move |ctx: &mut CallerWrap<Ext>| {
+            let termination_reason = match SystemBreakCode::try_from(code) {
+                Ok(system_break_code) => match system_break_code {
+                    SystemBreakCode::OutOfGas => Self::out_of_gas(ctx),
+                    SystemBreakCode::StackLimitExceeded => Self::stack_limit_exceeded(),
+                },
+                _ => TrapExplanation::Unknown.into(),
```
breathx, line 1327: aren't we supposed to panic here?
its only case if we failed up instrumentation process, isn't it? @grishasobol

core-backend/src/funcs.rs
```diff
@@ -1297,21 +1298,36 @@ where
...
+        ActorTerminationReason::from(current_counter).into()
+    }
 
-            let termination_reason: ActorTerminationReason = current_counter.into();
+    fn stack_limit_exceeded() -> UndefinedTerminationReason {
+        log::trace!(target: "syscalls", "system_break(StackLimitExceeded)");
+        TrapExplanation::StackLimitExceeded.into()
+    }
+
+    pub fn system_break(_gas: Gas, code: u64) -> impl Syscall<Ext> {
```
breathx, line 1320: btw why not u32? for codes-consistency with error codes

examples/signal-entry/src/lib.rs
```diff
@@ -18,15 +18,16 @@
 
 #![cfg_attr(not(feature = "std"), no_std)]
 
-use gstd::errors::SignalCode;
+// We can't depend on gstd because it declares panic handler, so we just use gcore.
```
breathx, line 21: you can still depend on gstd wout panic-handler feature, but it will cause tests compiling cargo issue

pallets/gear/Cargo.toml
```diff
@@ -105,7 +106,7 @@ demo-reserve-gas.workspace = true
 demo-rwlock.workspace = true
 demo-reservation-manager.workspace = true
 demo-send-from-reservation.workspace = true
-demo-signal-entry.workspace = true
+demo-signal-entry = { workspace = true, features = ["debug"] } # The 'debug' feature is required for some assertion in tests
```
breathx, line 109: fetch master, issue with debug gstd feature is solved afaik

pallets/gear/src/benchmarking/tests/syscalls_integrity.rs
```diff
@@ -180,7 +259,7 @@ where
             | SyscallName::OomPanic => {/* tests here aren't required, read module docs for more info */},
             SyscallName::Alloc => check_mem::<T>(false),
             SyscallName::Free => check_mem::<T>(true),
-            SyscallName::OutOfGas => { /*no need for tests */}
+            SyscallName::SystemBreak => { /*no need for tests */}
```
breathx, line 262: ... because tested in other bench test

utils/wasm-instrument/src/tests.rs
```diff
@@ -183,7 +191,8 @@ fn duplicate_export() {
     );
     let module = parse_wat(&wat);
 
-    assert!(inject(module, &ConstantCostRules::default(), "env").is_err());
+    let (gr_system_break_index, module) = inject_system_break_import(module, "env").unwrap();
```
breathx, line 194: don't you want to make them some kind of builder so nobody will missuse the thing that they always supposed to be inserted together one by one?

utils/wasm-instrument/src/lib.rs
```diff
@@ -48,45 +50,65 @@ pub const GLOBAL_NAME_FLAGS: &str = "gear_flags";
...
+
+    fn try_from(value: u64) -> Result<Self, Self::Error> {
+        SystemBreakCode::try_from(value as i64)
+    }
+}
+
+pub fn inject_system_break_import(
     module: elements::Module,
-    rules: &R,
     gas_module_name: &str,
```
breathx, line 87:
```suggestion
    break_module_name: &str,
```

breathx (approved):
@ark0f @grishasobol please review

Cargo.toml
StackOverflowExcept1on: Makes sense, I can rename the branch after review gear-tech/wasm-instrument#7

core-backend/src/funcs.rs
StackOverflowExcept1on: In fact, this is an unreachable branch, because we never use codes other than `0` and `1` at instrumentation level. Do you mean do `_ => unreachable!("msg")` instead of `Trap(Unknown)`?

StackOverflowExcept1on: I think `Trap(Unknown)` compiles to smaller binary compared to panic.

core-backend/src/funcs.rs
StackOverflowExcept1on: Btw `i64.const` and `i32.const` has same cost. If we can have the ability to extend syscall for free in future, then why not leave `i64`?

gcli/src/meta/executor.rs
```diff
@@ -140,7 +140,7 @@ mod env {
                 "gr_gas_available" => func!(store, i32),
                 "gr_leave" => func!(store),
                 "gr_message_id" => func!(store, i32),
-                "gr_out_of_gas" => func!(store),
+                "gr_system_break" => func!(store, u64),
```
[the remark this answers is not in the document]
breathx: > @clearloop can you look if this will cause any problems when the user wants to get metadata from a program that has not yet updated the instrumented version?

It will cause such problem, but for me its not necessary to support. Re-instrumentation will always be on first access to program on-chain

core-backend/src/funcs.rs
breathx: Just wondering atm. Anyway this call and its arg are inserted on instrumentation

core-backend/src/funcs.rs
breathx: > In fact, this is an unreachable branch, because we never use codes other than `0` and `1` at instrumentation level. Do you mean do `_ => unreachable!("msg")` instead of `Trap(Unknown)`?

Yes

breathx: > I think `Trap(Unknown)` compiles to smaller binary compared to panic. The `free` syscall uses the same approach instead of panicking.

But free can be called by program in its initial logic, while this call is only created already inside our runtime.

gcli/src/meta/executor.rs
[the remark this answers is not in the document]
StackOverflowExcept1on: I see that gcli uses the local wasm file and then applies instrumentation to it. In this case everything should work.

core-backend/src/funcs.rs
StackOverflowExcept1on: done in 3632f47607bee0bcb622b7bf088b36a26d2d33bb

utils/wasm-instrument/src/tests.rs
StackOverflowExcept1on: I can try to do it in this PR, but it looks like it will much  change the gear-wasm-instrument

StackOverflowExcept1on: done in last commits

Cargo.toml
breathx: @grishasobol @ark0f please reflect on mentioned pr above

breathx (approved):
Awesome

pallets/gear/src/schedule.rs
```diff
@@ -764,7 +764,7 @@ impl Default for Limits {
 impl<T: Config> Default for InstructionWeights<T> {
     fn default() -> Self {
         Self {
-            version: 10,
+            version: 11,
```
ark0f, line 767: @breathx, don't we rise version during releases?

utils/calc-stack-height/src/main.rs
```diff
@@ -35,8 +36,25 @@ fn main() -> anyhow::Result<()> {
...
+    struct Env;
+
+    impl WasmerEnv for Env {
+        fn init_with_instance(&mut self, _: &Instance) -> Result<(), HostEnvInitError> {
+            Ok(())
+        }
+    }
+
+    let ty = FunctionType::new(vec![Type::I32], vec![]);
+    let func = Function::new_with_env(&store, &ty, Env, |_, args| match SystemBreakCode::try_from(
```
ark0f, line 50:
```suggestion
    let ty = FunctionType::new(vec![Type::I32], vec![]);
    let func = Function::new(&store, &ty, |_, args| match SystemBreakCode::try_from(
```

utils/wasm-instrument/src/lib.rs
```diff
@@ -47,46 +50,173 @@ pub const GLOBAL_NAME_FLAGS: &str = "gear_flags";
...
+    /// Whether to insert a gas limiter into WASM module.
+    pub fn with_gas_limiter(&mut self, get_gas_rules: GetRulesFn) -> &mut Self {
+        self.gas_limiter = Some(get_gas_rules);
+        self
+    }
+
+    /// Performs instrumentation of a given WASM module depending
+    /// on the parameters with which the [`InstrumentationBuilder`] was created.
+    pub fn instrument(&mut self, module: Module) -> Result<Module, InstrumentationError> {
+        if !(self.stack_limiter.is_some() || self.gas_limiter.is_some()) {
```
ark0f, line 161: More readable
```suggestion
        if let (None, None) = (self.stack_limiter, self.gas_limiter) {
```

ark0f (approved):

utils/wasm-instrument/src/lib.rs
StackOverflowExcept1on: ok, but need to add ref: `&self.gas_limiter` because type is not `Copy`
done in 342ae11dfc127005aab291a0ce01668a131195ee

utils/calc-stack-height/src/main.rs
StackOverflowExcept1on: discussed in pm, done in c9099670212f0638bb7353add29b5675a33a3b94

Cargo.toml
grishasobol: I think we must relay only on branches, which updates only by approved PRs

utils/wasm-instrument/src/lib.rs
```diff
@@ -47,46 +50,173 @@ pub const GLOBAL_NAME_FLAGS: &str = "gear_flags";
...
 
     let mut mbuilder = builder::from_module(module);
 
-    // fn out_of_gas() -> ();
-    let import_sig = mbuilder.push_signature(builder::signature().build_sig());
+    // fn gr_system_break(code: u32) -> !;
+    let import_sig =
+        mbuilder.push_signature(builder::signature().with_param(ValueType::I32).build_sig());
 
     mbuilder.push_import(
```
grishasobol, line 216: please change this to 
```rust
    let module = mbuilder
        .import()
        .module(break_module_name)
        .field(SyscallName::SystemBreak.to_str())
        .external()
        .func(import_sig)
        .build()
        .build();
```

grishasobol (commented):
Very good work!The only one problem is your unsafe branch which you changed without  PRs: `v0.3.0-sign-ext-utils`. I think better to create new branch first from last approved commit https://github.com/gear-tech/wasm-instrument/commit/6fc94e779dcf5073b7937fa78ed50887305c5389  and then makes PR for later changes.

core-backend/src/funcs.rs
```diff
@@ -1317,21 +1318,36 @@ where
...
+            // since it didn't pass comparison to argument of `gas_charge()`
+            ext.decrease_current_counter_to(0);
+        }
+
+        ActorTerminationReason::from(current_counter).into()
+    }
 
-            let termination_reason: ActorTerminationReason = current_counter.into();
+    fn stack_limit_exceeded() -> UndefinedTerminationReason {
+        log::trace!(target: "syscalls", "system_break(StackLimitExceeded)");
```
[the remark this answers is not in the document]
StackOverflowExcept1on: just removed it for `stack_limit_exceeded` but still keeping for `out_of_gas`

grishasobol (approved):

pallets/gear/src/schedule.rs
breathx: nothings criminal here
could be bumped here as well for test purposes. each time re-checked on releases

PR #3517, opened 2023-11-22
chore(workspace): fix warnings from dapp template, add gstd as dep

StackOverflowExcept1on:
dapp-template throws some additional warnings during build: https://github.com/gear-foundation/dapp-template/blob/master/.cargo/config.toml#L5
pr also fixes `cargo build -p {demo-calc-hash, demo-meta-io}`
@gear-tech/dev

### DIFF
--- a/utils/wasm-builder/src/wasm_project.rs
+++ b/utils/wasm-builder/src/wasm_project.rs
@@ -211,7 +211,7 @@ impl WasmProject {
         // Copy original `Cargo.lock` if any
         let from_lock = self.original_dir.join("Cargo.lock");
         let to_lock = self.out_dir.join("Cargo.lock");
-        let _ = fs::copy(from_lock, to_lock);
+        drop(fs::copy(from_lock, to_lock));
 
         let mut source_code = "#![no_std] pub use orig_project::*;\n".to_owned();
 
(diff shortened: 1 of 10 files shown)


breathx (approved):

utils/wasm-builder/src/wasm_project.rs
```diff
@@ -211,7 +211,7 @@ impl WasmProject {
         // Copy original `Cargo.lock` if any
         let from_lock = self.original_dir.join("Cargo.lock");
         let to_lock = self.out_dir.join("Cargo.lock");
-        let _ = fs::copy(from_lock, to_lock);
+        drop(fs::copy(from_lock, to_lock));
```
shamilsan, line 214: As for me, the previous one was better. Does it cause a warning?

shamilsan (approved):

utils/wasm-builder/src/wasm_project.rs
StackOverflowExcept1on: yes, it uses `-Wlet-underscore` in `rustflags`

shamilsan: And what if we write:

```rust
_ = fs::copy(from_lock, to_lock);
```

StackOverflowExcept1on: it still causes warning
```
warning: non-binding let on a type that implements `Drop`
   --> /home/.../gear/utils/wasm-builder/src/wasm_project.rs:214:9
    |
214 |         _ = fs::copy(from_lock, to_lock);
```

PR #3527, opened 2023-11-27
feat(gstd): get panic message without `panic_info_message`

### LINKED ISSUE #3524: Proposal: get rid of nightly features in gstd
### Problem to Solve

Do we need nightly features in gstd at all?
1. Once we merge #3472, out of memory errors will become very rare, so there is no point in `feature(alloc_error_handler)`
2. Regarding the `feature(panic_info_message)`: `core::panic::PanicInfo` implements `Display` and will format the panic message: `panicked at {file}:{line}:{column}:\n{msg}`

   Current implementation of gstd uses custom format: `'{msg}', {file}:{line}:{column}`. Should we use `#[feature(panic_info_message)]` just to implement a custom format of panic message? I think we can parse the String in a standard format and convert it to a custom format instead of requiring nightly toolchain.

### Possible Solution

1. Convert the output of `<core::panic::PanicInfo as Display>` to our custom format.
    It's probably easy, but it's worth considering `Rust <1.73`:
    rust-lang/rust#112849
    
    Discussed:
    - `debug` feature: `Display` as is
    - no `debug` feature: get message from `Display`, w/o location

2. Make our custom format obsolete

### Notes

_No response_

StackOverflowExcept1on:
Resolves #3524

We are getting closer to `cargo build +stable -p gstd`

@gear-tech/dev


gstd/src/common/handlers.rs
```diff
@@ -45,29 +45,38 @@ mod panic_handler {
     #[cfg(feature = "debug")]
     #[panic_handler]
     pub fn panic(panic_info: &PanicInfo) -> ! {
-        use crate::prelude::format;
-        #[cfg(not(feature = "panic-messages"))]
-        let message = None::<&core::fmt::Arguments<'_>>;
-        #[cfg(feature = "panic-messages")]
-        let message = panic_info.message();
+        use crate::prelude::{borrow::Cow, format, ToString};
```
breathx, line 48: That's pretty similar approach we used to have before using nightly feature.
Please, keep the feature for that.

breathx (changes requested):

gstd/src/common/handlers.rs
StackOverflowExcept1on: done

breathx: I meant introduction of different feature, that is not nightly and can't be used with nightly one

StackOverflowExcept1on: but why I should create feature for that instead of just fallback to stable rust?

gstd/src/common/mod.rs
```diff
@@ -19,6 +19,6 @@
 //! Common modules for each Gear smart contract.
 
 pub mod errors;
-#[cfg(target_arch = "wasm32")]
+#[cfg(feature = "common-handlers")]
```
breathx, line 22: shouln't we support common handlers only for wasm32 arch btw?

StackOverflowExcept1on:
@breathx
> So what have we decided to do with possibility to throw whole display for not(feature(debug))?

I think it's quite expensive even for the nightly version. `core::fmt::write` uses about 2.5 KiB, and extra 2.5 KiB is used for `impl Display for u32` (it's quite large). Thus, contracts will increase by 5 KiB on the stable version and by 2,5 KiB on nightly.

> How is it working wout oom-handler and allocation error on stable panic message?

We've already discussed this. On the nightly version, `gr_oom_panic` syscall is called. On the stable version, `gr_panic` syscall is called with message like this: `'memory allocation of 65536 bytes failed', library/alloc/src/alloc.rs:411:13`. This creates bit difference in behavior, but in general it is unlikely that anyone will rely on it in tests.

breathx (approved):
Got your point. I'm good with the solution. Awesome work
Please review @ark0f

gstd/src/common/mod.rs
StackOverflowExcept1on: Yes, we should, but in this case we will not be able to run tests that check the message format. I think tests should be located in `handlers.rs`

ark0f:
I'd just print `PanicInfo` without any parsing. Do we have cons with this approach?

StackOverflowExcept1on:
@ark0f there are several reasons:
1. As user of a smart contract, it would be convenient for me to see an error like `'You are not admin', /path/to/lib.rs:1:2` instead of `panicked at /path/to/lib.rs:1:2:\nYou are not admin`. The user will see an error at the beginning of the line instead of long path to source at the beginning.
2. Different versions of rust have different default panic message formats. If you rely on this format in tests, then you will have to change the tests. This also applies to our pallet tests.

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,209 @@
...
-            (Some(msg), None) => format!("'{msg:?}'"),
-            (None, Some(loc)) => {
-                format!("{}:{}:{}", loc.file(), loc.line(), loc.column())
+        #[panic_handler]
+        pub fn panic(panic_info: &PanicInfo) -> ! {
+            use crate::prelude::{fmt::Write, mem::MaybeUninit, str};
+            use arrayvec::ArrayString;
+
+            let option = panic_info.message().zip(panic_info.location());
+            let (message, location) = unsafe { option.unwrap_unchecked() };
```
ark0f, line 116: Don't use `unchecked` functions

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,209 @@
...
+
+            let _ = debug_msg.try_push_str(PANIC_OCCURRED);
+            let _ = debug_msg.try_push_str("'");
+            let _ = write!(&mut debug_msg, "{message}");
+            let _ = debug_msg.try_push_str("', ");
+            let _ = debug_msg.try_push_str(location.file());
+            let _ = debug_msg.try_push_str(":");
+            let _ = debug_msg.try_push_str(itoa_u32(&mut buffer, location.line()));
+            let _ = debug_msg.try_push_str(":");
+            let _ = debug_msg.try_push_str(itoa_u32(&mut buffer, location.column()));
```
ark0f, line 146: Can be replaced with `write!()`

gstd/Cargo.toml
```diff
@@ -16,17 +17,22 @@ bs58 = { workspace = true, features = ["alloc"] }
...
 parity-scale-codec = { workspace = true, features = ["derive"] }
 primitive-types = { workspace = true, features = ["scale-info"] }
+rustversion = { version = "1.0", optional = true }
 scale-info = { workspace = true, features = ["derive"] }
 futures = { workspace = true, features = ["alloc"] }
 
 static_assertions.workspace = true
 
+[dev-dependencies]
+rustversion = "1.0"
```
ark0f, line 27: Why is it in dev deps?

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,209 @@
...
+                }
+                unsafe {
+                    str::from_utf8_unchecked(
+                        &*(buffer.get_unchecked(idx..) as *const [_] as *const _),
+                    )
+                }
+            }
+
+            let mut debug_msg = ArrayString::<{ PANIC_OCCURRED.len() + TRIMMED_MAX_LEN }>::new();
+            let mut buffer = unsafe { MaybeUninit::uninit().assume_init() };
```
ark0f, line 136:
```suggestion
            let mut buffer = [MaybeUninit::uninit(); 10];
```

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,209 @@
...
+            let _ = debug_msg.try_push_str(&PANIC_OCCURRED[..4]);
+            let _ = write!(&mut debug_msg, "{panic_info}");
+
+            let src = (&PANIC_OCCURRED[4..]).as_bytes();
+            let dest = unsafe {
+                debug_msg
+                    .as_bytes_mut()
+                    .get_unchecked_mut(4..PANIC_OCCURRED.len())
+            };
+            dest.copy_from_slice(src);
```
ark0f, line 176:
```
debug_msg[0..4]= "pani";
debug_msg[4..] = panic_info;
debug_msg[4..PANIC_OCCURRED.len()] = "c occurred: "
```

I guess these actions are redundant?

ark0f (changes requested):

gstd/Cargo.toml
StackOverflowExcept1on: it is also used for doc-tests

gstd/src/common/handlers.rs
StackOverflowExcept1on:
```
1) write with format: "panicked at '{MESSAGE}', {location}"
2) push `pani` to debug str: "panipanicked at "
3) replace "panicked at " with "c occurred: "
4) output: "panic occurred: '{MESSAGE}', {location}"

File: gstd/src/common/handlers.rs (hunk 4 above)
Comment (StackOverflowExcept1on) [reply gstd/src/common/handlers.rs:146]: `impl Display for u32` costs 2,5 KiB in binary VS `itoa_u32` 82 bytes

File: gstd/src/common/handlers.rs (hunk 3 above)
Comment (StackOverflowExcept1on) [reply gstd/src/common/handlers.rs:116]: it always `Some`
https://github.com/rust-lang/rust/blob/5b8bc568d28b2e922290c9a966b3231d0ce9398b/library/core/src/panic/panic_info.rs#L26-L29
https://github.com/rust-lang/rust/blob/5b8bc568d28b2e922290c9a966b3231d0ce9398b/library/std/src/panicking.rs#L643-L644

Comment (StackOverflowExcept1on) [reply gstd/src/common/handlers.rs:116]: this has not changed for the last 5 years

Comment (ark0f) [reply gstd/src/common/handlers.rs:116]: It doesn't matter `PanicInfo` always returns  `Some`. I'm talking about `unchecked` fns everywhere which can be considered UB in case of logic corruption

File: gstd/src/common/handlers.rs (hunk 4 above)
Comment (ark0f) [reply gstd/src/common/handlers.rs:146]: So we can leave `itoa_u32` but use still use macro

File: gstd/src/common/handlers.rs [hunk 8]
```diff
@@ -25,50 +25,209 @@
...
```
ark0f, line 176: Can be simplified
```suggestion
            debug_msg
                .as_bytes_mut()
                .get_unchecked_mut(4..PANIC_OCCURRED.len())
                .copy_from_slice(&PANIC_OCCURRED[4..])
```

gstd/src/common/handlers.rs
StackOverflowExcept1on: In this case, the size of the binary file will become larger (`2_273 -> `3_824`). Also you can't reuse `buffer` due to taking the mutable reference twice.
```rust
let mut buffer = [MaybeUninit::uninit(); 10];

let _ = write!(
    &mut debug_msg,
    "{PANIC_OCCURRED}'{message}', {}:{}:{}",
    location.file(),
    itoa_u32(&mut buffer, location.line()), //&mut #1
    itoa_u32(&mut buffer, location.column()) //&mut #2
);
```

gstd/src/common/handlers.rs
StackOverflowExcept1on: also `.as_bytes_mut()` and `get_unchecked_mut()` are `unsafe fn`
`.as_bytes()` is still needed
```
  173 |                 .copy_from_slice(&PANIC_OCCURRED[4..]);
      |                  --------------- ^^^^^^^^^^^^^^^^^^^^ expected `&[u8]`, found `&str`
```

StackOverflowExcept1on: b23ab7b7ae9cf6f4d3eb68e6215a88f72a526243

gstd/src/common/handlers.rs
StackOverflowExcept1on: I can add `SAFETY comment` to each unsafe block in that case

gstd/src/common/handlers.rs
StackOverflowExcept1on: 66be25db10710945d5eb639e0b7a9089a8ce8647

gstd/src/common/handlers.rs
StackOverflowExcept1on: 9b2b9adf1263c4378f1bfe24de0b264f602e9846

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,253 @@
...
+                            ),
+                            str::from_utf8_unchecked(
+                                msg_bytes.get_unchecked((pos + NEEDLE.len())..),
+                            ),
+                        )
+                    })
+            }
+
+            let option = parse_panic_msg(&default_panic_msg);
+            let (location, message) = unsafe { option.unwrap_unchecked() };
```
StackOverflowExcept1on, line 256: It can abort without panic message if `location` is very long.

ark0f: Abort is not recommended. Call of `gr_panic` should occur at any circumstances

StackOverflowExcept1on: f7e459bfe94c07abde45616da15d993728a6dd99

breathx (approved):

breathx:
@ark0f

ark0f (approved):

breathx (changes requested):
@StackOverflowExcept1on @ark0f @breathx 

```toml
# `gstd` doesn't compile on stable rust < 1.73 |
# _____________________________________________|

default = ["panic-message"]

nightly = ["oom-handler"]

oom-handler = [] # add oom_handler()

panic-handler = [] # add panic_handler()                                            [0]
panic-message = ["panic-handler"] # add panic_handler() + add message to gr_panic   [1]
panic-location = ["panic-message"] # add panic location to panic-message            [2]

debug = ["galloc/debug", "gcore/debug"] # gr_debug + debug!()

# _____________________________________________|
# panic: "message"
# panic: "message" at "path"
# _____________________________________________|

# #[cfg(feature = "panic-message")]
# #[cfg(feature = "nightly")]
# #![unstable feature]
# .message();

# #[cfg(feature = "panic-message")]
# #[cfg(not(feature = "nightly"))]
# parsing

#[cfg(not(feature = "panic-message"))]
# no info
# _____________________________________________|
```

gstd/src/common/mod.rs
```diff
@@ -19,6 +19,6 @@
 //! Common modules for each Gear smart contract.
 
 pub mod errors;
-#[cfg(target_arch = "wasm32")]
+#[cfg(feature = "panic-handler")]
```
breathx, line 22: that way `gstd = { version = "*", default-features = false, features = ["com-handler"] }` is no-op due to lack of panic handler feature

gstd/src/lib.rs
```diff
@@ -129,7 +129,11 @@
 #![no_std]
 #![warn(missing_docs)]
 #![cfg_attr(
-    all(target_arch = "wasm32", feature = "panic-messages",),
+    all(
+        target_arch = "wasm32",
+        feature = "panic-info-message",
+        feature = "panic-message"
```
breathx, line 135: I guess semantically it could be removed 
```suggestion
```

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,208 @@
...
+///     assert_eq!(panic_info.to_string(), expected_format(panic_info));
+/// }));
+///
+/// let result = panic::catch_unwind(|| {
+///     panic!("{MESSAGE}");
+/// });
+/// assert!(result.is_err());
+/// ```
+pub mod panic_handler {
+    #[cfg(target_arch = "wasm32")]
```
breathx, line 73: arch wasm should be on whole module, isn't it?

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,208 @@
...
-mod panic_handler {
-    use crate::ext;
-    use core::panic::PanicInfo;
+/// We currently support 3 panic handler profiles:
+/// - `panic-handler`: it displays `no info`
+/// - `panic-message`: it displays `'{message}'`
+///   - In nightly Rust, we use `#![feature(panic_info_message)]` and the
+///     [`write!`] macro.
+///   - In stable Rust, we need to modify the default panic handler message
+///     format.
```
breathx, line 41: move this below

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,208 @@
...
-    #[cfg(not(feature = "debug"))]
-    #[panic_handler]
-    pub fn panic(_: &PanicInfo) -> ! {
-        ext::panic("no info")
-    }
+        /// Minimal panic handler.
+        #[cfg(not(any(feature = "panic-message", feature = "panic-location")))]
+        #[panic_handler]
+        pub fn panic(_: &PanicInfo) -> ! {
+            ext::panic("no info")
```
breathx, line 82: Mb replace with `"<unknown>"`?

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,208 @@
...
+/// - `panic-message`: it displays `'{message}'`
+///   - In nightly Rust, we use `#![feature(panic_info_message)]` and the
+///     [`write!`] macro.
+///   - In stable Rust, we need to modify the default panic handler message
+///     format.
+/// - `panic-location`: it displays `'{message}', {location}`
+///
+/// Default panic handler message format (according to <https://github.com/rust-lang/rust/pull/112849>):
+/// - Rust  <1.73: `panicked at '{message}', {location}`
+/// - Rust >=1.73: `panicked at {location}:\n{message}`
```
breathx, line 46: We can evaluate that in order to compile rust on stable >= 1.73 (add this to lib.rs)

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,208 @@
...
     crate::ext::oom_panic()
 }
 
-#[cfg(feature = "panic-handler")]
-mod panic_handler {
-    use crate::ext;
-    use core::panic::PanicInfo;
+/// We currently support 3 panic handler profiles:
+/// - `panic-handler`: it displays `no info`
+/// - `panic-message`: it displays `'{message}'`
```
breathx, line 37: and below
```suggestion
/// - `panic-message`: it displays `panic occurred: '{message}'`
```

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,208 @@
...
+/// - `panic-location`: it displays `'{message}', {location}`
+///
+/// Default panic handler message format (according to <https://github.com/rust-lang/rust/pull/112849>):
+/// - Rust  <1.73: `panicked at '{message}', {location}`
+/// - Rust >=1.73: `panicked at {location}:\n{message}`
+///
+/// We parse the output of `impl Display for PanicInfo<'_>` and
+/// then convert it to custom format: `'{message}', {location}`.
+///
+/// Here is a test to verify that the default panic handler message format
```
breathx, line 51: should we use exactly doc tests for it?

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,208 @@
...
+        use crate::ext;
+        use core::panic::PanicInfo;
 
-    #[cfg(not(feature = "debug"))]
-    #[panic_handler]
-    pub fn panic(_: &PanicInfo) -> ! {
-        ext::panic("no info")
-    }
+        /// Minimal panic handler.
+        #[cfg(not(any(feature = "panic-message", feature = "panic-location")))]
```
breathx, line 79: For clarification
```suggestion
        #[cfg(feature = "panic-handler)]
        #[cfg(not(any(feature = "panic-message", feature = "panic-location")))]
```

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,208 @@
...
+        }
+
+        #[cfg(any(feature = "panic-message", feature = "panic-location"))]
+        mod constants {
+            /// Max amount of bytes allowed to be thrown as string explanation
+            /// of the error.
+            pub const TRIMMED_MAX_LEN: usize = 1024; //TODO: do not duplicate `gear_core::str::TRIMMED_MAX_LEN`
+            /// This prefix is used to print debug message:
+            /// `debug!("panic occurred: {msg}")`.
+            pub const PANIC_OCCURRED: &str = "panic occurred: ";
```
breathx, line 92: why only for debug?

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,208 @@
...
+            #[cfg(not(feature = "nightly"))]
+            pub const PANICKED_AT: &str = "panicked at ";
+        }
+
+        #[cfg(any(feature = "panic-message", feature = "panic-location"))]
+        use constants::*;
+
+        /// Panic handler for nightly Rust.
+        #[cfg(feature = "nightly")]
+        #[cfg(any(feature = "panic-message", feature = "panic-location"))]
```
breathx, line 109: It's not always whole nightly implied
```suggestion
        #[cfg(feature = "panic-info-message")]
        #[cfg(feature = "panic-handler")]
        #[cfg(any(feature = "panic-message", feature = "panic-location"))]
```

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,208 @@
...
+
+            let mut debug_msg = ArrayString::<ARRAY_STRING_MAX_LEN>::new();
+
+            #[cfg(feature = "debug")]
+            let _ = debug_msg.try_push_str(PANIC_OCCURRED);
+
+            let _ = match (panic_info.message(), panic_info.location()) {
+                #[cfg(all(feature = "panic-message", feature = "panic-location"))]
+                (Some(msg), Some(loc)) => write!(&mut debug_msg, "'{msg}', {loc}"),
+                #[cfg(all(feature = "panic-message", not(feature = "panic-location")))]
```
breathx, line 123: just match cfg(feat = location), or cfg(not(feat = location))

gstd/src/common/handlers.rs
breathx: same below

breathx (changes requested):

gstd/src/lib.rs
StackOverflowExcept1on: What's the point of enabling nightly feature for panic handler with "no info"?

gstd/src/common/handlers.rs
StackOverflowExcept1on: `panic occurred: '` is just prefix for `debug!`, not for panic message

gstd/src/common/handlers.rs
StackOverflowExcept1on: doc tests are convenient because we can use `std::`, since gstd is `no_std` it will not work in normal tests.

gstd/src/common/handlers.rs
StackOverflowExcept1on: See how it works now in the master branch. I didn't change anything.
```rust
gr_panic("'msg', /path/to/file:1:2");
gr_debug("panic occurred: 'msg', /path/to/file:1:2");
```

gstd/src/common/handlers.rs
StackOverflowExcept1on: this is due to doc tests

gstd/src/common/handlers.rs
StackOverflowExcept1on: It seems to me more convenient to have tests and implementation in one file

gstd/src/common/handlers.rs
StackOverflowExcept1on: this is due to doc tests
`panic_handler::internal` - wasm thing, `panic_handler::` - doc tests

gstd/src/common/mod.rs
StackOverflowExcept1on: ok i did
```rust
#[cfg(feature = "panic-handler")]
pub mod panic_handler { }
```

gstd/src/common/handlers.rs
StackOverflowExcept1on: 9d99503e38b4d15e7f77d27eff6da71ec6013f86

gstd/src/common/handlers.rs
StackOverflowExcept1on: 63a341bba9b0b32f6fdf76d767af17a22003f606

gstd/src/common/handlers.rs
StackOverflowExcept1on: fbeaefc20735744ce8f60bd42132c9a67f0cab7d

gstd/src/common/handlers.rs
StackOverflowExcept1on: ed7741062e067677412d4bd8f6c7103251f718ab

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,208 @@
...
     use crate::ext;
     use core::panic::PanicInfo;
 
-    #[cfg(not(feature = "debug"))]
+    mod constants {
+        /// This prefix is used before the panic message.
+        pub const PANIC_PREFIX: &str = "panicked with ";
+        /// This panic message is used in the minimal panic handler and when
+        /// internal errors occur.
+        #[cfg(any(feature = "panic-info-message", not(feature = "panic-message")))]
```
breathx, line 63:
```suggestion
        #[cfg(any(all(feature = "panic-info-message", feature = "panic-message"), not(feature = "panic-message")))]
```
🤔

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,208 @@
...
+        pub const UNKNOWN_REASON: &str = "<unknown>";
+
+        /// This prefix is used by `impl Display for PanicInfo<'_>`.
+        #[cfg(all(not(feature = "panic-info-message"), feature = "panic-message"))]
+        pub const PANICKED_AT: &str = "panicked at ";
+
+        /// Max amount of bytes allowed to be thrown as string explanation
+        /// of the error.
+        #[cfg(feature = "panic-message")]
+        pub const TRIMMED_MAX_LEN: usize = 1024; //TODO: do not duplicate `gear_core::str::TRIMMED_MAX_LEN`
```
breathx, line 73: trimmed max len should be 1024 - panic prefix len

breathx (approved):

examples/waiter/tests/utils.rs
```diff
@@ -13,10 +13,6 @@ pub fn assert_panicked(result: &RunResult, panic_msg: &str) {
         )))
     ));
     let payload = String::from_utf8(result.log()[0].payload().into())
```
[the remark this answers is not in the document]
StackOverflowExcept1on: #3668

gstd/Cargo.toml
```diff
@@ -27,18 +30,26 @@ futures = { workspace = true, features = ["alloc"] }
...
+default = ["panic-message"]
+
+#! ## Panic handler profiles
+
+## When enabled, a minimal panic handler is provided by this crate.
+## Instead of a panic message, "<unknown>" is displayed.
+panic-handler = ["const_format"]
+## When enabled, a panic handler will also display a panic message.
+panic-message = ["panic-handler", "arrayvec"]
+## When enabled, a panic handler will also display a panic message and location.
```
ark0f, line 42: Worth to mention it's not recommended to use in production to avoid sensitive path leak

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,208 @@
...
+
+        #[cfg(feature = "debug")]
+        let _ = ext::debug(&debug_msg);
+
+        ext::panic(&debug_msg)
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
ark0f, line 208: It would be good to test every formatter

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,208 @@
...
+    }
+
+    use constants::*;
+
+    /// Minimal panic handler.
+    #[cfg(not(feature = "panic-message"))]
     #[panic_handler]
     pub fn panic(_: &PanicInfo) -> ! {
-        ext::panic("no info")
+        const MESSAGE: &str = const_format::formatcp!("{PANIC_PREFIX}'{UNKNOWN_REASON}'");
```
ark0f, line 85: The whole dependency can be replaced with `concat!()`
```suggestion
        const MESSAGE: &str = concat!(PANIC_PREFIX, "'", UNKNOWN_REASON, "'");
```

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,208 @@
...
 
-        crate::debug!("panic occurred: {msg}");
-        ext::panic(&msg)
+        #[cfg(feature = "debug")]
+        let _ = ext::debug(&debug_msg);
+
+        ext::panic(&debug_msg)
+    }
+
+    /// Panic handler for stable Rust >=1.73.
```
ark0f, line 123:
```suggestion
    /// Panic handler for stable Rust.
```

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,208 @@
...
+///
+/// How we get the panic message in different versions of Rust:
+/// - In nightly Rust, we use `#![feature(panic_info_message)]` and the
+///   [`write!`] macro.
+/// - In stable Rust, we need to modify the default panic handler message
+///   format.
+///
+/// Default panic handler message format (according to <https://github.com/rust-lang/rust/pull/112849>):
+/// - Rust  <1.73: `panicked at '{message}', {location}`
+/// - Rust >=1.73: `panicked at {location}:\n{message}`
```
ark0f, line 47: Should be removed because we force Rust 1.73

ark0f (approved):

gstd/Cargo.toml
StackOverflowExcept1on: but the stable rust will leak file paths anyway

gstd/src/common/handlers.rs
StackOverflowExcept1on: https://play.rust-lang.org/?version=nightly&mode=release&edition=2021&gist=812db8d01033f6f02bef668f38193c46
it doesn't compile

gstd/src/common/handlers.rs
StackOverflowExcept1on: I'll remove <1.73 but keep the current format to document the code itself

gstd/Cargo.toml
```diff
@@ -27,18 +30,26 @@ futures = { workspace = true, features = ["alloc"] }
...
 #! ## Default features
 
-default = ["panic-handler"]
-## When enabled, a panic handler is provided by this crate.
-panic-handler = []
+default = ["panic-message"]
+
+#! ## Panic handler profiles
+
+## When enabled, a minimal panic handler is provided by this crate.
```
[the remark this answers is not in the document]
StackOverflowExcept1on: a9d9c3f0355447368dfe31c670d053387de10248

gstd/src/common/handlers.rs
breathx: this will require compiling wasms.
the existing test is about default rust panic message formatting (for the case of future changes as it did in 1.73)
I don't think it's that much necessary (at this moment at least). Or we can implement processing functions that we use inside different builders (`foo(&'_ PanicInfo) -> &'_ str`) and check them in unit test

gstd/src/common/handlers.rs
breathx: Docs about elder rust still are here

gstd/Cargo.toml
breathx: It's anyway good to mention I guess

gstd/src/common/handlers.rs
```diff
@@ -25,50 +25,208 @@
...
+            #[cfg(feature = "panic-location")]
+            (Some(msg), Some(loc)) => {
+                let _ = write!(&mut debug_msg, "'{msg}' at '{loc}'");
+            }
+            #[cfg(not(feature = "panic-location"))]
+            (Some(msg), _) => {
+                let _ = write!(&mut debug_msg, "'{msg}'");
+            }
+            _ => {
+                let _ = debug_msg.try_push_str(const_format::formatcp!("'{UNKNOWN_REASON}'"));
```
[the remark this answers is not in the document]
breathx: ignoring that case

breathx:
🥳

PR #3535, opened 2023-11-28
feat(pallet-gear): Actual weight correction in `Gear::run` extrinsic

### LINKED ISSUE #3521: Force gear::run return actual weight
### Problem to Solve

Explorers similar to polkadotjs are unable to correctly show block load due to lack of this weight 

### Possible Solution

Query gas allowance after gear_run and return actual weight with Pays::No

### Notes

_No response_

ukint-vs:
Resolves #3521 .

Use max_block for Gear::run weight.
Sutract max_block from the block weight to get the actual block weight without `Gear::run`.

We can only decrease weight in `PostDispatchInfo`.
After this fix `Gear::run` shows correct actual weight.

image

@gear-tech/dev


shamilsan (approved):
🔥

pallets/gear/src/lib.rs
```diff
@@ -1600,9 +1600,11 @@ pub mod pallet {
             // overlay and never be committed to storage.
             GearRunInBlock::<T>::set(Some(()));
 
-            let weight_used = <frame_system::Pallet<T>>::block_weight();
+            // Subtract current block weight for Normal dispatches from the maximum block weight.
             let max_weight = <T as frame_system::Config>::BlockWeights::get().max_block;
-            let remaining_weight = max_weight.saturating_sub(weight_used.total());
+            let remaining_weight = max_weight.saturating_sub(
```
ekovalev, line 1605: How about the inherents (like `Timestamp::set()`) which also use some weight (very little though - ~10,500,000, but still) and have `DispatchClass::Mandatory`?
Also, we might have `DispatchClass::Operational` extrinsics one day. Should we account for those, as well, when calculating the remaining weight?

pallets/gear/src/lib.rs
```diff
@@ -1600,9 +1600,12 @@ pub mod pallet {
             // overlay and never be committed to storage.
             GearRunInBlock::<T>::set(Some(()));
 
-            let weight_used = <frame_system::Pallet<T>>::block_weight();
+            // Subtract current block weight for Normal dispatches from the maximum block weight.
             let max_weight = <T as frame_system::Config>::BlockWeights::get().max_block;
-            let remaining_weight = max_weight.saturating_sub(weight_used.total());
+            let weight_used = <frame_system::Pallet<T>>::block_weight()
```
breathx, line 1605: Provide explaining comments here pls

breathx (approved):

pallets/gear/src/lib.rs
```diff
@@ -1579,7 +1579,7 @@ pub mod pallet {
 
         /// Process message queue
         #[pallet::call_index(6)]
-        #[pallet::weight((Weight::zero(), DispatchClass::Mandatory))]
+        #[pallet::weight((<T as frame_system::Config>::BlockWeights::get().max_block, DispatchClass::Mandatory))]
```
breathx, line 1582: Need to deal with exhausting block resources. Consider use weight as max_block - used. It will result in really used = used / 2?

breathx (changes requested):

pallets/gear/src/lib.rs
ukint-vs: Block can't exhaust cuz it's Mandatory `DispatchClass`

breathx (approved):

PR #3538, opened 2023-11-29
feat(gear-wasm-builder): Add pre-processor

NikVolf:


utils/wasm-builder/src/wasm_project.rs
```diff
@@ -64,6 +64,20 @@ pub struct WasmProject {
...
+pub struct PreProcessOutput {
+    /// Location of the artifact generated by pre-processor
+    pub path: PathBuf,
+    /// Content of the artifact generated by pre-processor
+    pub content: Vec<u8>,
+}
+
+/// Pre-processor hook for wasm generation
+pub trait PreProcessor {
+    fn pre_process(&self, original: PathBuf) -> Result<PreProcessOutput>;
```
shamilsan, line 80: Why not `&Path` or `impl AsRef<Path>` instead of `Path`?

utils/wasm-builder/src/wasm_project.rs
```diff
@@ -414,6 +433,15 @@ extern "C" fn metahash() {{
 
         fs::create_dir_all(&self.target_dir).context("Failed to create target directory")?;
 
+        for pre_processor in self.pre_processors.iter() {
```
shamilsan, line 436: No need to call iter imo
```suggestion
        for pre_processor in &self.pre_processors {
```

shamilsan (approved):

utils/wasm-builder/src/wasm_project.rs
NikVolf: because it can't be made into a trait object

PR #3541, opened 2023-11-29
refactor(gsys, wasm-instrument): Make params types for syscalls in `wasm-instrument` and `gsys` consistent

techraed:
- Makes signatures of syscalls in `gsys` and `wasm-instrument` consistent by:
    - Setting `SizedBufferStart` param for those syscalls that accept buffers of the known size (literally, every syscall except for `gr_read`), as it's done in `wasm-instrument`.
    - Using `Offset` type for buffer offset params. There was a mix of using `Length` and `Index`, now only new `Offset` is used.
- Fixes some errors in docs and syscalls params types in `gsys`.
- Adds test for checking mutable error pointer position in the syscall.

### DIFF
--- a/gsys/src/lib.rs
+++ b/gsys/src/lib.rs
@@ -51,7 +51,7 @@ pub type Handle = u32;
 /// Represents hash type.
 pub type Hash = [u8; 32];
 
-/// Represents index type.
+/// Represents offset type.
 pub type Offset = u32;
 
 /// Represents length type.


utils/wasm-gen/src/generator/syscalls.rs
```diff
@@ -239,68 +239,9 @@ impl InvocableSysCall {
...
-            | SysCallName::Random
-            | SysCallName::Size
-            | SysCallName::Source
-            | SysCallName::ValueAvailable
-            | SysCallName::Value
-            | SysCallName::WaitFor
-            | SysCallName::WaitUpTo
-            | SysCallName::Wait
-            | SysCallName::Alloc
-            | SysCallName::Free
```
StackOverflowExcept1on, line 268: `SysCallName::Alloc` and `SysCallName::Free` must be fallible too

mertwole: But they don't write error to the pointer, so it depends on the definition of `fallible` in current context

utils/wasm-instrument/src/syscalls.rs
```diff
@@ -599,16 +610,16 @@ impl SysCallName {
...
-    Handler,         // i32 handler number
-    Alloc,           // i32 alloc pages
-    Free,            // i32 free page
-    Version,         // i32 version number of exec settings
+    Length,              // i32 buffers length
+    Ptr(PtrInfo),        // i32 pointer
+    Gas,                 // i64 gas amount
+    Offset,              // i32 offset in the input buffer (message payload)
+    DurationBlockNumber, // i32 duration in blocks
+    DelayBlockNumber,    // i32 delay in blocks
```
mertwole, line 618:
```suggestion
    Duration, // i32 duration in blocks
    Delay,    // i32 delay in blocks
```
I think it's clear from the context that `delay` and `duration` are defined in the terms of blocks as this approach is used everywhere in gear

SabaunT: So if we agree on that https://github.com/gear-tech/gear/pull/3541#discussion_r1410546814, I think we can resolve this too.

shamilsan (approved):
Awesome 🔥

ark0f (approved):

PR #3350, opened 2023-09-24
feat(node): Two-phase block proposing to respect deadline

ekovalev:
A long-discussed feature of limiting the `Gear::run()` evaluation time by a timeout constant to maintain consistent blocks production and avoid the "block production took too long" consensus error.

The changes mostly have to do with the `node/authorship` crate modification. Also, some small changes in the Substrate fork were necessary.
The approach taken is described in detail in the crate's `README.md`.

**Heads up**: this change has to do with concurrent modification of single a block builder instance (which, originally, was deemed to be neither cloneable nor safe to send in another thread), therefore it has to be tested with extra care to provide bullet-proof guarantees the resulting behaviour is as expected.

PR #3558, opened 2023-12-01
feat(node): Two phase block proposing for release v1.0.3

ekovalev:
Based on the original pull request #3350 to be a part of the v1.0.3 release.

The detailed explanation of what we are trying to achieve by this change is given in the README.md.

Briefly, due to Gear-specific block structure (namely, that there is a mandatory "pseudo-inherent" extrinsic that has to be the last one in any block after all other normal/operational extrinsics), we need to guarantee that each phase has enough time to run in order to maintain the desired block partitioning (25/75). This ratio is defined in the `Runtime` through max weights adjustment for each `DispatchClass`, but has not really been enforced in the block authorship utility. As a result, a discrepancy between how much gas (that is, time, a.k.a. `GasAllowance`) the `Gear::run()` thought it had at its disposal and how much time it actually had could lead to the `Gear::run()` still being executed past the block proposal deadline and, subsequently, to the "Block production took too long" error.

The modifications being introduced are threefold:
- refactor the deadlines so that each phase (extrinsics application and the `Gear::run()` had their own hard deadlines in terms of time (not only weight);
- adjust the `max_gas` and pass it explicitly to the `Gear::run()` if there is a suspicion that the remaining block gas (as it is seen in the `Runtime`) exceeds the actual time before the ultimate deadline;
- make the `Gear::run()` itself execute against a timer to be able to drop it if it is still taking too long.

Now a questionable point:
By default the block creation procedure leaves a good 1/3 of the block time to the finalization. While it still takes time to convert the overlay into changes and calculate the state root etc., it still seems too much, considering that in Gear we don't have any heavy `on_finalize` hooks. It should make no harm if the `Gear::run()` could use some small portion of this time in case it needs it to complete.
As a result, a concept of "deadline slippage" is being introduced. We set it to 10% in the code initially, and allow both hard deadlines we now have during the proposing to be "relaxed" by this slippage percentage.

However, it may seem that the slippage concept is, in fact, redundant, and we could have the same effect by simply supplying a greater "ultimate" deadline to the proposer in the first place (because all the deadlines calculations inside the proposer are linear with respect to the max duration). Let's say, instead of the default `2/3` of the total proposal duration make it `3/4` or so. The downside of this, however, is that now in terms of time we'll have 1.125s to apply all the extrinsics a block while the `Runtime` will still think we have `10^12` weight units which correspond to 1s.

Feedback on this appreciated. 

TODO:
- [x] Update (simplify) old basic tests that only test deadlines, tx skipping, soft deadlines etc. by suppressing the `Gear::run()` to render the tests more deterministic;
- [ ] Introduce more tests that verify the correctness of Gear-specific flow.


common/src/lib.rs
```diff
@@ -340,3 +340,14 @@ impl<AccountId: Default, ProgramId, Balance> PaymentVoucher<AccountId, ProgramId
         unimplemented!()
     }
 }
+
+/// Trait that the RuntimeApi should implement in order to allow deconstruction and reconstruction
+/// to and from its components.
+#[cfg(any(feature = "std", test))]
+pub trait RuntimeApiExt<C> {
```
breathx, line 347: Consider renaming into `Constructable` or sth

examples/constructor/src/call.rs
```diff
@@ -317,6 +318,19 @@ mod wasm {
...
         }
 
+        fn write_n(self) -> Option<Vec<u8>> {
+            let Self::WriteN(count) = self else {
+                unreachable!()
+            };
+
+            let end = count.value();
+            for i in 0_u64..end {
+                unsafe { DATA.insert("value".into(), i.encode()) };
```
breathx, line 328: Key isn't intuitive
```suggestion
                unsafe { DATA.insert("last_written_n".into(), i.encode()) };
```

pallets/gear-debug/src/lib.rs
```diff
@@ -281,5 +282,26 @@ pub mod pallet {
...
+                *_fraction * max
+            } else {
+                Weight::zero()
+            }
+        })]
+        pub fn exhaust_block_resources(
+            _origin: OriginFor<T>,
+            _fraction: Percent,
+        ) -> DispatchResultWithPostInfo {
+            Ok(Pays::No.into())
```
breathx, line 304:
```suggestion
            ensure_root(origin)?;
            Ok(Pays::No.into())
```

node/authorship/src/authorship.rs
```diff
@@ -520,25 +549,87 @@ where
 
         self.transaction_pool.remove_invalid(&unqueue_invalid);
 
-        // Pushing a custom extrinsic that must run at the end of a block
-        let custom_extrinsic = block_builder.create_terminal_extrinsic(self.max_gas)?;
+        // Attempt to apply pseudo-inherent on top of the current overlay in a separate thread.
+        // In case the timeout is hit, we will proceed without it.
+        let client = self.client.clone();
+        let backend = self.backend.clone();
+        let pseudo_inherent = block_builder.create_terminal_extrinsic(self.max_gas)?;
```
breathx, line 556: shouldn't we create it already in update_block?

node/authorship/src/authorship.rs
```diff
@@ -520,25 +549,87 @@ where
...
-            Ok(()) => {
-                debug!(target: "gear::authorship", "⚙️  ... pushed to the block");
-            }
-            Err(ApplyExtrinsicFailed(Validity(e))) if e.exhausted_resources() => {
-                warn!(target: "gear::authorship", "⚠️  Dropping terminal extrinsic from an overweight block.")
+            rx.await?
+        }.boxed();
+        match futures::future::select(
+            update_block,
+            // TODO: consider adding some tolerance here like `deadline` + 20% or something.
```
breathx, line 600: I'd allow exceeding 15-20%

node/authorship/src/block_builder.rs
```diff
@@ -90,4 +226,96 @@ where
...
+            .api
+            .into_storage_changes(&state, self.parent_hash)
+            .map_err(sp_blockchain::Error::StorageChanges)?;
+
+        Ok(storage_changes)
+    }
+
+    /// Break a builder instance into its parts.
+    #[allow(clippy::type_complexity)]
+    pub fn deconstruct(
```
breathx, line 251:
```suggestion
    pub fn into_parts(
```

node/authorship/src/tests.rs
```diff
@@ -319,109 +414,58 @@ fn proposed_storage_changes_match_execute_block_storage_changes() {
...
+        Box::new(time::Instant::now),
+        timestamp,
+        1500_u64,
         None,
-    ))
-    .unwrap();
-
+        proposal
+    );
     // Terminal extrinsic rolled back, therefore only have 1 inherent + 6 normal
```
breathx, line 464: + 1 for bank?

node/authorship/src/tests.rs
```diff
@@ -702,57 +677,705 @@ fn terminal_extrinsic_discarded_from_txpool() {
...
+    // Assert that the original and the cloned block builders produce same storage changes
+    assert_eq!(
+        changes_1.transaction_storage_root,
+        changes_2.transaction_storage_root
+    );
+    // that are different from what builder created from scratch produces
+    assert_ne!(
+        changes_1.transaction_storage_root,
+        changes_3.transaction_storage_root
+    );
```
breathx, line 782: Construct back and assert eq

node/authorship/src/tests.rs
```diff
@@ -702,57 +677,705 @@ fn terminal_extrinsic_discarded_from_txpool() {
...
+            now,
+            Timestamp::current(),
+            max_duration,
+            None,
+            proposal
+        );
+
+        // then
+        // The block should have 2 extrinsics: the timestamp and one ordinary one.
+        // However, there is a tiny chance the `pseudo-inherent` can make it into the block
```
breathx, line 992: what if we decrease gear run complexity within the test?

node/authorship/src/tests.rs
```diff
@@ -702,57 +677,705 @@ fn terminal_extrinsic_discarded_from_txpool() {
...
+
+    #[test]
+    fn should_not_panic_when_deadline_is_reached() {
+        init_logger();
+
+        init!(client, backend, txpool, spawner, _genesis_hash);
+
+        let block_id = BlockId::number(0);
+        let cell = Mutex::new((false, time::Instant::now()));
+        // The `proposer.now()` that increments the `Instant` by 160s each time it's called
```
breathx, line 1017: and assert the there is not gear run

breathx (approved):

node/authorship/src/authorship.rs
breathx: btw maybe with some warn log if exceeds more than for 5%+ for internal monitoring?

node/authorship/src/authorship.rs
ekovalev: Eventually, got rid of this runtime api call at all by introducing an additional method `push_final` in the `BlockBuilder` which does both things: fetches the pseudo-inherent extrinsic (a.k.a. `Gear::run()`) and tris to apply it inside a single transactional call.

node/authorship/src/authorship.rs
ekovalev: A whole new concept has been introduce to allow this which needs to be re-reviewed.

node/authorship/src/tests.rs
ekovalev: Actually, the last part of this test is exactly about this. Since `BlockBuilder` instances (both reconstructed from parts) can't be tested for equality directly (not being `Eq`), we compare their derivatives - storage changes.
Or do you mean anything else?

breathx (approved):

breathx:
If CI passes, for me - ready to merge https://github.com/gear-tech/gear/actions/runs/7168775177

PR #3566, opened 2023-12-04
feat(pallet-gear-rpc): Introduce gas spent for vouchers

ukint-vs:
Resolves # .

new rpc call:
- `unstable_gear_voucher_calculateGas`

- bump api version to 3
- pass `gas_sponsor` to `calculate_gas_info`

rpc example:
```bash
curl --location '127.0.0.1:9944' \
--header 'Content-Type: application/json' \
--data '{
    "id": 1,
    "jsonrpc": "2.0",
    "method": "unstable_gear_voucher_calculateGas",
    "params": [
        "0x76ff436ab12395ba8380969990ad3b199cd45b5bcffaeb0a04983743f66acb15",
        {
            "Send": "0x61630d9020ba5e376a4a1f156b7b780759d6b43aa36be510af00ae93f3a4f850"
        },
        "0x50494E47",
        0,
        true
    ]
}'
```

@reviewer-or-team


pallets/gear/rpc/src/lib.rs
```diff
@@ -214,13 +237,26 @@ where
...
+                    Some(self.allowance_multiplier),
+                )
             } else {
                 api.calculate_gas_info(
                     at_hash,
                     source,
                     kind,
                     payload,
                     value,
+                    is_prepaid,
```
breathx, line 259: it's wrong approach. try to pass calls maybe. It was discussed before

breathx (changes requested):

pallets/gear/rpc/src/lib.rs
ukint-vs: added VoucherCall enum to pass msg_id or source id

breathx:
Reduntant api after #3653

PR #3567, opened 2023-12-04
refactor(pallets): Introduce `Origin::cast` fn

breathx:


SabaunT (approved):

breathx:
@shamilsan

common/src/lib.rs
```diff
@@ -84,6 +84,9 @@ pub type Gas = u64;
 pub trait Origin: Sized {
     fn into_origin(self) -> H256;
     fn from_origin(val: H256) -> Self;
+    fn cast<T: Origin>(self) -> T {
```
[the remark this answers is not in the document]
breathx: exactly
there is no possibility to impl exactly `Into` and if we name this fn `Origin::into` it will cause logical and compiler misunderstandings

ark0f, line 87: Isn't it just `into()`?

shamilsan: https://github.com/gear-tech/gear/pull/3567#discussion_r1422611930

Issue #3304, opened 2023-09-19
Introduce backtrace for node logs


clearloop:
what about node logs?

_Originally posted by @breathx in https://github.com/gear-tech/gear/pull/3170#discussion_r1328695890_

PR #3573, opened 2023-12-07
feat(gcli): support customized CLI derived from gcli

### LINKED ISSUE #3343: Construct API on demand in gcli
### Problem to Solve

https://github.com/gear-tech/gear/actions/runs/6267044926/job/17038072645?pr=3326

gcli tries to connect API even if it is not required in command `gcli program meta`

https://github.com/gear-tech/gear/blob/master/gcli/src/cmd/program.rs#L81

since it connects to the official RPC when no endpoint is specified, this command will be failed when the default RPC is down

### Possible Solution

wrap the API constructor and construct it on-demand

### Notes

ref #3326

### LINKED ISSUE #3572: Introduce shared interfaces for deriving CLI from gcli
### Problem to Solve

we currently have well designed modular gcli commands, in case users want to build customized cli for their programs, we provide a trait for it

### Possible Solution

```rust
//! on gcli side
pub trait  App: clap::Parser {
   fn exec(&self, api: gsdk::Api) -> Result<()> {
       // ...
   }

   // ...
}
```

```rust
//! in user's binary file

#[derive(Parser)]
enum MySubCommand {
    login: gcil::Login,
    cmd: MyCustomizedCommand,
}

#[derive(Debug, Parser)]
struct MyApp {
    #[clap(subcommand)]
    command: MyCommand,
}

impl gcli::App for MyApp {
    fn exec(&self, api: gsdk::Api) -> Result<()> {
        // ...
    }
}

```

### Notes

_No response_

clearloop:
Resolves #3572 
Resolves #3343

- [x] introduce trait `App` for deriving CLI for programs
- [x] refactor binary gcli with new `App` trait
- [x] example of using the App trait
- [x] construct API on demand

@gear-tech/dev


breathx (approved):
Offtop, but is it already possible to dump stdout from Node process? Seems like a good issue to solve

NikVolf (approved):
marvelous!

clearloop:
> Offtop, but is it already possible to dump stdout from Node process? Seems like a good issue to solve

will do it soon here https://github.com/gear-tech/gear/issues/3304

PR #3581, opened 2023-12-11
refactor(crates-io): use `toml_edit` instead of `cargo_toml`

### LINKED ISSUE #3575: Fix the documentation of gstd on crates.io
### Problem to Solve

docs.rs failed to build the docs of `gstd` build logs: https://docs.rs/crate/gstd/1.0.2/builds/1051841, bcz the `document-features` feature https://github.com/gear-tech/gear/blob/master/gstd/src/lib.rs#L32-L35

### Possible Solution

mb we need to provide

```toml
[features]
document-features = ["dep:document-features"]
```

or checkout the reason from the commands that docs.rs building our docs

### Notes

_No response_

clearloop:
Resolves #3575

after kicking out `cargo_toml`, the patching job could be just wrapped up with 

- adding local packages' versions to the workspace manifest
- rename the version of `sp-arithmetic`

which means, if we `toml_edit` the versions of our local packages on release, change the `sp-arithmetic` in our substrate fork, we don't need most of the logic in this tool on publishing our packages ( benefited from we have published a lot of our forks, `gear-wasm`, etc. during developing this tool), in case of the uncertain changes from the upstream, we can keep this tool until we have a perfect solution for version controling

@gear-tech/dev


utils/crates-io/src/version.rs
```diff
@@ -44,3 +45,20 @@ pub fn verify(name: &str, version: &str) -> Result<bool> {
...
+        .output()
+        .map_err(|e| anyhow!("failed to execute command git, {e}"))?
+        .stdout
+        .iter()
+        .filter_map(|&c| {
+            if c.is_ascii_whitespace() {
+                return None;
+            }
+            Some(c as char)
+        })
```
StackOverflowExcept1on, line 62:
```suggestion
        .filter_map(|&c| c.is_ascii_whitespace().then(c as char))
```

utils/crates-io/src/publisher.rs
```diff
@@ -62,39 +59,44 @@ impl Publisher {
...
             }
 
-            let mut manifest = workspace.manifest(&p.manifest_path)?;
-            rename::package(p, &mut manifest.manifest)?;
-            rename::deps(&mut manifest.manifest, self.index.keys().collect(), version)?;
-
             self.graph
-                .insert(self.index.get(&p.name).cloned(), manifest);
+                .insert(self.index.get(&p.name).cloned(), workspace.manifest(p)?);
         }
```
StackOverflowExcept1on, line 80:
```suggestion
        for package @ cargo_metadata::Package { name, .. } in &self.metadata.packages {
            if !index.contains(&name.as_ref()) {
                continue;
            }

            println!("Verifying {name}@{version} ...");
            if crate::verify(name, &version)? {
                println!("Package {name}@{version} already published .");
                continue;
            }

            self.graph
                .insert(self.index.get(name).cloned(), workspace.manifest(package)?);
        }
 ```

File: utils/crates-io/src/manifest.rs [hunk 3]
```diff
@@ -46,23 +48,141 @@ impl ManifestWithPath {
...
-        manifest
-            .complete_from_path_and_workspace(path.as_ref(), Some((&self.manifest, &self.path)))?;
+    /// the workspace manifest
+    pub fn manifest(&self, pkg: &Package) -> Result<Self> {
+        self.ensure_workspace()?;
+
+        // Complete documentation as from <https://docs.rs>
+        let mut manifest: Document = fs::read_to_string(&pkg.manifest_path)?.parse()?;
+        let name = pkg.name.clone();
+        manifest["package"]["documentation"] = toml_edit::value(format!("https://docs.rs/{name}"));
```
StackOverflowExcept1on, line 65: since `toml_edit` supports `serde` feature, maybe we can declare some `struct` with the contents of `Cargo.toml` and modify it instead of accesing keys?

clearloop: 3f03d09

utils/crates-io/src/publisher.rs
clearloop: 3f03d09

utils/crates-io/src/manifest.rs
StackOverflowExcept1on: iirc serde has flatten attr for that

StackOverflowExcept1on (approved):

utils/crates-io/src/manifest.rs
clearloop: https://github.com/gear-tech/gear/issues/3583

breathx (approved):
Im still confused why do we need comm hash in version. Why can't we just manually trigger it only on releases

clearloop:
> Im still confused why do we need comm hash in version. Why can't we just manually trigger it only on releases

It is for avoiding publishing the current version of our packages by accident ( I used to publish some packages by accident before ), the commit-hash-suffix version is not using in anywhere, just for testing for now

Issue #3588, opened 2023-12-14
Filter dependency features in wasm-builder 


clearloop:
### Problem

Dependency features are enabled unexpectedly in wasm-builder

location: https://github.com/gear-tech/gear/blob/master/utils/wasm-builder/src/lib.rs#L126

### Steps

Based on demo-ping

```toml
dependencies]
gstd.workspace = true
gcli = { workspace = true, optional = true }

[build-dependencies]
gear-wasm-builder.workspace = true

[features]
debug = ["gstd/debug"]
default = ["std"]
std = [ "gcli" ]
```
while compiling `demo-ping`, feature `gcli` will be enabled

### Relevant Log Output

<details><summary>Click to expand/collapse</summary>
<p>

```
<Paste your log here>
```

</p>
</details>

breathx:
you don't need gcli while compiling wasm
demo-ping goes to native when std, and for dependable lib with wasm bin while not

breathx:
I suppose issue contains no bug and is only misuse of features

clearloop:
> you don't need gcli while compiling wasm demo-ping goes to native when std, and for dependable lib with wasm bin while not

https://github.com/gear-tech/gear/issues/3587

like the feature in this issue, building specified CLI for programs, imo the expected behavior of passing features is like

```toml
[dependencies]
# this converts to a dependency feature automatically
foo = { version = "*", optional = true } 
```

expected behavior:

```
cargo build 

# nested
cargo build --target wasm32-unknown-unknown --no-default-features
```

but got

```
cargo build 

# nested
cargo build --target wasm32-unknown-unknown --no-default-features --features foo
```

shamilsan:
Could you please add more details about the issue?

If you add an optional dependency to the program, a generated Wasm-project's manifest will contain a feature:

```toml
[dependencies.orig-project]
default-features = false
package = "example"
path = "/path/to/orig-crate"

[features]
foo = ["orig-project/foo"]

[lib]
crate-type = ["cdylib"]
name = "example"

[package]
edition = "2021"
name = "example-wasm"
version = "0.1.0"

# ...
```

If you build your program with feature `foo`, it will be propagated to the generated crate:

```bash
cargo b -F foo # Will use `foo` dependency
```

If you add `foo`-feature as a default, it will be propagated by default:

```toml
[features]
default = ["foo"]
```

```bash
cargo b # Will use `foo` dependency
```

[...]

PR #3589, opened 2023-12-14
feat(gcli): embed programs in commands

### LINKED ISSUE #3587: Introduce features `embed` for gcli
### Problem to Solve

After #3573, we have introduced a interface that constructing CLI for programs, once we present a CLI based on program, we actually can get the compiled WASM of the program directly from the target folder

### Possible Solution

when `embed` feature enabled, skipping parsing program path from command line arguments but initializing related commands with the built binaries

### Notes

_No response_

### LINKED ISSUE #3592: Override WASM binary in embedded CLI
### Problem to Solve

for command `upload`, https://github.com/gear-tech/gear/pull/3589 currently just overrides the path of the built program for the compatible with the previous API, we should override the binary in the future in case of users renaming the path of the binaries.

### Possible Solution

mb adding a new field `overided_code` in command `Upload`, or wait for the implementation of **the commands for handle inputs**, and do it together in the future

### Notes

_No response_

clearloop:
Resolves #3587 
Resolves #3592 
Ref #3588

- [x] introduce tools for looking up wasm binaries
- [x] introduce features `embed`
- [x] nest binaries for related command if feature `embed` is enabled
- [x] example
- [x] tests

@gear-tech/dev


breathx (approved):

examples/messager/Cargo.toml
```diff
@@ -5,8 +5,14 @@ authors.workspace = true
 license.workspace = true
 edition.workspace = true
 
+[[bin]]
+name = "demo_messager"
```
shamilsan, line 9: Maybe `messenger`?

https://english.stackexchange.com/questions/211365/why-is-messenger-the-term-instead-of-messager

examples/messager/Cargo.toml
```diff
@@ -5,8 +5,14 @@ authors.workspace = true
...
+path = "src/bin/demo_messager.rs"
+required-features = ["gcli"]
+
 [dependencies]
 gstd.workspace = true
+gcli = { workspace = true, optional = true, features = [ "embed" ] }
```
shamilsan, line 15:
```suggestion
gcli = { workspace = true, optional = true, features = ["embed"] }
```

gcli/src/cmd/program.rs
```diff
@@ -60,6 +64,15 @@ pub enum Program {
 }
 
 impl Program {
+    /// Override metadata binary.
+    pub fn override_meta(&self, meta: Vec<u8>) -> Self {
```
shamilsan, line 68: Maybe we can reflect that we clone `self` in the function name.
```suggestion
    pub fn clone_with_meta_override(&self, meta: Vec<u8>) -> Self {
```

gcli/src/embed.rs
```diff
@@ -0,0 +1,89 @@
...
+            } else {
+                "release"
+            },
+            ancestors.next()?.to_str()?,
+        ];
+
+        let [bin, stem] = [
+            PathBuf::from(format!("{target}/wasm32-unknown-unknown/{profile}")),
+            PathBuf::from(name.replace('-', "_")),
+        ];
```
shamilsan, line 80: Why array with destructuring?
```suggestion
        let bin = PathBuf::from(format!("{target}/wasm32-unknown-unknown/{profile}"));
        let stem = PathBuf::from(name.replace('-', "_"));
```

gcli/src/embed.rs
```diff
@@ -0,0 +1,89 @@
...
+
+        let [opt, meta] = [
+            fs::read(bin.join(stem.with_extension("wasm"))).ok()?,
+            fs::read(bin.join(stem.with_extension("meta.wasm"))).unwrap_or_default(),
+        ];
+
+        Some(Self { opt, meta })
```
shamilsan, line 87: Also,
```suggestion
        Some(Self {
            opt: fs::read(bin.join(stem.with_extension("wasm"))).ok()?,
            meta: fs::read(bin.join(stem.with_extension("meta.wasm"))).unwrap_or_default(),
        })
```

shamilsan (approved):

gcli/src/cmd/program.rs
clearloop: makes sense, and this method should be used by users in common cases

examples/messager/Cargo.toml
clearloop: https://github.com/gear-tech/gear/issues/3621 will do this in another PR since this related to workspace changes

PR #3597, opened 2023-12-15
refactor(numerated): make refactoring for crate numerated

grishasobol:
Make refactoring to make code more suitable for @ark0f and @breathx
1) rename NonEmptyInterval -> Interval
2) rename Interval -> IntervalIterator
3) move all iterators to file iterators.rs
4) remove BoundValue (it's too difficult as live shows)


ukint-vs (approved):

common/numerated/src/interval.rs
```diff
@@ -16,57 +16,39 @@
...
+    /// Safe, if `start ≤ end`.
     #[track_caller]
     pub unsafe fn new_unchecked(start: T, end: T) -> Self {
-        debug_assert!(start <= end);
+        debug_assert!(
+            start <= end,
+            "Calling this method you must guarantee that `start ≤ end`"
+        );
         Self { start, end }
     }
```
breathx, line 54: https://github.com/gear-tech/gear/blob/1128b5cc172e6fe004eded3a34d47782f304f849/common/numerated/src/interval.rs#L56
`try_new() -> Result<Self>`

common/numerated/src/interval.rs
```diff
@@ -85,352 +67,251 @@ impl<T: Numerated> NonEmptyInterval<T> {
...
-    type Error = IntoNonEmptyIntervalError;
-
-    fn try_from(interval: &Interval<T>) -> Result<Self, Self::Error> {
-        interval
-            .end
-            .map(|end| unsafe {
-                // Guaranteed by `Self` that start ≤ end
-                NonEmptyInterval::new_unchecked(interval.start, end)
+    /// Returns new [Interval] with `start` = `start` + 1, if it's possible.
+    pub fn inc_start(&self) -> Option<Self> {
```
breathx, line 81: `inc_start(&mut self) -> Result<()>`
or
`with_increased_start(self) -> Option<Self>`

common/numerated/src/interval.rs
```diff
@@ -85,352 +67,251 @@ impl<T: Numerated> NonEmptyInterval<T> {
...
+        start.inc_if_lt(end).map(|start| {
+            debug_assert!(start <= end, "`T: Numerated` impl error");
+            Interval { start, end }
+        })
+    }
+
+    /// Trying to make [Interval] from `range`.
+    /// - If `range.start > range.end`, then returns [IncorrectRangeError].
+    /// - If `range.start == range.end`, then returns [EmptyRangeError].
+    pub fn try_from_range(range: Range<T>) -> Result<Self, IncorrectOrEmptyRangeError> {
```
breathx, line 93:
```suggestion
    pub fn try_from_range(range: Range<T>) -> Result<Self, TryFromRangeError> {
```

common/numerated/src/interval.rs
```diff
@@ -85,352 +67,251 @@ impl<T: Numerated> NonEmptyInterval<T> {
...
-        self.end.is_none()
-    }
-    /// Tries to convert into (start, end).
-    pub fn into_inner(self) -> Option<(T, T)> {
-        NonEmptyInterval::try_from(self).ok().map(Into::into)
-    }
-    /// Tries to convert into range inclusive.
-    pub fn into_range_inclusive(self) -> Option<RangeInclusive<T>> {
-        RangeInclusive::try_from(self).ok()
+impl<T: Numerated> PartialEq<RangeInclusive<T>> for Interval<T> {
```
breathx, line 108: compare by refs

common/numerated/src/interval.rs
```diff
@@ -85,352 +67,251 @@ impl<T: Numerated> NonEmptyInterval<T> {
...
-
-    fn try_from((start, end): (S, E)) -> Result<Self, Self::Error> {
-        use BoundValue::*;
+/// Trying to make interval from range where start > end.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct IncorrectRangeError;
+
+/// Trying to make interval from range where start > end or empty range.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, derive_more::From)]
+pub enum IncorrectOrEmptyRangeError {
```
breathx, line 161: Remove underlying structs in favour of just enum variants

common/numerated/src/interval.rs
```diff
@@ -85,352 +67,251 @@ impl<T: Numerated> NonEmptyInterval<T> {
...
-                    Ok(Self {
-                        start,
-                        end: Some(end),
-                    })
-                } else if start == end {
-                    Ok(Self { start, end: None })
-                } else {
-                    Err(IntoIntervalError)
-                }
+    fn try_from(range: RangeFrom<I>) -> Result<Self, Self::Error> {
```
breathx, line 171: Consider simplifying check in 1 fn (that it's not max value) instead of bound and following unbound.

common/numerated/src/interval.rs
```diff
@@ -85,352 +67,251 @@ impl<T: Numerated> NonEmptyInterval<T> {
...
+        match Into::<T::Bound>::into(range.start).unbound() {
```
breathx, line 172:
```suggestion
        match T::Bound::from(range.start).unbound() {
```

common/numerated/src/interval.rs
```diff
@@ -85,352 +67,251 @@ impl<T: Numerated> NonEmptyInterval<T> {
...
+    for Interval<T>
+{
+    type Error = EmptyRangeError;
 
-    fn try_from(range: Range<I>) -> Result<Self, Self::Error> {
-        Self::try_from((range.start, range.end))
+    fn try_from(range: RangeTo<I>) -> Result<Self, Self::Error> {
+        let end: T::Bound = range.end.into();
+
+        let Some(end) = end.unbound() else {
```
breathx, line 191:
```suggestion
        let Some(end) = T::Bound::from(range.end).unbound() else {
```

common/numerated/src/interval.rs
```diff
@@ -85,352 +67,251 @@ impl<T: Numerated> NonEmptyInterval<T> {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct OutOfBoundsError;
+
+/// Trying to make zero len or out of bounds interval.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, derive_more::From)]
+pub enum NewWithLenError {
+    /// Trying to make zero len interval.
+    ZeroLen,
+    /// Trying to make out of bounds interval.
+    OutOfBounds(OutOfBoundsError),
```
breathx, line 255: Same here
```suggestion
    OutOfBounds,
```

common/numerated/src/iterators.rs
```diff
@@ -0,0 +1,330 @@
...
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//! [IntervalIterator], [VoidsIterator], [DifferenceIterator] implementations.
```
breathx, line 19: Doc comments across the whole crate
```suggestion
//! [`IntervalIterator`], [`VoidsIterator`], [`DifferenceIterator`] implementations.
```

common/numerated/src/iterators.rs
```diff
@@ -0,0 +1,330 @@
...
+    T: Numerated + UpperBounded,
+    S: Into<T::Bound>,
+    E: Into<T::Bound>,
+{
+    type Error = IncorrectRangeError;
+
+    fn try_from(range: (S, E)) -> Result<Self, Self::Error> {
+        match Interval::try_from(range) {
+            Ok(interval) => Ok(interval.into()),
+            Err(IncorrectOrEmptyRangeError::EmptyRange(_)) => Ok(Self(None)),
```
breathx, line 105: add cast of range errors to interval iterator for all of these constructors

common/numerated/src/iterators.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+}
+
+impl<T: Numerated> Iterator for IntervalIterator<T> {
+    type Item = T;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.0.map(|interval| {
+            let start = interval.start();
+            self.0 = interval.inc_start();
```
breathx, line 133: mut access as stated above

common/numerated/src/iterators.rs
```diff
@@ -0,0 +1,330 @@
...
+    pub(crate) iter1: I,
+    pub(crate) iter2: I,
+    pub(crate) interval1: Option<Interval<T>>,
+    pub(crate) interval2: Option<Interval<T>>,
+}
+
+impl<T: Numerated, I: Iterator<Item = Interval<T>>> Iterator for DifferenceIterator<T, I> {
+    type Item = Interval<T>;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
breathx, line 179: Add comments on algo

common/numerated/src/iterators.rs
```diff
@@ -0,0 +1,330 @@
...
+        loop {
+            let interval1 = if let Some(interval1) = self.interval1 {
+                interval1
+            } else {
+                let interval1 = self.iter1.next()?;
+                self.interval1 = Some(interval1);
+                interval1
+            };
```
breathx, line 187: consider similar changes below
```suggestion
            self.interval1 = self.interval1.or_else(|| self.iter1.next());
            let interval1 = self.interval1?;
```

common/numerated/src/iterators.rs
```diff
@@ -0,0 +1,330 @@
...
+
+            let interval2 = if let Some(interval2) = self.interval2 {
```
breathx, line 189: matching seems more readable here

common/numerated/src/iterators.rs
```diff
@@ -0,0 +1,330 @@
...
+                interval2
+            } else if let Some(interval2) = self.iter2.next() {
+                interval2
+            } else {
+                self.interval1 = None;
+                return Some(interval1);
```
breathx, line 195: Same below
```suggestion
                return self.interval1.take();
```

common/numerated/src/iterators.rs
```diff
@@ -0,0 +1,330 @@
...
+                continue;
+            }
+
+            self.interval2 = Some(interval2);
+
+            if interval1.end() < interval2.start() {
+                self.interval1 = None;
+                return Some(interval1);
+            } else {
+                if let Some(new_start) = interval2.end().inc_if_lt(interval1.end()) {
```
breathx, line 209: for these algos you can also introduce intervals api like: `interval1.difference(interval2)`;
consider `interval1.cross(interval2) -> bool` or maybe `.meet()`/`.touches()`;
otherwise you have A LOT of repeated code that's not readable, while task is pretty obvious.

common/numerated/src/iterators.rs
```diff
@@ -0,0 +1,330 @@
...
+            }
+        }
+    }
+}
+
+/// Helper struct to iterate over voids in tree.
+///
+/// See also [`IntervalsTree::voids`].
+pub struct VoidsIterator<T: Numerated, I: Iterator<Item = Interval<T>>> {
+    pub(crate) inner: Option<(I, Interval<T>)>,
```
breathx, line 235: Shouldn't it be different Options or even only optional interval?

common/numerated/src/iterators.rs
```diff
@@ -0,0 +1,330 @@
...
+}
+
+impl<T: Numerated, I: Iterator<Item = Interval<T>>> Iterator for VoidsIterator<T, I> {
+    type Item = Interval<T>;
+
+    fn next(&mut self) -> Option<Self::Item> {
```
breathx, line 241: add comments on algo

common/numerated/src/iterators.rs
```diff
@@ -0,0 +1,330 @@
...
+        let (iter, interval) = self.inner.as_mut()?;
+        if let Some(next) = iter.next() {
+            let (start, end) = next.into_parts();
+
+            // Guaranties by tree: between two intervals always exists void.
+            debug_assert!(interval.start() < start);
+
+            let Some(void_end) = start.dec_if_gt(interval.start()) else {
+                debug_assert!(false, "`T: Numerated` impl error");
```
breathx, line 250: shouldn't we panic in prod for such cases?

common/numerated/src/iterators.rs
```diff
@@ -0,0 +1,330 @@
...
+        assert!(IntervalIterator::<u8>::new_with_len(None, 0)
+            .unwrap()
+            .is_empty());
+        assert!(IntervalIterator::<u8>::try_from(1..1).unwrap().is_empty());
+        assert!(IntervalIterator::<u8>::try_from(..0).unwrap().is_empty());
+        assert!(IntervalIterator::<u8>::try_from(None..).unwrap().is_empty());
+    }
+
+    #[test]
+    fn voids_iterator() {
```
breathx, line 297: this task could be solved by difference iterator, isn't it?, so voids iterator is kinda useless

common/numerated/src/lib.rs
```diff
@@ -16,28 +16,29 @@
 // You should have received a copy of the GNU General Public License
```
breathx, line 16: General thought on intervals and impls for iterators: Try to work with refs almost everything in order to keep it optimized.

common/numerated/src/numerated.rs
```diff
@@ -16,22 +16,11 @@
...
-pub enum BoundValue<T: Sized> {
-    /// The bound is a value. Contains `T` value.
-    Value(T),
-    /// The bound is an upper bound. Contains `T` max value.
-    Upper(T),
-}
+use core::cmp::Ordering;
+use num_traits::{One, PrimInt, Unsigned};
 
 /// For any type `T`, `Bound<T>` is a type, which has set of values bigger than `T` by one element.
```
breathx, line 25:
```suggestion
/// For any type `T`, `Bound<T>` is a type, which has set of values bigger than `T` by one.
```

common/numerated/src/numerated.rs
```diff
@@ -16,22 +16,11 @@
...
 /// - Each value from `T` has unambiguous mapping to `Bound<T>`.
```
breathx, line 26: Should we consider specific enum or struct for `Bound` to force it being a type instead of some strange trait?

common/numerated/src/tree.rs
```diff
@@ -77,10 +77,10 @@ use scale_info::{
...
-/// let other_tree: IntervalsTree<i32> = [3, 4, 5, 7, 8, 9].iter().collect();
-/// let and_not: Vec<_> = tree.and_not_iter(&other_tree).map(RangeInclusive::from).collect();
-/// assert_eq!(and_not, vec![1..=2, 6..=6]);
+/// // Difference iterator: iterate over intervals from `tree` which are not in `other_tree`.
+/// let other_tree: IntervalsTree<i32> = [3, 4, 5, 7, 8, 9].into_iter().collect();
+/// let difference: Vec<_> = tree.difference(&other_tree).map(RangeInclusive::from).collect();
+/// assert_eq!(difference, vec![1..=2, 6..=6]);
 /// ```
 ///
 /// # Possible panic cases
```
breathx, line 86: https://github.com/gear-tech/gear/blob/1128b5cc172e6fe004eded3a34d47782f304f849/common/numerated/src/tree.rs#L91
Add trait bound `T: Numerated`

breathx: Same for interval

common/numerated/src/tree.rs
```diff
@@ -124,31 +124,29 @@ impl<T: Copy> Default for IntervalsTree<T> {
...
-        )
+        f.debug_list().entries(self.iter()).finish()
     }
 }
 
 impl<T: Numerated> IntervalsTree<T> {
-    fn into_start_end<I: Into<Interval<T>>>(interval: I) -> Option<(T, T)> {
-        Into::<Interval<T>>::into(interval).into_inner()
+    fn into_start_end<I: Into<IntervalIterator<T>>>(interval: I) -> Option<(T, T)> {
+        Into::<IntervalIterator<T>>::into(interval)
```
breathx, line 135:
```suggestion
        IntervalIterator::<T>::from(interval)
```

common/numerated/src/tree.rs
```diff
@@ -124,31 +124,29 @@ impl<T: Copy> Default for IntervalsTree<T> {
...
+            Interval::<T>::new_unchecked(start, end)
         })
     }
 
     /// Returns true if for each `p` ∈ `interval` ⇒ `p` ∈ `self`, otherwise returns false.
-    pub fn contains<I: Into<Interval<T>>>(&self, interval: I) -> bool {
+    pub fn contains<I: Into<IntervalIterator<T>>>(&self, interval: I) -> bool {
         let Some((start, end)) = Self::into_start_end(interval) else {
             // Empty interval is always contained.
             return true;
```
breathx, line 152: https://github.com/gear-tech/gear/blob/1128b5cc172e6fe004eded3a34d47782f304f849/common/numerated/src/tree.rs#L154-L159
```rust
self.inner.range(..=end).next_back().map(|(s, e)| s <= start && e >= end).ok_or(false)
```

common/numerated/src/tree.rs
```diff
@@ -203,6 +200,7 @@ impl<T: Numerated> IntervalsTree<T> {
 
         if let Some(right_end) = right_end.inc_if_lt(start) {
```
breathx, line 201: comment all of the cases pls for whole algo with points alignment in docs like `start < right_start <= right_end < end`

common/numerated/src/tree.rs
```diff
@@ -203,6 +200,7 @@ impl<T: Numerated> IntervalsTree<T> {
...
             if right_end == start {
+                debug_assert!(right_start <= end, "Must be cause of method it was found");
                 self.inner.insert(right_start, end);
             } else {
```
breathx, line 205: btw couldn't remove/insert fns be optimized with bin search mb ?

breathx: case of tree like 1, 3, 5, 7, 9... etc

common/numerated/src/tree.rs
```diff
@@ -236,16 +234,16 @@ impl<T: Numerated> IntervalsTree<T> {
         let end = right_end.max(end);
 
         let Some((left_start, left_end)) = left_interval else {
+            debug_assert!(start <= end, "Must be cause of method it was found");
```
breathx, line 237: if we won't place panics here , than this check seems excess

common/numerated/src/tree.rs
```diff
@@ -236,16 +234,16 @@ impl<T: Numerated> IntervalsTree<T> {
...
         debug_assert!(left_start <= start);
         let Some(left_end) = left_end.inc_if_lt(right_start) else {
-            unreachable!(
-                "`T: Numerated` impl error: for each x: T, y: T, x < y ↔ x.inc_if_lt(y) == Some(_)"
-            );
+            debug_assert!(false, "`T: Numerated` impl error");
+            return;
         };
 
         if left_end >= start {
```
breathx, line 249: `self.inner.insert(left_start.max(start), end)`
maybe add specific private fn like `self.put(.., ..)` with internal debug assert and track caller

common/numerated/src/tree.rs
```diff
@@ -255,21 +253,24 @@ impl<T: Numerated> IntervalsTree<T> {
...
     /// Remove `interval` from tree.
     /// - if `interval` is empty, then nothing will be removed.
-    /// - if `interval` is not empty, then after removing for each `p` ∈ `interval` ⇒ `p` ∉ `self`.
+    /// - if `interval` is not empty, then after removing: for each `p` ∈ `interval` ⇒ `p` ∉ `self`.
     ///
     /// Complexity: `O(log(n) + m)`, where
     /// - `n` is amount of intervals in `self`
     /// - `m` is amount of intervals in `self` ⋂ `interval`
-    pub fn remove<I: Into<Interval<T>>>(&mut self, interval: I) {
+    pub fn remove<I: Into<IntervalIterator<T>>>(&mut self, interval: I) {
```
breathx, line 273: consider returning bool or rename to `clear() -> ()` mb

common/numerated/src/tree.rs
```diff
@@ -255,21 +253,24 @@ impl<T: Numerated> IntervalsTree<T> {
...
         let Some((start, end)) = Self::into_start_end(interval) else {
             // Empty interval - nothing to remove.
             return;
```
breathx, line 276: https://github.com/gear-tech/gear/blob/1128b5cc172e6fe004eded3a34d47782f304f849/common/numerated/src/tree.rs#L285-L287
its some kind of shortcut/fast-check, consider adding the same from `iter.start()`

breathx: https://github.com/gear-tech/gear/blob/1128b5cc172e6fe004eded3a34d47782f304f849/common/numerated/src/tree.rs#L290-L291
maybe you could `.fold()` it and similar places?

breathx: https://github.com/gear-tech/gear/blob/1128b5cc172e6fe004eded3a34d47782f304f849/common/numerated/src/tree.rs#L292-L296
wouldn't it be much easier and better introducing Ord for interval and fn like `interval1.contains(interval2)`?

common/numerated/src/tree.rs
```diff
@@ -405,10 +406,10 @@ impl<T: Numerated> IntervalsTree<T> {
     /// Number of points in tree set.
     ///
     /// Complexity: `O(n)`, where `n` is amount of intervals in `self`.
-    pub fn points_amount(&self) -> Option<T::N> {
-        let mut res = T::N::zero();
+    pub fn points_amount(&self) -> Option<T::Distance> {
```
breathx, line 409: If `Bound` was a type it would be much logical to return `Bound<T::Distance>` here

common/numerated/src/tree.rs
```diff
@@ -417,7 +418,7 @@ impl<T: Numerated> IntervalsTree<T> {
     pub fn points_iter(&self) -> impl Iterator<Item = T> + '_ {
```
breathx, line 418: de-facto you copy everything, isn't it better to wrap map iterator and expand point lazy: on `.next()` called?

common/numerated/src/tree.rs
```diff
@@ -427,131 +428,7 @@ impl<T: Numerated> IntervalsTree<T> {
...
-            }
-        }
-    }
-}
-
-impl<T: Numerated, D: Into<Interval<T>>> FromIterator<D> for IntervalsTree<T> {
+impl<T: Numerated, D: Into<IntervalIterator<T>>> FromIterator<D> for IntervalsTree<T> {
     fn from_iter<I: IntoIterator<Item = D>>(iter: I) -> Self {
         let mut tree = Self::new();
         for interval in iter {
```
breathx, line 434: https://github.com/gear-tech/gear/blob/1128b5cc172e6fe004eded3a34d47782f304f849/common/numerated/src/tree.rs#L448
isn't it test for `.insert()` instead of `.try_insert()`?
I guess same for test `remove`

breathx: https://github.com/gear-tech/gear/blob/1128b5cc172e6fe004eded3a34d47782f304f849/common/numerated/src/tree.rs#L625
it would be good to have more err-cases tests with asserting errors

breathx (changes requested):

common/numerated/src/tree.rs
ark0f: It will to force to write `T: Numerated` for every impl of each corresponding type

common/numerated/src/tree.rs
ark0f: Agreed. I'd remove these `try_` methods at all because their semantic is misleading as insertion, removal (and so on) are actually infallible

common/numerated/src/tree.rs
ark0f: Consider `is_subset` for interval

common/numerated/src/tree.rs
```diff
@@ -342,10 +346,10 @@ impl<T: Numerated> IntervalsTree<T> {
     /// Iterating complexity: `O(log(n) + m)`, where
     /// - `n` is amount of intervals in `self`
     /// - `m` is amount of intervals in `self` ⋂ `interval`
-    pub fn voids<I: Into<Interval<T>>>(
+    pub fn voids<I: Into<IntervalIterator<T>>>(
```
ark0f, line 349:
```suggestion
    pub fn gaps<I: Into<IntervalIterator<T>>>(
```

common/numerated/src/iterators.rs
ark0f: Consider `is_subset` like in other libraries

common/numerated/src/numerated.rs
```diff
@@ -90,81 +68,134 @@ pub trait Bound<T: Sized>: From<T> + Copy {
...
     /// Returns `true`, if `self` is enclosed by `a` and `b`.
     fn enclosed_by(self, a: &Self, b: &Self) -> bool {
         self <= *a.max(b) && self >= *a.min(b)
     }
 }
 
-impl<T> From<T> for BoundValue<T> {
+/// Bound type for `Option<T>`.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, derive_more::From)]
+pub struct OptionBound<T>(Option<T>);
```
ark0f, line 114: Almost every of the following impls can be just derived

common/numerated/src/tree.rs
ark0f: > maybe you could .fold() it and similar places?

I'd leave it as is for simplicity

common/numerated/src/tree.rs
ark0f: It looks like this check should not be here at all

common/numerated/src/tree.rs
```diff
@@ -161,20 +159,23 @@ impl<T: Numerated> IntervalsTree<T> {
...
+    /// The same as [`Self::contains`], but returns [`I::Error`] if `try_into` [IntervalIterator] fails.
+    pub fn try_contains<I: TryInto<IntervalIterator<T>>>(
+        &self,
+        interval: I,
+    ) -> Result<bool, I::Error> {
+        let interval: IntervalIterator<T> = interval.try_into()?;
         Ok(self.contains(interval))
     }
 
     /// Insert interval into tree.
```
ark0f, line 171: It's maybe good to describe interval merging and splitting in `insert` and `remove` functions becasue iterators behavior becomes unobvious

ark0f (changes requested):

common/numerated/src/interval.rs
```diff
@@ -85,352 +67,251 @@ impl<T: Numerated> NonEmptyInterval<T> {
...
-            NonEmptyInterval::new_unchecked(point, point).into()
-        }
-    }
-}
-
-impl<T: Numerated> From<&T> for Interval<T> {
-    fn from(point: &T) -> Self {
-        Self::from(*point)
+        let (start, end) = (point, point);
+        debug_assert!(start <= end, "`T: Ord` impl error");
```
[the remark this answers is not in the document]
grishasobol: Yes it tests Ord implementation as you can see in assertion message

common/numerated/src/lib.rs
breathx: > Numerated is required to be Copy

oh, missed it, and than closing comment about eq, partialeq, but **this** comment is more about intervals, iterators etc

common/numerated/src/interval.rs
breathx: > No need for result for such simple action. `with_` usually used for constructor or builder pattern

why don't you build it here

ark0f: Because type itself is not builder

common/numerated/src/interval.rs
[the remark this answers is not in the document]
ark0f: I'm not sure we should tests someone's `Ord` implementation

common/numerated/src/iterators.rs
breathx: Difference iterator, according to docs is "Helper struct to iterate over intervals from `tree1`, which are not in `tree2`.", in fact this should be "iterator of intervals" instead of tree1/2
BUT
voids iterator is difference of [..] interval and a given one, isn't it?

common/numerated/src/interval.rs
grishasobol: Sorry, I cannot understand your suggestion.

common/numerated/src/interval.rs
grishasobol: agree with @ark0f

common/numerated/src/interval.rs
grishasobol: agree with @ark0f

common/numerated/src/lib.rs
grishasobol: May be it's a right point in common situations, but here I decide to simplify implementation code. Also there cannot be any performance decrease when structure is consist of 2 or even 3 copyable elements like Interval and IntervalIterator.

common/numerated/src/numerated.rs
grishasobol: No, it must be trait to allow user to choose his own bound type. Also you can always use OptionBound, if you do not want to implement yours

common/numerated/src/interval.rs
grishasobol: Cannot be done because T::Bound  does not implement  From<I>

common/numerated/src/interval.rs
grishasobol: Cannot be done because T::Bound does not implement From

common/numerated/src/iterators.rs
grishasobol: We need to make 3 impl for EmptyRangeError, for IncorrectOrEmptyRangeError, for NewWithLenError, and use them in 4 places. Are you sure we have to do this ?

common/numerated/src/iterators.rs
grishasobol: Sorry, I cannot understand your suggestion.

common/numerated/src/iterators.rs
grishasobol: I cannot see any more readable possible solutions. If you have some more specific ideas I would be glad to consider them

common/numerated/src/iterators.rs
grishasobol: What the reason to make not like this ? It's the most simple solution

common/numerated/src/iterators.rs
grishasobol: We, can panic. In rust core you can find both ways when some `T` implementation is incorrect:
1) returns wrong result
2) panics
3) sometimes also return error in result (but it's rare case)

I choose finally in numerated `wrong result` in order to make probability of panics on migration as less as it's possible.

common/numerated/src/interval.rs
grishasobol: done

common/numerated/src/interval.rs
grishasobol: done

common/numerated/src/interval.rs
grishasobol: done

common/numerated/src/iterators.rs
grishasobol: done

common/numerated/src/iterators.rs
grishasobol: done

common/numerated/src/iterators.rs
grishasobol: done

common/numerated/src/iterators.rs
grishasobol: done

common/numerated/src/iterators.rs
grishasobol: done

common/numerated/src/iterators.rs
grishasobol: done

common/numerated/src/interval.rs
grishasobol: doen

common/numerated/src/numerated.rs
grishasobol: done

common/numerated/src/tree.rs
grishasobol: What the reason?

common/numerated/src/tree.rs
grishasobol: impossible

common/numerated/src/tree.rs
grishasobol: done

common/numerated/src/tree.rs
grishasobol: it's already optimised with bin search. see BTreeMap::range

common/numerated/src/tree.rs
grishasobol: removed and use put

common/numerated/src/tree.rs
grishasobol: nice, done

common/numerated/src/tree.rs
grishasobol: What should we return than?
1) at least one point from `self` has been removed
2) each point from `interval` has been removed from `self`

common/numerated/src/tree.rs
grishasobol: > https://github.com/gear-tech/gear/blob/1128b5cc172e6fe004eded3a34d47782f304f849/common/numerated/src/tree.rs#L285-L287
> 
> 
> its some kind of shortcut/fast-check, consider adding the same from `iter.start()`

What is `iter.start()`?

grishasobol: > https://github.com/gear-tech/gear/blob/1128b5cc172e6fe004eded3a34d47782f304f849/common/numerated/src/tree.rs#L292-L296
> 
> 
> wouldn't it be much easier and better introducing Ord for interval and fn like `interval1.contains(interval2)`?

How exactly you want this contains must work `interval1.contains(interval2)` ?  in this code fragment is not a check for `contains` it is partially intersection. Also we use both bool statements: one for break and second for `left_interval` initialisation

common/numerated/src/tree.rs
grishasobol: `T::Distance` does not guarantied to have Bound. `T` has  bound, but it cannot be returned because then we need restrictions `T: UpperBound + LowerBound`

common/numerated/src/tree.rs
grishasobol: > de-facto you copy everything
would be grate to see some proofs

common/numerated/src/tests.rs
```diff
@@ -19,72 +19,107 @@
...
 };
 
-fn rand_interval() -> impl Strategy<Value = Interval<i16>> {
-    any::<i16>()
-        .prop_flat_map(|start| (start..).prop_map(move |end| (start..=end).try_into().unwrap()))
-}
+macro_rules! any_numerated {
+    ($t:ty) => {
+        fn rand_interval() -> impl Strategy<Value = IntervalIterator<$t>> {
+            any::<$t>().prop_flat_map(|start| {
```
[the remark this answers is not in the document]
grishasobol: ok, changed

common/numerated/src/tree.rs
grishasobol: Ok, I'll thinking. May be better to remove them

common/numerated/src/interval.rs
```diff
@@ -85,352 +67,251 @@ impl<T: Numerated> NonEmptyInterval<T> {
...
+pub struct EmptyRangeError;
 
-impl<T: Numerated, S: Into<T::B>, E: Into<T::B>> TryFrom<(S, E)> for Interval<T> {
-    type Error = IntoIntervalError;
-
-    fn try_from((start, end): (S, E)) -> Result<Self, Self::Error> {
-        use BoundValue::*;
+/// Trying to make interval from range where start > end.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct IncorrectRangeError;
```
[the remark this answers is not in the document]
grishasobol: What?

common/numerated/src/tree.rs
```diff
@@ -124,31 +124,29 @@ impl<T: Copy> Default for IntervalsTree<T> {
...
         self.inner.iter().map(|(&start, &end)| unsafe {
             // Safe, because `Self` guaranties, that inner contains only `start` ≤ `end`.
-            NonEmptyInterval::<T>::new_unchecked(start, end)
+            Interval::<T>::new_unchecked(start, end)
         })
     }
 
     /// Returns true if for each `p` ∈ `interval` ⇒ `p` ∈ `self`, otherwise returns false.
-    pub fn contains<I: Into<Interval<T>>>(&self, interval: I) -> bool {
+    pub fn contains<I: Into<IntervalIterator<T>>>(&self, interval: I) -> bool {
```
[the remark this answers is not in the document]
grishasobol: What the reason?

common/numerated/src/tree.rs
```diff
@@ -236,16 +234,16 @@ impl<T: Numerated> IntervalsTree<T> {
...
             return;
         };
 
         debug_assert!(left_end < right_start);
         debug_assert!(left_start <= start);
         let Some(left_end) = left_end.inc_if_lt(right_start) else {
-            unreachable!(
-                "`T: Numerated` impl error: for each x: T, y: T, x < y ↔ x.inc_if_lt(y) == Some(_)"
-            );
+            debug_assert!(false, "`T: Numerated` impl error");
```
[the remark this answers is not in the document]
grishasobol: No, I choose this way in order to decrease probability of possible panics on migrations. Although it's common way in core collections how to work when impl of Ord, Eq, Copy, Clone  and etc are incorrect: they just work incorrect

common/numerated/src/interval.rs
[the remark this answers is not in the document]
grishasobol: impossible: conflicting implementations

common/numerated/src/interval.rs
```diff
@@ -85,352 +67,251 @@ impl<T: Numerated> NonEmptyInterval<T> {
...
+        start: S,
+        len: L,
+    ) -> Result<Interval<T>, NewWithLenError> {
+        let start: T::Bound = start.into();
+        let len: Option<T::Distance> = len.into();
+        match (start.unbound(), len) {
+            (_, Some(len)) if len.is_zero() => Err(NewWithLenError::ZeroLen),
+            (None, _) => Err(OutOfBoundsError.into()),
+            (Some(start), len) => {
+                // subtraction `len - 1` is safe, because `len != 0`
```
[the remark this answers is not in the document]
grishasobol: len is unsigned

common/numerated/src/iterators.rs
```diff
@@ -0,0 +1,330 @@
...
+}
+
+/// Helper struct to iterate over intervals from `tree1`, which are not in `tree2`.
+///
+/// See also [`IntervalsTree::difference`].
+pub struct DifferenceIterator<T: Numerated, I: Iterator<Item = Interval<T>>> {
+    pub(crate) iter1: I,
+    pub(crate) iter2: I,
+    pub(crate) interval1: Option<Interval<T>>,
+    pub(crate) interval2: Option<Interval<T>>,
```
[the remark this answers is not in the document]
grishasobol: cannot agree with this naming

common/numerated/src/iterators.rs
grishasobol: resolve and continue to discuss in @ark0f comment below

common/numerated/src/tree.rs
```diff
@@ -364,37 +368,34 @@ impl<T: Numerated> IntervalsTree<T> {
```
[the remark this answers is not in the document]
grishasobol: Wow, so cool suggestion!! I'll be glad to see proof of concept

common/numerated/src/tree.rs
grishasobol: Cannot see reasons to rename

common/numerated/src/numerated.rs
grishasobol: proof of concept

common/numerated/src/interval.rs
```diff
@@ -85,352 +67,251 @@ impl<T: Numerated> NonEmptyInterval<T> {
...
-                let c = c.map(|c| c - T::N::one()).unwrap_or(T::N::max_value());
-                s.add_if_enclosed_by(c, T::max_value())
-                    .map(|e| NonEmptyInterval::new(s, e).unwrap_or_else(|| {
-                        unreachable!("`T: Numerated` impl error: for each s: T, c: T::N ⇔ s.add_if_between(c, _) ≥ s")
-                    }).into())
+    /// Returns interval `start..=start + len - 1` if it's possible.
+    /// - if `len == None`, then it is supposed, that `len == T::Distance::max_value() + 1`.
+    /// - if `start + len - 1` is out of `T`, then returns [NewWithLenError::OutOfBounds].
+    /// - if `len == 0`, then returns [NewWithLenError::ZeroLen].
+    pub fn new_with_len<S: Into<T::Bound>, L: Into<Option<T::Distance>>>(
```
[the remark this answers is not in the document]
grishasobol: done

common/numerated/src/interval.rs
```diff
@@ -85,352 +67,251 @@ impl<T: Numerated> NonEmptyInterval<T> {
...
-impl<T: Display> Display for NonEmptyInterval<T> {
-    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
-        write!(f, "({}..={})", self.start, self.end)
+impl<T: Numerated + LowerBounded + UpperBounded> Interval<T> {
+    /// Returns `len: T::Distance` (amount of points in `self`) converting it to `T` point:
+    /// - If `len` is bigger than `T` possible elements amount, then returns `T::Bound` __upper__ value.
+    /// - Else returns for `len` corresponding by numeration `p: T::Bound`:
+    /// ```text
+    ///   { 1 -> T::Bound::from(T::min_value() + 1), 2 -> T::Bound::from(T::min_value() + 2), ... }
+    /// ```
```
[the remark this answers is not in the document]
grishasobol: change comment

common/numerated/src/iterators.rs
```diff
@@ -0,0 +1,330 @@
...
+        })
+    }
+}
+
+impl<T: Numerated + UpperBounded> IntervalIterator<T> {
+    /// Returns interval `start..start + len` if it's possible.
+    /// - if `len == None`, then it is supposed, that `len == T::Distance::max_value() + 1`.
+    /// - if `start + len - 1` is out of `T`, then returns [OutOfBoundsError].
+    /// - if `len` is zero, then returns empty interval.
+    pub fn new_with_len<S: Into<T::Bound>, L: Into<Option<T::Distance>>>(
```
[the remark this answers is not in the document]
grishasobol: done

common/numerated/src/iterators.rs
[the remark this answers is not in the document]
grishasobol: done

common/numerated/src/iterators.rs
```diff
@@ -0,0 +1,330 @@
...
+            };
+
+            let interval2 = if let Some(interval2) = self.interval2 {
+                interval2
+            } else if let Some(interval2) = self.iter2.next() {
+                interval2
+            } else {
+                self.interval1 = None;
+                return Some(interval1);
+            };
```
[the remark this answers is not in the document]
grishasobol: done

ark0f (approved):

common/numerated/src/tree.rs
```diff
@@ -405,10 +399,10 @@ impl<T: Numerated> IntervalsTree<T> {
     /// Number of points in tree set.
     ///
     /// Complexity: `O(n)`, where `n` is amount of intervals in `self`.
-    pub fn points_amount(&self) -> Option<T::N> {
-        let mut res = T::N::zero();
+    pub fn points_amount(&self) -> Option<T::Distance> {
```
breathx, line 402: Shouldn't we consider an option to return usize here and for similar?

breathx: its not the distance actually

breathx (approved):
Most of the comments mostly point author attention on some places and are not necessary to implement

Huge work done! 🎆

common/numerated/src/numerated.rs
```diff
@@ -90,81 +68,134 @@ pub trait Bound<T: Sized>: From<T> + Copy {
...
-            Some(value) => Self::Value(value),
-            None => Self::Upper(T::max_value()),
+impl<T> PartialOrd for OptionBound<T>
+where
+    T: PartialOrd,
+{
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        match (self.0.as_ref(), other.0.as_ref()) {
+            (None, None) => Some(Ordering::Equal),
+            (None, Some(_)) => Some(Ordering::Less),
```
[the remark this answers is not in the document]
grishasobol: Good catch)) `ord` implementations is unnecessary for `Bound`, so this error does not appear in tree and interval tests. So, I have add tests for option bound in numerated.rs.

common/numerated/src/interval.rs
```diff
@@ -85,352 +67,247 @@ impl<T: Numerated> NonEmptyInterval<T> {
...
+    fn try_from((start, end): (S, E)) -> Result<Self, Self::Error> {
+        let start: T::Bound = start.into();
+        let end: T::Bound = end.into();
 
-        distance.checked_add(&T::N::one())
+        match (start.unbound(), end.unbound()) {
+            (None, None) => Err(TryFromRangeError::EmptyRange),
+            (None, Some(_)) => Err(TryFromRangeError::IncorrectRange),
+            (start, None) => Self::try_from(start..).map_err(|_| TryFromRangeError::EmptyRange),
+            (Some(start), Some(end)) => Self::try_from_range(start..end),
```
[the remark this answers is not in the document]
grishasobol: We need to make difference between constructors and internal representation. Internal interval is start..=end, but it can be constructed from different ranges/objects. I choose `try_from(x: Bound, y: Bound)` to be not inclusive, because Bound it is supposed to be something not inclusive in upper bound. 

I have added NOTE comment to clarify this moment

common/numerated/src/interval.rs
```diff
@@ -439,67 +316,40 @@ mod tests {
...
+        assert_eq!(Interval::<u8>::from(..).len(), None);
+
+        assert_eq!(Interval::<u8>::try_from(1..7).unwrap().raw_len(), Some(6));
+        assert_eq!(Interval::<u8>::try_from(..1).unwrap().raw_len(), Some(1));
+        assert_eq!(Interval::<u8>::from(..=1).raw_len(), Some(2));
+        assert_eq!(Interval::<u8>::try_from(1..).unwrap().raw_len(), Some(255));
+        assert_eq!(Interval::<u8>::try_from(0..).unwrap().raw_len(), None);
+        assert_eq!(Interval::<u8>::from(..).raw_len(), None);
+
+        assert_eq!(Interval::<i8>::try_from(-1..1).unwrap().len(), -126); // corresponds to 2 numeration
```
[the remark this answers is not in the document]
grishasobol: len() is supposed to return `T`, so if user wanna to return unsigned, than it can use `raw_len() -> T::Distance`

common/numerated/src/iterators.rs
```diff
@@ -0,0 +1,364 @@
+// This file is part of Gear.
+
+// Copyright (C) 2023 Gear Technologies Inc.
```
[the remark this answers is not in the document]
grishasobol: A yes)) this was created a long time ago

common/numerated/src/iterators.rs
```diff
@@ -0,0 +1,364 @@
...
+    }
+}
+
+impl<T: Numerated> From<T> for IntervalIterator<T> {
+    fn from(point: T) -> Self {
+        Interval::from(point).into()
+    }
+}
+
+impl<T: Numerated + LowerBounded> From<RangeToInclusive<T>> for IntervalIterator<T> {
```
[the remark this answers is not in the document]
grishasobol: I think it's impossible, because of different `T` types constrains

common/numerated/src/tree.rs
```diff
@@ -124,99 +124,100 @@ impl<T: Copy> Default for IntervalsTree<T> {
...
-            // because there can be neighbor interval which must be merged with `interval`.
-            self.inner.range(..=point_after_end)
-        } else {
-            self.inner.range(..=end)
-        }
-        .map(|(&s, &e)| (s, e));
+        // If `end` < `last`, then we must take in account next point after `end`,
+        // because there can be neighbor interval which must be merged with `interval`.
+        let iter_end = end.inc_if_lt(last).unwrap_or(end);
+        let mut iter = self.inner.range(..=iter_end).map(|(&s, &e)| (s, e));
```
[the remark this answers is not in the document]
grishasobol: `cloned()` - would clone tuple of refs, not derefs parts of tuples

common/numerated/src/tree.rs
grishasobol: It's distance, because it's sum of distances.
`usize` - it's same as u32 or u64, and so, it's not suitable to count all points for corresponding u32 or u64 type.

common/numerated/src/iterators.rs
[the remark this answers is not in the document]
breathx: I mentioned only some of them. But yeah, just checked and found only two. It's not that interesting to implement

Issue #3452, opened 2023-10-26
Remove false positive clippy lint on next rust update


StackOverflowExcept1on:
### File Location(s)

`examples/constructor/src/builder.rs`
`scripts/src/clippy.sh`
rust-lang/rust-clippy#11545

### Proposal

```rust
    // TODO #3452: remove this on next rust update
    #[allow(clippy::useless_conversion)]
    pub fn add_from_iter(mut self, calls: impl Iterator<Item = Call>) -> Self { ... }
```

SabaunT:
That clippy bug is fixed in commits to rust from 21-st September, but the bug with `overflow evaluating the requirement` (see https://github.com/rust-lang/rust/issues/117211) is from commits to rust from 19th September. That's why in #3451 we have to just skip it.

PR #3602, opened 2023-12-17
feat(wasm-builder): New approach to build demos

ark0f:
This is a new approach to build Gear programs which can replace `gear-wasm-builder`.

Schematic architecture:
image

Pipeline:
image

Purpose of program build script is just write features to lock file and don't use any `cargo:rerun-if` instructions so default cargo heuristics are enabled.

Binaries crate build script list its build dependencies and corresponding reads lock files to obtain program features then it runs `cargo build -p program1 -p program2` so `cargo` builds WASM binaries effectively and concurrently. Then it writes to lock files to skip `cargo` invocation in next or in another binaries crate to avoid few seconds cache verification

Pros:
- Advanced feature tracking. 
Features are tracked automatically so we don't need to re-run build script each time like in `gear-wasm-builder` (can be mitigated by `cargo-metadata` and `Cargo.toml` tracking)
- Faster build. 
`gear-wasm-builder` builds program via new `cargo` command in the same target directory so each `cargo` process blocks other. `wasm-dep-builder` collects list of programs so only one `cargo` process runs.

Cons:
- Create and import binaries crate.
It is important to create binaries create that will list and build programs

Current problem:
image

When crate includes itself as dev dependency (any integration tests) with different feature set then cargo build and runs 2 different versions of build script - for workspace and for tests.

`gear-wasm-builder` case is that `cargo` builds and runs script just once with merged feature set.

So, how to resolve these features? Should we deliver different WASM binaries? Or just merge feature sets?
I vote for merging but we need some mechanism to obtain all of feature sets cargo resolved for each build script invocation of program before build script invocation of binaries crate if we don't want to waste time on WASM build with incomplete feature set


clearloop:
plz check this https://github.com/gear-tech/gear/issues/3588#issuecomment-1862789332 btw

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -33,6 +33,7 @@
 //! So, we suppose this instruction have weight 0.
 
 #![cfg(feature = "runtime-benchmarks")]
+#![allow(clippy::assertions_on_constants)]
```
StackOverflowExcept1on, line 36: this is bug in current clippy version (see #3452), plz delete it

ark0f: > plz delete it

It's just impossible to use something else but bash scripts which is inconvenient for my workflow

StackOverflowExcept1on: https://doc.rust-lang.org/cargo/reference/workspaces.html#the-lints-table

ark0f: Not supported by our pinned toolchain. Lints were merged 9 September. Our toolchain is 5 September

https://github.com/rust-lang/cargo/pull/12648#event-10323314352

ark0f:
@clearloop 

> plz check this https://github.com/gear-tech/gear/issues/3588#issuecomment-1862789332 btw

Not applicable. This approach resolves features via `CARGO_FEATURE_*` and default cargo tracking heuristics

ark0f:
A1 label because benchmarks compile to WASM and require WASM binaries inside so mechanism to get these binaries is required to be developed

gclient/binaries/build.rs
```diff
@@ -0,0 +1,21 @@
+// This file is part of Gear.
+//
+// Copyright (C) 2023 Gear Technologies Inc.
```
gshep, line 3: I suggest removing the year number from the copyright since it is not consistent across the source code and we will not update it later

utils/wasm-dep-builder/src/utils.rs
```diff
@@ -0,0 +1,129 @@
...
+
+impl PartialOrd for UnderscoreString {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for UnderscoreString {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.underscore().cmp(&other.underscore())
```
gshep, line 127: this triggers additional memory allocation however still doesn't affect any performance

gshep (approved):
Good job! :1st_place_medal:

gclient/binaries/build.rs
ark0f: I see most of files have year when just a few don't

examples/signal-wait/src/lib.rs
```diff
@@ -0,0 +1 @@
+
```
StackOverflowExcept1on, line 1: why `examples/signal-wait/**` is empty?

StackOverflowExcept1on (approved):

ark0f:
`cargo check` may be failed because of missing `.lock`. Investigating...

NikVolf:
Can we maybe put "dependency builder" inside gear-wasm-builder itself?

breathx:
> Can we maybe put "dependency builder" inside gear-wasm-builder itself?

It will replace gear_wasm_builder some day, so is there any reason to merge them now?

Just for notice: as it was mentioned, there are a few problems with these approach, and as far as we discussed with @ark0f they're already solved, but not pushed

ark0f:
> Can you describe the approach in the PR outline?

Done

breathx:
Temporary closing in favour of opened referencing issue as won't be implemented

PR #3606, opened 2023-12-19
feat(runtime): Revokable vouchers

breathx:
TODO:
- [x] benchmarks
- [x] docs
- [x] mark deprecated
- [x] additional test for call_legacy impl (?)


common/src/storage/complicated/counter.rs
```diff
@@ -44,6 +44,14 @@ pub trait Counter {
 
     /// Resets stored value by setting default/starting value.
     fn reset();
+
+    /// Increases and returns stored value.
+    ///
+    /// Should be safe from overflow.
+    fn inc_get() -> Self::Value {
```
ekovalev, line 51: Why not let `increase()` return the new counter value?

pallets/gear-voucher/src/lib.rs
```diff
@@ -100,151 +98,413 @@ pub mod pallet {
...
-        /// a caller's responsibility to ensure the consistency of the input parameters.
-        /// - `amount`: The voucher amount.
+        /// Deposits event `VoucherIssued`, that contains `VoucherId` to be
+        /// used by spender for balance-less on-chain interactions.
         ///
-        /// ## Complexity
-        /// O(Z + C) where Z is the length of the call and C its execution weight.
+        /// Arguments:
+        /// * spender:  user id that is eligible to use the voucher;
+        /// * balance:  voucher balance could be used for tx fees and gas,
```
gshep, line 211:
```suggestion
        /// * balance:  voucher balance could be used for transactions fees and gas;
```

pallets/gear-voucher/src/lib.rs
```diff
@@ -100,151 +98,413 @@ pub mod pallet {
...
+            T::Currency::transfer(
+                &owner,
+                &voucher_id.cast(),
+                balance,
+                ExistenceRequirement::KeepAlive,
+            )
+            .map_err(|_| Error::<T>::BalanceTransfer)?;
+
+            // Asserting non-zero validity.
+            ensure!(!validity.is_zero(), Error::<T>::ZeroValidity);
```
gshep, line 251: Better place this check before any access to storage, before asserting amount of programs or after for example.

pallets/gear-voucher/src/lib.rs
```diff
@@ -100,151 +98,413 @@ pub mod pallet {
...
-    impl<T: Config> Pallet<T> {
-        /// Derive a synthesized account ID from an account ID and a program ID.
-        pub fn voucher_id(who: &T::AccountId, program_id: &ProgramId) -> T::AccountId {
-            let entropy = (b"modlpy/voucher__", who, program_id).using_encoded(blake2_256);
-            Decode::decode(&mut TrailingZeroInput::new(entropy.as_ref()))
-                .expect("infinite length input; no invalid inputs for type; qed")
+        /// Revoke existing voucher.
+        ///
+        /// This extrinsic revokes existing voucher, if current block is greater
+        /// than expiration block of the voucher (its no longer valid).
```
gshep, line 310:
```suggestion
        /// than expiration block of the voucher (it is no longer valid).
```

common/src/storage/complicated/counter.rs
gshep: it is used only in one place so looks like the method is unnecessary

pallets/gear-voucher/src/tests.rs
```diff
@@ -1,82 +1,674 @@
...
-// GNU General Public License for more details.
+// // This program is distributed in the hope that it will be useful,
+// // but WITHOUT ANY WARRANTY; without even the implied warranty of
+// // MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// // GNU General Public License for more details.
 
-// You should have received a copy of the GNU General Public License
-// along with this program. If not, see <https://www.gnu.org/licenses/>.
+// // You should have received a copy of the GNU General Public License
+// // along with this program. If not, see <https://www.gnu.org/licenses/>.
```
gshep, line 17: unnecessary comments

pallets/gear/src/mock.rs
```diff
@@ -52,6 +56,62 @@ pub(crate) const USER_3: AccountId = 3;
 pub(crate) const LOW_BALANCE_USER: AccountId = 4;
 pub(crate) const BLOCK_AUTHOR: AccountId = 255;
 
+#[derive(Encode, Decode, Clone, Eq, PartialEq, Default, TypeInfo)]
+pub struct VoucherLegitimate;
+
+impl SignedExtension for VoucherLegitimate {
```
gshep, line 62: The code looks similar to other testing mocks. Can it be refactored anyhow?

gshep (commented):
Good job! :1st_place_medal:

gshep (approved):

pallets/gear-voucher/src/lib.rs
```diff
@@ -100,151 +98,413 @@ pub mod pallet {
...
+        #[pallet::weight(T::WeightInfo::issue())]
         pub fn issue(
             origin: OriginFor<T>,
-            to: AccountIdLookupOf<T>,
-            program: ProgramId,
-            value: BalanceOf<T>,
+            spender: AccountIdOf<T>,
+            balance: BalanceOf<T>,
+            programs: Option<BTreeSet<ProgramId>>,
+            validity: BlockNumberFor<T>,
```
techraed, line 225: isn't _duration_ a common name for that?

runtime/vara/src/lib.rs
```diff
@@ -1057,9 +1064,7 @@ pub struct DelegateFeeAccountBuilder;
 impl DelegateFee<RuntimeCall, AccountId> for DelegateFeeAccountBuilder {
     fn delegate_fee(call: &RuntimeCall, who: &AccountId) -> Option<AccountId> {
         match call {
-            RuntimeCall::GearVoucher(pallet_gear_voucher::Call::call { call }) => {
-                GearVoucher::sponsor_of(who, call)
-            }
+            RuntimeCall::GearVoucher(voucher_call) => voucher_call.sponsored_by(who.clone()),
```
techraed, line 1067: not related to this PR, but better to open a separate issue for the TODO over the `DelegateFee` trait impl.

pallets/gear-voucher/src/lib.rs
```diff
@@ -100,151 +98,413 @@ pub mod pallet {
...
-        payload: Vec<u8>,
-        gas_limit: u64,
-        value: Balance,
-        keep_alive: bool,
-    },
-}
+            // Optionally extends whitelisted programs with amount validation.
+            if let Some(extra_programs) = append_programs {
+                if let Some(mut new_programs) = extra_programs {
+                    if let Some(ref mut programs) = voucher.programs {
```
techraed, line 445: are there any security reasons for not setting `programs` if `voucher.programs.is_none()`?

pallets/payment/src/tests.rs
```diff
@@ -584,17 +599,17 @@ fn fee_payer_replacement_works() {
         let weight = Weight::from_parts(1_000, 0);
 
         let pre = CustomChargeTransactionPayment::<Test>::from(0)
-            .pre_dispatch(&ALICE, call, &info_from_weight(weight), len)
+            .pre_dispatch(&BOB, call, &info_from_weight(weight), len)
             .unwrap();
 
         let fee_weight = WeightToFeeFor::<Test>::weight_to_fee(&weight);
 
         // Alice hasn't paid fees
```
ekovalev, line 607: Should be `// Bob hasn't paid fees`, should it not?

pallets/gear-voucher/src/lib.rs
ekovalev: > isn't _duration_ a common name for that?

If not _validity_, it'd rather be _expiry_ than _duration_: this value determines until which block a voucher is valid, not for how many blocks it's going to be valid.

pallets/gear-voucher/src/internal.rs
```diff
@@ -0,0 +1,202 @@
...
+        self.programs
+            .as_ref()
+            .map(|v| v.contains(&program_id))
+            .unwrap_or(true)
+    }
+}
+
+/// Prepaid call to be executed on-chain.
+#[derive(Debug, Clone, Encode, Decode, TypeInfo, PartialEq, Eq, PartialOrd, Ord)]
+pub enum PrepaidCall<Balance> {
```
ekovalev, line 77: Couldn't we get away without this artificial type? It only serves for the purpose of validating a random call eligibility for being "prepaid". We could pass a usual call filter associated type that would filter out illegitimate calls, could we? Which, in my opinion, would be a more "Substrate-idiomatic" way. Or am I missing something?

ekovalev: This way there would be no need for a custom call dispatcher.

pallets/gear-voucher/src/internal.rs
```diff
@@ -0,0 +1,202 @@
...
+                Self::destination_program(&origin, call).ok_or(Error::<T>::UnknownDestination)?;
+
+            ensure!(
+                programs.contains(&destination),
+                Error::<T>::InappropriateDestination
+            );
+        }
+
+        ensure!(
+            <frame_system::Pallet<T>>::block_number() <= voucher.validity,
```
ukint-vs, line 165: Voucher validity should be checked before destination

pallets/gear-voucher/src/lib.rs
ukint-vs: the exact expiration block number seems more logical for using the voucher

runtime/vara/src/lib.rs
```diff
@@ -1108,6 +1114,62 @@ impl pallet_vesting::Config for Runtime {
     const MAX_VESTING_SCHEDULES: u32 = 28;
 }
 
+#[derive(Encode, Decode, Clone, Eq, PartialEq, Default, TypeInfo)]
+pub struct VoucherLegitimate;
+
+impl SignedExtension for VoucherLegitimate {
+    const IDENTIFIER: &'static str = "VoucherLegitimate";
```
grishasobol, line 1121: Could you explain please the reason of this additional validation? In order to save money for somebody who uses voucher wrong?

grishasobol (approved):

pallets/gear-voucher/src/tests.rs
```diff
@@ -1,82 +1,674 @@
...
+                Some(Some(set)),
+                // prolong validity
+                None,
+            ),
+            Error::<Test>::MaxProgramsLimitExceeded
         );
     });
 }
+
+mod utils {
```
techraed, line 622: It's a good practice when you read functions from top to bottom (from caller to callee), it really makes increases readability. But here and in `internal.rs` it's a reversed order.

techraed (commented):
Would be nice to test voucher pallet calls being consistent with each other, like 1) test that after calling dispatcher on voucher several times the free balance of a voucher decreases and 2) test that expected surplus is returned when revoke is called.

Can't see how vouchers are deleted. I guess it's planned to remove them in `revoke`. But what about outdated vouchers that aren't revoked or updated for too long?

pallets/gear-voucher/src/internal.rs
```diff
@@ -0,0 +1,202 @@
...
+    fn dispatch(
+        account_id: Self::AccountId,
+        sponsor_id: Self::AccountId,
+        call: PrepaidCall<Self::Balance>,
+    ) -> DispatchResultWithPostInfo;
+}
+
+impl<T: Config> Pallet<T> {
+    /// Derive a synthesized account ID from an account ID and a program ID.
+    pub fn voucher_id(who: &T::AccountId, program_id: &ProgramId) -> T::AccountId {
```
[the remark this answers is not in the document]
techraed: maybe move to separate `impl` - that would be more explicit

pallets/gear-voucher/src/internal.rs
```diff
@@ -0,0 +1,202 @@
...
+
+/// Type containing all data about voucher.
+#[derive(Debug, Encode, Decode, TypeInfo)]
+pub struct VoucherInfo<AccountId, BlockNumber> {
+    /// Owner of the voucher.
+    /// May be different to original issuer.
+    /// Owner manages and claims back remaining balance of the voucher.
+    pub owner: AccountId,
+    /// Set of programs this voucher could be used to interact with.
+    /// In case of None means any gear program.
```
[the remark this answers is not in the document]
thecaralice: What about this?

```rs
    /// In case of [`None`] means any gear program.
```

See [rustdoc documentation](https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html#linking-to-items-by-name)

pallets/gear-voucher/src/lib.rs
```diff
@@ -100,151 +98,413 @@ pub mod pallet {
...
+            append_programs: Option<Option<BTreeSet<ProgramId>>>,
+            prolong_validity: Option<BlockNumberFor<T>>,
+        ) -> DispatchResultWithPostInfo {
+            // Ensuring origin.
+            let origin = ensure_signed(origin)?;
+
+            // Querying voucher.
+            let Some(mut voucher) = Vouchers::<T>::get(spender.clone(), voucher_id) else {
+                return Err(Error::<T>::InexistentVoucher.into());
+            };
```
thecaralice, line 411:
```suggestion
            let mut voucher = Vouchers::<T>::get(spender.clone(), voucher_id).ok_or(Error::<T>::InexistentVoucher)?;
```

common/src/storage/complicated/counter.rs
thecaralice: +1

pallets/gear/src/mock.rs
breathx: Nope: due to inability to cast some specific `mock*::RuntimeCall` into ref to `pallet_gear_voucher::Call`

pallets/gear-voucher/src/lib.rs
breathx: > could be the case that some decided to restrict usage of the voucher

we're allowing only extending voucher functionality. if you're about to restrict it - revoke on expiration and issue to user a new one

required by the business logic to support trust-less interactions between issuers and users

pallets/gear-voucher/src/lib.rs
```diff
@@ -100,151 +98,413 @@ pub mod pallet {
...
+            origin: OriginFor<T>,
+            spender: AccountIdOf<T>,
+            voucher_id: VoucherId,
+        ) -> DispatchResultWithPostInfo {
+            // Ensuring origin.
+            let origin = ensure_signed(origin)?;
+
+            // Querying voucher data.
+            // NOTE: currently getting instead of taking value.
+            let Some(voucher) = Vouchers::<T>::get(spender.clone(), voucher_id) else {
```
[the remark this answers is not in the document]
breathx: Due to proper handling of gas tree on vouchers pallet side some gas can still return to the account that is attached to some voucher so it was decided not to remove voucher until we implement proper refcounting to it.

pallets/gear-voucher/src/lib.rs
```diff
@@ -100,151 +98,413 @@ pub mod pallet {
...
+            // Casting voucher id into voucher account id.
+            let voucher_id_acc = voucher_id.cast();
+
+            // Querying voucher account id balance.
+            let voucher_balance = T::Currency::free_balance(&voucher_id_acc);
+
+            // If balance of the voucher account is not zero, than transferring
+            // all the funds to voucher owner and depositing event
+            // `VoucherRevoked`, otherwise call is Noop.
+            if !voucher_balance.is_zero() {
```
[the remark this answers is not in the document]
breathx: It would be exactly dirty. No reason to spam events while operation was no-op on storage

pallets/gear-voucher/src/tests.rs
```diff
@@ -1,82 +1,674 @@
...
     new_test_ext().execute_with(|| {
-        let program_id = H256::from(b"some//quasy//random//program//id").cast();
-        let synthesized = Voucher::voucher_id(&BOB, &program_id);
+        let program_id = H256::random().cast();
 
+        let alice_balance = Balances::free_balance(ALICE);
+
+        let voucher_id = utils::issue(ALICE, BOB, program_id);
+        assert_ok!(voucher_id);
+        let voucher_id = voucher_id.unwrap();
```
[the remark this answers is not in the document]
breathx: Initially it had better ui and we use this approach some places else but ok

pallets/gear-voucher/src/lib.rs
```diff
@@ -100,151 +98,413 @@ pub mod pallet {
...
+            };
+
+            // NOTE: currently ensuring that owner revokes voucher.
+            ensure!(voucher.owner == origin, Error::<T>::BadOrigin);
+
+            // Ensuring voucher is no longer valid.
+            ensure!(
+                <frame_system::Pallet<T>>::block_number() > voucher.validity,
+                Error::<T>::IrrevocableYet
+            );
```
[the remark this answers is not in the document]
breathx: Same answer as for restricting voucher functionality: consumer/user of the voucher must have 100% guarantees that owner of the voucher is unable to revoke it anytime they want: imagine case when I paid for gas (voucher) you by fiat money etc

pallets/gear-voucher/src/benchmarking.rs
```diff
@@ -16,16 +16,14 @@
...
-#[allow(unused)]
-use crate::Pallet as GearVoucher;
 use crate::*;
 use common::{benchmarking, Origin};
-use frame_benchmarking::{benchmarks, impl_benchmark_test_suite};
+use frame_benchmarking::{benchmarks, impl_benchmark_test_suite, Zero};
 use frame_support::traits::Currency;
 use frame_system::RawOrigin;
-use sp_runtime::traits::UniqueSaturatedInto;
+use sp_runtime::traits::{One, UniqueSaturatedInto};
```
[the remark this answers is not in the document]
breathx: Can't see any problem of reading it cause its common reimports for pallets

pallets/gear-voucher/src/mock.rs
```diff
@@ -99,8 +113,18 @@ impl Mailbox for MailboxMock {
...
     }
-    fn peek(_key1: &Self::Key1, _key2: &Self::Key2) -> Option<Self::Value> {
-        unimplemented!()
+    fn peek(key1: &Self::Key1, key2: &Self::Key2) -> Option<Self::Value> {
+        if *key2 == MAILBOXED_MESSAGE {
+            Some(UserStoredMessage::new(
+                MAILBOXED_MESSAGE,
+                MAILBOXED_PROGRAM,
+                (*key1).cast(),
+                vec![].try_into().unwrap(),
```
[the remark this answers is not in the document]
breathx: https://github.com/gear-tech/gear/issues/3641

pallets/gear-voucher/src/internal.rs
breathx: Yeah, but it seems weird to accept the type of call where 98% are invalid and couldn't be used with voucher logic.
With approach of separate enum type we
1) increase ui for polkajs and similars;
2) specify exact calls that could be processed (instead of misleading to all of them)
3) (:c) have overhead of a custom call dispatcher, but it seems fine for me

Furthermore, it's considering for future to issue vouchers for some specific actions: e.g. only send_messages or anything else: there will be problems parsing runtime call every time (or implementing try into manually)

pallets/gear-voucher/src/mock.rs
```diff
@@ -117,6 +141,7 @@ impl pallet_gear_voucher::Config for Test {
     type WeightInfo = ();
     type CallsDispatcher = ();
```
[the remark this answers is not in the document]
breathx: It will require a lot of other thing to be implemented in the mock. Can't see any reason for that cause this pallet tests vouchers logic totally except processing of the messages it sends, but we have those tests in pallet gear (where processing is implemented)

pallets/gear-voucher/src/tests.rs
```diff
@@ -1,82 +1,674 @@
...
+            PrepaidCall::SendReply {
+                reply_to_id: MAILBOXED_MESSAGE,
+                payload: vec![],
+                gas_limit: 0,
+                value: 0,
+                keep_alive: false
+            },
+        ));
+
+        // Always ok.
```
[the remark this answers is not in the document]
breathx: added comment
it's not because of `()` impl

pallets/gear-voucher/src/internal.rs
ekovalev: Regarding the last argument, if we look at the Substrate's own `proxy` pallet, we'll see quite a similar use case: there they have a hierarchy of `ProxyTypes`, and filter calls based on what a specific proxy can or cannot do. It looks a bit complicated, I agree, but that's what they do.
Anyway, if you think your approach is better (apart from being more readable), then I'm fine.

ekovalev:
General comment: the original idea of having a separate pallet for vouchers as opposed to implementing the whole machinery using "in-built" Substrate proxies was to reduce complexity - that's why the original implementation was so light-weight, even though it lacked many features. Now I have a gut feeling that going the "proxy" way (and even having delegated fees with it) wouldn't seem too complex already. And if we continue introducing more functionality (like dedicated vouchers for specific actions) that'll make the similarity with `ProxyType`'s hierarchy even more obvious.
Just my two cents ).
Otherwise all is good.

pallets/gear-voucher/src/internal.rs
breathx: The topic is always open and I think we can change the solution implementation a bit later. Atm this is how ecosystem images their interactions with vouchers

ekovalev (approved):

pallets/gear-voucher/src/lib.rs
techraed: right, that's more trust-less, agree

techraed (commented):
> Can't see how vouchers are deleted. I guess it's planned to remove them in `revoke`. But what about outdated vouchers that aren't revoked or updated for too long?

@breathx could you address this?

gshep (approved):

breathx:
> > Can't see how vouchers are deleted. I guess it's planned to remove them in `revoke`. But what about outdated vouchers that aren't revoked or updated for too long?
> 
> @breathx could you address this?

https://github.com/gear-tech/gear/pull/3606#discussion_r1443014312

Outdated vouchers would be able to be revoked some time soon by anyone who's interested in it: for example, block producers to reduce storage consumption for free

techraed (approved):

PR #3609, opened 2023-12-20
fix(node): Take 'wait' syscall into account in calculate_gas_info

gshep:
`gr_wait` syscall greedily  consumes all available gas. So when the gas value can be divided by waitlist cost without remainder `calculate_gas_info` will return incorrect info.

Test case - https://github.com/gear-tech/gear/actions/runs/7274019037

@gear-tech/dev

### DIFF
--- a/pallets/gear/src/runtime_api.rs
+++ b/pallets/gear/src/runtime_api.rs
@@ -160,7 +160,6 @@ where
                 }
             };
 
-            let dispatch_id = queued_dispatch.id();
             let success_reply = queued_dispatch
                 .reply_details()
                 .map(|rd| rd.to_reply_code().is_success())
@@ -215,10 +214,14 @@ where
                     Some(remaining_gas) => {
                         min_limit = min_limit.max(initial_gas.saturating_sub(remaining_gas))
                     }
-                    None if matches!(note, JournalNote::WaitDispatch { .. }) => {
-                        min_limit = initial_gas
-                    }
-                    _ => (),
+                    None => match note {
+                        JournalNote::WaitDispatch { ref dispatch, .. }
+                            if from_main_chain(dispatch.id())? =>
+                        {
+                            min_limit = initial_gas
+                        }
+                        _ => (),
+                    },
                 }
 
                 match note {
--- a/pallets/gear/src/runtime_api.rs
+++ b/pallets/gear/src/runtime_api.rs
@@ -211,8 +211,14 @@ where
             for note in journal {
                 core_processor::handle_journal(vec![note.clone()], &mut ext_manager);
 
-                if let Some(remaining_gas) = get_main_limit() {
-                    min_limit = min_limit.max(initial_gas.saturating_sub(remaining_gas));
+                match get_main_limit() {
+                    Some(remaining_gas) => {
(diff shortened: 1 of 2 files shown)


pallets/gear/src/runtime_api.rs
```diff
@@ -211,8 +211,14 @@ where
...
                 core_processor::handle_journal(vec![note.clone()], &mut ext_manager);
 
-                if let Some(remaining_gas) = get_main_limit() {
-                    min_limit = min_limit.max(initial_gas.saturating_sub(remaining_gas));
+                match get_main_limit() {
+                    Some(remaining_gas) => {
+                        min_limit = min_limit.max(initial_gas.saturating_sub(remaining_gas))
+                    }
+                    None if matches!(note, JournalNote::WaitDispatch { .. }) => {
+                        min_limit = initial_gas
```
breathx, line 219: it's not always "infinite" gr_wait()

breathx (changes requested):

pallets/gear/src/runtime_api.rs
gshep: then it will go to the `Some(...)` branch since there will be some gas value in the node

StackOverflowExcept1on: Yes, but do we have problems with `gr_wait_for(dur)` and `gr_wait_up_to(dur)`? They should not consume all available gas because of limited duration.

gshep: Frankly speaking bounded syscalls can have an enough big duration argument so additional check wouldn't be redundant

pallets/gear/src/runtime_api.rs
```diff
@@ -211,8 +210,18 @@ where
...
                 core_processor::handle_journal(vec![note.clone()], &mut ext_manager);
 
-                if let Some(remaining_gas) = get_main_limit() {
-                    min_limit = min_limit.max(initial_gas.saturating_sub(remaining_gas));
+                match get_main_limit() {
+                    Some(remaining_gas) => {
+                        min_limit = min_limit.max(initial_gas.saturating_sub(remaining_gas))
+                    }
+                    None => match note {
+                        JournalNote::WaitDispatch { ref dispatch, .. }
```
breathx, line 218: Add comment pls

breathx: if from main chain, but finite wait (e.g. wait_for), it shouldn't result into initial gas

breathx (changes requested):

pallets/gear/src/runtime_api.rs
gshep: wait calls with bounded duration would go to the first match arm. However if the calls have such big duration then it won't matter how the min_limit calculated: it will be the same

breathx: > wait calls with bounded duration would go to the first match arm. However if the calls have such big duration then it won't matter how the min_limit calculated: it will be the same

oh, truth.
add comment anyway pls, it will be hard to understand wout context in future

breathx (approved):

StackOverflowExcept1on (approved):

PR #3610, opened 2023-12-20
feat(gstd): add the missing `dbg!` macro

shamilsan:
- Added `dbg!` macro to `gstd` prelude similar to the standard library's one (https://doc.rust-lang.org/std/macro.dbg.html).

@gear-tech/dev


holykol (approved):
cool

ukint-vs (approved):

PR #3611, opened 2023-12-20
chore: bump gear-dlmalloc for free_range support

holykol:
Bump allocator to version that uses new `free_range` call.

@gear-tech/dev

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -166,8 +166,8 @@ tempfile = "3.8.1"
 
 # Published deps
 #
-# https://github.com/gear-tech/dlmalloc-rust/tree/v0.1.4
-dlmalloc = { package = "gear-dlmalloc", version = "0.1.4" }
+# https://github.com/gear-tech/gear-dlmalloc/tree/0.2.0
+dlmalloc = { package = "gear-dlmalloc", version = "0.2.0" }
 # https://github.com/gear-tech/wasm-instrument/tree/gear-stable-v0.3.0
 gwasm-instrument = { version = "0.3.0", default-features = false }
 # https://github.com/gear-tech/wasm-utils/tree/v0.19.0
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4158,9 +4158,9 @@ dependencies = [
 
 [[package]]
 name = "gear-dlmalloc"
-version = "0.1.4"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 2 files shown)


shamilsan (approved):
Need to add a `0.1.5` tag there too: https://github.com/gear-tech/dlmalloc-rust/tags

clearloop (approved):

breathx:
> Need to add a `0.1.5` tag there too: https://github.com/gear-tech/dlmalloc-rust/tags

We've never tagged it before. Should we?

breathx (approved):

holykol:
@breathx
> We've never tagged it before. Should we?

I've already published `0.1.5` on crates.io, so tagging just makes it consistent. If anything, it should have been a major version change because it technically breaks backwards compatibility by relying on a new syscall

Cargo.toml
```diff
@@ -166,8 +166,8 @@ tempfile = "3.8.1"
 
 # Published deps
 #
-# https://github.com/gear-tech/dlmalloc-rust/tree/v0.1.4
-dlmalloc = { package = "gear-dlmalloc", version = "0.1.4" }
+# https://github.com/gear-tech/dlmalloc-rust/tree/v0.1.5
+dlmalloc = { package = "gear-dlmalloc", version = "0.1.5" }
```
shamilsan, line 170: Bump version to 0.2.0 due to breaking changes (see https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html)

shamilsan (changes requested):

Cargo.toml
holykol: Done

Cargo.toml
```diff
@@ -166,8 +166,8 @@ tempfile = "3.8.1"
 
 # Published deps
 #
-# https://github.com/gear-tech/dlmalloc-rust/tree/v0.1.4
-dlmalloc = { package = "gear-dlmalloc", version = "0.1.4" }
+# https://github.com/gear-tech/dlmalloc-rust/tree/v0.1.5
```
shamilsan, line 169: Please change this link too
```suggestion
# https://github.com/gear-tech/dlmalloc-rust/tree/0.2.0
```

shamilsan (approved):

breathx (approved):

Cargo.toml
holykol: thanks

PR #3612, opened 2023-12-21
fix(wasm-gen): copy data from `HashWithValue` before writing to `err`

StackOverflowExcept1on:
We currently have a problem passing the same pointers as mutable and immutable.

### DIFF
--- a/utils/wasm-gen/src/generator/syscalls/imports.rs
+++ b/utils/wasm-gen/src/generator/syscalls/imports.rs
@@ -341,6 +341,32 @@ impl<'a, 'b> SyscallsImportsGenerator<'a, 'b> {
         let pid_value_ptr = rid_pid_value_ptr + mem::size_of::<Hash>() as i32;
 
         let func_instructions = Instructions::new(vec![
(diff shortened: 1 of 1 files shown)


breathx (approved):

techraed (changes requested):
Seems odd.
I found the origin of the problem in `build_call_with_destination`, which also affects other cases.

breathx:
> Seems odd. I found the origin of the problem in `build_call_with_destination`, which also affects other cases.

So what to do with the PR?

techraed:
It can still be beneficial if there are syscalls invocations where each param has the same value. We faced the problem during tests when wasm from `[0; N]` buffer forming `Unstructured` was created.
However, that's a rare situation and tests can be re-written with some random seeded data input.

What do you think? @StackOverflowExcept1on

StackOverflowExcept1on:
@techraed I think we can merge this because at least 3 people came across this problem in tests. Rewriting using a seed is not always obvious when copying code from existing tests.

techraed:
Alright, for the sake of tests simplicity let's do that.

techraed (approved):

PR #3615, opened 2023-12-21
feat(wasm-gen): avoid infinite wait-wake for loose syscalls

### LINKED ISSUE #3273: Avoid infinite wait-wake patterns in `wasm-gen`
### Problem to Solve

Sometimes we have a code wich always invokes `wait_for` on every execution, until it exhausts the gas-limit. We should avoid such "deadlock" programs generation

StackOverflowExcept1on:
Resolves #3273

Also:
- added `MemoryLayout` struct that will never be overwritten during fuzzing
- fixed bug with mutable array

TODO:
- [x] think about limiting `wait` in the `init` function
- [x] gear-tech/wasm-tools#5


breathx (approved):

utils/wasm-gen/src/generator/entry_points.rs
```diff
@@ -164,8 +166,31 @@ impl<'a, 'b> EntryPointsGenerator<'a, 'b> {
             (module, func_type)
         });
 
-        let export_body_instructions =
+        let mut export_body_instructions =
             self.generate_export_body(export_body_call_idx, export_body_call_func_type)?;
+
+        // after initializing the program, we will write about this in a special pointer
+        if name == "init" {
```
techraed, line 173: seems that code must be in `generate_export_body`.

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -143,6 +143,20 @@ impl<'a, 'b> From<DisabledAdditionalDataInjector<'a, 'b>> for SyscallsInvocator<
 }
 
 impl<'a, 'b> SyscallsInvocator<'a, 'b> {
+    /// Returns the size of the memory in pages.
+    fn memory_size_pages(&self) -> u32 {
```
techraed, line 147: move that under `memory_size_in_bytes`

utils/wasm-gen/src/generator/entry_points.rs
```diff
@@ -164,8 +166,31 @@ impl<'a, 'b> EntryPointsGenerator<'a, 'b> {
...
+            if let Some(memory_size_pages) = self.module.initial_mem_size() {
+                let mem_size = Into::<WasmPageCount>::into(memory_size_pages).memory_size();
+
+                let upper_limit = mem_size.saturating_sub(1) as i32;
+                let wait_called_ptr = upper_limit.saturating_sub(50);
+                let init_called_ptr = wait_called_ptr + mem::size_of::<bool>() as i32;
```
techraed, line 179: These ptrs layouts are known in advance. I would move this code (and the same in _invocator.rs_) to some function/method.

utils/wasm-gen/src/config.rs
```diff
@@ -131,6 +131,11 @@ impl ConfigsBundle for (GearWasmGeneratorConfig, SelectableParams) {
 pub struct StandardGearWasmConfigsBundle {
     /// Externalities to be logged.
     pub log_info: Option<String>,
+    /// Frequency of wait syscalls.
+    ///
+    /// For example, if this parameter is 4, wait syscalls will be invoked
+    /// with probability 1/4.
+    pub waiting_frequency: Option<u32>,
```
techraed, line 138: I'm not sure that **frequency** is a suitable name for that.
You see, `SyscallInjectionType::Function` actually defines **frequency**.

techraed: I have here a suggestion to make this solution more general, which will make `wasm-gen` even better.

What about inserting not only the frequency, but the chance to be called in the `SyscallInjectionType::Function`?
So it will be `SyscallInjectionType::Function { call_chance: Option<u32>, frequency: RangeInclusive<u32> }`.

In this case when generating instructions for any syscall we will first process `call_chance`. If it's `None`, then no need in defining instructions like `if syscall_called_ptr & call_chance ...`.

The only downside here is that we must define a layout in memory which stores flags for every syscall, like it's done here for `wait`. Maybe there are some ways to avoid that gracefully.

All in all, that will make handling `wait` syscall invocation not that alien to `build_syscall_invoke_instructions` and the whole design.

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -315,6 +329,49 @@ impl<'a, 'b> SyscallsInvocator<'a, 'b> {
...
 
+        if let Some(wait_frequency) = self
+            .config
+            .waiting_frequency()
+            .filter(|_| invocable.is_wait_syscall())
+        {
+            let mem_size = self.memory_size_in_bytes();
+
+            let upper_limit = mem_size.saturating_sub(1) as i32;
+            let wait_called_ptr = upper_limit.saturating_sub(50);
```
techraed, line 340: This actually means that every syscall invocation must never touch that memory. Can we assert that?

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -143,6 +143,20 @@ impl<'a, 'b> From<DisabledAdditionalDataInjector<'a, 'b>> for SyscallsInvocator<
...
+    fn memory_size_pages(&self) -> u32 {
+        self.module
+            .initial_mem_size()
+            // To instantiate this generator, we must instantiate SyscallImportsGenerator, which can be
+            // instantiated only with memory import generation proof.
+            .expect("generator is instantiated with a memory import generation proof")
+    }
+
+    /// Returns the size of the memory in bytes.
+    fn memory_size_in_bytes(&self) -> u32 {
```
techraed, line 156: maybe like this as `memory_size_pages` is done?

```suggestion
    fn memory_size_bytes(&self) -> u32 {
```

utils/wasm-gen/src/generator/syscalls/invocator.rs
StackOverflowExcept1on: 805fa95b0bb72a0efba37a115abcb38d853964d3

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -143,6 +146,20 @@ impl<'a, 'b> From<DisabledAdditionalDataInjector<'a, 'b>> for SyscallsInvocator<
 }
 
 impl<'a, 'b> SyscallsInvocator<'a, 'b> {
+    /// Returns the size of the memory in bytes.
+    fn memory_size_bytes(&self) -> u32 {
```
techraed, line 150: I'd move both of methods under all others as "used" ones, not the "users". Or event to a separate `impl`

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -315,6 +332,49 @@ impl<'a, 'b> SyscallsInvocator<'a, 'b> {
         };
         instructions.append(&mut result_processing);
 
+        if let Some(waiting_probability) = self
```
techraed, line 335: can we move all this logic to a separate method and add a explanatory doc to it?

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -386,7 +441,8 @@ impl<'a, 'b> SyscallsInvocator<'a, 'b> {
                 }
                 ProcessedSyscallParams::MemoryArrayLength => {
                     let length;
-                    let upper_limit = mem_size.saturating_sub(1) as i32;
+                    let upper_limit =
+                        mem_size.saturating_sub(MemoryLayout::RESERVED_MEMORY_SIZE) as i32;
```
techraed, line 445: did it get it right, that if memory array ptr is chosen to be somewhere like `let random_ptr = mem_size.saturating_sub(MemoryLayout::RESERVED_MEMORY_SIZE).saturating_sub(2)`, then it's enough to have length greater than 2 to rewrite everything (in case we got syscall with `Mut*` like ptr`?

utils/wasm-gen/src/generator/syscalls/imports.rs
```diff
@@ -799,12 +799,13 @@ impl<'a, 'b> SyscallsImportsGenerator<'a, 'b> {
 
     /// Reserves enough memory build precise syscall.
     fn reserve_memory(&self) -> i32 {
-        self.memory_size_in_bytes()
-            .saturating_sub(Self::PRECISE_SYSCALL_MEMORY_SIZE) as i32
+        self.memory_size_bytes()
+            .saturating_sub(MemoryLayout::RESERVED_MEMORY_SIZE)
+            .saturating_sub(Self::PRECISE_SYSCALL_RESERVED_MEMORY_SIZE) as i32
```
techraed, line 804: I believe that should be included into `MemoryLayout`.

utils/wasm-gen/src/generator/syscalls/invocator.rs
techraed: it's possible not to pay attention to that if there's only few syscalls which has `Mut*` array ptr params. Can you check that please?

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -738,6 +797,41 @@ impl<'a, 'b> SyscallsInvocator<'a, 'b> {
     }
 }
 
+/// Represents memory layout that can be safely used between syscalls and instructions.
+pub struct MemoryLayout {
```
techraed, line 801: I'd move it from invocator. The type is not only related to invocator.

techraed (commented):
the idea with `MemoryLayout` is great!

utils/wasm-gen/src/generator/entry_points.rs
```diff
@@ -211,6 +214,24 @@ impl<'a, 'b> EntryPointsGenerator<'a, 'b> {
         }
         res.push(Instruction::Call(export_body_call_idx as u32));
         res.extend(results.iter().map(|_| Instruction::Drop));
+
+        // after initializing the program, we will write about this in a special pointer
+        if name == "init" {
+            if let Some(memory_size_pages) = self.module.initial_mem_size() {
```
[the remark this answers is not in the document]
StackOverflowExcept1on: The order is always as you say:
https://github.com/gear-tech/gear/blob/9a37ccb406de68af8ddc6c851faa0c90ba0f0918/utils/wasm-gen/src/generator.rs#L122-L128

I would change `if let Some(_)` to `.expect(...)`

utils/wasm-gen/src/generator/syscalls/imports.rs
StackOverflowExcept1on: Each precise syscall can use `MemoryLayout` to store some data. But I think in general we just want to allocate memory for the current precise syscall.

utils/wasm-gen/src/generator/syscalls/invocator.rs
StackOverflowExcept1on: No, it works something like this:
```
upper_limit = mem_size - 256

offset = int_in_range(0..=upper_limit)
length = int_in_range(0..=(upper_limit - offset))
```

For example, let's see what happens at the maximum:
```
upper_limit = mem_size - 256

offset = upper_limit
length = (upper_limit - offset) = 0
```

We can point to reserved memory, but with length 0.

utils/wasm-gen/src/config.rs
```diff
@@ -156,6 +161,7 @@ impl Default for StandardGearWasmConfigsBundle {
     fn default() -> Self {
         Self {
             log_info: Some("StandardGearWasmConfigsBundle".into()),
+            waiting_probability: Some(4),
```
breathx, line 164: I'd reduce it to 8-12 mb

utils/wasm-gen/src/config/syscalls.rs
```diff
@@ -114,6 +122,7 @@ pub struct SyscallsConfig {
     precise_syscalls_config: PreciseSyscallsConfig,
     error_processing_config: ErrorProcessingConfig,
     log_info: Option<String>,
+    waiting_probability: Option<u32>,
```
breathx, line 125: NonZeroU32?

breathx (approved):

utils/wasm-gen/src/config.rs
StackOverflowExcept1on: We have control instructions. I.e. wasm-gen can produce
```rust
if some_cond {
    while some_cond2 {
        if some_cond3 {
            gr_wait() //with call chance 1/8 or 1/12...?
        }
    }
}
```

utils/wasm-gen/src/generator/entry_points.rs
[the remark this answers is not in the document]
techraed: Right, it's if you do it from the main generator, but you can also do everything manually - all these types are public.

utils/wasm-gen/src/generator/entry_points.rs
```diff
@@ -99,9 +103,11 @@ impl<'a, 'b> EntryPointsGenerator<'a, 'b> {
     /// Returns disabled entry points generator and a proof that all entry points from config were generated.
     pub fn generate_entry_points(
         mut self,
+        mem_import_gen_proof: MemoryImportGenerationProof,
```
techraed, line 106: make the proof to be a requirement for `EntryPointsGenerator` instantiation

techraed: also adjust docs stating this

utils/wasm-gen/src/generator/entry_points.rs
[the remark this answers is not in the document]
techraed: I see now MemoryImportGenerationProof is used in `EntryPointsGenerator`.

utils/wasm-gen/src/memory_layout.rs
```diff
@@ -0,0 +1,63 @@
+// This file is part of Gear.
```
techraed, line 1: why here? you can move it to `wasm` module. There are related to memory entities too

techraed (approved):

utils/wasm-gen/src/generator/entry_points.rs
StackOverflowExcept1on: I'm not sure if this can be done semantically correctly since `EntryPointsGenerator` is created from `GearWasmGenerator`. This means that we need to move `MemoryImportGenerationProof` to `GearWasmGenerator` as well.

StackOverflowExcept1on: Done in the last two commits:
- now `generate_entry_points` is very similar to `generate_syscalls`
- new type `EntryPointsGeneratorInstantiator` has been introduced (`generate_syscalls` does that too)

techraed (approved):

PR #3616, opened 2023-12-22
chore(ci): Use `debug` & `release` profiles

ark0f:
Rust:
- Fix `gring`'s `verbose` and `vanity` flags have the same short `v` name
- Fix `gcli` uses incorrect type for `verbose` flag
- Fix `gcli` uses debug node for tests and invalid profile flag to build demo
- Make benchmark listing consistent between `pallet-gear` and `pallet-gear-voucher`
- Fix missing `common/runtime-benchmarks` feature in `pallet-gear-voucher`

CI:
- Use debug profile in some bash scripts
- Add labels to run native Windows and release builds in PR
- Add fast benchmark check before long running check
- Use dynamic matrix to run build in release and debug modes
- Make `macos` and `win-native` builds cancelable
- Remove obsolete caching


clearloop (approved):

clearloop (commented):
The reason of using release profile in gcli tests is that testing with release built node runs faster, so since under debug profile with `opt-level = 3` gcli can still pass the tests, everything still works

ukint-vs (approved):

gshep (approved):

shamilsan (approved):
Please write a couple of words in the PR description what is the reason of moving to debug in CI.

breathx (changes requested):
What's the point not to run it in release + decreasing of the node perf?

ark0f:
> What's the point not to run it in release

The only reason why we ever use release profile is slow block production in node. `debug` profile adds extended capabilities  in terms of assertions, coverage and debugging

> decreasing of the node perf

Decreasing is insignificant and tests are not about performance. They are about correctness

breathx (changes requested):
Needs changes according to collective discussion

breathx:
@clearloop review pls

.github/workflows/PR.yml
```diff
@@ -45,8 +45,8 @@ jobs:
       cache: ${{ needs.status.outputs.cache }}
 
   build:
-    needs: status
-    if: ${{ needs.status.outputs.build == 'true' }}
+    #needs: status
```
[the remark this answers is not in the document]
clearloop: I guess this PR is still under development, plz remove `A0-pleasereview` tag if it is @ark0f

.github/workflows/build.yml
```diff
@@ -28,8 +28,21 @@ env:
...
 
 jobs:
+  matrix:
+    runs-on: [kuberunner, github-runner-02]
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - id: set-matrix
+        run: |
+          echo '::set-output name=matrix::{"include": [{"cargo_profile": "", "cargo_target_dir": "debug"}, {"cargo_profile": "--release", "cargo_target_dir": "release"}]}'
```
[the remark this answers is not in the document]
clearloop: this seems like hard coded matrix inputs, https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations, still under development?

clearloop (changes requested):
plz remove `A0-pleasereview` or add `A5-dontmerge` if this pr is still under development

matrix for processing both `dev` and `release` looks overkill, add rustc flag `-g` to release build could be an option as well https://doc.rust-lang.org/rustc/codegen-options/index.html#debuginfo

breathx (commented):
- [ ] Remove trash branches and commented code
- [ ] Revert fix of authorship

.github/workflows/build-win-native.yml
```diff
@@ -0,0 +1,105 @@
...
+        if: ${{ matrix.profile == 'release' }}
+        run: cargo run -p calc-stack-height --release --locked
+
+      - name: "Test: gsdk tests"
+        run: ./scripts/gear.sh test gsdk ${{ matrix.profile_flags }}
+
+      - name: "Test: Client tests"
+        run: ./scripts/gear.sh test client ${{ matrix.profile_flags }}
+
+      - name: "Test: Lazy pages"
```
grishasobol, line 99: please rename here and in other places. maybe "workspace tests" or "pallet, lazy-pages and runtime tests". Actually I don't know whether we don't run all workspace tests here

grishasobol (approved):

gshep (approved):
Awesome! :1st_place_medal:

.github/workflows/build-win-native.yml
ark0f: I don't see reason to rename the step as its purpose is to test lazy-pages despite the step runs all of these tests

ark0f:
Awaiting for #3803

clearloop (approved):

.github/workflows/build-win-native.yml
grishasobol: So, it leads to misleading when some pallet tests will fail, because it won't be actually lazy-pages problem

grishasobol: Why we don't call `gear.sh test gear` here ?

PR #3624, opened 2023-12-26
feat(core, runtime): Builtin actors framework

ekovalev:
Following the discussion on reworking the original approach to builtin actors (#3504), a new extensible framework for Runtime builtin actors is proposed.

A set of `BuiltinActor`'s is defined as the `pallet_gear_builtin_actor::Config` trait associated type - a tuple consisting of custom types each of which must implement the `pallet_gear_builtin_actor::BuiltinActor` trait (similar to how migrations or signed extras are defined in the Runtime). Those can be pallets (if access to on-chain storage is required) or any other custom types.
To enable messages routing to a specific builtin actor, the latter must be registered with the `pallet_gear::Config::<T>::BuiltinRouter` associated type that implements the `pallet_gear::builtin::BuiltinRouter` trait and provides the `lookup()` function to derive the respective `BuiltinId` based on the message destination address (`ProgramId`).
This functionality is provided by the `gear_builtin_actor` pallet.

The `BuiltinId` must be unique across all actors: it is claimed by a builtin actor itself as the actor's type's associated constant. The `gear-builtin-actor` pallet provides the `register_actor()` function that would check if there are conflicting `BuiltinId`'s. This should be the preferred way to register an actor during Runtime upgrades.

This PR includes:
- [x] builtin actor framework implementation: new assoc. type in the `Gear` pallet and the `GearBuiltinActor` pallet
- [x] unit tests (more can be add in the future)
- [x] migration template to register new actors
- [x] RPC to query actors' `ProgramId`'s
- [x] docs
- [x] adjustment of the `calculate_gas_info()` for the case of a builtin actor.

Examples of actual "production" builtin actor (like `staking-proxy` one) are beyond the scope of this PR and will be added subsequently.


pallets/gear/src/builtin.rs
```diff
@@ -0,0 +1,66 @@
...
+use impl_trait_for_tuples::impl_for_tuples;
+use sp_runtime::DispatchError;
+
+/// A trait representing a registry that provides methods to lookup a builtin actor
+pub trait BuiltinLookup<ActorId> {
+    /// Returns available actors identifiers.
+    fn lookup(id: &ActorId) -> Option<BuiltinId>;
+}
+
+impl<ActorId> BuiltinLookup<ActorId> for () {
```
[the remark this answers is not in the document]
ekovalev: To avoid builtin actors self-assigning addresses that can potentially clash with `AccountId`'s, `ProgramId`'s and other 32 bit addresses we use in our system. Hence a separate id which serves as a seed. For instance, pallets have `PalletId` that is used to derive an actual account id - not hard-code the account id itself.

mertwole (commented):
Can we add one more call to built-in actor that will contain sender address?

gshep:
> Can we add one more call to built-in actor that will contain sender address?

Perhaps you can use `source` of the message

mertwole:
> Perhaps you can use `source` of the message

Yes, I meant that

pallets/gear-builtin-actor/src/lib.rs
```diff
@@ -65,6 +65,37 @@ pub use pallet::*;
...
+    destination: BuiltinId,
+    payload: Vec<u8>,
+}
+
+impl Dispatchable for SimpleBuiltinMessage {
+    type Payload = Vec<u8>;
+
+    fn source(&self) -> ProgramId {
+        self.source
+    }
```
mertwole, line 88: We need this not for all use-cases, so I think it can be removed from trait

pallets/gear-builtin-actor/src/lib.rs
```diff
@@ -65,6 +65,37 @@ pub use pallet::*;
 #[allow(dead_code)]
 const LOG_TARGET: &str = "gear::builtin_actor";
 
+pub trait Dispatchable {
+    type Payload: AsRef<[u8]>;
```
mertwole, line 69: Why do we need this associated type? We already have abstraction over payload in `payload_bytes` returning `&[u8]`

pallets/gear-builtin-actor/src/benchmarking.rs
```diff
@@ -0,0 +1,133 @@
...
+        let _ = BuiltinActorPallet::<T>::register_actor::<DummyActor1, _, _>();
+        let _ = BuiltinActorPallet::<T>::register_actor::<DummyActor2, _, _>();
+        let _ = BuiltinActorPallet::<T>::register_actor::<DummyActor3, _, _>();
+        let _ = BuiltinActorPallet::<T>::register_actor::<DummyActor4, _, _>();
+        let _ = BuiltinActorPallet::<T>::register_actor::<DummyActor5, _, _>();
+        let _ = BuiltinActorPallet::<T>::register_actor::<DummyActor6, _, _>();
+        let _ = BuiltinActorPallet::<T>::register_actor::<DummyActor7, _, _>();
+        let _ = BuiltinActorPallet::<T>::register_actor::<DummyActor8, _, _>();
+        let _ = BuiltinActorPallet::<T>::register_actor::<DummyActor9, _, _>();
+        let _ = BuiltinActorPallet::<T>::register_actor::<DummyActor10, _, _>();
```
mertwole, line 97: `handle` will not work with regiustered built-in actors as it takes `BuiltinId` as the input and therefore don't perform lookup

pallets/gear-builtin-actor/README.md
mertwole, line 1: Please specify here or somewhere else the workflow of adding my built-in actor to the runtime as this process is a little bit complicated

pallets/gear/src/queue.rs
```diff
@@ -202,6 +205,25 @@ where
...
+                        // All other cases result in no-execution outcome.
+                        let dispatch = dispatch.into_incoming(gas_limit);
+                        let system_reservation_ctx =
+                            SystemReservationContext::from_dispatch(&dispatch);
+                        process_non_executable(dispatch, program_id, system_reservation_ctx)
+                    }
+                };
+                core_processor::handle_journal(journal, &mut ext_manager);
+                continue;
+            }
```
gshep, line 225: I would like to suggest another implementation. The key point is that we don't need to keep builtin program ids in a storage: I've benched generation of ids and it took about `1_000_000` of gas that is much lesser than one DB read. So the idea is that gear's Config has `BuiltinRouterProvider` that creates an instance of `BuiltinRouter` which has an array with all available builtins.


```suggestion
            // In case the destination is a builtin actor, process the dispatch differently.
            let maybe_journal = match dispatch.kind() {
                DispatchKind::Handle => {
                    builtin_router.dispatch(builtin_id, dispatch.clone(), gas_limit)
                }
                _ => {
                    // Builtin actors can only execute `handle` messages.
                    // All other cases result in no-execution outcome.
                    let dispatch = dispatch.into_incoming(gas_limit);
                    let system_reservation_ctx =
                        SystemReservationContext::from_dispatch(&dispatch);

                    Some(process_non_executable(dispatch, program_id, system_reservation_ctx))
                }
            };
            
            if let Some(journal) = maybe_journal {
                core_processor::handle_journal(journal, &mut ext_manager);
                continue;
            }
```

So our `run_queue` would look like:
```rust
fn run_queue(initial_gas: Self::Gas) -> Self::Gas {
            let gas_for_builtin_router = T::BuiltinRouterProvider::required_gas();

            // Setting adjusted initial gas allowance
            GasAllowanceOf::<T>::put(initial_gas.saturating_sub(gas_for_builtin_router));
            
            let builtin_router = T::BuiltinRouterProvider::new();

            // Ext manager creation.
            // It will be processing messages execution results following its `JournalHandler` trait implementation.
            // It also will handle delayed tasks following `TasksHandler`.
            let mut ext_manager = Default::default();

            // Processing regular and delayed tasks.
            Self::process_tasks(&mut ext_manager);

            // Processing message queue.
            Self::process_queue(ext_manager, builtin_router);

            // Calculating weight burned within the block.
            initial_gas.saturating_sub(GasAllowanceOf::<T>::get())
        }
```

The next point is that `max_gas` isn't necessary: let builtins handle gas at their own. That was an issue with bls381: the maximum gas cannot be determined since it depends on the number of arguments which should be scale-decoded first.

Sorry for my vague thoughts, let's have a call to dicuss it ).

ekovalev: Thanks for the input!

Regarding the `max_gas`, the reason it is there is that, unlike in usual actors, we can't track gas spent as we progress, and stop processing once the gas has run out. In the case of `builtins`, once it's started processing it will have to finish it. Sure there is no chance of running processing indefinitely (because the code of builtin actors comes from us and we are certain nothing of that kind if possible) but the `payload`, when you decode it, may still require using more gas than the user may have provided. So we should either do message processing in two steps: decode the `payload` first and find out what the exact cost will be and then do actual processing, or charge the maximum possible gas upfront and then return the unused part.
Currently, I have opted for the latter approach.

pallets/gear-builtin-actor/src/lib.rs
```diff
@@ -0,0 +1,403 @@
...
+use sp_runtime::traits::{TrailingZeroInput, Zero};
+use sp_std::prelude::*;
+pub use weights::WeightInfo;
+
+pub use pallet::*;
+
+#[allow(dead_code)]
+const LOG_TARGET: &str = "gear::builtin_actor";
+
+pub trait Dispatchable {
```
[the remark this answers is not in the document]
ekovalev: > I think it's no need to introduce this trait as the message format will be dictated by this pallet anyway

We don't force a specific message type anywhere in the code. Everything a builtin actor should know about a message is that it has a source, a destination (supposedly, this actor's `BuiltinId`) and a payload convertible to bytes. I personally don't see a problem abstracting away the concrete message type.

Otherwise, your argument can be applied to a thousand different places in Substrate code (like, for instance, every pallet being the only implementation of its `Config` trait etc.).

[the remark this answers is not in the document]
ekovalev: > I think it's no need to introduce this trait as the message format will be dictated by this pallet anyway

We don't force a specific message type anywhere in the code. Everything a builtin actor should know about a message is that it has a source, a destination (supposedly, this actor's `BuiltinId`) and a payload convertible to bytes. I personally don't see a problem abstracting away the concrete message type.

Otherwise, your argument can be applied to a thousand different places in Substrate code (like, for instance, every pallet being the only implementation of its `Config` trait etc.).

pallets/gear-builtin-actor/src/benchmarking.rs
```diff
@@ -0,0 +1,133 @@
...
+impl_builtin_actor!(DummyActor6, 6);
+impl_builtin_actor!(DummyActor7, 7);
+impl_builtin_actor!(DummyActor8, 8);
+impl_builtin_actor!(DummyActor9, 9);
+impl_builtin_actor!(DummyActor10, 10);
+
+benchmarks! {
+    lookup {
+        // Populate the storage with maximum (16, as of today) builtin actors ids
+        for i in 0_u64..16 {
```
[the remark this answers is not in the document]
ekovalev: > magic number `16`

Just an uninformed guess of how many builtin actors we might have in the runtime.

pallets/gear-builtin-actor/src/benchmarking.rs
ekovalev: We are not benchmarking `lookup` with this test. The reason for registering a bunch of actors is to populate the storage to simulate iteration through a tuple elements that might be taking place in reality.
This test is rather useless anyway because with 16 actors it takes literally 0 weight to identify the type whose `handle` should be called. Keeping it exclusively for the sake of generality.

mertwole: Yes, I mean that `lookup` isn't executed, so you don't need to fill storage as it'll be not queried. What actually will be used are 2 of built-in actors you registered in mock

pallets/gear-builtin-actor/src/benchmarking.rs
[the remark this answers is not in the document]
mertwole: please introduce a constant for convinience

pallets/gear-builtin-actor/src/lib.rs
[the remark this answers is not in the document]
mertwole: then ok, if the purpose is to abstract message format from built-in actors themselves. But you have `type BuiltinActor: BuiltinActor<SimpleBuiltinMessage, u64>;` in the pallet `Config`

pallets/gear-builtin-actor/README.md
ekovalev: The workflow is standard for Substrate (same way we add migrations or signed extras):
- you implement the `BuiltinActor` trait with your type or pallet
- in the `pallet_bulitin_actor::Config` trait implementation for the `Runtime` add your type like so:
```
impl pallet_gear_builtin_actor::Config for Runtime {
    type BuiltinActor = (Actor1, Actor2, ..., YourActorType);
    ...
}
```

See the mock runtime for the `pallet_builtin_actor` tests for an example.

pallets/gear-builtin-actor/src/benchmarking.rs
ekovalev: Anyway, all this doesn't matter anymore, because in the latest version of the framework we completely outed out from storing known builtin ids in favour of deriving them in every block on the fly (which is at least 5 times cheaper based on our benchmarks comparing to fetching values from storage).

pallets/gear-builtin-actor/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+    type QueuedDispatch = StoredDispatch;
+    type Output = JournalNote;
+
+    fn lookup(&self, id: &ProgramId) -> Option<BuiltinId> {
+        self.registry.get(id).copied()
+    }
+
+    fn dispatch(&self, dispatch: StoredDispatch, gas_limit: u64) -> Option<Vec<JournalNote>> {
+        let actor_id = dispatch.destination();
+        if let Some(builtin_id) = self.registry.get(&actor_id) {
```
gshep, line 339: to decrease nesting:
```rust
let Some(builtin_id) = self.registry.get(&actor_id) else {
    return None;
}

...
```

core/src/ids.rs
```diff
@@ -243,6 +247,23 @@ impl ReservationId {
     }
 }
 
+/// A builtin actor type identifier. Serves as a seed for a builtin actor ID generation.
+#[derive(Clone, Copy, Debug, Decode, Encode, Eq, PartialEq, TypeInfo)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
```
gshep, line 252: does it really need to derive serde traits? If I'm not mistaken that functionality isn't used

pallets/gear/src/queue.rs
```diff
@@ -167,6 +167,9 @@ where
             T::DebugInfo::remap_id();
         }
 
+        // Create an instance of a builtin router
+        let builtin_router = T::BuiltinRouter::provide();
```
gshep, line 171: The last one thing I would like to mention is that we need to bench this method and decrease initial_gas for block accordingly. That idea was demonstrated in the previous comment with a sample.

gshep (approved):
Awesome! :100:

pallets/gear-builtin-actor/src/lib.rs
```diff
@@ -0,0 +1,383 @@
...
+}
+
+pub struct BuiltinRegistry<T: Config> {
+    pub registry: BTreeMap<ProgramId, BuiltinId>,
+    pub _phantom: sp_std::marker::PhantomData<T>,
+}
+impl<T: Config> BuiltinRegistry<T> {
+    fn new() -> Self {
+        let mut registry = BTreeMap::new();
+        let mut builtin_ids = Vec::new();
```
gshep, line 310: It is better to preallocate some space for items. Are we able to use count from an `impl_for_tuples` attribute for that?

core/src/ids.rs
ekovalev: It should be required by some trait bound - otherwise I wouldn't have made it derive those traits. But I'll double check.

pallets/gear-builtin-actor/src/lib.rs
ekovalev: > It is better to preallocate some space for items. Are we able to use count from an `impl_for_tuples` attribute for that?

We use a hardcoded constant - the max number of elements in a tuple (that is, the max supported number of builtin actors) currently set to 16. Could reuse it here.

pallets/gear/src/queue.rs
ekovalev: Implemented.

pallets/gear-builtin-actor/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+
+/// Built-in actor error type
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq, derive_more::Display)]
+pub enum BuiltinActorError {
+    /// Occurs if the underlying call has the weight greater than the `gas_limit`.
+    #[display(fmt = "Not enough gas supplied")]
+    InsufficientGas,
+    /// Occurs if the dispatch's message can't be decoded into a known type.
+    #[display(fmt = "Failure to decode message")]
+    UnknownMessageType,
```
breathx, line 109: More likely UnknownCall or sth

pallets/gear-builtin-actor/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+    /// Occurs if a builtin id doesn't belong to any registered actor.
+    #[display(fmt = "Unknown builtin id")]
+    UnknownBuiltinId,
```
breathx, line 112: state that its supposed to be unreachable

breathx: UnknownActor

pallets/gear-builtin-actor/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+pub trait Dispatchable {
+    type Payload: AsRef<[u8]>;
+
+    fn source(&self) -> ProgramId;
+    fn destination(&self) -> BuiltinId;
+    fn payload_bytes(&self) -> &[u8];
+}
+
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
+pub struct SimpleBuiltinMessage {
```
breathx, line 79: you can avoid this type by operating some config asoc: Dispatchable

pallets/gear-builtin-actor/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+
+pub type BuiltinResult<P> = Result<P, BuiltinActorError>;
+
+/// A trait representing an interface of a builtin actor that can receive a message
+/// and produce a set of outputs that can then be converted into a reply message.
+pub trait BuiltinActor<Message: Dispatchable, Gas: Zero> {
+    /// Handles a message and returns a result and the actual gas spent.
+    fn handle(message: &Message, gas_limit: Gas) -> (BuiltinResult<Message::Payload>, Gas);
+
+    fn get_ids(buffer: &mut Vec<BuiltinId>);
```
breathx, line 135: fn get_ids() -> BTreeSet<>?

pallets/gear-builtin-actor/src/lib.rs
breathx: as well as impl from into to message type used in send_dispatch ext_manager

pallets/gear-builtin-actor/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+
+    fn payload_bytes(&self) -> &[u8] {
+        self.payload.as_ref()
+    }
+}
+
+/// Built-in actor error type
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq, derive_more::Display)]
+pub enum BuiltinActorError {
+    /// Occurs if the underlying call has the weight greater than the `gas_limit`.
```
breathx, line 104: what if it failed call?

breathx: I guess builtin::Result may be `Result<Vec<u8>, Gas | Decode | DispatchFail(Vec<u8>)>` but we need to somehow guarantee here that there is no opportunity for UnknownId and remove it, cause it must be controlled by the caller

pallets/gear-builtin-actor/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+    #[pallet::hooks]
+    impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
+
+    impl<T: Config> Pallet<T> {
+        /// Generate an `actor_id` given a builtin ID.
+        ///
+        ///
+        /// This does computations, therefore we should seek to cache the value at the time of
+        /// a builtin actor registration.
+        pub fn generate_actor_id(builtin_id: BuiltinId) -> ProgramId {
```
breathx, line 200: we can do similar trick with that is used for vouchers hash(["bultin", idx]) so we will avoid using pallet id that could change from chain to chain based on gear

breathx: and avoid adding rpc api

pallets/gear-builtin-actor/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+            let entropy = (T::PalletId::get(), builtin_id).using_encoded(blake2_256);
+            let actor_id = Decode::decode(&mut TrailingZeroInput::new(entropy.as_ref()))
+                .expect("infinite length input; no invalid inputs for type; qed");
+            actor_id
+        }
+
+        pub fn process_success(
```
breathx, line 207: I'd reuse core_processor logic since we have those fns there

breathx: same for errors

breathx: btw, if I'm not mistaken we use payload bytes with string explanation only for userspace panic case and it has dedicated backend format, see [`TrapExplanation::Trap`]

pallets/gear-builtin-actor/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+
+            journal
+        }
+    }
+}
+
+impl<T: Config> BuiltinRouterProvider<StoredDispatch, JournalNote, u64> for Pallet<T> {
+    type Router = BuiltinRegistry<T>;
+
+    fn provide() -> BuiltinRegistry<T> {
```
breathx, line 305: why not just `create` or `compact` whatever?

pallets/gear-builtin-actor/src/lib.rs
breathx, line 207: this fns should be called across trait impls and call results, isn't it?

pallets/gear-builtin-actor/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+            });
+
+            journal.push(JournalNote::MessageConsumed(message_id));
+
+            journal
+        }
+    }
+}
+
+impl<T: Config> BuiltinRouterProvider<StoredDispatch, JournalNote, u64> for Pallet<T> {
```
breathx, line 302: we need impl for () somewhere around I guess

pallets/gear-builtin-actor/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+        BuiltinRegistry::<T>::new()
+    }
+
+    fn provision_cost() -> u64 {
+        <T as Config>::WeightInfo::provide().ref_time()
+    }
+}
+
+pub struct BuiltinRegistry<T: Config> {
+    pub registry: BTreeMap<ProgramId, BuiltinId>,
```
breathx, line 315: why do we need to key-value same types?

gbuiltin/src/lib.rs
```diff
@@ -0,0 +1,151 @@
...
+    #[derive(Encode, Decode, Clone, Debug, PartialEq, Eq, derive_more::Display)]
+    #[codec(crate = scale)]
+    pub enum DispatchErrorReason {
+        #[display(fmt = "Runtime internal error")]
+        RuntimeError,
+    }
+}
+
+// TODO: review and adjust
+pub mod staking {
```
breathx, line 96: these should be deleted for now

breathx: and qs to consider: in case of incompatibility of the api shouldn't we truly create a new builtin?

breathx (changes requested):
Didn't finish review yet, but these comments are enough to talk about for now

breathx:
And just some thoughts about future staking impl. I still can't figure out for myself which approach is better: take the value from message applied or directly from account. I guess we need some pros and cons list

pallets/gear-builtin-actor/src/lib.rs
ekovalev: It's a reverse mapping with respect to the original hashing from `BuiltinId` (a.k.a. `u64`) to `ProgramId`. These types are obviously not the same.

pallets/gear-builtin-actor/rpc/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+impl<C, Block> GearBuiltinActorApiServer<<Block as BlockT>::Hash, H256>
+    for GearBuiltinActor<C, Block>
+where
+    Block: BlockT,
+    C: 'static + ProvideRuntimeApi<Block> + HeaderBackend<Block>,
+    C::Api: GearBuiltinActorRuntimeApi<Block>,
+{
+    fn generate_actor_id(&self, builtin_id: u64) -> RpcResult<H256> {
```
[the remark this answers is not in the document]
ekovalev: We can get away without it, of course, if `ProgramIds` corresponding to builtin actors are published somewhere so people (and programs) can send messages to them.
But if you use an rpc like this one you always have a verified `ProgramId` (up-to-date), without having to worry if the hashing algorithm or seed have changed since the `ProgramIds` were published etc. The last thing a program developer wants is to realize they hardcoded a wrong address of a builtin actor when the program has already been uploaded.

pallets/gear-builtin-actor/src/lib.rs
ekovalev: It doesn't have to be a call. It's up to each actor `handle()` function to decode and deal with the message payload. In case of `staking` it would decode a call - that's true.

pallets/gear-builtin-actor/src/lib.rs
ekovalev: that makes sense.

pallets/gear-builtin-actor/src/lib.rs
ekovalev: not sure we can wrap this one inside `impl_for_tuples`. That's the reason of having it `C-style`. But I'll check.

pallets/gear-builtin-actor/src/lib.rs
ekovalev: Agree with the first comment. Still believe having an rpc is useful to make sure the respective `ProgramId` address is always up-to-date for a builtin id.

pallets/gear-builtin-actor/src/lib.rs
ekovalev: Because providers provide, don't they? Or do I not understand the question?

gbuiltin/src/lib.rs
NikVolf: This is indeed a better approach I think

pallets/gear-builtin-actor/src/lib.rs
ekovalev: > we need impl for () somewhere around I guess

There is - it's just after the trait definition.

pallets/gear-builtin-actor/src/lib.rs
ekovalev: Ok, will try to use the original `process_success` and `process_error` functions from the `core-processor`. It's just that those functions are too closely coupled with wasm execution process so I thought it'd be easier to implement separate versions for builtin flow than try to adapt the existing ones.
As you mention, to send back execution error reason I'd have to import the `TrapExplanation` from core-backend which is completely out of the context, whereas what I actually need is just send some bytes.
Anyway, let's see what I can do.

pallets/gear/src/builtin.rs
```diff
@@ -0,0 +1,70 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use super::*;
+use gear_core::ids::{BuiltinId, ProgramId};
+
+/// A trait representing a registry that provides methods to lookup a builtin actor
+pub trait BuiltinRouter {
+    type QueuedDispatch;
+    type Output;
```
breathx, line 25: I guess at least Output may be removed in favour of just JournalNote.
And anyway if you disagree it more likely that dispatch will -> Option<Self::Output> where Output = Vec<_>

pallets/gear/src/builtin.rs
```diff
@@ -0,0 +1,70 @@
...
+
+    /// Looks up a builtin actor by its actor id.
+    fn lookup(&self, id: &ProgramId) -> Option<BuiltinId>;
```
breathx, line 28: why do you need BuiltinId if you already have ProgramId?
Maybe just true/false?

pallets/gear/src/runtime_api.rs
```diff
@@ -180,27 +183,30 @@ where
...
-            let step = QueueStep {
-                block_config: &block_config,
-                ext_manager: &mut ext_manager,
-                gas_limit,
-                dispatch: queued_dispatch,
-                balance: balance.unique_saturated_into(),
-            };
-
-            let journal = Self::run_queue_step(step);
+                .map_err(|_| b"Internal error: unable to get gas limit".to_vec())?;
```
breathx, line 187: use `internal_err()` with formatting similar to other places

pallets/gear/src/lib.rs
```diff
@@ -257,6 +259,9 @@ pub mod pallet {
         /// rent is disabled.
         #[pallet::constant]
         type ProgramRentDisabledDelta: Get<BlockNumberFor<Self>>;
+
+        /// The builtin actors registry.
+        type BuiltinRouter: BuiltinRouterProvider<StoredDispatch, JournalNote, u64>;
```
breathx, line 264: I prefer named generics here

pallets/gear/src/lib.rs
```diff
@@ -805,11 +810,12 @@ pub mod pallet {
                 .unwrap_or(false)
         }
 
-        /// Returns true if id is a program and the program has active status.
+        /// Returns true if `program_id` is that of a in active status or the builtin actor.
         pub fn is_active(program_id: ProgramId) -> bool {
-            ProgramStorageOf::<T>::get_program(program_id)
-                .map(|program| program.is_active())
-                .unwrap_or_default()
+            T::BuiltinRouter::provide().lookup(&program_id).is_some()
```
breathx, line 815: can we not to `provide()` it here every time?
checkout the usage and maybe it will be even unnecessary, or pass the ref to router

pallets/gear/src/lib.rs
```diff
@@ -826,10 +832,12 @@ pub mod pallet {
                 .unwrap_or_default()
         }
 
-        /// Returns true if there is a program with the specified id (it may be paused).
+        /// Returns true if there is a program with the specified `program_id`` (it may be paused)
+        /// or this `program_id` belongs to the built-in actor.
         pub fn program_exists(program_id: ProgramId) -> bool {
             ProgramStorageOf::<T>::program_exists(program_id)
                 || ProgramStorageOf::<T>::paused_program_exists(&program_id)
+                || T::BuiltinRouter::provide().lookup(&program_id).is_some()
```
breathx, line 840: btw it's conceptually not a program, but actor, so consider removing here as well as above

pallets/gear/src/lib.rs
```diff
@@ -1842,8 +1850,12 @@ pub mod pallet {
...
 
         fn run_queue(initial_gas: Self::Gas) -> Self::Gas {
-            // Setting adjusted initial gas allowance
-            GasAllowanceOf::<T>::put(initial_gas);
+            // Take note of how much gas we will need for builtin router creation.
+            let gas_for_builtin_router = T::BuiltinRouter::provision_cost();
+
+            // Setting adjusted initial gas allowance taking into account builtin router creation
+            // cost that will be incurred inside the `process_queue()` function.
+            GasAllowanceOf::<T>::put(initial_gas.saturating_sub(gas_for_builtin_router));
```
breathx, line 1858: found the place of this charge, I'd really provide() here in place to keep it together for the eyes

runtime/vara/src/lib.rs
```diff
@@ -1173,6 +1191,7 @@ construct_runtime!(
         StakingRewards: pallet_gear_staking_rewards = 106,
         GearVoucher: pallet_gear_voucher = 107,
         GearBank: pallet_gear_bank = 108,
+        GearBuiltinActor: pallet_gear_builtin = 109,
```
breathx, line 1194: just GearBuiltin?

runtime/vara/src/lib.rs
```diff
@@ -1048,6 +1049,23 @@ impl pallet_gear_messenger::Config for Runtime {
...
+#[cfg(not(feature = "runtime-benchmarks"))]
+pub type BuiltinActors = ();
+#[cfg(feature = "runtime-benchmarks")]
+pub type BuiltinActors = pallet_gear_builtin::benchmarking::BenchmarkingBuiltinActor<Runtime>;
+
+parameter_types! {
+    pub const BuiltinActorPalletId: PalletId = PalletId(*b"py/biact");
+}
+
+impl pallet_gear_builtin::Config for Runtime {
```
breathx, line 1062: btw couldn't find changes to gsdk generated code. are they not necessary here?

pallets/gear-builtin/src/benchmarking.rs
```diff
@@ -0,0 +1,141 @@
+// This file is part of Gear.
+
+// Copyright (C) 2022-2023 Gear Technologies Inc.
```
breathx, line 3: copyright (C) 2024 everywhere in new crates

pallets/gear-builtin/src/benchmarking.rs
```diff
@@ -0,0 +1,141 @@
...
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//! Benchmarks for the gear-built-in-actor pallet
```
breathx, line 19:
```suggestion
//! Benchmarks for the `pallet-gear-builtin`
```

breathx (changes requested):
`pallets/gear-builtin` is still under review and will be submitted asap

runtime/vara/src/lib.rs
ekovalev: apparently, they are not. At least the script didn't generate any new metadata. Anyway, will re-run it.

pallets/gear/src/lib.rs
ekovalev: The `program_exists` function must be aware of the Builtin destinations. Agree, that "providing" every time is sub-optimal, but I don't see how we could remove this check. Because all "handle" messages destinations have to meet this requirement, do they?

pallets/gear/src/builtin.rs
ekovalev: A message address is a `ProgramId`, but in order to find the correct actor who is going to process it we must "reverse-map" it to a `BuiltinId` since builtin actors only know their own `BuiltinId`.

pallets/gear/src/builtin.rs
ekovalev: Probably this makes sense: abstracting away the `JournalNote` type is unnecessary here.

pallets/gear-builtin/rpc/runtime-api/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use sp_core::H256;
+
+sp_api::decl_runtime_apis! {
+    pub trait GearBuiltinActorApi {
+        /// Calculate `ProgramId` (a.k.a. actor id) given a `BuiltinId`.
+        fn generate_actor_id(builtin_id: u64) -> H256;
```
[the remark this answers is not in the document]
ekovalev: It's just all other 32-byte ids used as input or output types in RPC calls are represented by H256. Haven't found a single RPC call that would return a core id type back to the calle. Maybe I need to look more carefully.

node/authorship/src/tests.rs
```diff
@@ -530,6 +530,8 @@ fn block_max_gas_works() {
 
     // Amount of gas burned in each block (even empty) by default
     const FIXED_BLOCK_GAS: u64 = 25_000_000;
+    // Gas cost of a built router instance creation
+    const BUILTIN_ROUTER_COST: u64 = 8_000_000;
```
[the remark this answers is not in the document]
ekovalev: right, it can be now. A leftover from when it wasn't in the WeightInfo.

breathx (changes requested):
`pallets/gear-builtin/src/tests/*` to be reviewed after suggested changes applied/discussed

core-processor/src/precharge.rs
```diff
@@ -185,6 +185,8 @@ pub fn calculate_gas_for_code(read_cost: u64, per_byte_cost: u64, code_len_bytes
     read_cost.saturating_add(code_len_bytes.saturating_mul(per_byte_cost))
 }
 
+/// Possible variants of the `DispatchResult` if the latter contains value.
+#[allow(missing_docs)]
```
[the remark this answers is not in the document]
ekovalev: exporting this enum will require documenting its every variant. A bit over the top in this case so I figured adding one extra line vs three would be less typing and keep the code more readable.

pallets/gear-builtin/Cargo.toml
```diff
@@ -0,0 +1,70 @@
...
+pallet-gear-gas = { workspace = true, features = ["std"] }
+pallet-gear-messenger = { workspace = true, features = ["std"] }
+pallet-gear-program = { workspace = true, features = ["std"] }
+pallet-gear-scheduler = { workspace = true, features = ["std"] }
+frame-executive = { workspace = true, features = ["std"] }
+frame-support-test = { workspace = true, features = ["std"] }
+env_logger.workspace = true
+hex-literal.workspace = true
+
+[features]
```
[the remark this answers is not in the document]
ekovalev: `try-runtime` definitely not - there is no any conditional code at all that is guarded by this feature in the `pallet-gear`.
For the `runtime-benchmarks`, it does enables some extra functions, but I don't see any obvious need for it in the context of the builtin pallet.

pallets/gear-builtin/rpc/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+    C: 'static + ProvideRuntimeApi<Block> + HeaderBackend<Block>,
+    C::Api: GearBuiltinActorRuntimeApi<Block>,
+{
+    fn generate_actor_id(&self, builtin_id: u64) -> RpcResult<H256> {
+        let api = self.client.runtime_api();
+        let best_hash = self.client.info().best_hash;
+
+        fn map_err(err: impl std::fmt::Debug, desc: &'static str) -> JsonRpseeError {
+            CallError::Custom(ErrorObject::owned(
+                Error::RuntimeError.into(),
```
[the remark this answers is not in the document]
ekovalev: > doesnt we return 9000 by default?

It's 8000 apparently. Aligned with the gear pallet rpc.

pallets/gear/src/builtin.rs
ekovalev: Another update: now we actually do need a value as the output of the `lookup` in order to enable the check whether a dispatch can be handled by a builtin actor before consuming it to avoid 
[excessive cloning of a dispatch](https://github.com/gear-tech/gear/pull/3624/files/6f271facf645b83cbab4ac57b5c24895cdfee62a#diff-25fc00efe7529b336fe02426d3e7fab7ab580dbb076b03e9e8b17cf632d2c35e).

pallets/gear-builtin/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+    fn from(err: BuiltinActorError) -> Self {
+        match err {
+            BuiltinActorError::InsufficientGas => {
+                ActorExecutionErrorReplyReason::Trap(TrapExplanation::GasLimitExceeded)
+            }
+            BuiltinActorError::UnknownMessageType => ActorExecutionErrorReplyReason::Trap(
+                TrapExplanation::Panic("Message decoding error".to_string().into()),
+            ),
+            // This should convey a message of hitting an unreachable state
+            BuiltinActorError::UnknownActor => {
```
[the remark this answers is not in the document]
ekovalev: Another update: now we actually do need a value as the output of the `lookup` in order to enable the check whether a dispatch can be handled by a builtin actor before consuming it (to avoid excessive cloning of a dispatch as discussed above in reference to the `queue.rs` from the `pallet_gear`).

pallets/gear-builtin/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+                    .to_vec()
+                    .try_into()
+                    .unwrap_or_else(|_| unreachable!("Response message is too large"));
+                let packet = ReplyPacket::new(payload, 0);
+
+                // Mark reply as sent
+                if let Ok(_reply_id) = message_context.reply_commit(packet.clone(), None) {
+                    let (outcome, context_store) = message_context.drain();
+
+                    dispatch_result.context_store = context_store;
```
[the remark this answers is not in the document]
ekovalev: Yes. It switches off auto-reply generation.

pallets/gear/src/lib.rs
ekovalev: **Update**: now these two actions (creating a `BuiltinDispatcher` and calculating the cost of this operation) are tied together.

pallets/gear-builtin/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+            destination: builtin_id,
+            payload: dispatch.payload_bytes().to_vec(),
+            _phantom: Default::default(),
+        })
+    }
+}
+
+/// Built-in actor error type
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq, derive_more::Display)]
+pub enum BuiltinActorError {
```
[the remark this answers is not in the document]
ekovalev: Ideally, every builtin actor should be allowed to return their own `Error` from the `handle()` function (which is generic over the error type). However, in current implementation we have to enforce a single error type to be shared by all builtin actors in the system if we want to use Substrate's macros for implementing traits for tuples.
Hence this common error type.

pallets/gear-builtin/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+
+    fn try_from(dispatch: StoredDispatch) -> Result<Self, Self::Error> {
+        let builtin_id =
+            <Pallet<T> as BuiltinRouterProvider<StoredDispatch, JournalNote, u64>>::provide()
+                .lookup(&dispatch.destination())
+                .ok_or(BuiltinActorError::UnknownActor)?;
+        Ok(Self {
+            source: dispatch.source(),
+            destination: builtin_id,
+            payload: dispatch.payload_bytes().to_vec(),
```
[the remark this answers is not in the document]
ekovalev: Not anymore relevant.

pallets/gear-builtin/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+
+        let mut gas_counter = GasCounter::new(gas_limit);
+
+        let (res, gas_spent) = <T as Config>::BuiltinActor::handle(&builtin_message, gas_limit);
+
+        // We rely on a builtin actor having performed the check for gas limit consistency
+        // and having reported an error if the `gas_limit` was to have been exceeded.
+        // However, to avoid gas tree corruption error, we must not report as spent more gas than
+        // the amount reserved in gas tree (that is, `gas_limit`). Hence (just in case):
+        let gas_spent = gas_spent.min(gas_limit);
```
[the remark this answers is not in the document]
ekovalev: Gas allowance doesn't apply here, I think. Once a builtin actor gets a message it will process it till end since we don't have any rolling gas meter with builtins.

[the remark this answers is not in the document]
ekovalev: This check is exclusively to avoid gas tree corruption error - we can't spend more than had been reserved in the gas tree. And with builtins that's totally possible, because there is no way to break execution half way.

pallets/gear-builtin/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+        // dispatches can end up here (TODO: elaborate).
+        if dispatch.kind() != DispatchKind::Handle {
+            unreachable!("Only handle dispatches can end up here");
+        }
+
+        let mut gas_counter = GasCounter::new(gas_limit);
+
+        let (res, gas_spent) = <T as Config>::BuiltinActor::handle(&builtin_message, gas_limit);
+
+        // We rely on a builtin actor having performed the check for gas limit consistency
```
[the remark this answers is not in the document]
ekovalev: > for example, actor shouldn't charge almost anything if it has bench for some call A and call A is received with insufficient gas

This is exactly what is happening. The problem is that a builtin actor can execute more than one action (like with staking, for example, where you have bonding, unbonding, nominating etc.). Which one is going to be executed each time only becomes known after a message has been decoded. Hence the gas limit sufficiency check has been moved inside the actor itself. An alternative could be what it used to be in the very first implementation - we check the `gas_limit` against the max possible cost, and then refund what has been paid in excess. Or break message processing in 2 stages: decoding message and checking gas limit first, and, if success, proceed to execution.
But it would really be very twisted. The current implementation is optimal given all the constraints (provided the actor is smart enough to change the gas limit immediately after decoding a message and before launching any further processing).

pallets/gear-builtin/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+        // and having reported an error if the `gas_limit` was to have been exceeded.
+        // However, to avoid gas tree corruption error, we must not report as spent more gas than
+        // the amount reserved in gas tree (that is, `gas_limit`). Hence (just in case):
+        let gas_spent = gas_spent.min(gas_limit);
+
+        // This should always return `ChargeResult::Enough` now thanks to the above check.
```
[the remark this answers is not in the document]
ekovalev: Actually, we don't care. Even if it is (for some reason, although it's impossible) we have `NotEnough` here, we proceed anyway. I'll just remove the confusing comment that implies the result of `charge()` matters.

pallets/gear-builtin/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+                };
+
+                Some(process_success(
+                    SuccessfulDispatchResultKind::Success,
+                    dispatch_result,
+                ))
+            }
+            Err(err) => {
+                log::debug!(target: LOG_TARGET, "Builtin actor error: {:?}", err);
+                let system_reservation_ctx = SystemReservationContext::from_dispatch(&dispatch);
```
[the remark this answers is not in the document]
ekovalev: what is the problem then? The `from_dispatch()` will always be returning default reservation context - exactly as we expect it, will it?

pallets/gear-builtin/src/lib.rs
```diff
@@ -0,0 +1,368 @@
...
+
+                let payload = response_bytes
+                    .as_ref()
+                    .to_vec()
+                    .try_into()
+                    .unwrap_or_else(|_| unreachable!("Response message is too large"));
+                let packet = ReplyPacket::new(payload, 0);
+
+                // Mark reply as sent
+                if let Ok(_reply_id) = message_context.reply_commit(packet.clone(), None) {
```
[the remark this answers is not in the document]
ekovalev: Error is virtually impossible here: it can only occur if there are some previous reply chunks already being stored for this dispatch, so when you try to recollect them all together they might exceed the size of the `LimitedVec`. But in our case we only have one packet.
Even if there was an error (again, impossible case) I guess we would just have ended up with auto-reply instead of a reply message with the payload generated by the builtin actor.

breathx (approved):

breathx:
Awesome and huge work done!

PR #3633, opened 2023-12-28
fix(vara): Add staking admin origin to pallet-staking

ukint-vs:
Staking Admin should be able to manage staking pallet as well as the root

@reviewer-or-team


breathx (approved):

runtime/vara/src/lib.rs
```diff
@@ -568,7 +568,7 @@ impl pallet_election_provider_multi_phase::Config for Runtime {
     type Fallback = onchain::OnChainExecution<OnChainSeqPhragmen>;
     type GovernanceFallback = onchain::OnChainExecution<OnChainSeqPhragmen>;
     type Solver = SequentialPhragmen<AccountId, SolutionAccuracyOf<Self>, ()>;
-    type ForceOrigin = EnsureRoot<AccountId>;
+    type ForceOrigin = AdminOrigin;
```
grishasobol, line 571: Could you explain please why do we need stacking admin in elections?

ukint-vs: To control this pallet. Like emergency solution result, min untrusted score, governance fallback

grishasobol: As I can see in doc StakingAdmin is only for slashes
https://github.com/gear-tech/gear/blob/e982f460bc17b0fdebd382f899ed80d0c353825c/runtime/vara/src/governance/origins.rs#L37C1-L39

ukint-vs: updated

grishasobol (approved):

PR #3642, opened 2024-01-07
refactor(afit)!: Replace `async_trait` with AFIT

thecaralice:
## Motivation

[Rust 1.75][rust-1.75] stabilized [AFIT and RPITIT][afit-rpitit].
These allow removing unnecessary indirections and allocations: while [`async_trait`][async_trait] desugars `async fn`s into`-> Pin<Box<dyn Future<Output=…> + Send + '_>>`, AFIT uses proper RPIT (`-> impl Future<Output=…>`), which has already been used for `async fn` outside of traits.

## Caveats

- Traits with RPITIT are **not** object safe.
  If a trait with `async fn` is supposed to be used as a trait object (`dyn Trait`), `async_trait` should be used.
- Additional `Future` bounds like `Send`/`Sync` are a bit painful to add -- that requires manual desugaring of the function in question (see `gcli/src/app.rs`)
- Has MSRV of 1.75

Because of these, `rustc` currently warns about AFIT usage (see `async_fn_in_trait` lint):
```rs
warning: use of `async fn` in public traits is discouraged as auto trait bounds cannot be specified
 --> src/lib.rs:7:5
  |
7 |     async fn fetch(&self, url: Url) -> HtmlBody;
  |     ^^^^^
  |
help: you can desugar to a normal `fn` that returns `impl Future` and add any desired bounds such as `Send`, but these cannot be relaxed without a breaking API change
  |
7 -     async fn fetch(&self, url: Url) -> HtmlBody;
7 +     fn fetch(&self, url: Url) -> impl std::future::Future<Output = HtmlBody> + Send;
  |
```
(example from the [announcement post][afit-rpitit])

## TODO

- [ ] Decide what to do with `gcli::async_trait` re-export
- [ ] Replace other usages of dynamic dispatch in traits with RPITIT.
      I was unable to find any traits with functions returning `Box<dyn …>`; however, I recall discussing that with someone a month ago.

[rust-1.75]: https://blog.rust-lang.org/2023/12/28/Rust-1.75.0.html
[afit-rpitit]: https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html
[async_trait]: https://docs.rs/async-trait/latest/async_trait/

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "nightly-2023-09-05"
+channel = "nightly-2023-10-31"
 components = [ "llvm-tools" ]
 targets = [ "wasm32-unknown-unknown" ]
 profile = "default"
--- a/examples/messager/src/bin/demo_messager.rs
+++ b/examples/messager/src/bin/demo_messager.rs
@@ -18,7 +18,7 @@
 #![cfg(feature = "gcli")]
 
 use gcli::{
-    anyhow, async_trait,
+    anyhow,
     clap::{self, Parser},
     cmd::Upload,
     color_eyre, tokio, App,
@@ -35,7 +35,6 @@ pub struct Messager {
     command: Command,
 }
 
-#[async_trait]
 impl App for Messager {
     async fn exec(&self) -> anyhow::Result<()> {
         let lookup = gcli::lookup!();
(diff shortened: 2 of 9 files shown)


rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "nightly-2023-09-05"
+channel = "nightly-2023-10-31"
```
StackOverflowExcept1on, line 2: We are unable to update to the latest version due to an issue with fuzzer coverage.

breathx (changes requested):
We're not ready to update compiler yet. Please wait until building project with latest stable toolchain

shamilsan:
Let's reopen this PR after the compiler version update.

PR #3647, opened 2024-01-09
feat(core): Add From<[T; N]> for LimitedVec

### LINKED ISSUE #3641: Consider From for LimitedVec
Might be a separate issue, `LimitedVec` should support `From<[T; N]>` where `N` is not greater than `LimitedVec` capacity

_Originally posted by @thecaralice in https://github.com/gear-tech/gear/pull/3606#discussion_r1439018651_

mqxf:
Resolves #3641 

- There is no way to compare M <= N in trait bounds
- static_assert!() doesnt work on const generics
- One potential solution I think could work uses ```where [(); {N - M}]:```, however, it needs `#![feature(generic_const_expr)]`. If you think this is better I can change it to that

@reviewer-or-team

### DIFF
--- a/core/src/buffer.rs
+++ b/core/src/buffer.rs
@@ -37,6 +37,13 @@ use scale_info::{
 #[derive(Clone, Default, Eq, Hash, Ord, PartialEq, PartialOrd, Decode, Encode, TypeInfo)]
 pub struct LimitedVec<T, E, const N: usize>(Vec<T>, PhantomData<E>);
 
+impl<T, E, const M: usize, const N: usize> From<[T; M]> for LimitedVec<T, E, N> {
+    fn from(value: [T; M]) -> Self {
+        assert!(M <= N);
+        Self(value.into(), Default::default())
+    }
+}
+
 /// Formatter for [`LimitedVec`] will print to precision of 8 by default, to print the whole data, use `{:+}`.
 impl<T: Clone + Default, E: Default, const N: usize> Display for LimitedVec<T, E, N>
 where


mqxf:
Just threw this little test together:

`#![feature(generic_const_exprs)]

#[derive(Clone, Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
pub struct LimitedVec<T, const N: usize>(Vec<T>);

impl<T, const M: usize, const N: usize> From<[T; M]> for LimitedVec<T, N>
    where [(); {N - M}]:
{
    fn from(value: [T; M]) -> Self {
        Self(value.into())
    }
}

fn main() {
    let a = [1, 2, 3];
    let b: LimitedVec<_, 5> = a.into();
    let c: LimitedVec<_, 3> = a.into();
    //let d: LimitedVec<_, 2> = a.into();
}`

Where if the line is uncommented, compilation fails with this error:
`error[E0080]: evaluation of `<LimitedVec<!0, 2> as std::convert::From<[!0; 3]>>::{constant#0}` failed
 --> src/main.rs:7:17
  |
7 |     where [(); {N - M}]:
  |                 ^^^^^ attempt to compute `2_usize - 3_usize`, which would overflow`

I personally think this would be better than runtime assertion, but I do not know whether I can use this feature since it is techincally unstable, though should not arise any problems in this scenario.

thecaralice:
Another option is using `#![feature(associated_const_equality)]` instead, which is, while technically still unstable, not as unstable as `#![feature(generic_const_exprs)]` is.
This is what I got while tinkering with this issue:
```rs
#![feature(associated_const_equality)]
use core::marker::PhantomData;

pub struct LimitedVec<T, E, const N: usize>(Vec<T>, PhantomData<E>);

macro_rules! const_conditions {
    ($($name:ident = |$($arg:ident: $argtype:ty),*| $result:expr;)*) => {
        $(
            enum $name<$(const $arg: $argtype),*> {}

            impl<$(const $arg: $argtype),*> $crate::ConstCondition for $name<$($arg),*> {
                const RESULT: bool = $result;
            }
        )*
    }
}

trait ConstCondition {
    const RESULT: bool;
}

const_conditions! {
    LessThan = |L: usize, R: usize| L < R;
}

impl<T, E, const M: usize, const N: usize> From<[T; M]> for LimitedVec<T, E, N>
where
    LessThan<N, M>: ConstCondition<RESULT = false>,
{
    fn from(value: [T; M]) -> Self {
        Self(value.into(), PhantomData)
    }
}
```
Or, without any macros,
```rs
#![feature(associated_const_equality)]

trait ConstCondition {
    const RESULT: bool;
}

enum LessThan<const L: usize, const R: usize> {}
impl<const L: usize, const R: usize> ConstCondition for LessThan<L, R> {
    const RESULT: bool = L < R;
}

impl<T, E, const M: usize, const N: usize> From<[T; M]> for LimitedVec<T, E, N>
where
    LessThan<N, M>: ConstCondition<RESULT = false>,
{
    fn from(value: [T; M]) -> Self {
        Self(value.into(), PhantomData)
    }
}
```

mqxf:
That would probably be a better solution.

thecaralice:
Note that there is a drawback: you can not just `#[cfg]` it out like in #3470, because the compiler will still complain about ACE usage ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=e38e0946926f7f5a16fb34021be0bbcd)).

core/src/buffer.rs
```diff
@@ -37,6 +37,13 @@ use scale_info::{
 #[derive(Clone, Default, Eq, Hash, Ord, PartialEq, PartialOrd, Decode, Encode, TypeInfo)]
 pub struct LimitedVec<T, E, const N: usize>(Vec<T>, PhantomData<E>);
 
+impl<T, E, const M: usize, const N: usize> From<[T; M]> for LimitedVec<T, E, N> {
```
techraed, line 40: I'm against of assertions in runtime for such a type, that's mainly used in **non-test** code. It's not worth adding potentially not checked to be unreachable by tests panic in the trait impl that's needed only for tests.  
If there's no other stable solution, then it seems better to close the PR and wait for stable.

techraed (changes requested):

core/src/buffer.rs
breathx: +1

StackOverflowExcept1on:
Try this: https://github.com/nvzqz/static-assertions/issues/40#issuecomment-1458897730

core/src/buffer.rs
```diff
@@ -37,6 +37,13 @@ use scale_info::{
...
+    fn from(value: [T; M]) -> Self {
+        assert!(M <= N);
+        Self(value.into(), Default::default())
+    }
+}
+
```
StackOverflowExcept1on, line 46:
```suggestion
struct LessThenOrEqual<const M: usize, const N: usize> {}

impl<const M: usize, const N: usize> LessThenOrEqual<M, N> {
    const ASSERTION: () = assert!(M <= N);
}

impl<T, E, const M: usize, const N: usize> From<[T; M]> for LimitedVec<T, E, N> {
    fn from(value: [T; M]) -> Self {
        let _ = LessThenOrEqual::<M, N>::ASSERTION;
        Self(value.into(), Default::default())
    }
}
```

StackOverflowExcept1on: please also replace places where `::from` can be used

mqxf: this is still runtime assertion isnt it?

mqxf:
static assertions unfortunately cant work on const generics

core/src/buffer.rs
breathx: > this is still runtime assertion isnt it?

No

PR #3649, opened 2024-01-09
feat(wasm-builder): Check all `gear` requirements for code at compile time

### LINKED ISSUE #2819: Check all `gear` requirements for code at build
### Problem to Solve

Atm we entrance of init and handle functions on compile time using gear_wasm_builder, but furthermore we can check availability to instantiate and execute module. For example, we have several rules on available for each entry point syscalls (e.g. you can't call `gr_source` in `handle_signal`) - such things may be caught as well

### Possible Solution

/ .. /

### Notes

_No response_

mqxf:
Resolves #2819

@reviewer-or-team


utils/wasm-builder/src/checks.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    let exports = module.export_section().ok_or(Error::ExportSectionMissing)?;
+    let imports = module.import_section().ok_or(Error::ImportSectionMissing)?;
+    let functions = module.function_section().ok_or(Error::FunctionSectionMissing)?;
+    let types = module.type_section().ok_or(Error::TypeSectionMissing)?;
+    let code = module.code_section().ok_or(Error::CodeSectionMissing)?;
+
+    let imported_functions = imports.functions();
+
+    // While this is checked in Code::new, we might as well check it when building wasm program
```
StackOverflowExcept1on, line 25: This should have already been checked by wasm-builder. Your recent changes in `get_exports` are also checked.
https://github.com/gear-tech/gear/blob/c2733ecc57ddad27ce7c23a72789258276f89c2f/utils/wasm-builder/src/optimize.rs#L132-L134

mqxf: Should I perhaps do this in Code::try_new fully instead, that way it gets checked there instead of in wasm-builder

core/src/code.rs
```diff
@@ -42,15 +43,41 @@ use scale_info::{
...
+}
 
 /// Parse function exports from wasm module into [`DispatchKind`].
-fn get_exports(
-    module: &Module,
-    reject_unnecessary: bool,
-) -> Result<BTreeSet<DispatchKind>, CodeError> {
-    let mut exports = BTreeSet::<DispatchKind>::new();
+fn check_code(module: &Module, config: &TryNewCodeConfig) -> Result<(), CodeError> {
+    //return Ok(());
```
StackOverflowExcept1on, line 80: do not forget to cleanup

mqxf:
Still need to know whether I should implement a check for calling forbidden syscalls from handle_signal

core/src/code.rs
```diff
@@ -62,35 +87,71 @@ fn get_exports(
...
+            return Err(CodeError::RequiredExportFnNotFound);
+        }
+    }
+
+    if config.check_imports {
+        for import in imports {
+            if let External::Function(i) = import.external() {
+                let Type::Function(types) = &types[*i as usize];
+                // We can likely improve this by adding some helper function in SyscallName
+                let syscall = SyscallName::all()
```
StackOverflowExcept1on, line 131: 1. use `SyscallName::instrumentable` instead of `SyscallName::all()` (because of `SystemBreak` syscall)
2. `instrumentable` will return `BTreeSet<_>` and you can do `syscalls_set.contains(syscall_name)` which is more efficient than iteration for each import
3. prohibit importing syscall more than once (this is what embedded_executor does when instantiating a module)

StackOverflowExcept1on:
@mqxf `handle_signal` check is difficult to implement due to dynamic functions in wasm format

core/src/code.rs
```diff
@@ -62,35 +87,71 @@ fn get_exports(
...
+        let mut entry = false;
+        for export in exports {
+            if let Internal::Function(i) = export.internal() {
                 // Index access into arrays cannot panic unless the Module structure is invalid
                 let type_id = funcs[*i as usize - import_count].type_ref();
                 let Type::Function(ref f) = types[type_id as usize];
                 if !f.params().is_empty() || !f.results().is_empty() {
                     return Err(CodeError::InvalidExportFnSignature);
                 }
+                if !ALLOWED_EXPORTS.contains(&export.field()) && config.check_exports {
```
StackOverflowExcept1on, line 112: `config.check_exports` is already `true`
```suggestion
                if !ALLOWED_EXPORTS.contains(&export.field()) {
```

core/src/code.rs
```diff
@@ -62,35 +87,71 @@ fn get_exports(
...
+                    .ok_or(CodeError::UnknownImport)?;
+                let signature = syscall.signature();
+
+                let params = signature
+                    .params()
+                    .iter()
+                    .copied()
+                    .map(Into::<ValueType>::into)
+                    .collect::<Vec<_>>();
+                if params != types.params() {
```
StackOverflowExcept1on, line 142: I think you can create 2 iterators and compare using `.eq()` it instead of `.collect::<Vec<_>>()`
https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.eq

core/src/code.rs
mqxf: One issue here is that we dont have a from_str on SyscallName, so I cannot just search the Set for it, instead having to iterate again and check which one matches the to_string() signature. Perhaps I could make some kind of Map<String, SyscallName> by iterating over the set once, then mapping the str to each one. That is what I will do for now, if there is a better solution, I will replace it with that.

StackOverflowExcept1on: - plz remove `// We can likely improve this by adding some helper function in SyscallName`
- `from_str` doesn't seem to be appropriate because we only want to call it for instrumentable syscalls
- it would be nice to implement point 3 from previous comment

core/src/code.rs
```diff
@@ -124,14 +124,19 @@ fn check_code(module: &Module, config: &TryNewCodeConfig) -> Result<(), CodeErro
     }
 
     if config.check_imports {
-        let syscalls = SyscallName::instrumentable_map();
+        let syscalls = SyscallName::all_map();
```
StackOverflowExcept1on, line 127: plz revert commit 947bd0920ad8a1cfa1a5d35e5a54231d61af3bb7
I think someone forgot to add `SystemReserveGas` and `EnvVars` to `instrumentable()` and that's why tests fails

mqxf: ah ok

StackOverflowExcept1on (approved):

core/src/code.rs
```diff
@@ -62,35 +87,76 @@ fn get_exports(
...
+                    .iter()
+                    .copied()
+                    .map(Into::<ValueType>::into);
+                if !params.eq(types.params().iter().copied()) {
+                    return Err(CodeError::InvalidImportFnSignature);
+                }
+
+                let results = signature.results().unwrap_or(&[]);
+                if results != types.results() {
+                    return Err(CodeError::InvalidImportFnSignature);
```
NikVolf, line 153: I understand that those errors that already there don't specify much, so this `InvalidImportFnSignature` is created same

Can you refactor them all in next PR? So that every error contained an exact item (function/import/export) that caused it? And create issue now

NikVolf (approved):

PR #3655, opened 2024-01-12
refactor(rpc-call): Never return `<wasm:stripped>` from rpc calls

breathx:
These changes formalise error output of calculate gas calls:
* Internal errors (that are supposed to be unreachable) if occur will be thrown as "Internal error: entered unreachable code '`{ERROR EXPLANATION}`'" 
* 🥳 Rpc calls no longer return `<wasm:stripped>`, that happened to be in case of underlying extrinsic failure. Instead, it returns name of the error occurred.
* In case of non pallet-gear error returned it will contain pallet/module index within the runtime of the chain to resolve name collision.

### Before
```shell
curl http://testnet.vara.network:9944 -H "Content-Type:application/json;charset=utf-8" -d "{
    \"jsonrpc\":\"2.0\",
    \"id\":1,
    \"method\":\"gear_calculateInitCreateGas\",
    \"params\": [
        \"0x0000000000000000000000000000000000000000000000000000000000000000\",
        \"0x0000000000000000000000000000000000000000000000000000000000000000\",
        \"\",
        0,
        true
    ]
}"
{"jsonrpc":"2.0","error":{"code":8000,"message":"Runtime error","data":"\"Internal error: create_program failed with '<wasm:stripped>'\""},"id":1}%   
```

### Now
```shell
curl http://localhost:9944 -H "Content-Type:application/json;charset=utf-8" -d "{
    \"jsonrpc\":\"2.0\",
    \"id\":1,
    \"method\":\"gear_calculateInitCreateGas\",
    \"params\": [
        \"0x0000000000000000000000000000000000000000000000000000000000000000\",
        \"0x0000000000000000000000000000000000000000000000000000000000000000\",
        \"\",
        0,
        true
    ]
}"
{"jsonrpc":"2.0","error":{"code":8000,"message":"Runtime error","data":"Extrinsic `gear.create_program` failed: 'CodeDoesntExist'"},"id":1}
```

### DIFF
--- a/gsdk/tests/rpc.rs
+++ b/gsdk/tests/rpc.rs
@@ -49,7 +49,7 @@ async fn pallet_errors_formatting() -> Result<()> {
         CallError::Custom(ErrorObject::owned(
             8000,
             "Runtime error",
-            Some("\"Extrinsic `gear.upload_program` failed: 'ProgramConstructionFailed'\""),
+            Some("Extrinsic `gear.upload_program` failed: 'ProgramConstructionFailed'"),
         )),
     ))));
 
--- a/gsdk/tests/rpc.rs
+++ b/gsdk/tests/rpc.rs
@@ -29,6 +29,36 @@ use utils::{alice_account_id, dev_node, node_uri};
 mod utils;
 
 #[tokio::test]
+async fn pallet_errors_formatting() -> Result<()> {
+    let node = dev_node();
+    let api = Api::new(Some(&node_uri(&node))).await?;
+
+    let err = api
+        .calculate_upload_gas(
+            [0u8; 32].into(),
+            /* invalid code */ vec![],
+            vec![],
+            0,
+            true,
+            None,
+        )
+        .await
+        .expect_err("Must return error");
+
+    let expected_err = Error::Subxt(SubxtError::Rpc(RpcError::ClientError(Box::new(
+        CallError::Custom(ErrorObject::owned(
+            8000,
+            "Runtime error",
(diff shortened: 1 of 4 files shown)


gshep (approved):

clearloop (approved):

gsdk/tests/rpc.rs
```diff
@@ -28,6 +28,36 @@ use utils::{alice_account_id, dev_node, node_uri};
...
+
+    let expected_err = Error::Subxt(SubxtError::Rpc(RpcError::ClientError(Box::new(
+        CallError::Custom(ErrorObject::owned(
+            8000,
+            "Runtime error",
+            Some("Extrinsic `gear.upload_program` failed: 'ProgramConstructionFailed'"),
+        )),
+    ))));
+
+    assert_eq!(format!("{err}"), format!("{expected_err}"));
```
shamilsan, line 56: Can we compare error types instead of strings?

pallets/gear/src/runtime_api.rs
```diff
@@ -69,16 +72,45 @@ where
...
+            {
+                Some(index as usize)
+            } else {
+                None
+            };
+
+            let error_message: &'static str = e.into();
+
+            let gear_module_idx =
+                <<T as frame_system::Config>::PalletInfo as PalletInfo>::index::<Pallet<T>>()
```
shamilsan, line 88: Might be replaced with a type alias.
```suggestion
                PalletIdexOf<T>()
```

shamilsan (approved):

gsdk/tests/rpc.rs
breathx: Lack of derives

PR #3656, opened 2024-01-12
refactor(workspace): Use hard pin to `gear-dlmalloc`

breathx:
Inspired by the fear of poorly held semver in future


shamilsan:
It is better to respect semver instead of pinning the dep's version.

NikVolf:
yeah generally unless dependency is known to violate this, better avoid

breathx:
Won't be applied

PR #3662, opened 2024-01-13
refactor(core): Add `gr_user_break` syscall to replace all non returning syscalls.

### LINKED ISSUE #3216: Proposal: introduce gr_break

mqxf:
Resolves #3216 

@reviewer-or-team

### DIFF
--- a/core-backend/src/env.rs
+++ b/core-backend/src/env.rs
@@ -190,16 +190,12 @@ where
         builder.add_func(CreateProgram, wrap_syscall!(create_program));
         builder.add_func(CreateProgramWGas, wrap_syscall!(create_program_wgas));
         builder.add_func(Debug, wrap_syscall!(debug));
-        builder.add_func(Panic, wrap_syscall!(panic));
-        builder.add_func(OomPanic, wrap_syscall!(oom_panic));
-        builder.add_func(Exit, wrap_syscall!(exit));
         builder.add_func(ReplyCode, wrap_syscall!(reply_code));
         builder.add_func(SignalCode, wrap_syscall!(signal_code));
         builder.add_func(ReserveGas, wrap_syscall!(reserve_gas));
         builder.add_func(ReplyDeposit, wrap_syscall!(reply_deposit));
         builder.add_func(UnreserveGas, wrap_syscall!(unreserve_gas));
         builder.add_func(GasAvailable, wrap_syscall!(gas_available));
-        builder.add_func(Leave, wrap_syscall!(leave));
         builder.add_func(MessageId, wrap_syscall!(message_id));
         builder.add_func(PayProgramRent, wrap_syscall!(pay_program_rent));
         builder.add_func(ProgramId, wrap_syscall!(program_id));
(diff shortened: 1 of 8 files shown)


breathx:
First of all, the issue is already solved. No idea why GitHub shows only reference.
Secondly, you can't just remove support for some already existing on-chain calls

mqxf:
The issue wasn't solved, since he mentioned two syscalls, gr_system_break, which was solved, and gr_user_break, which wasn't. I asked around before making this issue and I am not removing any support, we are just moving any syscalls that dont return into one syscall, since they don't take many arguments.

breathx:
> The issue wasn't solved, since he mentioned two syscalls, gr_system_break, which was solved, and gr_user_break, which wasn't. 

https://github.com/gear-tech/gear/pull/3512#pullrequestreview-1756139921 It was discussed that other changes are redundant and incorrect. 

> I asked around before making this issue and I am not removing any support, we are just moving any syscalls that dont return into one syscall, since they don't take many arguments.

As I can see, you've removed support for previously existing programs, because old signatures of syscalls no longer exist.

mqxf:
I see. A bit of a stretch, but if this is useful enough, could some sort of "migrations" be used, where when a program is loaded, I could write some code which would check if it imports syscalls and replace the uses of those syscalls with the new break syscall? Probably not, but just an idea.

PR #3670, opened 2024-01-17
feat(gtest): Add option to check for if a message panicked on RunResult

### LINKED ISSUE #3668: gtest: add func/method `assert_panic`
### Problem to Solve

https://github.com/gear-tech/gear/pull/3527#discussion_r1452533240

### Possible Solution

demo-waiter defines its own function, but we can move it into gtest itself

### Notes

_No response_

mqxf:
Resolves #3668

@reviewer-or-team


examples/waiter/tests/mx_lock_access.rs
```diff
@@ -39,14 +39,11 @@ fn access_mx_lock_guard_from_different_msg_fails(
...
-            "Mutex guard held by message {} is being accessed by message {}",
-            lock_msg_id,
-            lock_access_result.sent_message_id()
-        ),
-    );
+    assert!(lock_access_result.panicked_with(&format!(
+        "Mutex guard held by message {} is being accessed by message {}",
+        lock_msg_id,
+        lock_access_result.sent_message_id()
+    ),));
```
shamilsan, line 46:
```suggestion
    )));
```

gtest/src/log.rs
```diff
@@ -396,6 +396,28 @@ impl RunResult {
...
+        self.log.len() == 1
+            && matches!(
+                self.log[0].reply_code(),
+                Some(ReplyCode::Error(ErrorReplyReason::Execution(
+                    SimpleExecutionError::UserspacePanic
+                )))
+            )
+    }
+
+    /// If the main message panicked with a given message.
```
techraed, line 411: I would add that this test is the best to check the `msg` to be a substring of the error payload. So if it returns false, it doesn't necessarily mean that there wasn't panic.

techraed (approved):
Please be sure that you have checked all the examples which roughly use a similar check

gtest/src/log.rs
```diff
@@ -396,6 +396,28 @@ impl RunResult {
...
+    pub fn panicked_with(&self, msg: &str) -> bool {
+        if self.panicked() {
+            let payload = String::from_utf8(self.log[0].payload().into())
+                .expect("Unable to decode panic message");
+            payload.contains(&format!("panicked with '{msg}'"))
+        } else {
+            false
+        }
+    }
```
StackOverflowExcept1on, line 420:
```suggestion
    /// Trying to get the panic log.
    fn panic_log(&self) -> Option<&CoreLog> {
        let [core_log] = &self.log[..] else {
            return None;
        };
        let is_panic = matches!(
            core_log.reply_code(),
            Some(ReplyCode::Error(ErrorReplyReason::Execution(
                SimpleExecutionError::UserspacePanic
            )))
        );
        is_panic.then_some(core_log)
    }

    /// If the main message panicked.
    pub fn panicked(&self) -> bool {
        self.panic_log().is_some()
    }

    /// Asserts that the main message panicked with a given message.
    #[track_caller]
    pub fn assert_panicked_with(&self, msg: impl Into<String>) {
        if let Some(log) = self.panic_log() {
            let msg = msg.into();
            let payload =
                String::from_utf8(log.payload().into()).expect("Unable to decode panic message");

            assert!(
                payload.contains(&format!("panicked with '{msg}'")),
                "expected panic message that contains `{msg}`, but the actual panic message is `{payload}`"
            );
        }
    }
```

examples/waiter/tests/mx_lock_access.rs
```diff
@@ -39,14 +39,11 @@ fn access_mx_lock_guard_from_different_msg_fails(
...
-            "Mutex guard held by message {} is being accessed by message {}",
-            lock_msg_id,
-            lock_access_result.sent_message_id()
-        ),
-    );
+    assert!(lock_access_result.panicked_with(&format!(
+        "Mutex guard held by message {} is being accessed by message {}",
+        lock_msg_id,
+        lock_access_result.sent_message_id()
+    )));
```
StackOverflowExcept1on, line 46:
```suggestion
    lock_access_result.assert_panicked_with(format!(
        "Mutex guard held by message {} is being accessed by message {}",
        lock_msg_id,
        lock_access_result.sent_message_id(),
    ));
```

examples/waiter/tests/rw_lock_access.rs
```diff
@@ -83,15 +83,12 @@ fn access_rw_lock_guard_from_different_msg_fails(
...
-            lock_msg_id,
-            lock_access_result.sent_message_id()
-        ),
-    );
+    assert!(lock_access_result.panicked_with(&format!(
+        "{:?} lock guard held by message {} is being accessed by message {}",
+        lock_type,
+        lock_msg_id,
+        lock_access_result.sent_message_id()
+    ),));
```
StackOverflowExcept1on, line 91:
```suggestion
    lock_access_result.assert_panicked_with(format!(
        "{lock_type:?} lock guard held by message {lock_msg_id} is being accessed by message {}",
        lock_access_result.sent_message_id(),
    ));
```

gtest/src/log.rs
mqxf: I see this is a nicer solution, put perhaps then we should also make public the panic_log?

StackOverflowExcept1on (approved):

gtest/src/log.rs
```diff
@@ -396,6 +396,40 @@ impl RunResult {
             .flat_map(DecodedCoreLog::try_from_log)
             .collect()
     }
+
+    /// Trying to get the panic log.
+    fn panic_log(&self) -> Option<&CoreLog> {
```
techraed, line 401: I would move that below, under the `panicked` and `assert_panicked_with` to make code more readable (top to bottom read style)

techraed (approved):
Please don't forget to mention that if `assert_panicked_with` didn't panic, it doesn't necessarily mean that there was a panic with the provided log message, so API users won't be confused and use it properly

gtest/src/log.rs
StackOverflowExcept1on: I'm not sure who will use this, although we can provide `-> Option<String>` instead of `-> Option<&CoreLog>` and make it public

Cargo.lock
```diff
@@ -143,9 +143,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
```
breathx, line 145: Revert `.lock` changes please

gtest/src/log.rs
```diff
@@ -396,6 +396,40 @@ impl RunResult {
...
+        self.panic_log().is_some()
+    }
+
+    /// Asserts that the main message panicked with a given message.
+    #[track_caller]
+    pub fn assert_panicked_with(&self, msg: impl Into<String>) {
+        if let Some(log) = self.panic_log() {
+            let msg = msg.into();
+            let payload =
+                String::from_utf8(log.payload().into()).expect("Unable to decode panic message");
```
breathx, line 425: it will be not that obvious if we panic here

gtest/src/log.rs
```diff
@@ -396,6 +396,40 @@ impl RunResult {
...
+            core_log.reply_code(),
+            Some(ReplyCode::Error(ErrorReplyReason::Execution(
+                SimpleExecutionError::UserspacePanic
+            )))
+        );
+        is_panic.then_some(core_log)
+    }
+
+    /// If the main message panicked.
+    pub fn panicked(&self) -> bool {
```
breathx, line 415: redundant, you have `self.main_failed() -> bool` doing the same

gtest/src/log.rs
```diff
@@ -396,6 +396,40 @@ impl RunResult {
             .flat_map(DecodedCoreLog::try_from_log)
             .collect()
     }
+
+    /// Trying to get the panic log.
+    fn panic_log(&self) -> Option<&CoreLog> {
+        let [core_log] = &self.log[..] else {
```
breathx, line 402: why?

breathx: it's not necessary the only one log to be here

gtest/src/log.rs
```diff
@@ -396,6 +396,40 @@ impl RunResult {
...
+            return None;
+        };
+        let is_panic = matches!(
```
breathx, line 405: for that case you may return wrong message, not the main/primary one

gtest/src/log.rs
```diff
@@ -396,6 +396,40 @@ impl RunResult {
...
+    /// Asserts that the main message panicked with a given message.
+    #[track_caller]
+    pub fn assert_panicked_with(&self, msg: impl Into<String>) {
+        if let Some(log) = self.panic_log() {
+            let msg = msg.into();
+            let payload =
+                String::from_utf8(log.payload().into()).expect("Unable to decode panic message");
+
+            assert!(
+                payload.contains(&format!("panicked with '{msg}'")),
```
breathx, line 428: starts with

gtest/src/log.rs
```diff
@@ -396,6 +396,40 @@ impl RunResult {
...
+                "expected panic message that contains `{msg}`, but the actual panic message is `{payload}`"
+            );
+        }
+    }
 }
```
breathx, line 433: some tests?

breathx (changes requested):

gtest/src/log.rs
mqxf: this is how the implementation that existed that I removed did it, assuming that there is only one log. But if there is more that one, can we assume that it will always be the last one?

gtest/src/log.rs
mqxf: But isn't main failed not just panics? It could be OutOfGas or StackOverflow or any other error that could make the message fail?

gtest/src/log.rs
mqxf: Fucntion got changed, now works differently

gtest/src/log.rs
mqxf: Is there not something before this, like "Thread ... panicked with"?

gtest/src/log.rs
StackOverflowExcept1on: Good point, I missed this. Maybe change the pattern?
```suggestion
        let [.., core_log] = &self.log[..] else {
```

gtest/src/log.rs
breathx: "Panic occurred: panicked with '{message}'[ at '{loc}']"
Just for notice, I don't see any reason to assert paths

Cargo.lock
```diff
@@ -14879,4 +14879,4 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 1.0.109",
-]
+]
```
breathx, line 14882: newline

examples/waiter/tests/utils.rs
```diff
@@ -1,18 +1 @@
-use gstd::errors::{ErrorReplyReason, ReplyCode, SimpleExecutionError};
-use gtest::RunResult;
-
 pub const USER_ID: u64 = 10;
```
breathx, line 1: redundant module now

gtest/src/log.rs
```diff
@@ -396,6 +396,57 @@ impl RunResult {
...
+
+    /// If the main message panicked.
+    pub fn panicked(&self) -> bool {
+        if self.log.is_empty() {
+            return false;
+        }
+        matches!(
+            self.log
+                .last()
+                .expect("Checked for empty log previously")
```
breathx, line 408: you don't have any ordering guarantees.

breathx: search by reply_to(main_message)

breathx (changes requested):

gtest/src/log.rs
mqxf: log.id == result.message_id?

breathx: > log.id == result.message_id?

reply_to(), not log.id

mqxf: I cannot find that function in Log or CoreLog?

breathx: > I cannot find that function in Log or CoreLog?

As I see we cut reply details into just code while converting core type into gtest Log, we mustn't do it: I guess this PR should solve it. 
As well as we can't just take any reply_code.is_some() message, because it may be other one.

mqxf: Oh so should I find where this is a thing in the actual execution and add it onto CoreLog? The reply_to: MessageId

mqxf: Ah yes I found it its super simple I see

gtest/src/log.rs
mqxf: I'm pretty sure there are test which somehow assert that a program panicked, if I find those tests and use this in them, does that count as tests for this?

techraed:
@breathx take a look please. Seems to be ready to be merged.

gtest/src/log.rs
```diff
@@ -34,6 +34,7 @@ pub struct CoreLog {
     destination: ProgramId,
     payload: Payload,
     reply_code: Option<ReplyCode>,
```
breathx, line 36: you can just use core type ReplyDetails

breathx: on the other hand it may decrease flexibility if I want to assert that message replied to my message and I dont care about the reply code, so could be left as is atm

gtest/src/log.rs
```diff
@@ -396,6 +425,47 @@ impl RunResult {
...
+        .expect("Unable to decode panic message");
+
+        assert!(
+            payload.starts_with(&format!("Panic occurred: panicked with '{msg}'")),
+            "expected panic message that contains `{msg}`, but the actual panic message is `{payload}`"
+        );
+    }
+
+    /// Trying to get the panic log.
+    fn panic_log(&self) -> Option<&CoreLog> {
```
breathx, line 456:
```suggestion
    fn main_panic_log(&self) -> Option<&CoreLog> {
```

breathx (approved):

PR #3676, opened 2024-01-19
refactor(gtest): Remove ambiguous API and incorrectly named functions

### LINKED ISSUE #3224: Ambiguous API in gtest
### File Location(s)

https://github.com/gear-tech/gear/blob/fa2dfd893898d972180ae8352ff7c31b78814663/gtest/src/system.rs#L133C13-L133C13
https://github.com/gear-tech/gear/blob/fa2dfd893898d972180ae8352ff7c31b78814663/examples/program-factory/src/lib.rs#L195

### Proposal

If you create a program from a program using a nonexistent `CodeId`, no new program will be created in runtime, however, `is_active_program` will return true on a `ProgramId` made from the same nonexistent `CodeId`.

The test `test_non_existing_code_hash` uses ambiguous API as the results asserts that a fictional program id is an active program, which is not correct.

mqxf:
Resolves #3224

Previously, is_user, the only function, would return true is an actor was a `TestActor::User` OR if the ProgramId was not a valid actor. This caused some issues, like is_active_program, which would check if a program existed, would return true if a ProgramId didn't exist. I fixed this by removing that extra branch, now is_user tests only is the actor is `TestActor::User`. Aswell as that, I added is_program, which tests if it is `TestActor::Active` or `TestActor::Inactive`, and finally is_active_program.

@reviewer-or-team


techraed (approved):
seems ok, need final review from original author @breathx

gtest/src/manager.rs
```diff
@@ -577,8 +577,21 @@ impl ExtManager {
     }
```
breathx, line 578: This should still be refactored:
```rust
use TestActor::*;
matches!(x, Initialized(..), Uninitialized(..), Dormant) // is_program
matches!(x, Initialized(..), Uninitialized(..)) // is_active_program
matches!(x, Initialized(..)) // is_initialised_program
actors.get(id).is_none() || matches!(x, User) // is_user
```

breathx (changes requested):

breathx (approved):

PR #3685, opened 2024-01-22
refactor(node, testing): Use minimal runtime for authorship tests

ekovalev:
Replace the full-blown Vara `runtime` used in `gear-authorship` tests with a minimal working runtime to counteract the effect of multiple runtime wasm modules allocations which has started tremendously slowing down tests set-up since the native runtime execution was removed from Substrate.

This doesn't completely solve the problem. A better and more scalable solution would be to try to run all test cases in a single `#[test]` function. This will require more work though and may not make it into this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3777,6 +3777,7 @@ dependencies = [
  "futures-timer",
  "gear-common",
  "gear-core",
+ "gear-minimal-test-runtime",
  "gear-node-testing",
  "gear-runtime-common",
  "gear-runtime-interface",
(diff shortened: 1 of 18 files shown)


ukint-vs:
Is it ready for review?

ekovalev:
> Is it ready for review?

As far as this minimalistic runtime is concerned then yes, no more changes would have been made.
However, after discussions with @breathx we agreed that we should probably try to implement a more robust way (squeezing all test cases into a single #[test] block) so that we don't have to maintain two runtimes and always be sure we run tests against the real runtime.

PR #3687, opened 2024-01-23
refactor(gcli): switch resolved API to gclient

### LINKED ISSUE #3574: Switch to gclient in gcli
### Problem to Solve

for gathering customized apis

### Possible Solution

after #3572

### Notes

_No response_

clearloop:
Resolves #3574 

- [x] switch resolved API to gclient
-

@reviewer-or-team


gcli/src/cmd/send.rs
```diff
@@ -58,30 +53,33 @@ pub struct Send {
 }
 
 impl Send {
-    pub async fn exec(&self, signer: Signer) -> Result<()> {
-        let tx = signer
-            .calls
-            .send_message(
+    pub async fn exec(&self, app: &impl App) -> Result<()> {
+        let signer = app.signer().await?;
+        let gas_limit = if self.gas_limit == 0 {
```
breathx, line 58: what if user truly wants to send message with 0 gas limit?
for example, it's valid case for reply nowadays

breathx: I wouldn't do that

breathx (approved):
Frankly speaking, can't see any reason to keep building very strong dependent on each other crates like gsdk -> gcli/gclient, gclient -> gcli etc

Furthermore, I'd hide "signer" type at all if somebody wants to use gear runtime fns there should be predefined api and some explicit and not that easy access to signer for case of specific and rare calls like governance and similar

gcli/src/cmd/send.rs
clearloop: https://github.com/gear-tech/gear/issues/3709

breathx:
> > Frankly speaking, can't see any reason to keep building very strong dependent on each other crates like gsdk -> gcli/gclient, gclient -> gcli etc
> 
> it makes sense to merge `gsdk` into `gclient` nowadays since we are focus on one network only, the dynamic API provided in gsdk is not required:
> 
> Most of the logic in `gsdk` are for the dynamic API which is for adapting different versions of our node and even networks (previous vara and gear ), and since we have version managements on crates-io now, it's more convenient to find the matched api version for gear nodes

But it's still useful because we can use the latest version of the packages to interact with both main net and test net which may differ in runtime versions. Furthermore, it would be a lot of work to come back to generic api in future if we remove it

ukint-vs:
@CodiumAI-Agent /review

ukint-vs (approved):

PR #3688, opened 2024-01-23
chore(docs): replace "contract" by "program"

shamilsan:
@gear-tech/dev

### DIFF
--- a/gstd/Cargo.toml
+++ b/gstd/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "gstd"
 description = "Gear programs standard library"
-keywords = ["gear", "std", "no-std", "wasm", "smart-contracts"]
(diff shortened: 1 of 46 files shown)


gstd/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "gstd"
 description = "Gear programs standard library"
-keywords = ["gear", "std", "no-std", "wasm", "smart-contracts"]
+keywords = ["gear", "std", "no-std", "wasm", "programs"]
```
mertwole, line 4: I think `"programs"` doesn't make sense in this context

mertwole (approved):

gstd/Cargo.toml
shamilsan: Agree, reverted in 7bde400

PR #3691, opened 2024-01-25
remove initialized field from ExecutableActorData

grishasobol:


breathx (approved):
Just for clarification: it's all because we now never use Option<ExecutableData> and suppose that ExecutableData is always for processing some message: inactive programs checked in queue before?

grishasobol:
> Just for clarification: it's all because we now never use Option and suppose that ExecutableData is always for processing some message: inactive programs checked in queue before?

Actually, I don't know answer on your question)) but I found that after queue refactoring this field become useless, so just remove it.

ark0f (approved):

PR #3698, opened 2024-01-28
test(gtest): show bug with reservations in `gtest`

shamilsan:
Showcase for https://github.com/gear-tech/gear/issues/3699

@gear-tech/dev


NikVolf (commented):
Should we fix it right away?

breathx:
> Should we fix it right away?

As I see it's not 1-line fix, so I guess better to keep an issue open changing its urgency to asap and merge showcase

examples/custom/src/reserver.rs
```diff
@@ -0,0 +1,41 @@
...
+    use gstd::{msg, prelude::*, ReservationId};
+
+    #[derive(Default)]
+    pub(crate) struct State {
+        reservation_id: Option<ReservationId>,
+    }
+
+    pub(crate) fn init() -> State {
+        Default::default()
+    }
```
breathx, line 30: useless

breathx (approved):

examples/custom/src/reserver.rs
shamilsan: It is not the `init` entrypoint. This function returns state to be used here: 

https://github.com/gear-tech/gear/pull/3698/files#diff-7338aaa9606870091a816d7873106f089b9cc8465c00bd8dfc476ced27002d55R87

shamilsan: But we can replace it by creating `State` in place indeed.

shamilsan: [af7e257](https://github.com/gear-tech/gear/pull/3698/commits/af7e25764c7fe81df19b160a3d1665be54f0ac3b)

techraed (approved):

PR #3704, opened 2024-01-29
feat(crates-io): update substrate deps on crates-io

### LINKED ISSUE #3697: `gtest` v1.1.0: Can't compile tests on Linux
### Problem

When using `gtest` v1.1.0 from crates.io in the project, it is unable to build tests on Linux due to incompatible dependencies in deps tree (especially `wasmtime`). Moving to `gtest` from `gear` repo with `v1.1.0` tag helps.

https://github.com/gear-foundation/dapps/actions/runs/7637058675/job/20805177232

### Steps

1. Add `gtest = "1.1.0"` to dev-dependencies in the program's crate.
2. Run `cargo build` on Linux.
3. Get a linking error.

### Possible Solution

Reupload the `gtest` library to crates.io with a fixed dependencies versions.

clearloop:
Resolves #3697 

|package| test of compiling gtest in dapp |
|-|-|
|https://crates.io/crates/gtest/1.1.1-rc.0| https://github.com/clearloop/docker-build/actions/runs/7699331047/job/20980683093|

- [x] update substrate deps
- [x] test with gtest with candidate crates-io packages on crates-io check

@gear-tech/dev


StackOverflowExcept1on (approved):

utils/crates-io/src/publisher.rs
```diff
@@ -114,6 +114,11 @@ impl Publisher {
             panic!("Packages {failed:?} failed to pass the check ...");
         }
 
+        // Test if gtest works
+        if !crate::test("demo-syscall-error")?.success() {
```
StackOverflowExcept1on, line 118: it would be nice to test `gtest` and `gclient` together

clearloop: we currently don't have examples using `gclient` for testing and since `gclient` has no substrate dependencies, it is safe from the dependency issue, it works when it can be compiled

StackOverflowExcept1on: we have substrate dependencies in dependency tree: `gclient -> gsdk -> sp-core`. what if we change `gsdk` and it uses some unpatched dependency?

clearloop: makes sense, actually I can embed a test of `gclient` in `crates-io check` to verify it as well

breathx (approved):

PR #3713, opened 2024-02-05
refactor(pallet): Deny absence of active program data

grishasobol:


core-processor/src/processing.rs
```diff
@@ -250,17 +282,22 @@ pub fn process_error(
...
+                    origin,
+                    reason: err_payload,
+                },
+                _ => DispatchOutcome::MessageTrap {
+                    program_id,
+                    trap: err_payload,
+                },
+            }
+        }
+        ProcessErrorCase::NonExecutable(_) => DispatchOutcome::NoExecution,
```
breathx, line 300: so now for case when init message dies in waitlist we deposit event with NoExecution status instead of InitFailure?

breathx: so we dont clean its waiting init queue, dont mark it inactive etc?

breathx: we need test for that

core-processor/src/processing.rs
```diff
@@ -207,28 +230,37 @@ pub fn process_error(
...
 
-    if system_reservation_ctx.has_any() {
-        if !dispatch.is_error_reply()
+    if let ProcessErrorCase::ExecutionFailed {
+        reason,
+        allow_send_signal,
+    } = &case
+    {
+        // TODO: consider to handle error reply and init #3701
+        if *allow_send_signal
```
breathx, line 239: why do you need this allowance if you keep checking kinds right in place

core-processor/src/processing.rs
```diff
@@ -155,26 +159,45 @@ where
...
         }),
-        Err(ExecutionError::Actor(e)) => Ok(process_error(
+        Err(ExecutionError::Actor(e)) => Ok(process_execution_error(
             dispatch,
             program_id,
             e.gas_amount.burned(),
             SystemReservationContext::default(),
             e.reason,
-            true,
+            false,
```
breathx, line 168: why false

core-processor/src/precharge.rs
```diff
@@ -421,27 +406,22 @@ pub fn precharge_for_memory(
...
                 SystemReservationContext::from_dispatch(&context.data.dispatch);
-            return Err(process_error(
+            return Err(process_execution_error(
                 context.data.dispatch,
                 context.data.destination_id,
                 context.data.gas_counter.burned(),
                 system_reservation_ctx,
-                reason,
+                ActorExecutionErrorReplyReason::PreChargeGasLimitExceeded(op),
                 false,
```
breathx, line 425: why false

breathx (changes requested):

gtest/src/manager.rs
```diff
@@ -1203,4 +1206,8 @@ impl JournalHandler for ExtManager {
     fn reply_deposit(&mut self, _message_id: MessageId, future_reply_id: MessageId, amount: u64) {
         self.gas_limits.insert(future_reply_id, amount);
     }
+
+    fn waiting_init_message(&mut self, _dispatch: IncomingDispatch, _destination: ProgramId) {
+        panic!("Waiting init message is used for on-chain logic only");
```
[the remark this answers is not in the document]
grishasobol: we have panic in gtest manager for all unimplemented cases, so I would like to stay like this

pallets/gear/src/manager/mod.rs
```diff
@@ -404,42 +403,27 @@ where
...
+    fn process_failed_init(program_id: ProgramId, origin: ProgramId) {
         // Some messages addressed to the program could be processed
         // in the queue before init message. For example, that could
         // happen when init message had more gas limit then rest block
         // gas allowance, but a dispatch message to the program was
         // dequeued. The other case is async init.
         Self::clean_waitlist(program_id);
 
-        ProgramStorageOf::<T>::update_program_if_active(program_id, |p, bn| {
+        let _ = ProgramStorageOf::<T>::update_program_if_active(program_id, |p, bn| {
```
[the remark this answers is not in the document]
grishasobol: you can invent new invariants infinitely. if you use some function param only for panics and also create an additional field for in some structures, than it's bad solution, because developer while reading the code suppose that it is some necessary for logic param.

Also `executed` is a bad name and sometimes `executed` has `true` value even in cases nothing was executed.

core-processor/src/processing.rs
grishasobol: 1) I did not change this logic
2) https://github.com/gear-tech/gear/blob/69cdfeab61f932c87e9e982333c524ba593754ef/pallets/gear/src/manager/task.rs#L177-L180

core-processor/src/processing.rs
grishasobol: because we does not send signal in case of ExecutionError::Actor, but still need to handle this this error as execution. I is also possible to make additional variant in enum if you want

core-processor/src/processing.rs
grishasobol: because it was false. I have changed logic in this PR insignificantly. The logic is that if Actor error occurs than it means that it's impossible to instantiate env for execution by some reason, so we just skip signal sending, because it won't be able to be executed also.

core-processor/src/precharge.rs
grishasobol: this is a bug, thank you

grishasobol: must be true

pallets/gear/src/queue.rs
```diff
@@ -17,42 +17,33 @@
...
 where
-    T: Config,
     T::AccountId: Origin,
-    F: FnOnce(
-        PrechargedDispatch,
-    ) -> Result<(PrechargedDispatch, Option<ExecutableActorData>), PrechargedDispatch>,
 {
-    pub(crate) fn execute(self) -> Result<Vec<JournalNote>, QueueStepError> {
-        let Self {
+    pub(crate) fn queue_step(queue_step: QueueStep<'_, T>) -> Vec<JournalNote> {
```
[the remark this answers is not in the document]
grishasobol: Have any ideas?

gtest/src/manager.rs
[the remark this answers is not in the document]
breathx: The only entry for misused panic is `stop_processing` where `unimplemented` should be as well.

pallets/gear/src/queue.rs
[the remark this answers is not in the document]
breathx: execute_step(step: QueueStep)
run_step(step: QueueStep) 
🤔

pallets/gear/src/manager/mod.rs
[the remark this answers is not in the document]
breathx: Yes you can, and you must make sure that each of your brand-new ideas fit in the runtime logic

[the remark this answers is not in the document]
breathx: At least, put log::error, but I still think the panic there is better if we didn't broke anything on refactor

gtest/src/manager.rs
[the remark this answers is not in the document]
grishasobol: ok

core-processor/src/processing.rs
breathx: ExecutionError::Actor doesnt anything (and wasn't, I double checked after your answer) due to empty reservation context (its default used), so this field is unnecessary I guess ?

grishasobol: You can have not empty context. If your message was sent before runtime upgrade, which by some reason now will rejects your program on instantiation, then you need to unreserve system reservation, but no need to send signal.

gtest/src/manager.rs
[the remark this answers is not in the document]
grishasobol: done

pallets/gear/src/queue.rs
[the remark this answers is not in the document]
grishasobol: renamed

pallets/gear/src/manager/mod.rs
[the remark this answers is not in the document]
grishasobol: as I have already tell `executed` is not really executed it is just set to true or false in order to fit this panic cases, so this panics are useless

core-processor/src/processing.rs
grishasobol: new issue created

core-processor/src/precharge.rs
grishasobol: removed, now we send signal always when we can do it

pallets/gear/src/queue.rs
```diff
@@ -61,49 +51,96 @@ where
...
+                );
+            }
+
+            let (dispatch, gas, _) = precharged_dispatch.into_parts();
+            return vec![
+                JournalNote::GasBurned {
+                    message_id: dispatch.id(),
+                    amount: gas.burned(),
+                },
+                JournalNote::WaitingInitMessage {
```
ark0f, line 98: Looks like this is the only place where this journal note is used. Seems to be very ad-hoc-ish to me because journal notes expected to be created on `core-processor` side and handled on `pallet` side

ark0f (changes requested):

pallets/gear/src/manager/journal.rs
```diff
@@ -536,4 +537,27 @@ where
...
+
+    fn waiting_init_message(&mut self, dispatch: IncomingDispatch, destination: ProgramId) {
+        let (kind, message, context) = dispatch.into();
+
+        log::trace!(
+            "Append message {:?} to the waiting init messages list",
+            message.id()
+        );
+
+        let dispatch = StoredDispatch::new(kind, message.into_stored(destination), context);
```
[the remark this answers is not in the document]
grishasobol: Yes, it can but what the reason to do this?

pallets/gear/src/queue.rs
```diff
@@ -61,49 +51,96 @@ where
...
+        if matches!(program.state, ProgramState::Uninitialized { message_id }
+            if message_id != dispatch_id && dispatch_kind != DispatchKind::Reply)
+        {
+            if dispatch_kind == DispatchKind::Init {
+                // Panic is impossible, because gear protocol does not provide functionality
+                // to send second init message to any existing program.
+                unreachable!(
+                    "Init message {dispatch_id:?} is not the first init message to the program {destination_id:?}"
+                );
+            }
```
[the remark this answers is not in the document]
grishasobol: New function?

pallets/gear/src/queue.rs
grishasobol: So, I can make fn process_waiting_init() -> Vec<JournalNote> in core-processor, but this is not make scence

pallets/gear/src/queue.rs
```diff
@@ -61,49 +51,96 @@ where
...
             Ok(dispatch) => dispatch,
-            Err(journal) => return Ok(journal),
+            Err(journal) => return journal,
+        };
+
+        let Some(program) =
+            ProgramStorageOf::<T>::get_program(destination_id).and_then(|p| match p {
+                Program::Active(p) => Some(p),
+                _ => None,
+            })
```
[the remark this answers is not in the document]
grishasobol: nice, done

core-processor/src/precharge.rs
```diff
@@ -421,28 +402,22 @@ pub fn precharge_for_memory(
...
-                    return Err(process_allowance_exceed(
-                        context.data.dispatch,
-                        context.data.destination_id,
-                        context.data.gas_counter.burned(),
-                    ));
-                }
-                PrechargeError::GasExceeded(op) => {
-                    ActorExecutionErrorReplyReason::PreChargeGasLimitExceeded(op)
-                }
+            let PrechargeError::GasExceeded(op) = err else {
```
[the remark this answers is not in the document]
grishasobol: ok, return back to match

pallets/gear/src/manager/journal.rs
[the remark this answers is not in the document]
ark0f: I don't see reason to glue parts of `StoredDispatch` inside this function. I think it should be callee responsibility

pallets/gear/src/queue.rs
ark0f: The same problem like I mentioned here :/

https://github.com/gear-tech/gear/pull/3714#discussion_r1480781414

breathx (approved):

ark0f (approved):

grishasobol:
continue to work in https://github.com/gear-tech/gear/pull/3690

PR #3714, opened 2024-02-05
refactor!(pallet): Make re-instrumentation fallible

grishasobol:


core-errors/src/simple.rs
```diff
@@ -183,6 +183,10 @@ pub enum ErrorReplyReason {
     #[display(fmt = "removal from waitlist")]
     RemovedFromWaitlist = 3,
 
+    /// Program re-instrumentation failed.
+    #[display(fmt = "program re-instrumentation failed")]
+    Reinstrumentation = 4,
```
breathx, line 188:
```suggestion
    ReinstrumentationFailure = 4,
```

core-errors/src/simple.rs
```diff
@@ -204,7 +208,10 @@ impl ErrorReplyReason {
         match self {
             Self::Execution(error) => bytes[1..].copy_from_slice(&error.to_bytes()),
             Self::FailedToCreateProgram(error) => bytes[1..].copy_from_slice(&error.to_bytes()),
-            Self::InactiveProgram | Self::RemovedFromWaitlist | Self::Unsupported => {}
+            Self::InactiveProgram
```
breathx, line 211: add from_bytes impl

pallets/gear/src/tests.rs
```diff
@@ -9652,6 +9652,90 @@ fn test_reinstrumentation_works() {
...
+        let mut old_version = 0;
+        let _reset_guard = DynamicSchedule::mutate(|schedule| {
+            // Insert new original code to cause re-instrumentation failure.
+            let wasm = ProgramCodeKind::Custom("(module)").to_bytes();
+            <<Test as Config>::CodeStorage as CodeStorage>::OriginalCodeStorage::insert(
+                code_id, wasm,
+            );
+
+            old_version = schedule.instruction_weights.version;
+            schedule.instruction_weights.version = 0xdeadbeef;
```
breathx, line 9673: instead of mutating schedule version you can just downgrade instrumentation version for code

breathx (approved):

core-errors/src/simple.rs
grishasobol: done

pallets/gear/src/tests.rs
grishasobol: what the reason if I have already do like this?)

breathx: Code simplicity for future generations :)
But you can keep it as it is ofc

pallets/gear/src/queue.rs
```diff
@@ -160,7 +160,15 @@ where
                     Err(journal) => return journal,
                 };
 
-                (Pallet::<T>::reinstrument_code(code_id, &schedule), context)
+                let code = match Pallet::<T>::reinstrument_code(code_id, &schedule) {
+                    Ok(code) => code,
+                    Err(e) => {
+                        log::debug!("Re-instrumentation error for code {code_id:?}: {e:?}");
+                        return core_processor::process_reinstrumentation_error(context);
```
ark0f, line 167: And we have blurring of responsibility between `core-processor` and `pallet-gear` of journal note creation-handling chain. I don't think we can do anything here without further deeper refactoring :/

ark0f (approved):

pallets/gear/src/queue.rs
grishasobol: Yes, the correct way is to provide some generic in order to give core-processor ability to have limited access to storage

pallets/gear/src/tests.rs
grishasobol: It's more simple solution, because downgrading version means that we are sure it's not 0. About 0xdeadbeaf we can be sure it won't be reached

PR #3720, opened 2024-02-06
fix(pallet-gear): Remove incorrect `log::error`

### LINKED ISSUE #3707: Some messages from waiting from init list can be removed from wait list
### Problem

We have `log::error` in case some message id is in waiting init list, but does not have message in wait list:
https://github.com/gear-tech/gear/blob/24d905e3068f9f70219be4ab17428d966bdbae88/pallets/gear/src/manager/journal.rs#L62-L75

Even if we forget about  #3705, here can be some  cases, when this can appear:
1) Message has been already removed from wait list because of hold in wait list expiring
2) Message has been already woken by program inside init.
3) Message has been woken by program inside init, then goes to message queue and then appended to waiting init list once more, so we will have two equal message ids in waiting init list, but only for first id message will be found in wait list.  

### Possible Solution

We should decide whether we consider this case as error or not. Actually I didn't found any problems, which can occur  in that case.

breathx:
Resolves #3707.

Truly not an error. Similar to waking some random message id.

### DIFF
--- a/pallets/gear/src/manager/journal.rs
+++ b/pallets/gear/src/manager/journal.rs
@@ -58,19 +58,17 @@ where
     ) {
         use CoreDispatchOutcome::*;
 
-        let wake_waiting_init_msgs = |p_id: ProgramId| {
-            ProgramStorageOf::<T>::waiting_init_take_messages(p_id)
+        let wake_waiting_init_msgs = |program_id: ProgramId| {
+            ProgramStorageOf::<T>::waiting_init_take_messages(program_id)
                 .into_iter()
-                .for_each(|m_id| {
-                    if let Some(m) = Pallet::<T>::wake_dispatch(
-                        p_id,
-                        m_id,
+                .for_each(|message_id| {
+                    if let Some(dispatch) = Pallet::<T>::wake_dispatch(
+                        program_id,
+                        message_id,
                         MessageWokenSystemReason::ProgramGotInitialized.into_reason(),
                     ) {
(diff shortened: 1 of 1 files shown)


StackOverflowExcept1on (approved):

pallets/gear/src/manager/journal.rs
```diff
@@ -58,19 +58,17 @@ where
...
+                    if let Some(dispatch) = Pallet::<T>::wake_dispatch(
+                        program_id,
+                        message_id,
                         MessageWokenSystemReason::ProgramGotInitialized.into_reason(),
                     ) {
-                        QueueOf::<T>::queue(m)
+                        QueueOf::<T>::queue(dispatch)
                             .unwrap_or_else(|e| unreachable!("Message queue corrupted! {:?}", e));
-                    } else {
-                        log::error!("Cannot find message in wl")
```
techraed, line 73: why not just change in to `trace`? I'm ok with removing, but that gives more useful logs in case of debugging.

techraed (approved):

pallets/gear/src/manager/journal.rs
breathx: I dont see any point for that. Furthermore, waiting init list etc is about to be deprecated

techraed: removing waiting init list makes it good point to remove log, however I'd still vote for leaving the log to be `trace`.

PR #3725, opened 2024-02-09
feat(runtime): Add ability to decline given voucher

breathx:


pallets/payment/src/tests.rs
```diff
@@ -1389,6 +1389,185 @@ fn voucher_call_deprecated_upload_payer_err() {
...
+            BOB,
+            voucher_initial_balance,
+            None,
+            false,
+            100,
+        ));
+        let voucher_id = get_last_voucher_id();
+        let voucher_account_id = voucher_id.cast::<AccountIdOf<Test>>();
+
+        System::set_block_number(System::block_number() + 1_000);
```
shamilsan, line 1493: Isn't it better to call `run_to_block`?

pallets/gear-voucher/src/lib.rs
```diff
@@ -563,14 +571,57 @@ pub mod pallet {
...
+        /// Decline existing and not expired voucher.
+        ///
+        /// This extrinsic expires voucher of the caller, if it's still active,
+        /// allowing it to be revoked.
+        ///
+        /// Arguments:
+        /// * voucher_id:   voucher id to be declined.
+        #[pallet::call_index(5)]
+        #[pallet::weight(T::DbWeight::get().reads_writes(2, 4))]
+        pub fn decline(origin: OriginFor<T>, voucher_id: VoucherId) -> DispatchResultWithPostInfo {
```
shamilsan, line 597: Maybe better to `cancel` or `discontinue` instead of `decline`?

pallets/gear-voucher/src/lib.rs
```diff
@@ -563,14 +571,57 @@ pub mod pallet {
...
+            let current_bn = <frame_system::Pallet<T>>::block_number();
+
+            // Validating that voucher is not expired.
+            ensure!(current_bn < voucher.expiry, Error::<T>::VoucherExpired);
+
+            // Set voucher into expired state.
+            voucher.expiry = current_bn;
+
+            // Updating voucher in storage.
+            // TODO: consider revoke here once gas counting implemented.
```
shamilsan, line 615: `TODO` without an issue?

runtime/vara/src/lib.rs
```diff
@@ -1091,7 +1091,7 @@ impl pallet_gear_voucher::Config for Runtime {
     type Currency = Balances;
     type PalletId = VoucherPalletId;
     type WeightInfo = weights::pallet_gear_voucher::SubstrateWeight<Runtime>;
-    type CallsDispatcher = Gear;
+    type CallsDispatcher = pallet_gear::DefaultDispatcher<Runtime>;
```
shamilsan, line 1094: What's the difference?

pallets/payment/src/tests.rs
breathx: There we dont care about any hooks or any other activities, except changing bn. Similar hack used in many other places

breathx: Lets say, test time optimisation

runtime/vara/src/lib.rs
breathx: Previously call dispatcher trait was on GearPallet, but GearPallet doesn't have access to GearVoucher fns, since its Config doesn't require `: pallet_gear_voucher: Config`, so this wrap accepts runtime with both impls, furthermore such approach allows partially implement dispatcher for some other tests avoiding bringing all of the pallets at a time

pallets/gear-voucher/src/lib.rs
breathx: I was thinking about `reject` but it a little more rude that neutral `decline`
`discontinue` feels ultra weird
`cancel` sounds like I terminate voucher that was issued by ME. while `decline` shows that somebody "offered" me the voucher.
So I'd go with this naming atm or consider any other possibilities

shamilsan: Ah, it is declined by the user, not the voucher issuer. Make sense!

breathx: > Ah, it is declined by the user, not the voucher issuer. Make sense!

yeah, by its spender*

pallets/gear-voucher/src/internal.rs
```diff
@@ -144,6 +145,7 @@ pub trait PrepaidCallsDispatcher {
     fn dispatch(
         account_id: Self::AccountId,
         sponsor_id: Self::AccountId,
+        voucher_id: VoucherId,
```
grishasobol, line 148: `sponsor_id` is now exceeded and must be gotten inside `dispatch `function for corresponding calls:
`let sponsor_id = voucher_id.cast()`

pallets/gear/src/lib.rs
```diff
@@ -1774,7 +1777,10 @@ pub mod pallet {
         }
     }
 
-    impl<T: Config> PrepaidCallsDispatcher for Pallet<T>
+    /// Dispatcher for all types of prepaid calls: gear or gear-voucher pallets.
+    pub struct DefaultDispatcher<T: Config + pallet_gear_voucher::Config>(PhantomData<T>);
```
ark0f, line 1781:
```suggestion
    pub struct PrepaidCallDispatcher<T: Config + pallet_gear_voucher::Config>(PhantomData<T>);
```
Too generic nam

pallets/gear-voucher/src/internal.rs
grishasobol: deprecated vouchers can also use only `voucher_id`

grishasobol (approved):

pallets/gear-voucher/src/lib.rs
```diff
@@ -563,14 +571,57 @@ pub mod pallet {
...
+
+        /// Decline existing and not expired voucher.
+        ///
+        /// This extrinsic expires voucher of the caller, if it's still active,
+        /// allowing it to be revoked.
+        ///
+        /// Arguments:
+        /// * voucher_id:   voucher id to be declined.
+        #[pallet::call_index(5)]
+        #[pallet::weight(T::DbWeight::get().reads_writes(2, 4))]
```
ark0f, line 596: Why don't we bench it?

pallets/gear/src/lib.rs
```diff
@@ -1792,12 +1798,14 @@ pub mod pallet {
                 PrepaidCall::UploadCode { code } => {
                     <T as Config>::WeightInfo::upload_code(code.len() as u32 / 1024)
                 }
+                PrepaidCall::DeclineVoucher => T::DbWeight::get().reads_writes(2, 4),
```
ark0f, line 1801: Duplicated weight from `#[pallet::weight(...)]` on call itself

ark0f (changes requested):

pallets/gear-voucher/src/lib.rs
AndrePanin: The operation leads to voucher expiration, as a result it is not possible to use it, right? Maybe `Deactivate` ? `Decline` is good too.

pallets/gear-voucher/src/lib.rs
```diff
@@ -563,14 +571,57 @@ pub mod pallet {
                 !matches!(call, PrepaidCall::UploadCode { .. }),
                 Error::<T>::CodeUploadingDisabled
             );
+            ensure!(
+                !matches!(call, PrepaidCall::DeclineVoucher),
+                Error::<T>::InexistentVoucher
+            );
```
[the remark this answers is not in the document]
shamilsan: Why?

pallets/gear-voucher/src/internal.rs
breathx: At the moment it makes sense, but future changes to vouchers will bring us back to 3 params because sponsor won't be eq voucher_id.cast(), so I'd keep it atm in favour of more clear signature and usage of dispatcher

PR #3731, opened 2024-02-12
fix(gear-authorship): increase time limit in tests

ukint-vs:
Resolves # .

-
-
-

@reviewer-or-team

### DIFF
--- a/node/authorship/src/tests.rs
+++ b/node/authorship/src/tests.rs
@@ -462,7 +462,7 @@ fn queue_remains_intact_if_processing_fails() {
         BlockId::number(0),
         Box::new(time::Instant::now),
         timestamp,
-        1500_u64,
+        2000_u64,
         None,
         proposal
     );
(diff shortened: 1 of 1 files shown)


node/authorship/src/tests.rs
```diff
@@ -462,7 +462,7 @@ fn queue_remains_intact_if_processing_fails() {
         BlockId::number(0),
         Box::new(time::Instant::now),
         timestamp,
-        1500_u64,
+        2000_u64,
```
grishasobol, line 465: what does this limit mean? And how we can calculate it using chain block production time limit?

ekovalev: This is the actual time limit for block production: 1/2 of a slot duration (3s), which will be later reduced again by 1/3 inside the proposing routine so that we have that 1s for extrinsics processing that is expected by the runtime.

techraed: did it actually help resolve CI issue? problem occurs on master

breathx: this actually shouldn't be increased just for passing tests, because its time dependent. separate CI step for only this ignored test, aggregating all of them in just one thread

breathx (changes requested):
Check comment above

breathx:
see https://github.com/gear-tech/gear/pull/3685#issuecomment-1927486445

Issue #3718, opened 2024-02-06
does not make system unreserve in case `ExecutionError::Actor` is returned


grishasobol:
### Problem

Currently we use default system reservation context in case `ExecutionError::Actor` is returned from `execute_wasm`. This is not correct in case message had been executed and has reservation context, but by some reasons after wake `ExecutionError::Actor` occurs. This can happen for example if we would make runtime upgrade with some breaking changes in wasm executor environment instantiation. And message was waked after runtime upgrade.
So, we must handle this case, in order to avoid possible panic from this place:
https://github.com/gear-tech/gear/blob/69cdfeab61f932c87e9e982333c524ba593754ef/common/src/gas_provider/internal.rs#L580-L584

ark0f:
Worth to mention system reservation __have__ to be unreserved after execution in any circumstances

PR #3743, opened 2024-02-21
fix!(pallet-gear): append outgoing bytes limit to prevent runtime allocator overflow

grishasobol:
Resolves https://github.com/gear-tech/core-audit/issues/25 and #3718

1) Appends `OutgoingBytesLimit` runtime constant and corresponding `outgoing_bytes_limit` to `ContextSettings`.
2) `MessageContext` now takes in account `outgoing_bytes_limit` in `send_commit`, `send_push`, `send_push_input`.

P.S. special thanks to @gshep @techraed


core-processor/src/common.rs
```diff
@@ -498,6 +498,9 @@ pub enum SystemExecutionError {
     /// Error during `into_ext_info()` call
     #[display(fmt = "`into_ext_info()` error: {_0}")]
     IntoExtInfo(MemoryError),
+    /// Incoming dispatch store has too many outgoing messages total bytes.
+    #[display(fmt = "Incoming dispatch store has too many outgoing messages total bytes")]
+    MessageStoreOutgoingBytesOverflow,
```
grishasobol, line 503: This error will be returned from `core_processor::process`

core/src/message/context.rs
```diff
@@ -34,74 +34,36 @@ use scale_info::{
     TypeInfo,
 };
 
-use super::{DispatchKind, IncomingDispatch};
+use super::{DispatchKind, IncomingDispatch, Packet};
 
 /// Context settings.
-#[derive(Copy, Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Decode, Encode, TypeInfo)]
+#[derive(Copy, Clone, Debug, Default)]
 pub struct ContextSettings {
```
grishasobol, line 41: NOTE: Make fields pub

core/src/message/context.rs
```diff
@@ -121,7 +83,7 @@ pub struct ContextOutcomeDrain {
 /// Context outcome.
 ///
 /// Contains all outgoing messages and wakes that should be done after execution.
-#[derive(Default, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Decode, Encode, TypeInfo)]
+#[derive(Default, Debug)]
```
grishasobol, line 86: NOTE: remove useless implementations

core/src/message/context.rs
```diff
@@ -224,31 +186,47 @@ impl ContextStore {
...
 impl MessageContext {
-    /// Create new MessageContext with given ContextSettings.
+    /// Create new message context.
+    /// Returns `None` if outgoing messages bytes limit exceeded.
     pub fn new(
         dispatch: IncomingDispatch,
         program_id: ProgramId,
         settings: ContextSettings,
-    ) -> Self {
+    ) -> Option<Self> {
```
grishasobol, line 206: NOTE: now returns Option<Self>. May be better return an error, but I do not see the difference

core/src/message/mod.rs
```diff
@@ -78,6 +77,14 @@ impl Display for PayloadSizeError {
 /// Payload type for message.
 pub type Payload = LimitedVec<u8, PayloadSizeError, MAX_PAYLOAD_SIZE>;
 
+impl Payload {
+    /// Get payload length as u32.
+    pub fn len_u32(&self) -> u32 {
```
grishasobol, line 82: NOTE: add this method in oder to be able to use payload len as u32 without panics in gear-core/src/context.rs

pallets/gear/src/lib.rs
```diff
@@ -165,10 +165,14 @@ pub mod pallet {
         #[pallet::constant]
         type Schedule: Get<Schedule<Self>>;
 
-        /// The maximum amount of messages that can be produced in single run.
+        /// The maximum amount of messages that can be produced in one message processing.
```
grishasobol, line 168: NOTE: actually is not limit  for single run, but for message all execution, because `outgoing` map is not cleaned between executions

pallets/gear/src/queue.rs
```diff
@@ -182,7 +182,7 @@ where
             (context, code, balance).into(),
             (random.encode(), bn.unique_saturated_into()),
         )
-        .unwrap_or_else(|e| unreachable!("core-processor logic invalidated: {e}"))
+        .unwrap_or_else(|e| unreachable!("{e}"))
```
grishasobol, line 185: NOTE: was incorrect info message, because errors can appear in case `pallet-gear` provides for `core-processor` incorrect arguments or suitable environment for execution has not been provided.

NikVolf (approved):

runtime/vara/src/lib.rs
```diff
@@ -973,6 +973,9 @@ parameter_types! {
     pub const DispatchHoldCost: u64 = 100;
 
     pub const OutgoingLimit: u32 = 1024;
+    // 64 MB, must be less than max runtime heap memory.
+    // NOTE: currently runtime heap memory is 1 GB (see https://shorturl.at/DET45)
+    pub const OutgoingBytesLimit: u32 = 64 * 1024 * 1024;
```
grishasobol, line 978: You opinion here is welcome. May be we should set bigger limit, may be smaller. Actually current memory is 32 mb, so I think is impossible to send more than that value, because of gas consumption&

core/src/message/context.rs
```diff
@@ -352,10 +345,18 @@ impl MessageContext {
 
     /// Pushes payload into stored payload by handle.
     pub fn send_push(&mut self, handle: u32, buffer: &[u8]) -> Result<(), Error> {
+        let bytes_amount = u32::try_from(buffer.len()).map_err(|_| Error::MaxMessageSizeExceed)?;
```
grishasobol, line 348: Actually both `OutgoingMessagesBytesLimitExceeded` and  `MaxMessageSizeExceed` are suitable to be returned. But I chose the old one to be more suitable for legacy

core-processor/src/common.rs
grishasobol: **IMPORTANT**: By the way, this error can happen if some already existed message exceed outgoing bytes limit, so this will cause panic in runtime... So may be we must process this case as message execution error.

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -179,8 +179,9 @@ fn default_processor_context<T: Config>() -> ProcessorContext {
         message_context: MessageContext::new(
```
shamilsan, line 179: We could derive `Default` for `MessageContext` and use `Default::default` instead of `MessageContext::new(...)`.

In general, it is good practice to derive `D`-`D`-`C` (`Debug`, `Default`, `Clone`) to all public structs.

grishasobol: actually `MessageContext` is not public because truly it uses only inside `core-processor`

grishasobol: so, it should be in core-processor but we define it in gear-core I don't know why

NikVolf:
Can you move your PR notes to the code itself @grishasobol ?

grishasobol:
> Can you move your PR notes to the code itself @grishasobol ?
> 
> And resolve the rest

Ok, some of them truly can be added in code. About _possible_ gear_run panics  - I'm thinking what is better to do to prevent them

core-processor/src/common.rs
grishasobol: I've change this temporary: instead of returning `SystemExecutionError::MessageStoreOutgoingBytesOverflow`, we return now `ActorExecutionError::UnsupportedMessage`. After next runtime upgrade we should return back to system error.

breathx (changes requested):
Pls wait for my review

NikVolf:
@breathx please 

> Pls wait for my review

core-processor/src/executor.rs
```diff
@@ -182,7 +182,14 @@ where
         AllocationsContext::new(allocations.clone(), static_pages, settings.max_pages);
 
     // Creating message context.
-    let message_context = MessageContext::new(dispatch.clone(), program_id, msg_ctx_settings);
+    let Some(message_context) = MessageContext::new(dispatch.clone(), program_id, msg_ctx_settings)
```
breathx, line 185: whoa, I was missing it. I'll open an issue from this comment to remove cloning here because it could be pretty harmful

breathx (changes requested):

core-processor/src/common.rs
grishasobol: Yes

core-processor/src/executor.rs
```diff
@@ -380,30 +387,33 @@ where
...
+                    .try_into()
+                    .map_err(|e| format!("Failed to create payload: {e:?}"))?,
+                gas_limit,
+                Default::default(),
+                Default::default(),
+            ),
+            None,
+        ),
+        program.id(),
+        Default::default(),
```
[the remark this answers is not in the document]
grishasobol: first comment ? What the reason to define not zero limits if message sending is restricted?

core/src/message/context.rs
```diff
@@ -241,31 +201,47 @@ impl ContextStore {
...
+                let mut counter = 0u32;
+                for payload in store.outgoing.values().filter_map(|x| x.as_ref()) {
+                    counter = counter.checked_add(payload.len_u32())?;
+                }
+                counter
+            }
+            None => 0,
+        };
+
+        (outgoing_bytes_counter <= settings.outgoing_bytes_limit).then_some(())?;
```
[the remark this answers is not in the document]
grishasobol: you mean `.then(Some(()))` ?

core/src/message/context.rs
```diff
@@ -241,31 +201,47 @@ impl ContextStore {
...
+
+        Some(Self {
             kind,
             outcome: ContextOutcome::new(program_id, message.source(), message.id()),
             current: message,
             store: store.unwrap_or_default(),
             settings,
-        }
+            outgoing_bytes_counter,
```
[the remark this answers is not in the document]
grishasobol: > it saves us both from reallocs on journal notes creation + stored context creation

It does not save as from allocator overflow. You can make a lot of send_push without send_commit in first execution before wait. After wake we should takes in account these bytes because they take runtime memory.

core-processor/src/executor.rs
grishasobol: I don't think it's harmful, but pretty stupid yes

core/src/message/context.rs
```diff
@@ -687,8 +847,9 @@ mod tests {
         let mut message_context = MessageContext::new(
             Default::default(),
             Default::default(),
-            ContextSettings::new(0, 0, 0, 0, 0, 1024),
-        );
+            ContextSettings::with_outgoing_limits(1024, u32::MAX),
+        )
+        .unwrap();
```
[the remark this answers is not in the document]
grishasobol: done

core-processor/src/common.rs
grishasobol: commented

runtime/vara/src/lib.rs
grishasobol: Yes 512 it's too big currently. We must have at lease 2 GB runtime heap size for that by default.

core-processor/src/executor.rs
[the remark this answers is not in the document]
breathx: Message sending there isn't restricted. Fns purpose is to collect the reply

core/src/message/context.rs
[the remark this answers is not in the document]
breathx: I mean replacing with common `if`

core/src/message/context.rs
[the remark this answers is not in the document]
breathx: Got you point, overlooking it

core-processor/src/executor.rs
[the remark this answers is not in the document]
grishasobol: What do you want exactly?

core/src/message/context.rs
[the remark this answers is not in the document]
grishasobol: done

core/src/message/context.rs
```diff
@@ -328,6 +307,12 @@ impl MessageContext {
         delay: u32,
         reservation: Option<ReservationId>,
     ) -> Result<MessageId, Error> {
+        let new_outgoing_bytes = self
```
breathx, line 310: to consider: move inside success branch. For example, this check is excess if Error::LateAccess scenario

breathx (changes requested):
Everything is good, thanks for your patience!

Requesting changes due to lack of impls and tests for `reply_push`, `reply_push_input` and `reply_commit`

core-processor/src/executor.rs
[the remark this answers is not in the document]
breathx: To be able to send messages in `fn execute_for_reply`. Tests dont fail due to lack of impls, see lates review

grishasobol:
> Everything is good, thanks for your patience!
> 
> Requesting changes due to lack of impls and tests for `reply_push`, `reply_push_input` and `reply_commit`

We don't need to implement this for replys because this is `outgoing messages bytes limit`. Reply is limited by max message size and we do not need to append additional counter for replies.

core/src/message/context.rs
```diff
@@ -34,76 +34,36 @@ use scale_info::{
...
 };
 
-use super::{DispatchKind, IncomingDispatch};
+use super::{DispatchKind, IncomingDispatch, Packet};
 
 /// Context settings.
-#[derive(
-    Copy, Clone, Default, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Decode, Encode, TypeInfo,
-)]
+#[derive(Copy, Clone, Debug, Default)]
```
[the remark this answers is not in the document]
grishasobol: May be. Actually it's not a big problem

core/src/message/context.rs
grishasobol: Agree. But I think it's more wide problem - because we have a bug: after payload `take()` in `send_commit` and `reply_commit` we do not set payload back in case an error occurs.

grishasobol: done. issue created #3779

techraed (approved):

core/src/message/mod.rs
```diff
@@ -214,6 +221,9 @@ pub trait Packet {
     /// Packet payload bytes.
     fn payload_bytes(&self) -> &[u8];
 
+    /// Payload len
+    fn payload_len(&self) -> u32;
```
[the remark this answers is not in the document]
grishasobol: Current implementations has type-level guaranties: 
https://github.com/gear-tech/gear/blob/06ec57ada5c8746807519938fb01c3524b57b160/core/src/message/mod.rs#L56-L57
https://github.com/gear-tech/gear/blob/06ec57ada5c8746807519938fb01c3524b57b160/core/src/message/mod.rs#L81-L85

breathx (approved):

PR #3747, opened 2024-02-22
chore(test, gsdk): Add a test case showing how to query a storage

gshep:
Demonstrate how to query various program counters from the program storage. The test is ignored by default in order not to increase overall testing time.

@gear-tech/dev


clearloop (approved):
This test is actually out of the general `rpc` concept, would be better renaming it to `tests/storage.rs`

PR #3750, opened 2024-02-26
fix(gring): remove ss58 version configuration

### LINKED ISSUE #3749: The default SS58 prefix is not initialized successfully in gring
### Problem

code: https://github.com/gear-tech/gear/blob/master/utils/gring/src/ss58.rs#L34
interface: https://github.com/gear-tech/gear/blob/master/utils/gring/src/ss58.rs#L102

The atomic storage is defined in module `ss58`, when calling `default_ss58_version()`:

- it works in tests
- it doesn't work for the code in other modules
---

EDITED: this is caused by https://github.com/gear-tech/gear/blob/master/utils/gring/src/keyring.rs#L78

### Steps

just calling `default_ss58_version` in different places, run the command:

```
gring list
| Name               | Address                                           |
| ------------------ | ------------------------------------------------- |
| uuc                | 13zZKyRapef1PBqo4nyP8m865kCNwL7wiHKo6L7JoiWoiQtv |
| _gring_test_new-1  | 167fCpVQWgirXJyoWmkt7HYGJnmy6pzyT83q6uhf3SzarDus |
| _dev_alice         | 15oF4uVJwmo4TdGW7VfQxNLavjCXviqxT9S1MgbjMNHr6Sp5 |
```

you'll see the encoded addresses are formatted with prefix `0`

write a test, it will be formatted with prefix `137`, vara addresses should starts with `kG`

### Possible Solution

_No response_

### Notes

_No response_

### Relevant Log Output

<p>

```
<Paste your log here>
```

</p>
</details>

clearloop:
Resolves #3749 

after thinking over about it, ss58 version config is actually not necessary in the keyring config, it is previous used for adding the ss58 address to the new created addresses

```json
{
  "encoded": "...",
  "address": "this-is-the-field-for-ss58-address"
}
```

1. gring should generate wallets just for vara by default, and even if the caller libraries really want to change the default ss58 version, they still can do it via `set_default_ss58_version` interface before calling gring but not using the config file
2. whatever the ss58 prefix is, the decoded public key will always be the same ( the decoding algorithm of ss58 address skips the ss58 prefix )
3. people can <kbd>add</kbd> their wallets exported from the polkadot-js extension, and the addresses listed by gring <kbd>list</kbd> should follow the original address defined in the keystore without any modifications

- [x] pin the ss58 version of gring to vara prefix
- [x] introduce command `add`

@gear-tech/dev


NikVolf (approved):
agree with motivation!

PR #3753, opened 2024-02-27
feat(gcli): make gas limit optional argument

### LINKED ISSUE #3709: Make gas limit optional arguments in gcli
what if user truly wants to send message with 0 gas limit?
for example, it's valid case for reply nowadays

_Originally posted by @breathx in https://github.com/gear-tech/gear/pull/3687#discussion_r1474350796_

clearloop:
Resolves #3709 
@gear-tech/dev

### DIFF
--- a/gcli/src/cmd/reply.rs
+++ b/gcli/src/cmd/reply.rs
@@ -40,6 +40,8 @@ pub struct Reply {
     #[arg(short, long, default_value = "0x")]
     payload: String,
     /// Reply gas limit
+    ///
+    /// Use estimated gas limit automatically if not set.
     #[arg(short, long)]
     gas_limit: Option<u64>,
     /// Reply value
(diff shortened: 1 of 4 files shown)


gcli/src/cmd/reply.rs
```diff
@@ -52,7 +52,9 @@ impl Reply {
         let signer = app.signer().await?;
         let reply_to_id = self.reply_to_id.to_hash()?;
 
-        let gas_limit = if self.gas_limit == 0 {
+        let gas_limit = if let Some(gas_limit) = self.gas_limit {
+            gas_limit
+        } else {
```
shamilsan, line 57: I'd use `unwrap_or_else`.
```suggestion
        let gas_limit  = self.gas_limit.unwrap_or_else(|| {
```

shamilsan (approved):

gcli/src/cmd/reply.rs
clearloop: calculating gas requires `async` methods, so it is hard to be placed in the `else` closure even it is cleaner

shamilsan: Ah, agree, missed it.

breathx (approved):

PR #3755, opened 2024-02-27
feat(pallet-gear): test program with large indexes

### LINKED ISSUE #3754: Test program with large element indexes
### Problem to Solve

https://github.com/casper-network/casper-wasm/pull/1

We have WASM deserializing logic in our runtime https://github.com/gear-tech/gear/blob/master/core/src/code/mod.rs#L120, so if this problem exists, we need to update our `parity-wasm` fork as well.

### Possible Solution

1. write a test with our current implementation
2. cherry pick capser-wasm if it can solve the problem

### Notes

_No response_

clearloop:
Resolves #3754

This PR provides a testing program filled by empty functions to reach the indexes limit and the code length limit to confirm we are not affected by the deserialization problem of parity-wasm

From my point of view I prefer not cherry picking the fix from `casper-wasm` at the moment:

1. we have the code size limit `512 * 1024 (512kb)`,  and the deserializing logic slows down significantly on my 16GM memory machine when the indexes reaches `268_435_441`, at least code size `1024 * 1024 * 1024 (1gb)`
   1. I can dig the specific result of the cost between `Vec` and `BTreeMap` as the `IndexMap` of `parity-wasm`, if anybody like to see it.
3. the stability of `parity-wasm` has been verified by the whole industry for years, picking the update may lead us to new problems since `parity-wasm` has already been archived
4. instead of picking the update, I'd vote on switching to [wasm-tools](https://github.com/bytecodealliance/wasm-tools) since its code base is still under active development

On the other hand:

this issue is also considerable for the runtime upgrades or WASM modules used for the chain configuration, but it could be discovered on general tests of launching a new runtime upgrades

- [x] add program test with large indexes
- [x] add index-map test with large indexes
- [ ] ~~update parity-wasm if it can solve the problem~~

@gear-tech/dev

### DIFF
--- a/pallets/gear/src/schedule.rs
+++ b/pallets/gear/src/schedule.rs
@@ -1440,16 +1440,16 @@ mod test {
 
         // NOTE:
         //
-        // For triggering the bug, assigning `max_idx` to `u32::MAX`.
+        // For triggering the bug, assigning `u32::MAX` to `max_idx`.
         let max_idx = ((code_limit - empty_program_len) / empty_fn_len + 1) as u32;
 
         // NOTE:
         //
-        // We are not generating the wasm module in memory
-        // because it takes too much.
+        // We are not generating the wasm module in memory because it
+        // takes too much.
         //
         // Mock the max idx in the index map of parity-wasm, cherry-pick
-        // https://github.com/casper-network/casper-wasm/pull/1 not our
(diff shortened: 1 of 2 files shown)


breathx (approved):

breathx:
Agree here. Programs size limit could be expanded to a much bigger one without leading to this error. What do you think, @NikVolf ?

PR #3762, opened 2024-02-28
fix(tools): split pre-processing, optimization and code validation

NikVolf:
Changes:
- rework of pre-processors
- add `code_validator` module to `gear-wasm-builder`
- split logic into several parts


StackOverflowExcept1on (changes requested):
we should probably rework the pre-processor and optimizer to fix the original problem

breathx:
> we should probably rework the pre-processor and optimizer to fix the original problem

What is the original problem?

NikVolf:
> > we should probably rework the pre-processor and optimizer to fix the original problem
> 
> What is the original problem?

wasm builder is not very kind to complex custom builds with multiple steps

utils/wasm-builder/src/wasm_project.rs
```diff
@@ -427,26 +430,51 @@ extern "C" fn metahash() {{
...
+                        pre_processed_path,
+                        format!("{file_base_name}_{pre_processor_name}"),
+                    ));
+                }
+            }
+        }
 
-            let mut pre_processed_path = self.wasm_target_dir.clone();
-            pre_processed_path.push(output.path);
+        if wasm_files.is_empty() {
```
NikVolf, line 457: original always should be pushed, why skip it?

or there should be the way for pre-processor to preserve the original compilation artifact under original name

StackOverflowExcept1on: done in 454cf15d50072d43d3db7ba8f0395ed5f5a10907

StackOverflowExcept1on (approved):

PR #3767, opened 2024-02-29
ci(PR): manage validator checks in `PR.yml`

clearloop:
Resolves #3765 
Closes #3820 

triggers validator-checks inside of `PR.yml` in case of creating empty check suite in github UI

@gear-tech/dev

PR #3769, opened 2024-02-29
ci(PR): gather pull request actions in `PR.yml`

### LINKED ISSUE #3768: Call docs and typos checks in PR.yml
### File Location(s)

https://github.com/gear-tech/gear/blob/master/.github/workflows/docs.yml

### Proposal

- move the content of `typos.yml` to `PR.yml`
- make docs a workflow call and redirect it to `PR.yml`

after these, we can gather all of the workflows in to 1 namespace, make everything cleaner

clearloop:
Resolves #3768

after this PR and #3767, we'll only have one name space `PR` in the checks of pull requests

- [x] rename `typos.yml`
- [x] call docs inside of `PR.yml`

@gear-tech/dev


StackOverflowExcept1on (approved):

.github/actions/message/action.yml
```diff
@@ -12,6 +12,8 @@ outputs:
     description: If enable cache.
   check:
     description: If trigger check.
+  docs:
```
breathx, line 15: docs always should be triggered I guess

breathx (approved):
Same as #3767  needs master fetch

.github/workflows/PR.yml
```diff
@@ -52,6 +52,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: ./.github/actions/docs
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
StackOverflowExcept1on, line 56: iirc secrets starting with "GITHUB_" are not supported

clearloop: > iirc secrets starting with "GITHUB_" are not supported


what do you mean ?

https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret

and it works in fc33a96 as well

breathx:
@clearloop ready to merge?

PR #3771, opened 2024-02-29
feat(node): Implement builtin actor providing bls12-381

### LINKED ISSUE #3513: Built-in actor providing bls381
### Problem to Solve

inspired by #3504 

### Possible Solution

Implement built-in actor that is able to compute miller loop and exponentiation for bls381.

### Notes

_No response_

gshep:
Resolves #3513 

- `decode_vec` benchmark since it's required to scale-decode vectors by hand for gas counting
- crate `gear-builtins-bls381` declaring entities for bls12_381 crypto
- benchmarks for all bls12-381 funcs and corresponding update of weights
- implementation of the builtin actor functionality

@gear-tech/dev


gbuiltins/bls381/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+#[cfg(not(feature = "std"))]
+extern crate alloc;
+
+#[cfg(not(feature = "std"))]
+use alloc::vec::Vec;
+
+use codec::{Decode, Encode};
+
+/// Constant defines codec index of [`Request::MultiMillerLoop`].
+pub const REQUEST_MULTI_MILLER_LOOP: u8 = 0;
```
NikVolf, line 30: Why do you need those constants, especially in public?

gbuiltins/bls381/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+    ///   - `base`: `ArkScaleProjective<G2Projective>`.
+    ///   - `scalar`: `ArkScale<Vec<u64>>`.
+    #[codec(index = 5)]
+    ProjectiveMultiplicationG2 { base: Vec<u8>, scalar: Vec<u8> },
+}
+
+/// The enumeration represents possible common errors for all requests.
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
+#[codec(crate = codec)]
+pub enum CommonError {
```
NikVolf, line 96: Maybe rename to just `Error` ?

gbuiltins/bls381/src/lib.rs
NikVolf: if absolutely required to get hint about payload, put enum + const generation under `macro_rules!` to avoid duplication

gbuiltins/bls381/src/lib.rs
gshep: ok

gbuiltins/bls381/src/lib.rs
gshep: We need to decode payload manually in builtin actor for precise gas counting since there is no instrumentation similar to gear programs.

Good point for macro, thanks! However enum variants can't be generated by the Rust macro so to implement such kind of logic one has to use push-down accumulation, incremental TT munchers and other approaches. So I'm afraid the macro will be over complicated and take more lines of code than the current implementation.

Therefore if you don't mind I would stay with that solution of defining constants/variants by hands.

pallets/gear-builtin/src/benchmarking.rs
```diff
@@ -112,6 +134,199 @@ benchmarks! {
...
+    }
+
+    bls12_381_multi_miller_loop {
+        let c in 0 .. 100;
+
+        let count = c as usize;
+
+        let mut rng = ark_std::test_rng();
+
+        let messages = {
```
mertwole, line 164:
```suggestion
        let messages = (0..count).map(|_| G1::rand(&mut rng).into()).collect::<Vec<G1Affine>>();
```
is far shorter

mertwole (commented):
Some notes about benches

breathx:
rebase and set as ready for review pls @gshep

gbuiltins/bls381/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+#[derive(Encode, Clone, PartialEq, Eq, Debug)]
+#[codec(crate = codec)]
+pub enum Request {
+    /// Request to pairing multi Miller loop for *BLS12-381*.
+    ///
+    /// Encoded:
+    ///   - `a`: `ArkScale<Vec<G1Affine>>`.
+    ///   - `b`: `ArkScale<Vec<G2Affine>>`.
+    #[codec(index = 0)]
+    MultiMillerLoop { a: Vec<u8>, b: Vec<u8> },
```
ekovalev, line 52: What is the problem accepting `ArkScale<_>` types themselves as parameters?
In any case the caller would have to create `a` and `b` as `ArkScale` instances and then encode those into bytes.
Isn't there a way to charge gas for decoding on the receiving end based on the total payload size only?

breathx:
to be reviewed asap

gbuiltins/bls381/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// Result of the projective multiplication.
+    ProjectiveMultiplication(Result<Vec<u8>, ()>),
+}
+
+/// Result of the multi Miller loop computation.
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
+#[codec(crate = codec)]
+pub enum MultiMillerLoopResult {
+    /// Encoded: `ArkScale<Bls12_381::TargetField>`.
+    Ok(Vec<u8>),
```
ekovalev, line 129: In my opinion returning a byte vec to the caller affects the usability of the contract by a lot (hint: not in a good way). If I were a user of this builtin actor, I'd be confused: how am I supposed to decode the bytes into the result of some paired crypto operation?
Same with encoding and decoding in other requests.
If dealing with `ArkScale<_>` struct prevents accurate gas estimation I'd sacrifice the precision in favour of usability.

gbuiltins/bls381/src/lib.rs
gshep: Yep, we are able to go that way however there are some points we have to take into consideration:
1. the `sp-crypto-ec-utils` API requires to get encoded `ArkScale<_>` as byte vectors. It means that even if we would require `ArkScale<_>` in the Request enum so the client code will not encode the arguments we will have to encode them by ourselves. Moreover we would have to bench such encoding operation to properly charge the gas;
2. the other argument is that we also would have to bench SCALE-decoding of `ArkScale<_>` instance for the same purpose of gas counting because we need to know `G1/2`-items count to charge the gas accordingly;
3. another one point is that there is a bug in the ark-works libs - https://github.com/arkworks-rs/algebra/issues/707 so we have to check items count on equality.

So I decided to decode vectors by hand to mitigate all the mentioned issues. Should be quite optimal )

gbuiltins/bls381/src/lib.rs
gshep: The answer is the same as for the other remark: it is a consequence of the `sp-crypto-ec-utils` API.

ekovalev:
Would be good to have a simple calling contract in examples to see how it works in practice (specifically, encoding/decoding of the `ArcScale` struct in user program)

gbuiltins/bls381/src/lib.rs
ekovalev: Still, the purpose of this crate is to abstract away encoding details from users. Otherwise one might argue that gear programs API requires the `handle` payload as a vector so let the users build the payload as they like - why bother creating any helper crates like this one then?

gshep: Sorry, but cannot agree. The purpose of the builtin is just to re-export `sp-crypto-ec-utils` API to the gear-programs and that API makes some constraints. Yep, we are able to make API less noisy but that would impact gas consumption.

gshep:
> Would be good to have a simple calling contract in examples to see how it works in practice (specifically, encoding/decoding of the `ArcScale` struct in user program)

Yep, there is already on but  as a standalone project - https://github.com/LouiseMedova/bn254_verification/tree/builtin-bls381 .  If you don't mind I would like to implement it in another PR as a gclient test in our repo.

gbuiltins/bls381/src/lib.rs
ekovalev: > Sorry, but cannot agree. The purpose of the builtin is just to re-export `sp-crypto-ec-utils` API to the gear-programs and that API makes some constraints. Yep, we are able to make API less noisy but that would impact gas consumption.

How would this impact gas consumption if the same set of operations is performed in the end? The user who engages the calling contract will pay for all the gas regardless where exactly the encoding is being done.

But the way it is now is more error-prone (except being less usable): if the calling contract somehow screws up correctly encoding the arguments into bytes it would still make a valid `MultiMillerLoop` request. If we take care of encoding it'd be impossible. Not to mention API versioning

gbuiltins/bls381/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+
+[dependencies]
+codec = { workspace = true, features = ["derive"] }
+derive_more.workspace = true
+
+[features]
+default = ["std"]
```
ekovalev, line 14: Is it necessary to define default feature for this crate?

gbuiltins/bls381/src/lib.rs
breathx: Agree with @gshep
Anyway we're passing everything as bytes. Program -> builtin -> host -> builtin -> program

gbuiltins/bls381/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+pub const REQUEST_PROJECTIVE_MULTIPLICATION_G2: u8 = 5;
+
+/// Type that should be used to create a message to the bls12_381 builtin actor.
+#[derive(Encode, Clone, PartialEq, Eq, Debug)]
+#[codec(crate = codec)]
+pub enum Request {
+    /// Request to pairing multi Miller loop for *BLS12-381*.
+    ///
+    /// Encoded:
+    ///   - `a`: `ArkScale<Vec<G1Affine>>`.
```
breathx, line 49: And reference crate inside hyperlink. Maybe consider types re-import
```suggestion
    ///   - `a`: [`ArkScale<Vec<G1Affine>>`].
```

gbuiltins/bls381/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+/// Constant defines codec index of [`Request::FinalExponentiation`].
+pub const REQUEST_FINAL_EXPONENTIATION: u8 = 1;
+/// Constant defines codec index of [`Request::MultiScalarMultiplicationG1`].
+pub const REQUEST_MULTI_SCALAR_MULTIPLICATION_G1: u8 = 2;
+/// Constant defines codec index of [`Request::MultiScalarMultiplicationG2`].
+pub const REQUEST_MULTI_SCALAR_MULTIPLICATION_G2: u8 = 3;
+/// Constant defines codec index of [`Request::ProjectiveMultiplicationG1`].
+pub const REQUEST_PROJECTIVE_MULTIPLICATION_G1: u8 = 4;
+/// Constant defines codec index of [`Request::ProjectiveMultiplicationG2`].
+pub const REQUEST_PROJECTIVE_MULTIPLICATION_G2: u8 = 5;
```
breathx, line 40: To consider an option 
```suggestion
pub const PROJECTIVE_MULTIPLICATION_G2_DESCRIMINANT: u8 = 5;
```

gbuiltins/bls381/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+    DecodeItemCount,
+}
+
+/// The enumeration contains result to a request.
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq, derive_more::From)]
+#[codec(crate = codec)]
+pub enum Response {
+    /// Common error.
+    #[from]
+    Error(Error),
```
breathx, line 111: error shouldn't be in Response
builtin framework supports -> Result<YOUR_OK_BYTES, Error> where Error includes your custom variant

gbuiltins/bls381/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+    ///   - `base`: `ArkScaleProjective<G2Projective>`.
+    ///   - `scalar`: `ArkScale<Vec<u64>>`.
+    #[codec(index = 5)]
+    ProjectiveMultiplicationG2 { base: Vec<u8>, scalar: Vec<u8> },
+}
+
+/// The enumeration represents possible common errors for all requests.
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
+#[codec(crate = codec)]
+pub enum Error {
```
breathx, line 96: all of these errors seems like decoding error so I'd merge them and used default DecodeError type of error from builtin framework

breathx: caller doesnt care of which type of encoding failed - it will just fail because it relies on its "valid" data

gbuiltins/bls381/src/lib.rs
breathx: Supporting @ekovalev care about type safety and @gshep idea of optimised passing by different instances, I'd like to suggest you implement one more type that will be used inside programs for exact Decoding

breathx: Same for encoding mb?

breathx: And these will have "Raw" postfix

gbuiltins/bls381/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+}
+
+/// Result of the multi scalar multiplication.
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
+#[codec(crate = codec)]
+pub enum MultiScalarMultiplicationResult {
+    // Encoded: `ArkScaleProjective<ark_bls12_381::G(1,2)Projective>`.
+    Ok(Vec<u8>),
+    /// Computation error.
+    Error,
```
breathx, line 152: no need in these extra errors. I'd also consider their concretisation as with decoding errors from above

gbuiltins/bls381/src/lib.rs
```diff
@@ -0,0 +1,240 @@
...
+mod tests {
+    extern crate alloc;
+
+    use super::*;
+    use alloc::vec;
+
+    // The standard Decode implementation cannot be used for precise gas charging.
+    // The following test checks that scale codec indexes of variants are set correctly.
+    #[test]
+    fn codec_enum_indexes() {
```
breathx, line 176: whoa, please, impl closure to be used)) for each case
and you need to write a test or extend this about these constants matching some specific number, as for backward compatibility and modifying by accident

pallets/gear-builtin/src/bls12_381.rs
```diff
@@ -0,0 +1,333 @@
+// This file is part of Gear.
```
breathx, line 1: NOTE: This file will be reviewed once other comments addressed

pallets/gear-builtin/src/benchmarking.rs
```diff
@@ -21,8 +21,18 @@
 #[allow(unused)]
```
breathx, line 21: NOTE: This file will be reviewed once other comments addressed

breathx (changes requested):

breathx:
Forget to mention in the PR review, at least MINIMAL demo + test are necessary within the PR

gbuiltins/bls381/src/lib.rs
gshep: discussed: use BuiltinError::Custom with the error description

gbuiltins/bls381/src/lib.rs
gshep: discussed: use current implementation

pallets/gear-builtin/Cargo.toml
```diff
@@ -19,13 +19,22 @@ primitive-types = { workspace = true, features = ["scale-info"] }
 log.workspace = true
 derive_more.workspace = true
 impl-trait-for-tuples.workspace = true
+ark-std = { workspace = true, optional = true }
```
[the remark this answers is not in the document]
gshep: optional deps are required for benchmarks. In the same time we need ark-scale/serialize to implement the actor

ekovalev (approved):

pallets/gear-builtin/src/benchmarking.rs
```diff
@@ -101,6 +122,191 @@ benchmarks! {
...
+        let encoded_scalars = ark_scalars.encode();
+
+        let bases = (0..count).map(|_| G2::rand(&mut rng)).collect::<Vec<_>>();
+        let bases = G2::batch_convert_to_mul_base(&bases);
+        let ark_bases: ArkScale<Vec<G2Affine>> = bases.clone().into();
+        let encoded_bases = ark_bases.encode();
+
+        let mut _result: Result<Vec<u8>, ()> = Err(());
+    }: {
+        _result = sp_crypto_ec_utils::bls12_381::host_calls::bls12_381_msm_g2(encoded_bases, encoded_scalars);
```
breathx, line 255: isn't it much more productive to bench exactly call including decoding instead of just host calls?

pallets/gear-builtin/src/benchmarking.rs
```diff
@@ -112,6 +134,199 @@ benchmarks! {
...
+
+    bls12_381_msm_g1 {
+        let c in 0 .. 1_000;
+
+        let count = c as usize;
+
+        let mut rng = ark_std::test_rng();
+
+        let scalars = (0..count)
+            .map(|_| <G1 as Group>::ScalarField::rand(&mut rng))
```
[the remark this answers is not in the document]
breathx: any feedback here @gshep ?

pallets/gear-builtin/src/tests/bls381.rs
```diff
@@ -0,0 +1,695 @@
...
+        pallet_gear::manager::HandleKind::Handle(builtin_id),
+        payload,
+        0,
+        true,
+        None,
+        None,
+    )
+    .expect("calculate_gas_info failed");
+    rollback_transaction();
+
```
breathx, line 64: could you pls add here assertion that res.min_gas and res.burned != 0 && != block gas limit (or < block gas limit * 90%)

pallets/gear-builtin/src/bls12_381.rs
```diff
@@ -0,0 +1,378 @@
...
+    mut payload: &[u8],
+    gas_limit: u64,
+) -> (Result<Response, BuiltinActorError>, u64) {
+    let (gas_spent, result) = decode_vec::<T, _>(gas_limit, 0, &mut payload);
+    let a = match result {
+        Some(Ok(array)) => array,
+        Some(Err(e)) => return (Err(e), gas_spent),
+        None => return (Err(BuiltinActorError::InsufficientGas), gas_spent),
+    };
+
```
breathx, line 112: I guess such things could be simplified. To keep going open an issue of "beautification" of this module pls. Atm ok

breathx: and about errors: I'd still create some enum here that `: Into<LimitedStr>` to operate with types inside fns only

pallets/gear-builtin/src/bls12_381.rs
```diff
@@ -0,0 +1,378 @@
...
+        Ok(count_b) if count_b != count => {
+            return (
+                Err(BuiltinActorError::Custom(LimitedStr::from_small_str(
+                    "Multi Miller loop: non equal item count",
+                ))),
+                gas_spent,
+            )
+        }
+        Err(_) => {
+            log::debug!(
```
breathx, line 145: excess logs. I'd cover all branches that case

pallets/gear-builtin/src/bls12_381.rs
breathx: and caller to callee refactor flow

breathx (approved):
Overall looks good to me, ready to go

pallets/gear-builtin/src/bls12_381.rs
breathx: #3841

breathx:
Merge after #3836 and #3814 are merged

pallets/gear-builtin/src/benchmarking.rs
gshep: Nope, since encoded have different length depending on bls381 entities. It means it would be quite difficult to find such items so we will get encoded length 100, 101, 102, etc

gbuiltins/bls381/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+/// Use the following crates to construct a request:
+///  - `ark-scale`: <https://docs.rs/ark-scale/>;
+///  - `ark-bls12-381`: <https://docs.rs/ark-bls12-381/>.
+#[derive(Encode, Clone, PartialEq, Eq, Debug)]
+#[codec(crate = codec)]
+pub enum Request {
+    /// Request to pairing multi Miller loop for *BLS12-381*.
+    ///
+    /// Encoded:
+    ///   - `a`: `ArkScale<Vec<G1Affine>>`.
```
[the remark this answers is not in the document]
gshep: ok, will reexport corresponding libraries

pallets/gear-builtin/src/benchmarking.rs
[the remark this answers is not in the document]
gshep: changed as suggested

breathx (approved):
🔥

PR #3777, opened 2024-03-03
feat(pallet): Get rid of `without_storage_info`, use Bounded structs

### LINKED ISSUE #3727: Consider getting rid of `#[pallet::without_storage_info]`
### File Location(s)

pallets/*

### Proposal

According to the docs, using the `#[pallet::without_storage_info]` is not recommended.

> Because the `#[pallet::without_storage_info]` macro applies to all storage items in your pallet, you should only use it in a test or development environment. You should never use the `#[pallet::without_storage_info]` attribute macro in a production environment.

https://docs.substrate.io/reference/frame-macros/#palletwithout_storage_info

- We could mark `!MaxEncodedLen` data structures with `#[pallet::unbounded]`
- We could use limited data structures
- We can borrow something from a similar issue: https://github.com/paritytech/polkadot-sdk/issues/323

ukint-vs:
Resolves #3727  .

- 
-
-

@reviewer-or-team


breathx:
Needs more investigation. Temporary closing PR until renewed

PR #3778, opened 2024-03-03
fix(tools): Generate .binpath only on primary artifact

NikVolf:


utils/wasm-builder/src/wasm_project.rs
```diff
@@ -456,6 +460,7 @@ extern "C" fn metahash() {{
         fs::create_dir_all(&self.target_dir).context("Failed to create target directory")?;
 
         let mut wasm_files = vec![(original_wasm_path.clone(), file_base_name.clone())];
+        self.generate_bin_path(file_base_name)?;
```
StackOverflowExcept1on, line 463: for better readability since the `wasm_files` vec is created in loop
```suggestion
        self.generate_bin_path(file_base_name)?;

        let mut wasm_files = vec![(original_wasm_path.clone(), file_base_name.clone())];
```

StackOverflowExcept1on (approved):

PR #3781, opened 2024-03-05
refactor(node): Hold the instance of BuiltinDispatcher in ExtManager

gshep:
While working on #3513 found that BuiltinCache is not useful and also triggers additional access to DB. Another case is that we need to clean it from time to time to keep the list of builtins actual: every block creation or on runtime upgrade but it is easily to forget.

- remove BuiltinCache, adjust to the changes
- refactor `is_active`, `exists` methods to take a reference to `BuiltinDispatcher`
- hold the instance of `BuiltinDispatcher` in `ExtManager`

@gear-tech/dev


grishasobol (commented):
Actually I think is only one db read per block, because after that it will be gotten from overlay (which also save reads).
Could you please show the place where builtin cache was set before your changes?

pallets/gear/src/lib.rs
```diff
@@ -822,8 +820,8 @@ pub mod pallet {
         }
 
         /// Returns true if `program_id` is that of a in active status or the builtin actor.
-        pub fn is_active(program_id: ProgramId) -> bool {
-            T::BuiltinCache::exists(&program_id)
```
gshep, line 826: @grishasobol here and in `program_exists` too

gshep:
> Actually I think is only one db read per block, because after that it will be gotten from overlay (which also save reads). Could you please show the place where builtin cache was set before your changes?

You're absolutely correct however to build the cache we need to spend some weight/gas and it doesn't get counted

pallets/gear/src/lib.rs
grishasobol: A ok see now lazy initialisation. Was it a bug ? Because in case of builtins list update this cache will stay not updated.

Actually I don't like refactoring your have made)) but agree with you that this cache must be updated each block maybe in on_initialize()

breathx:
> > Actually I think is only one db read per block, because after that it will be gotten from overlay (which also save reads). Could you please show the place where builtin cache was set before your changes?
> 
> You're absolutely correct however to build the cache we need to spend some weight/gas and it doesn't get counted

Nobody cares about one-two extra read-writes in on_initialize sections, they have their dedicated extra time on stage of balancing block weight

breathx (commented):
I'll keep going with the point that its cool to have special storage we gonna operate in RAM only by using creation in on_initialize hook and deletion in on_finalize, because it allows us to call this set at any time and any place from code.

So I'd focus on such functionality. Would be cool to see others opinion. If mostly agreed on implementing it this way - I'll go through changes itself

ekovalev:
I don't mind this change. This `BuiltinCache` was introduced with the only purpose to do a quick look-up inside the `program_exists` and `is_active` routines without having to create an instance of the `BuiltinDispatcher` at every call. Since these functions can now do that in a more natural way I'm totally up for removing the cache as a redundant concept.

ekovalev (approved):

ukint-vs (approved):

PR #3790, opened 2024-03-06
feat(rpc-call): Implement new call to send message and collect reply from on-chain state

breathx:
### Working in progress.

TODOs:
- [x] Add pallet gear unit tests
- [x] Rename new functions
- [x] Remove duplicated code from calculate gas info implementation in pallet gear


breathx:
@osipov-mit kindly take a look and confirm if it works as expected

### Current API
```rust
fn gear_readOnlySendMessage(
    at: H256, // optional block hash at execution required
    origin: H256,
    destination: H256,
    payload: Vec<u8>,
    gas_limit: u64,
    value: u128,
) -> RpcResult<ReadOnlyReply>;
```

### Resulting struct
```rust
/// The struct contains results of read only send message RPC call.
pub struct ReadOnlyReply {
    /// Payload of the reply.
    pub payload: Vec<u8>,
    /// Value sent with reply.
    pub value: u128,
    /// Reply code of the reply.
    pub code: ReplyCode,
}
```

### `enum ReplyCode`
https://github.com/gear-tech/gear/blob/f9e03aeb253a6feb8ffd18030381df6aa2bf743f/core-errors/src/simple.rs#L44

### Curl usage over uploaded and initialised dapps/ping example:
telegram-cloud-photo-size-2-5425130417901656101-y

pallets/gear/src/runtime_api.rs
```diff
@@ -40,11 +42,204 @@ impl<T: Config> Pallet<T>
...
+            message.to_string()
+        )
+    }
+
+    // Returns none if queue is empty, otherwise - associated journal and bool,
+    // defining was it processed by builtin actor or not.
+    fn dequeue_head_and_run(
+        ext_manager: &mut ExtManager<T>,
+        builtin_dispatcher: &impl BuiltinDispatcher,
+    ) -> Result<Option<(Vec<JournalNote>, bool)>, String> {
```
grishasobol, line 127: What the reason to return this bool ?

core/src/message/mod.rs
```diff
@@ -230,3 +230,15 @@ pub trait Packet {
     /// Packet value.
     fn value(&self) -> Value;
 }
+
+/// The struct contains results of read only send message RPC call.
+#[derive(Clone, Debug, Decode, Encode, PartialEq, Eq, TypeInfo)]
+#[cfg_attr(feature = "std", derive(serde::Serialize, serde::Deserialize))]
+pub struct ReadOnlyReply {
```
grishasobol, line 237: Why do we need it in core if you use it only in pallet. Even more - reimport it in pallet.

pallets/gear/rpc/runtime-api/src/lib.rs
```diff
@@ -20,14 +20,17 @@
...
 use sp_core::H256;
 use sp_runtime::traits::Block as BlockT;
 use sp_std::vec::Vec;
 
 sp_api::decl_runtime_apis! {
-    #[api_version(2)]
+    #[api_version(3)]
     pub trait GearApi {
+        #[allow(clippy::too_many_arguments)]
+        fn read_only_send_message(origin: H256, destination: H256, payload: Vec<u8>, gas_limit: u64, value: u128, allowance_multiplier: u64) -> Result<ReadOnlyReply, Vec<u8>>;
```
grishasobol, line 32: to consider:  I think better name is `send_message_read_only`, because we have already `send_message` and this is the same action (but rpc) with some specific - `read_only`.

pallets/gear/src/runtime_api.rs
breathx: I'll reuse this code in calculate gas impl, because its 99% similar, but calculate gas needs to know who was processing it

core/src/message/mod.rs
breathx: Same as GasInfo. It's introduced in core to be shared between: pallets, runtime-apis and generated clients e.g. gsdk etc, where pallet will be excess

pallets/gear/rpc/runtime-api/src/lib.rs
breathx: Yeah, good suggestion, but actually I dont like this name at all, but couldn't imagine any better xd

breathx:
send_noncommitable_message  (?)

pallets/gear/rpc/runtime-api/src/lib.rs
```diff
@@ -20,14 +20,17 @@
...
 
-pub use pallet_gear::{manager::HandleKind, GasInfo};
+pub use pallet_gear::{manager::HandleKind, GasInfo, ReadOnlyReply};
 use sp_core::H256;
 use sp_runtime::traits::Block as BlockT;
 use sp_std::vec::Vec;
 
 sp_api::decl_runtime_apis! {
-    #[api_version(2)]
+    #[api_version(3)]
```
breathx, line 29: revert

pallets/gear/src/runtime_api.rs
```diff
@@ -40,81 +42,134 @@ impl<T: Config> Pallet<T>
...
+                    {
+                        if replied_to == message_id {
+                            return Ok(ReadOnlyReply {
+                                payload: dispatch.payload_bytes().to_vec(),
+                                value: dispatch.value(),
+                                code,
+                            });
+                        }
+                    }
+                }
```
gshep, line 110: insubstantial but for readability
```suggestion
                let JournalNote::SendDispatch { dispatch, .. } = note else {
                    continue;
                };
                
                let Some((replied_to, code)) =
                    dispatch.reply_details().map(|d| d.into_parts()) else
                {
                    continue;
                };
                
                if replied_to == message_id {
                    return Ok(ReadOnlyReply {
                        payload: dispatch.payload_bytes().to_vec(),
                        value: dispatch.value(),
                        code,
                    });
                }
```

gshep (approved):
Great job! :fire:

grishasobol (approved):

PR #3796, opened 2024-03-07
fix(gtest): Replace opinionated submit_code with non-opinionated

NikVolf:


gtest/src/system.rs
```diff
@@ -230,13 +230,22 @@ impl System {
     /// stores the code in storage.
     #[track_caller]
     pub fn submit_code<P: AsRef<Path>>(&self, code_path: P) -> CodeId {
+        let code = fs::read(&path).unwrap_or_else(|_| panic!("Failed to read file {:?}", path));
+        self.0.borrow_mut().store_new_code(&code)
```
StackOverflowExcept1on, line 234:
```suggestion
        let code = fs::read(&code_path)
            .unwrap_or_else(|_| panic!("Failed to read file {:?}", code_path.as_ref()));
        self.0.borrow_mut().store_new_code(&code)
```

PR #3800, opened 2024-03-08
chore(examples): introduce program creation test in async-tester

### LINKED ISSUE #3797: Extend async messages sending tests
Add tests with async programs creation

clearloop:
Resolves #3797 

met no problem in this test, everything works

for cases that never continue execution after the lock, imo the reason could be

- invalid arguments
- wrong code id or the target program failed to reply the the creator program

- [x] extend program create in async-tester
- [x] add test to pallet-gear
- [x] pass the test

@gear-tech/dev

### DIFF
--- a/examples/async-tester/src/lib.rs
+++ b/examples/async-tester/src/lib.rs
@@ -22,4 +22,5 @@ pub enum Kind {
     SendCommit,
     SendCommitWithGas(u64),
     CreateProgram(CodeId),
+    CreateProgramWithGas(CodeId, u64),
 }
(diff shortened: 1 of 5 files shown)


breathx:
Could you pls as well intro gas full and gasless scenarios

breathx (changes requested):
Due to comments above.

And remove draft mark from pr if it's ready for review. Having tests that doesnt fail is okay, original issue could hide in gtest or any other env

breathx (approved):

PR #3803, opened 2024-03-09
refactor(tests): Revamp gear-authorship tests

ekovalev:
**A problem**: gear-authorship test run against the actual Vara runtime; each test creates an instance of the client, which, in turn, creates the runtime executor. The latter would populate its runtime cache with two versioned Vara wasm runtimes. This entails a lot of unnecessary allocations that run in parallel, thereby causing tremendous slow-down.

**The proposed solution**: a static global structure has been introduced that creates the executor that is being reused by all tests, so that it only allocates memory for runtimes code once.

This alone, however, would only work well with the default `cargo test` runner. On CI we use the `cargo nextest` that runs the test binary for each test in a separate process thereby rendering the shared global state useless.
To mitigate that we keep the only `#[test]` macro in the `tests` module and explicitly spin a thread for each test case within the same process.

**Update**: it's tempting to use higher level parallelism primitives like those from `rayon` crate for the sake of the code being more concise and elegant. Unfortunately `rayon`'s `par_iter()` gets stuck half way so it will take a while to dig deeper to see what's wrong. There's an issue so we don't forget to get back to it some time in the future: #3809.


ukint-vs (approved):

Makefile
```diff
@@ -194,7 +194,6 @@ test-gear: # Crates except gclient, gcli, gsdk are excluded to significantly dec
 		--exclude gclient \
 		--exclude gcli \
 		--exclude gsdk \
-		--exclude gear-authorship \
```
breathx, line 197: they're still pretty long and heavy for pre-commit, isn't it?

node/authorship/src/tests.rs
```diff
@@ -222,208 +241,172 @@ pub(crate) fn init_logger() {
...
+        Box::new(proposed_storage_changes_should_match_execute_block_storage_changes) as TestCase,
+        Box::new(should_not_remove_invalid_transactions_when_skipping) as TestCase,
+        Box::new(should_cease_building_block_when_block_limit_is_reached) as TestCase,
+        Box::new(should_keep_adding_transactions_after_exhausts_resources_before_soft_deadline)
+            as TestCase,
+        Box::new(should_only_skip_up_to_some_limit_after_soft_deadline) as TestCase,
+    ];
+
+    let handles: Vec<_> = tests
+        .into_iter()
```
breathx, line 355: you can simply use par_iter from rayon

node/authorship/src/tests.rs
```diff
@@ -794,14 +789,14 @@ fn block_builder_cloned_ok() {
     );
 }
 
-#[test]
+// Test case 6
```
breathx, line 792: I'd name the functions than like `fn test_blah_blah`

node/authorship/src/tests.rs
```diff
@@ -222,208 +241,172 @@ pub(crate) fn init_logger() {
...
+                test();
+            })
+        })
+        .collect();
+
+    let mut output = vec![];
+    for handle in handles {
+        output.push(handle.join());
+    }
+    for result in output {
```
breathx, line 367: + its better to define which one failed if failed

breathx (approved):

Makefile
ekovalev: On par with some prop tests. About 40-ish seconds

node/authorship/src/tests.rs
ekovalev: it prints the stack trace in the std err with a link to the exact location of the error. Actually, to all of them if there is more than one error.
That's the reason for collecting all the outputs.
Is it what you are talking about?

node/authorship/src/tests.rs
ekovalev: I didn't do that just in case we want (for debug purpose, for example), run each test individually with `cargo test` again. Such renaming would be an overkill.
But if we don't then yes, maybe it'd make sense to have uniform naming like `test_...`

breathx: I guess remove prefix from 10fns isn't that hard if one day we return to `#[test]`

Makefile
ekovalev: If it's not too important for pre-commit then maybe we can keep excluding it. After all, changes in block production do not come that often comparing to runtime

node/authorship/src/tests.rs
ekovalev: Have tried it. Somehow rayon stumbles upon some data race, apparently - it's not even able to complete the test. Even with half of the test cases it is already too slow.
The std::thread, on the other hand, is consistent and almost doesn't depend on how many tests are run in parallel.

Will look into it again though because it's really tempting - the code is more elegant that way due to higher level parallelisation primitives

PR #3808, opened 2024-03-12
feat(gtest): use real weights

### LINKED ISSUE #3799: Use real weights in gtest
### Problem to Solve

Currently it uses constant const rules for instrumenting

### Possible Solution

Consider including the file from pallet weights 

### Notes

_No response_

StackOverflowExcept1on:
Resolves #3799

`utils/wasm-instrument/src/gas_metering/schedule.rs` is auto-generated file that can be compiled without depending on substrate


breathx (approved):
Lgtm as current implementation. But it would be great to open an issue (maybe even suggest a design) how could we remove all of these diffrerent files/weights/weightinfos/schedules and reuse there, lets say, two modules everywhere in code: test libs, pallets etc

StackOverflowExcept1on:
@breathx it's possible, but the problem is that we need to depend on frame-support with ~200 substrate dependencies.

breathx:
> @breathx it's possible, but the problem is that we need to depend on frame-support with ~200 substrate dependencies.

Sure, Im about any solution that wouldn't involve substrate deps there

utils/wasm-instrument/src/gas_metering/schedule.rs
```diff
@@ -0,0 +1,713 @@
+// This file is part of Gear.
```
gshep, line 1: It would be better to notify in the header that the file is autogenerated.

gshep (approved):
LGTM!

scripts/weight-dump.sh
```diff
@@ -0,0 +1,14 @@
...
+set -e
+
+dump_path="weight-dumps"
+mkdir -p "$dump_path"
+
+current_branch=$(git branch --show-current)
+
+dump_path1="$dump_path/${current_branch//\//-}.json"
+cargo run --package gear-weight-diff --release -- dump "$dump_path1" --label "$current_branch"
+cargo run --quiet --package gear-weight-diff --release codegen "$dump_path1" vara > utils/wasm-instrument/src/gas_metering/schedule_tmp.rs
```
ark0f, line 12:
```suggestion
cargo run --quiet --package gear-weight-diff --release -- codegen "$dump_path1" vara > utils/wasm-instrument/src/gas_metering/schedule_tmp.rs
```

.github/workflows/benchmarks.yml
```diff
@@ -44,6 +44,8 @@ jobs:
           # apply some patches for `pallets/gear/src/weights.rs`
           cp runtime/vara/src/weights/pallet_gear.rs pallets/gear/src/weights.rs
           sed -i -E 's/\w+::WeightInfo for SubstrateWeight/WeightInfo for SubstrateWeight/' pallets/gear/src/weights.rs
+          # generate code for lightweight scheduler that is used in gtest and other crates
+          ./scripts/weight-dump.sh
```
ark0f, line 48: I think this should be in `check.yml` because nobody will actually remember to manually update lightweight scheduler

utils/weight-diff/src/main.rs
```diff
@@ -266,5 +340,102 @@ fn main() {
...
+                serde_json::from_str(&fs::read_to_string(path).unwrap()).unwrap();
+            let raw_schedule = match runtime {
+                Runtime::Vara => serde_json::to_value(dump.vara_schedule).unwrap(),
+            };
+
+            let file =
+                syn::parse_file(&fs::read_to_string("pallets/gear/src/schedule.rs").unwrap())
+                    .unwrap();
+
+            let mut visitor = StructuresVisitor::default();
```
ark0f, line 354: Could you please clarify the purpose of visitor if all structure names are `DeserializableSchedul`'s fields

ark0f (approved):

.github/workflows/benchmarks.yml
StackOverflowExcept1on: If you want to update weights for your branch this should be done via github actions workflow.

utils/weight-diff/src/main.rs
StackOverflowExcept1on: Because we would like to keep the original data types for structure fields.

ukint-vs (approved):

PR #3812, opened 2024-03-14
refactor(pallet-gear): Remove waiting init list

### LINKED ISSUE #3705: Fix waiting init message is not appended to wait list
### Problem

There can be some cases when waiting init message is not going to be appended into wait list because of lock of gas.
https://github.com/gear-tech/gear/blob/9a37ccb406de68af8ddc6c851faa0c90ba0f0918/pallets/gear/src/internal.rs#L338-L348
This `log::error` can be called in case message have not enough gas to be added in the `wait list`. In most cases this is restricted by core-processor checks, but it does not protect cases when message will be added to `waiting init` list.
So, if some program or user will send error to uninitialised program with specific amount of gas, than message will be forgotten. So, it will be never consumed and gas node will stay forever in storage.

### Steps

You can try this test in pallet-gear:
```rust
#[test]
fn calculate_gas_waiting_init() {
    let wat = r#"
        (module
            (import "env" "memory" (memory 1))
            (import "env" "gr_wait_up_to" (func $wait (param i32)))
            (import "env" "gr_program_id" (func $program_id (param i32)))
            (import "env" "gr_send" (func $send (param i32 i32 i32 i32 i32)))
            (export "init" (func $init))
            (func $init
                (call $program_id (i32.const 0))
                (call $send (i32.const 0) (i32.const 0) (i32.const 1) (i32.const 0) (i32.const 1000))
                (call $wait (i32.const 1))
            )
        )"#;

    init_logger();
        let pid = Gear::upload_program(
            RuntimeOrigin::signed(USER_1),
            ProgramCodeKind::Custom(wat).to_bytes(),
            DEFAULT_SALT.to_vec(),
            EMPTY_PAYLOAD.to_vec(),
            210_105_500,
            0,
            false,
        )
        .map(|_| get_last_program_id())
        .expect("Program uploading failed");

        run_to_block(4, None);
    });
}
```
Gas limit amount `210_105_500` has been found, so that the case would happen.

### Possible Solution

1) Remove waiting init list and reply with error for all messages, that are sent to uninitialised program.
2) In case we cannot put message to the `wait list` then reply with an error.

ark0f:
resolves #3705

**Release notes:**

Waiting init list has been removed. Message sent to uninitialized program was stored in list and actually sent to the program when it is become initialized. Now messages sent to uninitialized (waited) program will be responded with `ErrorReplyReason::InactiveActor`.


pallets/gear/src/migrations.rs
```diff
@@ -0,0 +1,312 @@
...
+                        if let Some(dispatch) = Pallet::<T>::wake_dispatch(
+                            program_id,
+                            message_id,
+                            MessageWokenSystemReason::WaitInitListMigration.into_reason(),
+                        ) {
+                            // remove from waitlist
+                            weight = weight.saturating_add(T::DbWeight::get().writes(1));
+
+                            QueueOf::<T>::queue(dispatch).unwrap_or_else(|e| {
+                                unreachable!("Message queue corrupted! {:?}", e)
```
grishasobol, line 73: no panics in migrations

gshep:
:cry:

gshep (approved):
lgtm!

breathx:
to be reviewed asap

pallets/gear/src/migrations.rs
```diff
@@ -0,0 +1,311 @@
...
+        let mut weight = T::DbWeight::get().reads(1);
+
+        if current == 4 && onchain == 3 {
+            waiting_init_list::WaitingInitStorage::<T>::translate(
+                |program_id, messages: Vec<MessageId>| {
+                    // read and remove an element
+                    weight = weight.saturating_add(T::DbWeight::get().reads_writes(1, 1));
+
+                    for message_id in messages {
+                        if let Some(dispatch) = Pallet::<T>::wake_dispatch(
```
grishasobol, line 64: I think we should remove old tasks on migration also, because this can lead to panics in runtime: 
https://github.com/gear-tech/gear/blob/6316fc9fd1f8e52c3f383d93b16a45bd2415811b/pallets/gear/src/manager/task.rs#L112-L120

grishasobol (changes requested):

common/src/event.rs
```diff
@@ -162,6 +158,8 @@ pub enum MessageWokenSystemReason {
     TimeoutHasCome,
     /// Message can no longer pay rent for holding in storage (see #646).
     OutOfRent,
+    /// Waiting init list has been removed, so we just wake all the messages from it.
```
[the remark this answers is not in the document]
breathx: According to comment above (on `enum MessageWaitedSystemReason`), please don't introduce any new dumb variants for temp purposes.

You have variant `ProgramGotInitialized`, that used previously. We can use it with note that it's incorrect naming blah blah. And this variant needs deprecation as well, isn't it?

core-errors/src/simple.rs
```diff
@@ -181,9 +181,10 @@ pub enum ErrorReplyReason {
     #[display(fmt = "fail in program creation ({_0})")]
     FailedToCreateProgram(SimpleProgramCreationError) = 1,
 
-    /// Destination actor become inactive program and can't process the message.
-    #[display(fmt = "inactivity of destination program")]
-    InactiveProgram = 2,
+    /// Destination actor is in wait-in-init state
+    /// or become inactive program, so it can't process the message.
```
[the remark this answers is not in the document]
breathx: I'd separate these errors: Keep variant `InactiveProgram` only with renaming into `InactiveActor` and add a new one `UninitialisedProgram`

And address @grishasobol's comment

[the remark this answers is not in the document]
breathx: Processing wasn't halted in fact, it resulted into an error due to some reasons

core-processor/src/processing.rs
```diff
@@ -178,6 +178,7 @@ where
 }
 
 enum ProcessErrorCase {
+    WaitInit,
```
breathx, line 181: extend NonExecutable case with options: if its exited/inactive (due to rent, for example) or not yet initialised

core-processor/src/processing.rs
```diff
@@ -222,44 +223,6 @@ fn process_error(
         amount: gas_burned,
     });
 
-    // We check if value is greater than zero to don't provide
```
breathx, line 225: why was it moved?

core-processor/src/processing.rs
```diff
@@ -369,6 +373,25 @@ pub fn process_non_executable(
     )
 }
 
+/// Helper function for journal creation in case of waiting during init.
+pub fn process_wait_init(
```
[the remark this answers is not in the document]
breathx: agreed, but my proposal is to include it into `non_executable` case mb

examples/init-wait/src/wasm.rs
```diff
@@ -29,15 +29,6 @@ static mut STATE: State = State::NotInited;
 static mut INIT_MESSAGE: MessageId = MessageId::new([0; 32]);
 static mut TEST_DYNAMIC_MEMORY: BTreeMap<u32, ()> = BTreeMap::new();
 
-#[no_mangle]
```
breathx, line 32: grinding affected lines count?)

pallets/gear/src/queue.rs
```diff
@@ -85,17 +85,7 @@ where
...
-                JournalNote::GasBurned {
-                    message_id: dispatch.id(),
-                    amount: gas.burned(),
-                },
-                JournalNote::WaitingInitMessage {
-                    dispatch,
-                    destination: destination_id,
-                },
-            ];
+            return core_processor::process_wait_init(precharged_dispatch, destination_id);
```
breathx, line 88: its not wait init

breathx: its especially obvious in the context

pallets/gear/src/queue.rs
```diff
@@ -85,17 +85,7 @@ where
                 );
             }
 
-            let (dispatch, gas, _) = precharged_dispatch.into_parts();
-            return vec![
```
breathx, line 89: as I can guess, move of transferring value back inside core-processor was due to expectations for similar behaviour this place?
it's incorrect, message should be gone after this branch

pallets/gear/src/queue.rs
```diff
@@ -85,17 +85,7 @@ where
...
-                JournalNote::GasBurned {
-                    message_id: dispatch.id(),
-                    amount: gas.burned(),
```
breathx, line 92: very important stuff and space for small refactoring: please make sure that

Kind::Init is impossible to be sent in uninitialised program - there is a panic
Kind::Handle is acceptable for this error-returning case
Kind::Reply is acceptable for run
Kind::HandleSignal is impossible - paste panic as well

breathx: and speaking of those: it should be impossible to call system_reservation within the init, correct?
looks like this comment should be used as issue template

runtime/vara/src/migrations.rs
```diff
@@ -23,6 +23,8 @@ pub type Migrations = (
     /* release v1.2.0 */
     // migration for stored contexts
     pallet_gear_messenger::migrations::MigrateToV3<Runtime>,
+    // migration for removed waiting init list
+    pallet_gear::migrations::MigrateWaitingInitList<Runtime>,
     // check for existence of the rent pool account
     pallet_gear_staking_rewards::migrations::CheckRentPoolId<Runtime>,
```
breathx, line 29:
```suggestion
    // check for existence of the rent pool account
    pallet_gear_staking_rewards::migrations::CheckRentPoolId<Runtime>,
    /* not yet executed */
    // migration for removed waiting init list
    pallet_gear::migrations::MigrateWaitingInitList<Runtime>,
```

breathx: all migrations should be pushed back

pallets/gear/src/tests.rs
```diff
@@ -15856,25 +15890,38 @@ mod utils {
 
     #[track_caller]
     pub(super) fn assert_responses_to_user(user_id: AccountId, assertions: Vec<Assertion>) {
```
breathx, line 15892: 21 lines into 34 lines, no idea why to apply this refactor. readability for me just the same.

pallets/gear/src/lib.rs
```diff
@@ -41,6 +41,7 @@ mod mock;
 #[cfg(test)]
 mod tests;
 
+pub mod migrations;
```
breathx, line 44: why not impl this migration directly in pallet-gear-program crate?

pallets/gear/src/migrations.rs
breathx: Totally agree. You have the key for pallet-scheduler from waitlist - it keeps deadline in there next to message itself.

pallets/gear/src/migrations.rs
```diff
@@ -0,0 +1,311 @@
...
+
+            let mut compared = 0;
+            for (destination, dispatches) in dispatches.clone() {
+                for dispatch in dispatches {
+                    let LinkedNode {
+                        value: queued_dispatch,
+                        ..
+                    } = Dispatches::<Test>::get(dispatch.id()).unwrap();
+                    assert_eq!(queued_dispatch, dispatch);
+                    assert_eq!(queued_dispatch.destination(), destination);
```
breathx, line 303: for test purposes you may run queue here to assert errors in logs, if we've chosen pallet gear for migration place

pallets/gear/src/migrations.rs
```diff
@@ -0,0 +1,311 @@
+// This file is part of Gear.
```
breathx, line 1: This file to be reviewed deeply once all other comments addressed.

breathx (changes requested):

common/src/event.rs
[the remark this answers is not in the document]
ark0f: > dumb

Rude

> According to comment above (on enum MessageWaitedSystemReason), please don't introduce any new dumb variants for temp purposes.

I don't see any comments you are talking about

> You have variant ProgramGotInitialized, that used previously. We can use it with note that it's incorrect naming blah blah.

Variant is still used. Why don't introduce new one? No breaking changes because of re-fetched metadata

core-errors/src/simple.rs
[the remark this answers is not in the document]
ark0f: Reason is either uninitialized or non-executable (only inactive program) so I choose `InactiveProgram` and `UninitializedProgram`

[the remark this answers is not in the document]
ark0f: `UninitalizedProgram` sounds like subset of `InactiveActor`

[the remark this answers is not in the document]
breathx: > `UninitalizedProgram` sounds like subset of `InactiveActor`

Uninitialised actor within the protocol is active actor, that could be used as new messages destination, but it hasn't yet passed initialisation process, while inactive - temporary or permanently unavailable even for messages sending through extrinsics

pallets/gear/src/migrations.rs
ark0f: `wake_dispatch` automatically removes task

common/src/event.rs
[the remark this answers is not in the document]
ark0f: We need formalize allowance of breaking changes for types which structure can be obtained via metadata. Looks like `Event` is sensitive

core-errors/src/simple.rs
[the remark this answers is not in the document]
ark0f: > while inactive - __temporary__ unavailable

`UninitializedProgram`  is temporal, too. It looks like mix of semantics of each variant

common/src/event.rs
```diff
@@ -130,11 +130,7 @@ impl From<MessageWaitedType> for MessageWaitedRuntimeReason {
 #[derive(Debug, Encode, Decode, Clone, PartialEq, Eq, TypeInfo, SystemReason)]
 #[codec(crate = codec)]
 #[scale_info(crate = scale_info)]
-pub enum MessageWaitedSystemReason {
-    /// Program hadn't finished initialization and can not
-    /// process received message yet.
-    ProgramIsNotInitialized,
-}
+pub enum MessageWaitedSystemReason {}
```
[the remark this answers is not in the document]
ark0f: > I'd keep this variant, but marked deprecated.

I'd not even mark it. I'd leave it as is for the future

core-processor/src/processing.rs
ark0f: New error case makes early return so it doesn't need these notes

pallets/gear/src/lib.rs
ark0f: I need access to queue and `pallet-gear` has `pallet-gear-program` as dependency so `pallet-gear-program` can't' depend on `pallet-gear` because it would be recursively dependent

core-processor/src/processing.rs
ark0f: What for?

examples/init-wait/src/wasm.rs
ark0f: I've broken my brain because of non-standard function position

pallets/gear/src/tests.rs
ark0f:
```rust
if messages.len() != assertions.len() {
```

Before these lines if assertions amount is less than amount of events then you get `out of bounds` panic and no info about already converted events

runtime/vara/src/migrations.rs
ark0f: Wait. Were previous migrations executed?

pallets/gear/src/migrations.rs
grishasobol: Yep, see now. Could you please append the test for that

pallets/gear/src/queue.rs
ark0f: > it should be impossible to call system_reservation within the init

It looks like signal is the only case for system reserve. But I don't see any reason to deny this behavior

ark0f: > Kind::Init is impossible to be sent in uninitialised program - there is a panic

`if` branches are introduced by @grishasobol

> Kind::Handle is acceptable for this error-returning case

`messages_to_uninitialized_program_wait`

> Kind::Reply is acceptable for run

`wake_messages_after_program_inited`

> Kind::HandleSignal is impossible - paste panic as well

Possible only pre-charging and actual execution

common/src/event.rs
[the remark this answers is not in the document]
breathx: Can't see any reason for now. Anytime it could be un-deprecated))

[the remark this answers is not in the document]
breathx: Just not to misuse nowadays

runtime/vara/src/migrations.rs
breathx: yes

pallets/gear/src/queue.rs
breathx: > > it should be impossible to call system_reservation within the init
> 
> It looks like signal is the only case for system reserve. But I don't see any reason to deny this behavior

because handle signal relies on memory and program data which is wasted on failing init. I'll open an issue to consider wiping program after signal and unsuccessful init if deposit exists

breathx: #3842 
I guess in case of opened issue its up to you whether to check it rn or not

runtime/vara/src/migrations.rs
grishasobol: So?

grishasobol (approved):
LGTM

pallets/gear/src/queue.rs
ark0f: > because handle signal relies on memory and program data which is wasted on failing init. 

I mean, when we were designing system reservation there was assumption it will have more use-cases. So that's why I don't want to forbid it

common/src/event.rs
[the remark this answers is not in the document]
ark0f: > Can't see any reason for now.

We have empty `SystemTerminationReason` for the same reasons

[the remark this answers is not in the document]
osipov-mit: I think it's okay to drop this variant now. We haven't tracked the MessageWaited event anywhere and I almost sure nobody has done it.

runtime/vara/src/migrations.rs
```diff
@@ -27,4 +27,7 @@ pub type Migrations = (
...
     /* unapplied migrations */
     pallet_gear_program::migration::AppendStackEndMigration<Runtime>,
+    /* not yet executed */
+    // migration for removed waiting init list
+    pallet_gear::migrations::MigrateWaitingInitList<Runtime>,
```
breathx, line 32: note: on merging master (should be done once again) please put it in the end

pallets/gear/src/migrations.rs
```diff
@@ -0,0 +1,372 @@
...
+            log::info!("Successfully migrated storage");
+        } else {
+            log::info!("❌ Migration did not execute. This probably should be removed");
+        }
+
+        weight
+    }
+
+    #[cfg(feature = "try-runtime")]
+    fn pre_upgrade() -> Result<Vec<u8>, TryRuntimeError> {
```
breathx, line 96: please make pre_upgrade and post_upgrade dependent on storage version
ref: https://github.com/gear-tech/gear/pull/3869/commits/6253cea6c1a57337f25dd1fa295e250444b0525d

breathx (changes requested):
LGTM

Last thing to figure out: finishing this work why don't we want to specify if actor inactive or uninitialised? Its theoretically different cases of error replies that may (or must) be handled in userspace differently

pallets/gear/src/migrations.rs
ark0f: I don't get why

common/src/event.rs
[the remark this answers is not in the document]
ark0f: > but this way isn't preferable

So should `Event` be treated as sensitive to changes or not?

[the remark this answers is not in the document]
breathx: yep, always

pallets/gear/src/migrations.rs
breathx: because latest versions of try-runtime cli we use to check runtime migrations has step of checking indempotency of the migrations and fail without this check

breathx (approved):

Tru2thisNu (approved):

PR #3822, opened 2024-03-19
chore(wasm-builder): Test features tracking

ark0f:
Also skip WASM compilation during Intellij project sync because Intellij itself cannot yet.

Actually this was attempt to get rid of constant build script re-run :/

Idea was to use `CARGO_FEATURE_*` environment variables listing (`cargo_metadata`) and tracking (`cargo:rerun-if-env-changed`) [but](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-env-changed):
> Note that the environment variables here are intended for global environment variables like CC and such, it is not possible to use this for environment variables like TARGET that [Cargo sets for build scripts](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts). The environment variables in use are those received by cargo invocations, not those received by the executable of the build script.

### DIFF
--- a/utils/wasm-builder/tests/smoke.rs
+++ b/utils/wasm-builder/tests/smoke.rs
@@ -146,7 +146,7 @@ fn features_tracking() {
 
     CargoRunner::new().args(["build", "--features=a"]).run();
     assert!(read_export_entry("handle_reply").is_some());
-
+    assert!(read_export_entry("handle_signal").is_none());
     CargoRunner::new().args(["build", "--features=b"]).run();
     assert!(read_export_entry("handle_signal").is_some());
 }
--- a/utils/wasm-builder/tests/smoke.rs
+++ b/utils/wasm-builder/tests/smoke.rs
@@ -121,3 +121,32 @@ fn no_infinite_build() {
 
     CargoRunner::new().args(["build"]).run();
 }
+
+#[ignore]
+#[test]
+fn features_tracking() {
+    #[track_caller]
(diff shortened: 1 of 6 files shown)


breathx (approved):

utils/wasm-builder/tests/smoke.rs
```diff
@@ -121,3 +121,32 @@ fn no_infinite_build() {
...
+        .unwrap()
+        .entries()
+        .iter()
+        .find(|entry| entry.field() == name)
+        .cloned()
+    }
+
+    CargoRunner::new().args(["build", "--features=a"]).run();
+    assert!(read_export_entry("handle_reply").is_some());
+
```
clearloop, line 149:
```suggestion
    assert!(read_export_entry("handle_reply").is_some());
    assert!(read_export_entry("handle_signal").is_none());
```

utils/wasm-builder/tests/smoke.rs
```diff
@@ -121,3 +121,32 @@ fn no_infinite_build() {
...
+    CargoRunner::new().args(["build", "--features=b"]).run();
+    assert!(read_export_entry("handle_signal").is_some());
```
clearloop, line 151:
```suggestion
    assert!(read_export_entry("handle_signal").is_some());
    assert!(read_export_entry("handle_reply").is_none());
```

clearloop (approved):

PR #3824, opened 2024-03-20
fix: adding required values to get bond proposal

kehiy:
the get method needs 2 values. a self and a proposal id (I think). we need to pass it on to them. currently, there is only an editor error.

@ukint-vs


breathx (changes requested):
It couldn't need 2 arguments since `<Runtime as pallet_treasury::Config>::ProposalBond` is a pallet constant inside `trait Config` of `pallet-treasury`:
```rust
#[pallet::constant]
type ProposalBond: Get<Permill>;
```

And declaration of `Get` trait goes below:
```rust
pub trait Get<T> {
    fn get() -> T;
}
```

breathx:
Closing as inapplicable

PR #3833, opened 2024-03-25
feat(gstd): Add ability to get payload of error replies in async messaging

### LINKED ISSUE #3798: Allow accessing payload of the error reply in gstd async
### Problem to Solve

Atm it only returns ReplyCode in case of error reply received. There should be an option to return payload as well 

### Possible Solution

Impl it!

### Notes

_No response_

breathx:
Resolves #3798


gstd/src/common/errors.rs
```diff
@@ -87,6 +115,33 @@ impl From<CoreError> for Error {
     }
 }
 
+/// Error type returned by gstd API while using invalid arguments.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum UsageError {
+    /// This error occurs when providing zero duration to waiting functions
+    /// (e.g. see `exactly` and `up_to` functions in
+    /// [CodecMessageFuture](crate::msg::CodecMessageFuture)).
```
StackOverflowExcept1on, line 123:
```suggestion
    /// [`CodecMessageFuture`](crate::msg::CodecMessageFuture)).
```

gstd/src/common/errors.rs
```diff
@@ -87,6 +115,33 @@ impl From<CoreError> for Error {
...
+    EmptyWaitDuration,
+    /// This error occurs when providing zero gas amount to system gas reserving
+    /// function (see
+    /// [Config::set_system_reserve](crate::Config::set_system_reserve)).
```
StackOverflowExcept1on, line 127:
```suggestion
    /// [`Config::set_system_reserve`](crate::Config::set_system_reserve)).
```

StackOverflowExcept1on (approved):

gstd/src/common/errors.rs
```diff
@@ -33,50 +34,77 @@ pub type Result<T, E = Error> = core::result::Result<T, E>;
...
+    /// Note: this error returns from `gstd` APIs in case of invalid arguments.
+    Gstd(UsageError),
+
+    /* Business logic errors */
+    /// Received error reply while awaited response from another actor.
+    ///
+    /// NOTE: this error could only be returned from async messaging.
+    ErrorReply(String, ErrorReplyReason),
+
+    // TODO: consider should we meant unsupported replies as success.
```
NikVolf, line 65: No TODOs without an issue

gstd/src/msg/async.rs
```diff
@@ -51,15 +53,24 @@ where
...
             // Remove lock after waking.
             async_runtime::locks().remove(msg_id, waiting_reply_to);
 
-            if !reply_code.is_success() {
-                return Poll::Ready(Err(Error::ReplyCode(reply_code)));
-            }
+            match reply_code {
+                ReplyCode::Success(_) => Poll::Ready(f(payload)),
+                ReplyCode::Error(reason) => {
+                    let err = String::from_utf8(payload)
```
NikVolf, line 63: at this level, you should not consider it a string

Better wrap it in different error elsewhere

For example, put payload it in wrapped struct and add  `fn as_string() -> Option<String>` for it

gstd/src/common/errors.rs
breathx: that's the thing I was about to discuss on review: compatibility here is depend on "status" codes we use. I assume, that error codes will change a lot in future, while success ones are kinda constant and unlikely to be changed, so I think it's incorrect to mean unsupported replies as success ones. Wanna to hear other opinions

gstd/src/msg/async.rs
breathx: But payloads of error replies are always strings atm 
https://github.com/gear-tech/gear/blob/5922c049a78ac14ce37f839fa30c0647080ceef9/core-processor/src/processing.rs#L264-L270

breathx: Or you just want to avoid extra calculations in case of somebody ignore err?

NikVolf: also this, but better not to rely that something is string

protocol itself does not dictate strings in the payload

gstd/src/common/errors.rs
breathx: @ark0f could you pls share your thoughts here ?

gstd/src/msg/async.rs
NikVolf: What the problem with debug? You can convert it in display / debug impl...

StackOverflowExcept1on (approved):

gstd/src/common/errors.rs
```diff
@@ -33,50 +34,77 @@ pub type Result<T, E = Error> = core::result::Result<T, E>;
...
-            Error::Decode(e) => write!(f, "Decoding codec bytes error: {e}"),
-            Error::ReplyCode(e) => write!(f, "Reply came with non success reply code {e:?}"),
-            Error::EmptyWaitDuration => write!(f, "Wait duration can not be zero."),
-            Error::ZeroSystemReservationAmount => {
-                write!(f, "System reservation amount can not be zero in config.")
+            Error::Decode(e) => write!(f, "Scale codec decoding error: {e}"),
+            Error::Gstd(e) => write!(f, "`Gstd` API error: {e:?}"),
+            Error::ErrorReply(err, reason) => write!(f, "Received reply '{err}' due to {reason:?}"),
+            Error::UnsupportedReply(payload) => {
+                write!(f, "Received unsupported reply 0x'{}'", hex::encode(payload))
```
StackOverflowExcept1on, line 103: to match with `impl Display for ErrorReplyPayload`
```suggestion
                write!(f, "Received unsupported reply '0x{}'", hex::encode(payload))
```

gstd/src/msg/async.rs
```diff
@@ -51,15 +52,17 @@ where
...
             async_runtime::locks().remove(msg_id, waiting_reply_to);
 
-            if !reply_code.is_success() {
-                return Poll::Ready(Err(Error::ReplyCode(reply_code)));
+            match reply_code {
+                ReplyCode::Success(_) => Poll::Ready(f(payload)),
+                ReplyCode::Error(reason) => {
+                    Poll::Ready(Err(Error::ErrorReply(payload.into(), reason)))
+                }
+                ReplyCode::Unsupported => Poll::Ready(Err(Error::UnsupportedReply(payload))),
```
ark0f, line 64: I actually still have no idea how to deal with this error in case of new reply code variant to not break older program. Maybe we need to discuss it with our superiors? I want to ask you to add clear module documentation about this behavior at least

PR #3837, opened 2024-03-25
feat(wasm-builder): Another approach to build demos

ark0f:
Proof Of Concept of another approach to build WASM demos like #3602

Just uses wrapper utility around `cargo` instead of complicated lock-files approach. Still requires build script in demo/pallet to obtain binary itself.

Check `demo-async-init`, `demo-ping` and `pallet-gear` to see how it may look


breathx:
Temporary closing in favour of opened referencing issue as won't be implemented

PR #3840, opened 2024-03-26
feat(gear-bank): impl on-finalize transfers; optimise block-production gas payouts

breathx:


NikVolf (approved):

pallets/gear/src/mock.rs
```diff
@@ -245,11 +245,13 @@ pub fn run_to_block_maybe_with_queue(
 ) {
     while System::block_number() < n {
         System::on_finalize(System::block_number());
+        GearBank::on_finalize(System::block_number());
```
NikVolf, line 248: can we put this mockery into a macro?

breathx: It depends on pallets set used in the (mock) runtime, so it differs from crate to crate.
For proper generation it will require patching-reintroducing `construct_runtime!` which is in fact a good idea too keep mocks closer to reality

breathx: #3844

gshep (approved):
:+1:

PR #3843, opened 2024-03-28
feat(runtime): Add a builtin actor for staking operations

ekovalev:
The change includes:
- another builtin actor added to the Runtime to enable staking-related logic in programs (e.g. liquid staking etc.). The actor converts incoming messages into dispatchable calls and relays the latter to the `Staking` pallet;
- a set of types facilitating interaction with the staking builtin actor collected in a crate called `gbuiltin-staking`;
- a simple example of how contracts could interact with the builtin in question - for demo purpose only as real life scenarios would require more attention to details.

### DIFF
--- a/common/src/pallet_tests.rs
+++ b/common/src/pallet_tests.rs
@@ -212,7 +212,12 @@ macro_rules! impl_pallet_staking {
             }
         }
 
-        type X = (AccountId, BlockNumber, pallet_staking::Pallet<Test>, ConstU32<100>);
+        type DataProviderInfo = (
+            AccountId,
+            BlockNumber,
+            pallet_staking::Pallet<Test>,
+            ConstU32<100>,
+        );
 
         #[allow(dead_code)]
         type StakingConfigEraPayout = DummyEraPayout;
@@ -223,9 +228,14 @@ macro_rules! impl_pallet_staking {
         #[allow(dead_code)]
         type StakingConfigNextNewSession = ();
         #[allow(dead_code)]
-        type StakingConfigElectionProvider = frame_election_provider_support::NoElection<X>;
+        type StakingConfigElectionProvider =
+            frame_election_provider_support::NoElection<DataProviderInfo>;
         #[allow(dead_code)]
-        type StakingConfigGenesisElectionProvider = frame_election_provider_support::NoElection<X>;
+        type StakingConfigGenesisElectionProvider =
+            frame_election_provider_support::NoElection<DataProviderInfo>;
+        #[allow(dead_code)]
+        type StakingConfigMaxNominatorRewardedPerValidator =
+            ConstU32<256>;
 
         mod pallet_tests_staking_config_impl {
             use super::*;
@@ -244,7 +254,6 @@ macro_rules! impl_pallet_staking_inner {
             // 8 eras for unbonding
             pub const BondingDuration: u32 = 8;
             pub const SlashDeferDuration: u32 = 7;
-            pub const MaxNominatorRewardedPerValidator: u32 = 256;
             pub const OffendingValidatorsThreshold: Perbill = Perbill::from_percent(17);
             pub const HistoryDepth: u32 = 84;
             pub const MaxNominations: u32 = 16;
@@ -268,7 +277,7 @@ macro_rules! impl_pallet_staking_inner {
             type SessionInterface = ();
             type EraPayout = StakingConfigEraPayout;
             type NextNewSession = StakingConfigNextNewSession;
-            type MaxNominatorRewardedPerValidator = MaxNominatorRewardedPerValidator;
+            type MaxNominatorRewardedPerValidator = StakingConfigMaxNominatorRewardedPerValidator;
             type OffendingValidatorsThreshold = OffendingValidatorsThreshold;
             type VoterList = pallet_staking::UseNominatorsAndValidatorsMap<Self>;
             type TargetList = pallet_staking::UseValidatorsMap<Self>;
@@ -315,6 +324,12 @@ macro_rules! impl_pallet_staking_inner {
         $runtime:ty, GenesisElectionProvider = $genesis_election_provider:ty $(, $( $rest:tt )*)?
     ) => {
         type StakingConfigGenesisElectionProvider = $genesis_election_provider;
+
+        $crate::impl_pallet_staking_inner!($runtime, $($( $rest )*)?);
+    };
+
+    ($runtime:ty, MaxNominatorRewardedPerValidator = $max_rewarded:ty $(, $( $rest:tt )*)?) => {
+        type StakingConfigMaxNominatorRewardedPerValidator = $max_rewarded;
 
         $crate::impl_pallet_staking_inner!($runtime, $($( $rest )*)?);
     };
--- a/common/src/pallet_tests.rs
+++ b/common/src/pallet_tests.rs
@@ -195,3 +195,127 @@ macro_rules! impl_pallet_authorship_inner {
         $crate::impl_pallet_authorship_inner!($runtime, $($( $rest )*)?);
     };
 }
+
+#[macro_export]
+macro_rules! impl_pallet_staking {
+    ($( $tokens:tt )*) => {
+        #[allow(dead_code)]
+        pub struct DummyEraPayout;
+        impl pallet_staking::EraPayout<u128> for DummyEraPayout {
+            fn era_payout(
+                _total_staked: u128,
+                total_issuance: u128,
+                _era_duration_millis: u64,
+            ) -> (u128, u128) {
+                // At each era have 1% `total_issuance` increase
+                (Permill::from_percent(1) * total_issuance, 0)
+            }
+        }
+
+        type X = (AccountId, BlockNumber, pallet_staking::Pallet<Test>, ConstU32<100>);
+
+        #[allow(dead_code)]
+        type StakingConfigEraPayout = DummyEraPayout;
+        #[allow(dead_code)]
+        type StakingConfigSlash = ();
+        #[allow(dead_code)]
+        type StakingConfigReward = ();
+        #[allow(dead_code)]
+        type StakingConfigNextNewSession = ();
+        #[allow(dead_code)]
+        type StakingConfigElectionProvider = frame_election_provider_support::NoElection<X>;
+        #[allow(dead_code)]
+        type StakingConfigGenesisElectionProvider = frame_election_provider_support::NoElection<X>;
+
+        mod pallet_tests_staking_config_impl {
+            use super::*;
+
+            $crate::impl_pallet_staking_inner!($( $tokens )*);
+        }
+    };
+}
+
+#[macro_export]
+macro_rules! impl_pallet_staking_inner {
+    ($runtime:ty$(,)?) => {
+        parameter_types! {
+            // 6 sessions in an era
+            pub const SessionsPerEra: u32 = 6;
+            // 8 eras for unbonding
+            pub const BondingDuration: u32 = 8;
+            pub const SlashDeferDuration: u32 = 7;
+            pub const MaxNominatorRewardedPerValidator: u32 = 256;
+            pub const OffendingValidatorsThreshold: Perbill = Perbill::from_percent(17);
+            pub const HistoryDepth: u32 = 84;
+            pub const MaxNominations: u32 = 16;
+        }
+
+        impl pallet_staking::Config for Test {
+            type Currency = Balances;
+            type UnixTime = Timestamp;
+            type CurrencyBalance = <Self as pallet_balances::Config>::Balance;
+            type CurrencyToVote = ();
+            type ElectionProvider = StakingConfigElectionProvider;
+            type GenesisElectionProvider = StakingConfigGenesisElectionProvider;
+            type RewardRemainder = ();
+            type RuntimeEvent = RuntimeEvent;
+            type Slash = StakingConfigSlash;
+            type Reward = StakingConfigReward;
+            type SessionsPerEra = SessionsPerEra;
+            type BondingDuration = BondingDuration;
+            type SlashDeferDuration = SlashDeferDuration;
+            type AdminOrigin = frame_system::EnsureRoot<AccountId>;
+            type SessionInterface = ();
+            type EraPayout = StakingConfigEraPayout;
+            type NextNewSession = StakingConfigNextNewSession;
+            type MaxNominatorRewardedPerValidator = MaxNominatorRewardedPerValidator;
+            type OffendingValidatorsThreshold = OffendingValidatorsThreshold;
+            type VoterList = pallet_staking::UseNominatorsAndValidatorsMap<Self>;
+            type TargetList = pallet_staking::UseValidatorsMap<Self>;
+            type NominationsQuota = pallet_staking::FixedNominationsQuota<16>;
+            type MaxUnlockingChunks = ConstU32<32>;
+            type HistoryDepth = HistoryDepth;
+            type EventListeners = ();
+            type WeightInfo = ();
+            type BenchmarkingConfig = pallet_staking::TestBenchmarkingConfig;
+        }
+    };
+
+    ($runtime:ty, EraPayout = $era_payout:ty $(, $( $rest:tt )*)?) => {
+        type StakingConfigEraPayout = $era_payout;
+
+        $crate::impl_pallet_staking_inner!($runtime, $($( $rest )*)?);
+    };
+
+    ($runtime:ty, Slash = $slash:ty $(, $( $rest:tt )*)?) => {
+        type StakingConfigSlash = $slash;
+
+        $crate::impl_pallet_staking_inner!($runtime, $($( $rest )*)?);
+    };
+
+    ($runtime:ty, Reward = $reward:ty $(, $( $rest:tt )*)?) => {
+        type StakingConfigReward = $reward;
+
+        $crate::impl_pallet_staking_inner!($runtime, $($( $rest )*)?);
+    };
+
+    ($runtime:ty, NextNewSession = $next_new_session:ty $(, $( $rest:tt )*)?) => {
+        type StakingConfigNextNewSession = $next_new_session;
+
+        $crate::impl_pallet_staking_inner!($runtime, $($( $rest )*)?);
+    };
+
+    ($runtime:ty, ElectionProvider = $election_provider:ty $(, $( $rest:tt )*)?) => {
+        type StakingConfigElectionProvider = $election_provider;
+
+        $crate::impl_pallet_staking_inner!($runtime, $($( $rest )*)?);
+    };
+
+    (
+        $runtime:ty, GenesisElectionProvider = $genesis_election_provider:ty $(, $( $rest:tt )*)?
+    ) => {
+        type StakingConfigGenesisElectionProvider = $genesis_election_provider;
+
+        $crate::impl_pallet_staking_inner!($runtime, $($( $rest )*)?);
+    };
+}
--- a/common/src/pallet_tests.rs
+++ b/common/src/pallet_tests.rs
@@ -272,7 +272,7 @@ macro_rules! impl_pallet_staking_inner {
             type BondingDuration = BondingDuration;
             type SlashDeferDuration = SlashDeferDuration;
             type AdminOrigin = frame_system::EnsureRoot<AccountId>;
-            type SessionInterface = Self;
+            type SessionInterface = ();
             type EraPayout = StakingConfigEraPayout;
             type NextNewSession = StakingConfigNextNewSession;
             type MaxExposurePageSize = MaxExposurePageSize;
--- a/examples/staking-broker/src/wasm.rs
+++ b/examples/staking-broker/src/wasm.rs
@@ -22,7 +22,7 @@
 //! implementation of a liquid staking contract, for example, would be more complex.
 
 use gbuiltin_staking::*;
-use gstd::{debug, msg, prelude::*, ActorId};
+use gstd::{debug, errors::Error, msg, prelude::*, ActorId};
 use hashbrown::HashMap;
 use hex_literal::hex;
 use parity_scale_codec::Encode;
@@ -41,7 +41,7 @@ struct StakingBroker {
     /// Registry of bonded deposits
     bonded: HashMap<ActorId, u128>,
     /// Reward payee account id
-    reward_account: AccountId,
+    reward_account: ActorId,
 }
 
 static mut STATE: Option<StakingBroker> = None;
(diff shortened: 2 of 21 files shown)


examples/staking-broker/src/wasm.rs
```diff
@@ -0,0 +1,266 @@
...
+use gbuiltin_staking::*;
+use gstd::{debug, msg, prelude::*, ActorId};
+use hashbrown::HashMap;
+use hex_literal::hex;
+use parity_scale_codec::Encode;
+
+// Staking proxy builtin actor program id (hardcoded for all runtimes)
+const BUILTIN_ADDRESS: ActorId = ActorId::new(hex!(
+    "77f65ef190e11bfecb8fc8970fd3749e94bed66a23ec2f7a3623e785d0816761"
+));
```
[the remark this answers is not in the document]
ekovalev: > This is different from
> 
> https://github.com/gear-tech/gear/blob/4638ed4f9cc9d7fab848bebc5d6e7465d5275ddd/gbuiltins/staking/src/lib.rs#L40-L42
> 

Right, the current `BuiltinId` is generated using a different seed comparing to the old way (now we don't use any `PalletId` whatsoever, hence the difference). To start with, I'll align the comment with the actual ID.

Not quite sure about hardcoding the address in `gbuiltin` crate: sure all builtin ids has to be published somewhere (#3854), but inherently they are not hardcoded but rather generated in the runtime. Most likely it is not going to be the case but it is still possible that the ID generation algo changes, which will make the `gbulitin` go out of sync with the runtime if it has this constant hardcoded.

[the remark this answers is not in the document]
clearloop: > Most likely it is not going to be the case but it is still possible that the ID generation algo changes

Yeah, for this, I was thinking about calculating it in `build.rs` and export, etc, seems dirty in development but if we have some kind of this implementations ( `gbuiltin::STAKING` ), it would be super neat for developers, they can import the addresses it directly from our library instead of copying it from somewhere and updating it manually

btw if the addresses of our builtin programs are `versioned`, if it is not versioned and we generated different addresses in different builds, it could cause a lot of programs in the development of gear programs for users

common/src/pallet_tests.rs
```diff
@@ -174,3 +174,127 @@ macro_rules! impl_pallet_authorship_inner {
...
+                _total_staked: u128,
+                total_issuance: u128,
+                _era_duration_millis: u64,
+            ) -> (u128, u128) {
+                // At each era have 1% `total_issuance` increase
+                (Permill::from_percent(1) * total_issuance, 0)
+            }
+        }
+
+        type X = (AccountId, BlockNumber, pallet_staking::Pallet<Test>, ConstU32<100>);
```
gshep, line 194: I understand but perhaps it would be better to give it a longer name

pallets/gear-builtin/src/staking.rs
```diff
@@ -0,0 +1,198 @@
+// This file is part of Gear.
+
+// Copyright (C) 2021-2023 Gear Technologies Inc.
```
gshep, line 3: copyright year

gshep (approved):
Good job! :1st_place_medal:

common/src/pallet_tests.rs
ekovalev: Sure, it came across my mind as well before I forgot about it ))

examples/staking-broker/src/wasm.rs
```diff
@@ -0,0 +1,266 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+//
+// This is a very basic implementation of a staking broker program for demo purpose only.
+// It accepts users messages and bond, unbond or nominate some validators on their behalf.
+// It is not concerned with some important constraints like unbonding period etc., so a real
+// implementation of a liquid staking contract, for example, would be more complex.
```
breathx, line 22:
```suggestion

//! This is a very basic implementation of a staking broker program for demo purpose only.
//! It accepts users messages and bond, unbond or nominate some validators on their behalf.
//! It is not concerned with some important constraints like unbonding period etc., so a real
//! implementation of a liquid staking contract, for example, would be more complex.
```

examples/staking-broker/src/wasm.rs
```diff
@@ -0,0 +1,266 @@
...
+use hex_literal::hex;
+use parity_scale_codec::Encode;
+
+// Staking proxy builtin actor program id (hardcoded for all runtimes)
+const BUILTIN_ADDRESS: ActorId = ActorId::new(hex!(
+    "77f65ef190e11bfecb8fc8970fd3749e94bed66a23ec2f7a3623e785d0816761"
+));
+
+// Runtime AccountId type
+pub type AccountId = [u8; 32];
```
breathx, line 36: this can use just ActorId type since their encoding matches

breathx: moreover a few lines below you use both ActorId and this AccountId :/

examples/staking-broker/src/wasm.rs
```diff
@@ -0,0 +1,266 @@
...
+        .await
+    {
+        Ok(_) => {
+            debug!("[StakingBroker] Success reply from builtin actor received");
+            on_success();
+            msg::reply_bytes(b"Success", 0).expect("Failed to send reply");
+        }
+        Err(e) => {
+            debug!("[StakingBroker] Error reply from builtin actor received: {e:?}");
+            msg::reply_bytes(b"Error in upstream program", 0).expect("Failed to send reply");
```
breathx, line 65: its better to panic for such case since reply will be with error reply code

gbuiltins/staking/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+#[codec(crate = scale)]
+pub enum Request {
+    /// Version 1 of the staking built-in actor protocol.
+    V1(RequestV1),
+}
+
+#[derive(Encode, Decode, Clone, PartialEq, Eq, Debug)]
+#[codec(crate = scale)]
+pub enum RequestV1 {
```
breathx, line 79: it seems that we've agreed to support versioning on actor model layer instead of within the same actor, so this versioning should be removed

gbuiltins/staking/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+    Chill,
+    /// Request stakers payout for the given era.
+    PayoutStakers {
+        validator_stash: AccountId,
+        era: u32,
+    },
+    /// Rebond a portion of the sender's stash scheduled to be unlocked.
+    Rebond { value: u128 },
+    /// Set the reward destination.
+    SetPayee { payee: RewardAccount },
```
breathx, line 103: it could be Option if you really wish but actually can't see any reason not to accept just address here

gbuiltins/staking/src/lib.rs
```diff
@@ -0,0 +1,128 @@
...
+/// Message processing output
+pub type Response = ();
+
+/// Type representing a dispatched Runtime call internal error.
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq, derive_more::Display)]
+#[codec(crate = scale)]
+pub enum DispatchErrorReason {
+    #[display(fmt = "Runtime internal error")]
+    RuntimeError,
+}
```
breathx, line 128: these two types looks useless for me too

pallets/gear-builtin/src/benchmarking.rs
```diff
@@ -311,6 +315,24 @@ benchmarks! {
         let standard = <ark_bls12_381::g2::Config as SWCurveConfig>::mul_projective(&base, &bigint);
         assert_eq!(standard, result.0);
     }
+
+    staking_estimate_decode {
+        let l in 1 .. MAX_PAYLOAD_SIZE as u32;
+
+        // Longest payload corresponds to the `RequestV1::Nominate` variant.
```
breathx, line 322: consider limitations on payload length and pre-check of it, since amount of nominees is limited with 16 (I'd even consider using LimitedVec for 16 hashes, but seems too much since this param is dictated by the runtime and governance)
so max payload len for this is 1 + 4 + 18 * 32 bytes atm

breathx: it will remove need in benching anything there

pallets/gear-builtin/src/staking.rs
```diff
@@ -0,0 +1,198 @@
...
+//! The BuiltIn Actors pallet provides a set of unique accounts that are treated as built-in
+//! actors ids (`ProgramId`).
+//!
+//! - [`Config`]
+//!
+//! ## Overview
+//!
+//! The pallet implements a builtin actor that handles messages from the message queue that
+//! request some sort of staking-related logic (like tokens bonding, validators nominating etc.).
+//!//!
```
breathx, line 30:
```suggestion
//!
```

pallets/gear-builtin/src/staking.rs
```diff
@@ -0,0 +1,198 @@
...
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//! # Gear BuiltIn Actors Pallet
```
breathx, line 19: should the start be removed from **staking** mod docs?

pallets/gear-builtin/src/tests/staking.rs
```diff
@@ -0,0 +1,868 @@
...
+
+use crate::{
+    self as pallet_gear_builtin,
+    mock::{
+        BLOCK_AUTHOR, ENDOWMENT, EXISTENTIAL_DEPOSIT, MILLISECS_PER_BLOCK, SIGNER, UNITS,
+        VAL_1_STASH, VAL_2_STASH, VAL_3_STASH,
+    },
+    staking::Actor as StakingBuiltin,
+};
+use common::Origin;
```
breathx, line 27: please move everything "operational" to the bottom into utils mod, let tests go first

pallets/gear-builtin/src/tests/staking.rs
```diff
@@ -0,0 +1,868 @@
...
+            })
+            .encode(),
+            10_000_000_000,
+            0,
+            false,
+        ));
+
+        run_to_next_block();
+    });
+}
```
breathx, line 868: any error-case?

pallets/gear-builtin/src/staking.rs
```diff
@@ -0,0 +1,198 @@
...
+        match res {
+            Ok(_post_info) => {
+                log::debug!(
+                    target: LOG_TARGET,
+                    "Call dispatched successfully",
+                );
+                (Ok(()), actual_gas)
+            }
+            Err(e) => {
+                log::error!(target: LOG_TARGET, "Error disptaching call: {:?}", e);
```
breathx, line 83:
```suggestion
                log::error!(target: LOG_TARGET, "Error dispatching call: {:?}", e);
```

breathx: its not worth error log

pallets/gear-builtin/src/staking.rs
```diff
@@ -0,0 +1,198 @@
...
+                (
+                    Err(BuiltinActorError::Custom(
+                        DispatchErrorReason::RuntimeError.to_string().into(),
```
breathx, line 86: that usage seems supporting my initial idea that this type is unnecessary

pallets/gear-builtin/src/staking.rs
```diff
@@ -0,0 +1,198 @@
...
+                    }
+                };
+                Self::dispatch_call(origin, call, gas_limit)
+            }
+            _ => (Err(BuiltinActorError::DecodingError), Default::default()),
+        };
+
+        (
+            result.map(|response| {
+                response
```
breathx, line 190: may be just default?

pallets/gear-builtin/src/staking.rs
```diff
@@ -0,0 +1,198 @@
...
+    fn handle(dispatch: &StoredDispatch, gas_limit: u64) -> (Result<Payload, Self::Error>, u64) {
+        let message = dispatch.message();
+        let origin: T::AccountId = dispatch.source().cast();
+        let mut payload = message.payload_bytes();
+
+        let decoding_cost =
+            <T as Config>::WeightInfo::staking_estimate_decode(payload.len() as u32).ref_time();
+
+        // Decode the message payload to derive the desired action
+        let (result, gas_spent) = match Request::decode(&mut payload) {
```
breathx, line 112: especially due to type complexity this code blob feels hard to read

pallets/gear-builtin/src/staking.rs
```diff
@@ -0,0 +1,198 @@
...
+
+impl<T: Config> BuiltinActor for Actor<T>
+where
+    T::AccountId: Origin,
+{
+    const ID: u64 = 2;
+
+    type Error = BuiltinActorError;
+
+    fn handle(dispatch: &StoredDispatch, gas_limit: u64) -> (Result<Payload, Self::Error>, u64) {
```
breathx, line 103: BTW, for each operation you must make sure that remaining balance after execution isn't less than ED or eq 0, as we do for any sending, because it will cause a lot of different parts of the code, for example payout from `gear-bank` ⚠️

breathx (changes requested):

examples/staking-broker/src/wasm.rs
ekovalev: These are conceptually two different things, aren't they - a program address in actor space and an ordinary account id?
However, you're right that this `AccountId` declaration is redundant: we should reuse the `AccountId` from the `gbuiltin_staking` crate - the type that is used in API functions declarations. Removed it.

pallets/gear-builtin/src/benchmarking.rs
ekovalev: That's correct that a valid nominate request should comply with Runtime settings, however nothing stops a caller contract send an ill-formed request (that will inevitably be rejected in the upstream extrinsic) which we still need to decode. So unless we use a `LimitedVec` in the API (as you suggest, with some reasonable limit that we will never exceed in the runtime) estimating the decoding cost based on the payload size seems appropriate, is it not?

pallets/gear-builtin/src/staking.rs
ekovalev: Certainly. This is a leftover from the earlier implementation. Removed.

gbuiltins/staking/src/lib.rs
ekovalev: Are we talking about having a dedicated enum `RewardAccount` indicating the reward destination here?
Generally, it is supposed to be a mirror type for the [RewardDestination<AccountID>](https://github.com/paritytech/polkadot-sdk/blob/f13408d57f7ac2c1f49cc9a2f95393c0ad4b9a35/substrate/frame/staking/src/lib.rs#L401) which is more than a wrapper around an address.
Initially, I only kept two variants of that one (thinking there wouldn't be much use for the remaining 3 in our current context) but now I'm more convinced that even with the `Controller` variant having been deprecated we still should stick to one-to-one correspondence with the target type to support the whole variety of possible use cases that may arise in applications later.

pallets/gear-builtin/src/staking.rs
ekovalev: What's wrong with having a dedicated type as a wrapper around some string literal? We have similar error types all over the place, don't we?

gbuiltins/staking/src/lib.rs
ekovalev: Just abstractions. With zero runtime overhead. These particular ones might seem trivial but they could be more complex.

pallets/gear-builtin/src/staking.rs
ekovalev: Maybe this statement could be true for some builtin actor that deals with transfers in some way (because the ED checks belongs to balances logic whereas gear-bank stuff - to the `gear` pallet), but in this case all the actor does is relaying messages (converting them into dispatchable calls) and telling the caller how much it will cost (in terms of gas, because actors only operate in terms of gas). So any balances related logic is not this actor concern. I don't see where any extra checks could be applied here.

breathx:
Blocked by #3924

gbuiltins/staking/src/lib.rs
```diff
@@ -0,0 +1,111 @@
...
+//!
+
+#![no_std]
+
+extern crate alloc;
+
+use alloc::vec::Vec;
+use scale_info::scale::{self, Decode, Encode};
+
+pub type AccountId = [u8; 32];
```
breathx, line 64: use `gprimitives::ActorId`
```suggestion
```

pallets/gear-builtin/src/staking.rs
```diff
@@ -0,0 +1,178 @@
...
+        let message = dispatch.message();
+        let origin: T::AccountId = dispatch.source().cast();
+        let mut payload = message.payload_bytes();
+
+        let decoding_cost =
```
breathx, line 85: shouldn't gas limit be decreased by decoding cost here?

pallets/gear-builtin/src/staking.rs
```diff
@@ -0,0 +1,178 @@
...
+            <T as Config>::WeightInfo::staking_estimate_decode(payload.len() as u32).ref_time();
+
+        // Decode the message payload to derive the desired action
+        let (result, gas_spent) = match Request::decode(&mut payload) {
```
breathx, line 89: `let Ok(_) = expr else { return (Err(), decoding_cost) }` will increase readability imho

pallets/gear-builtin/src/staking.rs
```diff
@@ -0,0 +1,178 @@
...
+                            })
+                            .collect(),
+                    }
+                    .into(),
+                    Request::Chill => pallet_staking::Call::<T>::chill {}.into(),
+                    Request::PayoutStakers {
+                        validator_stash,
+                        era,
+                    } => {
+                        let stash_id = ProgramId::try_from(&validator_stash[..])
```
breathx, line 140: using gprimitives will allow direct casts
otherwise ActorId impls From<[u8; 32]>

pallets/gear-builtin/src/staking.rs
```diff
@@ -0,0 +1,178 @@
...
+        let decoding_cost =
+            <T as Config>::WeightInfo::staking_estimate_decode(payload.len() as u32).ref_time();
+
+        // Decode the message payload to derive the desired action
+        let (result, gas_spent) = match Request::decode(&mut payload) {
+            Ok(request) => {
+                // Handle staking requests
+                let gas_limit = gas_limit.saturating_sub(decoding_cost);
+                let call = match request {
+                    Request::Bond { value, payee } => {
```
breathx, line 94: I'd move all of it into separate "cast" function below

pallets/gear-builtin/src/lib.rs
```diff
@@ -158,7 +163,13 @@ pub mod pallet {
     pub struct Pallet<T>(PhantomData<T>);
 
     #[pallet::config]
-    pub trait Config: frame_system::Config {
+    pub trait Config: frame_system::Config + pallet_staking::Config {
```
breathx, line 166: can we somehow avoid depending on staking pallet explicitly? if yes, place todo here pls

pallets/gear-builtin/src/benchmarking.rs
```diff
@@ -286,6 +290,24 @@ benchmarks! {
     } verify {
         assert!(ArkScale::<G2Affine>::decode(&mut &_result.unwrap()[..]).is_ok())
     }
+
+    staking_estimate_decode {
```
breathx, line 294: nomination targets are limited by staking pallet
it could be used for benchmarks. moreover, it'd allow rejecting incoming requests by payload size if they're supposed to be invalid mb

pallets/gear-builtin/src/tests/staking.rs
```diff
@@ -0,0 +1,968 @@
...
+            10_000_000_000,
+            0,
+            false,
+        ));
+
+        run_to_next_block();
+        // No staking-related events should have been emitted
+        assert_no_staking_events();
+        // Event with user message has been triggered
+        assert_user_message_sent();
```
breathx, line 217: I'd assert error payload for such cases

breathx (changes requested):
A few last discussion and good to go. Awesome work done 👍🏻

pallets/gear-builtin/src/staking.rs
ekovalev: It should and it is (just a couple of lines further down before deriving a RuntimeCall).

pallets/gear-builtin/src/lib.rs
ekovalev: It certainly can be done, but the only way I see right now is to explicitly declare the `Config::RuntimeCall` type being `From<gbuiltin_staking::Request>` and then defining the conversion at the runtime level (where the main `RuntimeCall` enum is defined in `vara/runtime/src/lib.rs`). But that would be much uglier choice and would require exposing internal staking builtin logic at the higher runtime level.

In my opinion declaring the pallet's `Config` trait implementing also `pallet_staking::Config` to be able to convert the `pallet_staking::Call<T>` type to the builtin pallet's `RuntimeCall` is not a problem at all - this is Substrate-idiomatic approach and should not be avoided.

ekovalev: **Update:** removed the dependency.

pallets/gear-builtin/Cargo.toml
```diff
@@ -66,13 +73,15 @@ sha2 = { workspace = true, features = ["std"] }
...
+	"common/std",
 	"frame-benchmarking?/std",
 	"frame-support/std",
 	"frame-system/std",
 	"gear-core/std",
 	"gear-runtime-interface/std",
 	"pallet-gear/std",
-	"parity-scale-codec/std",
+	"pallet-staking/std",
+    "parity-scale-codec/std",
```
breathx, line 84: tab

pallets/gear-builtin/src/lib.rs
```diff
@@ -95,6 +99,9 @@ impl From<BuiltinActorError> for ActorExecutionErrorReplyReason {
             BuiltinActorError::DecodingError => ActorExecutionErrorReplyReason::Trap(
                 TrapExplanation::Panic("Message decoding error".to_string().into()),
             ),
+            BuiltinActorError::PayloadTooLarge => ActorExecutionErrorReplyReason::Trap(
```
breathx, line 102: I'd avoid this. Too large payload is kind of case of decoding error

breathx (commented):
Now need to merge actual master with substrate 1.4.0 and good to go!

breathx (approved):

PR #3848, opened 2024-03-30
fix(gtest): allow zero balance for `exec::exit()`

### LINKED ISSUE #3847: gtest is broken for exec::exit()
any exit leads to:

```
thread 'main' panicked at /Users/nikolayvolf/.cargo/git/checkouts/gear-bdc2b7d48ef3acbd/529cdef/gtest/src/manager.rs:602:13:
An attempt to mint value (0) less than existential deposit (10000000000000)
```

StackOverflowExcept1on:
Resolves #3847

@gear-tech/dev

### DIFF
--- a/gtest/src/manager.rs
+++ b/gtest/src/manager.rs
@@ -598,8 +598,8 @@ impl ExtManager {
         )
     }
 
-    pub(crate) fn mint_to(&mut self, id: &ProgramId, value: Balance) {
-        if value < crate::EXISTENTIAL_DEPOSIT {
+    pub(crate) fn mint_to(&mut self, id: &ProgramId, value: Balance, keep_alive: bool) {
+        if keep_alive && value < crate::EXISTENTIAL_DEPOSIT {
             panic!(
                 "An attempt to mint value ({}) less than existential deposit ({})",
                 value,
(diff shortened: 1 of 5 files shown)


breathx (approved):

breathx:
It's not actually about keeping something alive, but solves the issue

gtest/src/manager.rs
```diff
@@ -598,8 +598,8 @@ impl ExtManager {
         )
     }
 
-    pub(crate) fn mint_to(&mut self, id: &ProgramId, value: Balance) {
-        if value < crate::EXISTENTIAL_DEPOSIT {
+    pub(crate) fn mint_to(&mut self, id: &ProgramId, value: Balance, keep_alive: bool) {
```
gshep, line 601: Perhaps you don't know yet that I personally don't like primitive types especially booleans =) .

This remark is not sufficient but just to mention. Better introduce the new enumuration, for example `MintMode` with two variants `KeepAlive`, `AllowDeath`. Then calls to the method would look like `self.mint_to(&value_destination, balance, MintMode::AllowDeath/KeepAlive);` instead of not much informative `self.mint_to(&value_destination, balance, false/true);`

gshep (approved):
LGTM!

PR #3880, opened 2024-04-12
runtime-fuzzer(tests): Add test cases which could cause previously fixed panics

techraed:
TODO:
- [x] write unit-tests using these inputs
- [x] revert changes in voucher paller


StackOverflowExcept1on:
It would be nice to add not just fuzzer inputs, which may become outdated, but tests with fixed order of syscalls to reproduce panic in our runtime. Now it can be done via `.with_keeping_insertion_order(true)`: https://github.com/gear-tech/gear/blob/89f94fe7d83d8ecc4e94950588fd55a2c4362ce0/utils/wasm-gen/src/tests.rs#L456

StackOverflowExcept1on (approved):

techraed:
> It would be nice to add not just fuzzer inputs, which may become outdated, but tests with fixed order of syscalls to reproduce panic in our runtime. Now it can be done via `.with_keeping_insertion_order(true)`:
> 
> https://github.com/gear-tech/gear/blob/89f94fe7d83d8ecc4e94950588fd55a2c4362ce0/utils/wasm-gen/src/tests.rs#L456

discussed.

PR #3888, opened 2024-04-15
chore(scripts): exclude g-tools in test scripts by default

clearloop:
Resolves # .

- [x] exclude `gcli`, `gsdk`, `gclient` in `./scripts/gear.sh test`

@gear-tech/dev


NikVolf (approved):

Makefile
```diff
@@ -189,11 +189,8 @@ test-doc:
 	@ ./scripts/gear.sh test docs
 
 .PHONY: test-gear
-test-gear: # Crates except gclient, gcli, gsdk are excluded to significantly decrease time.
```
breathx, line 192: why?

clearloop: these three packages are nested in `test_workspace`, so imo we just need to clarify for the following packages in the command so we can remove the annotation for these three packages here? or what about 

```
The following crates are excluded to significantly decrease time.
```

breathx (approved):

PR #3897, opened 2024-04-17
refactor(core): move `validate_memory_params` to `AllocationsCotntext…

### LINKED ISSUE #3813: move `validate_memory_params` to `AllocationsCotntext::new`
### File Location(s)

core-processor/src/executor.rs
core/src/memory.rs

### Proposal

Presently we do not check `AllocationsContext` for values consistency in `new(...)`, but this should be done. See  `validate_memory_params ` to understand what checks should be added

Also `IncorrectAllocationDataError` can be removed after that, because AllocationsContext will guaranties that this is impossible.

ByteNacked:
Resolves #3813 .

Move `validate_memory_params` into `AllocationsContext` constructor and make it capable of failing.

Some tests in core/src/memory.rs were updated due to new invariants established by the validation of memory parameters upon `AllocationsContext` creation.

@reviewer-or-team

### DIFF
--- a/core/src/memory.rs
+++ b/core/src/memory.rs
@@ -253,6 +253,48 @@ impl GrowHandler for NoopGrowHandler {
     fn after_grow_action(self, _mem: &mut impl Memory) {}
 }
 
+/// Inconsistency in memory parameters provided for wasm execution.
+#[derive(Debug, Clone, PartialEq, Eq, derive_more::Display)]
+pub enum MemorySetupError {
+    /// Memory size exceeds max pages
+    #[display(fmt = "Memory size {memory_size:?} must be less than or equal to {max_pages:?}")]
+    MemorySizeExceedsMaxPages {
+        /// Memory size
+        memory_size: WasmPagesAmount,
+        /// Max allowed memory size
+        max_pages: WasmPagesAmount,
+    },
+    /// Insufficient memory size
+    #[display(fmt = "Memory size {memory_size:?} must be at least {static_pages:?}")]
+    InsufficientMemorySize {
+        /// Memory size
+        memory_size: WasmPagesAmount,
+        /// Static memory size
+        static_pages: WasmPagesAmount,
+    },
+    /// Stack end is out of static memory
+    #[display(fmt = "Stack end {stack_end:?} is out of static memory 0..{static_pages:?}")]
+    StackEndOutOfStaticMemory {
+        /// Stack end
+        stack_end: WasmPage,
+        /// Static memory size
+        static_pages: WasmPagesAmount,
+    },
+    /// Allocated page is out of allowed memory interval
+    #[display(
+        fmt = "Allocated page {page:?} is out of allowed memory interval {static_pages:?}..{memory_size:?}"
+    )]
+    AllocatedPageOutOfAllowedInterval {
+        /// Allocated page
+        page: WasmPage,
+        /// Static memory size
+        static_pages: WasmPagesAmount,
+        /// Memory size
+        memory_size: WasmPagesAmount,
+    },
+}
+
+// TODO: This error type used for temporary solution, should be removed in #3791.
 /// Incorrect allocation data error
 #[derive(Debug, Clone, Eq, PartialEq, derive_more::Display)]
 #[display(fmt = "Allocated memory pages or memory size are incorrect")]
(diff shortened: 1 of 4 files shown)


core/src/memory.rs
```diff
@@ -329,7 +437,9 @@ impl AllocationsContext {
             match Interval::<WasmPage>::try_from(start..end) {
                 Ok(interval) if interval.len() >= pages => break,
                 Err(TryFromRangeError::IncorrectRange) => {
-                    return Err(IncorrectAllocationDataError.into())
+                    unreachable!(
+                        "Allocated memory pages or memory size are incorrect, but we checked it in validate memory param, start: {start:?}, end: {end:?}"
+                    )
```
ByteNacked, line 442: This invariant should be upheld by `valdidate_memory_params` upon creation of object.

core/src/pages.rs
```diff
@@ -56,7 +56,7 @@ const _: () = assert!(WASM_PAGE_SIZE % GEAR_PAGE_SIZE == 0);
     Default,
     derive_more::Into,
 )]
-pub struct PagesAmount<const SIZE: u32>(u32);
+pub struct PagesAmount<const SIZE: u32>(pub(crate) u32);
```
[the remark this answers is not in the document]
grishasobol: It's not safe to do this, because `gear-core` is big enough, so developers can simply make bugs using this constructor. `PagesAmount` supports `Into<u32>`, `From<u16>` and `TryFrom<u32>` this must be enough for any purposes in tests.

core/src/pages.rs
```diff
@@ -186,7 +186,7 @@ impl<const SIZE: u32> PartialOrd<PagesAmount<SIZE>> for Page<SIZE> {
     Default,
     derive_more::Into,
 )]
-pub struct Page<const SIZE: u32>(u32);
+pub struct Page<const SIZE: u32>(pub(crate) u32);
```
grishasobol, line 189: The same  unsafe

core/src/pages.rs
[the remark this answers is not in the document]
grishasobol: Or you can make it `pub` for tests only

grishasobol (changes requested):

core/src/pages.rs
ByteNacked: Fixed, added method `[cfg(test)] fn into_inner() -> u32`

core/src/memory.rs
```diff
@@ -602,19 +898,26 @@ mod tests {
...
-                allocations in allocations(),
-                max_pages in wasm_pages_amount(),
-                mem_size in wasm_pages_amount(),
+                mem_params in combined_memory_params(),
                 actions in actions(),
             ) {
                 let _ = env_logger::try_init();
 
-                let mut ctx = AllocationsContext::new(allocations, static_pages, max_pages);
+                if check_if_dev_profile() {
```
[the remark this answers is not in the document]
ByteNacked: The problem with the slow property test was the poor performance of the  `AllocationContext::alloc` with opt = 1 in dev profile in happy case.

```rust
pub fn alloc<G: GrowHandler>(
{
      ...
        self.allocations.extend(interval.iter()); << HERE
       ...
}
```

For some reason, `IntervalIterator::next` works very slowly with opt = 1 in the development profile. I'm not sure if it should be fixed, but I made a workaround by reducing the chance of successful allocation in the prop-test (Same behavior as it was before this PR).

core/src/pages.rs
grishasobol: I cannot understand what the reason to add this method ? you can use `u32::from(page)`

core/src/memory.rs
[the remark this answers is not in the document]
ByteNacked: In the previous version of the property test, `AllocationContext::alloc` rarely succeeds due to the way its strategy generates input parameters. (Almost never executes `self.allocations.extend(interval.iter());`) 

After I changed the strategy, `AllocationContext::alloc`  started to succeeded more frequently (thus executing `self.allocations.extend(interval.iter());` more frequently) . That led to prop-test slow down.

Now it fixed by prop-test's strategy generating input parameters in such way which decrease chance of `AllocationContext::alloc` succession. (Same behavior as before PR).

core/src/pages.rs
ByteNacked: Missed it, fixed

core/src/memory.rs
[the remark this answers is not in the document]
ByteNacked: it seems that actual problem was in `BTree::extend` call.
Confirmed this with a snippet:
```rust
        //self.allocations.extend(interval.iter());
        for i in interval.iter() {
            core::hint::black_box(i);
        }
```

After discussion changed in prop-test number of generated pages in action per allocation. 
```rust
        fn actions() -> impl Strategy<Value = Vec<Action>> {
            let action = prop_oneof![
                // Allocate smaller number (0..32) of pages due to `BTree::extend` significantly slows down prop-test.
                wasm_pages_amount_with_range(0, 32).prop_map(|pages| Action::Alloc { pages }),
        ....
```
It solves problem with prop-test slow down too.

core/src/memory.rs
```diff
@@ -578,6 +837,44 @@ mod tests {
             })
         }
 
+        fn wasm_pages_amount() -> impl Strategy<Value = WasmPagesAmount> {
+            wasm_pages_amount_with_range(0, u16::MAX as u32 + 1)
+        }
+
+        // This high-order strategy generates valid memory parameters in a specific way that allows passing `AllocationContext::validate_memory_params` checks.
+        fn combined_memory_params() -> impl Strategy<
+            Value = (
```
grishasobol, line 846: Please append new struct here `MemoryParams`

grishasobol (approved):
Ok. Everything looks good. But before commit I would ask you to make local run of memory property tests with bigger amount of iterations in order to check everything is fine there

core/src/memory.rs
ByteNacked: I have a similar idea, but it seems impossible to create a combined strategy that uses a `struct` for `Value` instead of a tuple. The `prop_flat_map` combinator transforms a tuple of strategies into a strategy of tuples, but it cannot do the same for a struct.

grishasobol: Hm, yes. May be the most correct way is to implement `Arbitrary` for `struct MemoryParams`.
You can see examples in common/numerated/src/mock.rs. But may be better to make it in other PR

ByteNacked: Created issue #3906 3906

ByteNacked:
Waiting for 1 more reviewer's approval

grishasobol:
So, I think this review hangs on enough of time and can be merged

PR #3901, opened 2024-04-18
fix(core): Fix send_commit updating data on operation failure, potential fix for #3881

### LINKED ISSUE #3881: `impl Environment for Ext` in atomic way
### Problem to Solve

If some function ends up in error, it should has no affect on state

### Possible Solution

1. Go through all of the functions that is called by backend
2. Impl test set for each function with different outcomes
( for example we have fn:
```
fn foo() {
    do_a()?; // <- for each of these end points of the function \
    do_b()?; // <- implement test, that checks no-op on state \
    do_c()?; // <- of the ext itself in case of err
}
```
3. Fix found bugs 

### Notes

_No response_

playX18:
Partially resolves #3881. Adds `ChangeBuilder` which allows to accumulate necessary changes performing necessary checks and then apply them all-at-once after the operation without failing. 
This commit definitely fixes `send_commit` and `reservation_send_commit` being broken. Needs more investigation on how to update other methods. 

@breathx


NikVolf:
is it for review? or in progress still?

breathx:
WIP

core/src/gas.rs
```diff
@@ -150,6 +150,18 @@ impl GasCounter {
             burned: self.burned,
         }
     }
+
+    /// Clone the counter
+    ///
+    /// # Safety
+    ///
+    /// Use only when it's absolutely necessary to cloen the counter i.e atomic implementation of `Ext`.
```
breathx, line 158: Same for other copy-pasted clone impls
```suggestion
    /// Clone the counter.
    ///
    /// # Safety
    ///
    /// Use only when it's absolutely necessary to clone the counter i.e atomic implementation of `Ext`.
```

core-processor/src/ext.rs
```diff
@@ -319,6 +319,134 @@ impl GrowHandler for LazyGrowHandler {
...
+/// incrementally builds list of changes.
+///
+/// Changes are applied after operation is completed
+struct ExtChanges {
+    gas_counter: GasCounter,
+    gas_allowance_counter: GasAllowanceCounter,
+    value_counter: ValueCounter,
+    outgoing_glassless: u64,
+    mailbox_threshold: u64,
+    // TODO: Maybe replace with `SmallVec` to reduce heap usage?
```
breathx, line 339: Nice idea, I'd apply it right here
```suggestion
```

core-processor/src/ext.rs
```diff
@@ -319,6 +319,134 @@ impl GrowHandler for LazyGrowHandler {
...
+///
+/// Copies some counters into itself and performs operations on them and
+/// incrementally builds list of changes.
+///
+/// Changes are applied after operation is completed
+struct ExtChanges {
+    gas_counter: GasCounter,
+    gas_allowance_counter: GasAllowanceCounter,
+    value_counter: ValueCounter,
+    outgoing_glassless: u64,
```
breathx, line 337: gas-less (wout gas applied)
```suggestion
    outgoing_gasless: u64,
```

core-processor/src/ext.rs
```diff
@@ -319,6 +319,134 @@ impl GrowHandler for LazyGrowHandler {
...
+}
+
+/// Used to atomically update `Ext` which prevents some errors
+/// when data was updated but operation failed.
+///
+/// Copies some counters into itself and performs operations on them and
+/// incrementally builds list of changes.
+///
+/// Changes are applied after operation is completed
+struct ExtChanges {
```
breathx, line 333: just to consider: what do you think about changes just own counters and keeps amount to be burned/spent and on each try to charge something it sums them from underlying store with burn/charge argument and call something like `counter.can_charge()`?
that will allow us not to mutate, clone and replace anything?

core-processor/src/ext.rs
breathx: overall at this point it seems that changes could contain only one change so I'd remove it until needed

core-processor/src/ext.rs
```diff
@@ -319,6 +319,134 @@ impl GrowHandler for LazyGrowHandler {
...
+    }
+
+    fn non_zero_or_delay_gasfull(&mut self) -> Result<(), FallibleExtError> {
+        let prev_gasless_fee = self
+            .outgoing_glassless
+            .saturating_mul(self.mailbox_threshold);
+        self.reduce_gas(prev_gasless_fee)?;
+        self.outgoing_glassless = 0;
+        Ok(())
+    }
```
breathx, line 409: this two fns feels weird and uninformative here

breathx: lets maybe expand them in place back?

core-processor/src/ext.rs
```diff
@@ -319,6 +319,134 @@ impl GrowHandler for LazyGrowHandler {
...
+            return Err(ChargeError::GasAllowanceExceeded);
+        }
+        Ok(())
+    }
+
+    fn apply(mut self, ext: &mut Ext) {
+        for change in self.changes.drain(..) {
+            match change {
+                Change::MarkUsed(id) => {
+                    assert!(ext.context.gas_reserver.mark_used(id).is_ok());
```
breathx, line 434: replace with debug assert, performing operation out of macro pls

core-processor/src/ext.rs
```diff
@@ -319,6 +319,134 @@ impl GrowHandler for LazyGrowHandler {
...
+                }
+            }
+        }
+
+        core::mem::swap(&mut ext.context.value_counter, &mut self.value_counter);
```
breathx, line 439: why not to just `ext.context.value_counter = self.value_counter`

core-processor/src/ext.rs
```diff
@@ -611,11 +706,22 @@ impl Ext {
...
-    fn charge_sending_fee(&mut self, delay: u32) -> Result<(), ChargeError> {
+    fn charge_sending_fee(
+        &mut self,
+        changes: &mut ExtChanges,
+        delay: u32,
+    ) -> Result<(), ChargeError> {
         if delay == 0 {
-            self.charge_gas_if_enough(self.context.message_context.settings().sending_fee)
+            changes.charge_gas_if_enough(self.context.message_context.settings().sending_fee)
+            //self.charge_gas_if_enough(self.context.message_context.settings().sending_fee)
```
breathx, line 716: rm comments

core-processor/src/ext.rs
```diff
@@ -611,11 +706,22 @@ impl Ext {
...
 
-    fn charge_sending_fee(&mut self, delay: u32) -> Result<(), ChargeError> {
+    fn charge_sending_fee(
+        &mut self,
+        changes: &mut ExtChanges,
+        delay: u32,
+    ) -> Result<(), ChargeError> {
         if delay == 0 {
-            self.charge_gas_if_enough(self.context.message_context.settings().sending_fee)
+            changes.charge_gas_if_enough(self.context.message_context.settings().sending_fee)
```
breathx, line 715: what if we impl charge_sending_fee directly on `ExtChanges` (btw I'd consider renaming into `ExtOverlay`)?

core-processor/src/ext.rs
```diff
@@ -1766,10 +1894,7 @@ mod tests {
             0,
         );
         // replace the following code with `assert!(res.is_ok());`
```
breathx, line 1896: rm comments

core-processor/src/ext.rs
```diff
@@ -1218,12 +1343,15 @@ impl Externalities for Ext {
...
     ) -> Result<(), Self::FallibleError> {
-        self.reduce_gas(amount)?;
+        let mut changes = ExtChanges::new(self);
+        changes.reduce_gas(amount)?;
 
         self.context
             .message_context
             .reply_deposit(message_id, amount)?;
 
+        changes.apply(self);
```
breathx, line 1353: Im wondering how to guarantee that we didn't forget to call changes.apply?

breathx (changes requested):

breathx:
Please take a look @techraed

core-processor/src/ext.rs
playX18: I've initially used this approach but it resulted in not so good looking code, I could try to make it look readable again and push the changes

core-processor/src/ext.rs
playX18: Because we cannot move the `value_counter` out of `self`, at least rustc errors this

core-processor/src/ext.rs
playX18: One way is to introduce `applied: bool` field and if it's false on `Drop for ExtChanges` panic that there's unapplied changes. Another way is to somehow have `Changes::with_changes(|changes| { ... })` and it could store both `Ext` we modify and changes that we're about to make

core-processor/src/ext.rs
breathx: Let's give it another try in different pr and focus to apply these changes in any manner

breathx: Feel free to share some thoughts in issue not to lose context

core-processor/src/ext.rs
breathx: Discussed in dm

core-processor/src/ext.rs
playX18: Can leave it as is right now and in separate PR where we get rid of cloning counters it won't be a problem

breathx:
@techraed reminder for review

core-processor/src/ext.rs
```diff
@@ -962,18 +1143,26 @@ impl Externalities for Ext {
...
-                .syscalls
-                .gr_reply_push_input_per_byte
-                .cost_for(range.len().into()),
-        )?;
-
-        self.context.message_context.reply_push_input(range)?;
-
-        Ok(())
+        self.with_changes(|mutator| {
+            let range = mutator
```
techraed, line 1147: just for consistency: either move it out of `with_changes` or insert the same `CheckedRange` instantiation call into `with_changes` for `send_push_input`

core-processor/src/ext.rs
```diff
@@ -1016,18 +1205,21 @@ impl Externalities for Ext {
...
-        self.charge_gas_if_enough(
-            self.context
-                .costs
-                .syscalls
-                .gr_read_per_byte
-                .cost_for(len.into()),
-        )?;
-        PayloadSliceLock::try_new((at, end), &mut self.context.message_context)
-            .ok_or_else(|| FallibleExecutionError::ReadWrongRange.into())
+        self.with_changes(|mutator| {
```
techraed, line 1208: it seems this approach with `with_changes` won't work for `gr_read` syscall. It'd discuss that in pm

core-processor/src/ext.rs
```diff
@@ -474,7 +726,40 @@ impl BackendExternalities for Ext {
 }
 
 impl Ext {
-    fn check_message_value(&mut self, message_value: u128) -> Result<(), FallibleExtError> {
+    fn with_changes<F, R>(&mut self, callback: F) -> Result<R, FallibleExtError>
```
techraed, line 729: seems we can introduce generic impl for `with_changes` by abstracting error type, can't we?

techraed (commented):
Amazing! Great approach! :fire:

Unfortunately, can't see any tests. We definitely need them for this PR. The tests approach is desribed in the original issue #3881. Must state that for all the syscalls and their implementations on the `core-processor` level the mutations must be performed after all the checks (same quideline for extrinsics as well).

breathx (changes requested):
Waiting tests for mutator as discussed personally before re-review

techraed (approved):

core-processor/src/ext.rs
```diff
@@ -474,7 +724,40 @@ impl BackendExternalities for Ext {
...
+    where
+        F: FnOnce(&mut ExtMutator) -> Result<R, FallibleExtError>,
+    {
+        let mut changes = ExtChanges::new(self);
```
breathx, line 731: repeated code may be flattened in 1 fn reused under 3 different signatures

core-processor/src/ext.rs
```diff
@@ -331,6 +331,256 @@ impl<Context> GrowHandler<Context> for LazyGrowHandler {
...
+        Ok(())
+    }
+
+    fn apply(self) {
+        if let Some(reservation) = self.changes.reservation_to_mark.take() {
+            let result = self.ext.context.gas_reserver.mark_used(reservation);
+            debug_assert!(result.is_ok());
+        }
+
+        core::mem::swap(
```
breathx, line 567: it could be directly set, isn't it?

core-processor/src/ext.rs
```diff
@@ -331,6 +331,256 @@ impl<Context> GrowHandler<Context> for LazyGrowHandler {
...
+/// Used to atomically update `Ext` which prevents some errors
+/// when data was updated but operation failed.
+///
+/// Copies some counters into itself and performs operations on them and
+/// incrementally builds list of changes.
+///
+/// Changes are applied after operation is completed
+struct ExtMutator<'a, 'b> {
+    ext: &'a mut Ext,
+    changes: &'b mut ExtChanges,
```
breathx, line 343: can't see any reason not to have changes right in this struct

breathx (changes requested):
Looks great!

Requesting changes only to resolve opened within the review discussions I'm worried of.

Could you also please check logic of `Ext::send`, because it's default-implemented in externalities trait by calling init + commit. I think it won't work with current solution, so it needs to figure out if it's critical atm or we can merge this as is and open an asap issue further 

Also, I'd rather prefer "overlay" naming instead of "mutator"

breathx:
@ark0f could you pls soft-check it for crucial bugs maybe?

playX18:
> Looks great!
> 
> Requesting changes only to resolve opened within the review discussions I'm worried of.
> 
> Could you also please check logic of `Ext::send`, because it's default-implemented in externalities trait by calling init + commit. I think it won't work with current solution, so it needs to figure out if it's critical atm or we can merge this as is and open an asap issue further
> 
> Also, I'd rather prefer "overlay" naming instead of "mutator"

I would merge this asap since this commit here is for a long time already and then look into `Ext::send` in a separate issue just like getting rid of `clone` on counters.

As for overlay vs mutator I feel like mutator states the purpose better and fits Rust memory semantics too: there's always only one mutator and it modifies the state while overlay has no such connotation

core-processor/src/ext.rs
```diff
@@ -334,6 +334,237 @@ impl<Context> GrowHandler<Context> for LazyGrowHandler {
...
 }
 
+/// Used to atomically update `Ext` which prevents some errors
+/// when data was updated but operation failed.
+///
+/// Copies some counters into itself and performs operations on them and
+/// incrementally builds list of changes.
+///
+/// Changes are applied after operation is completed
+struct ExtMutator<'a> {
```
ark0f, line 344: `#[must_use]` would be nice

core-processor/src/ext.rs
```diff
@@ -334,6 +334,237 @@ impl<Context> GrowHandler<Context> for LazyGrowHandler {
...
+                    self.reservation_to_mark.is_none(),
+                    "attempt to mark another reservation"
+                );
+                self.reservation_to_mark = Some(reservation_id);
+                Ok(())
+            }
+        } else {
+            Err(ReservationError::InvalidReservationId)
+        }
+    }
```
ark0f, line 494: Duplicate of important logic of `GasReserver`. I'd create almost the same common function `check_used`, so `GasReserver::mark_used` will just set flag to `true` and `ExtMutator::mark_reservation_used` will set `reservation_to_mark`

ark0f (changes requested):

core/src/reservation.rs
```diff
@@ -254,6 +254,20 @@ impl GasReserver {
...
+    pub fn check_used(&mut self, id: ReservationId) -> Result<bool, ReservationError> {
+        if let Some(
+            GasReservationState::Created { used, .. } | GasReservationState::Exists { used, .. },
+        ) = self.states.get_mut(&id)
+        {
+            Ok(*used)
+        } else {
+            Err(ReservationError::InvalidReservationId)
+        }
+    }
```
ark0f, line 269: I actually expected something like this:
```suggestion
    pub fn check_not_used(&mut self, id: ReservationId) -> Result<&mut bool, ReservationError> {
        if let Some(
            GasReservationState::Created { used, .. } | GasReservationState::Exists { used, .. },
        ) = self.states.get_mut(&id)
        {
            if *used {
            	return Err(ReservationError::InvalidReservationId)
            } else {
	        Ok(used)
	    }
        } else {
            Err(ReservationError::InvalidReservationId)
        }
    }
```
So function looks exactly like original `mark_used` except it doesn't set flag.

So function in `ExtMutator` becomes:
```rust
    fn mark_reservation_used(
        &mut self,
        reservation_id: ReservationId,
    ) -> Result<(), ReservationError> {
        let _used = self.ext.context.gas_reserver.check_not_used(reservation_id)?;
	self.reservation_to_mark = Some(reservation_id);
    }
```

And function in `GasReserver` becomes:
```rust
pub fn mark_used(&mut self, id: ReservationId) -> Result<(), ReservationError> {
	let used = self.check_not_used(reservation_id)?;
	*used = true;
	Ok(())
}
```

ark0f (changes requested):

core/src/reservation.rs
playX18: I suppose I implemented the suggestions too literally. Pushed the proper clean change

core/src/reservation.rs
```diff
@@ -239,15 +239,23 @@ impl GasReserver {
     /// for sending a new message from execution of `message_id`
     /// of current gas reserver.
     pub fn mark_used(&mut self, id: ReservationId) -> Result<(), ReservationError> {
+        let used = self.check_not_used(id)?;
+        *used = true;
+        Ok(())
+    }
+
+    /// Check if reservation is used.
```
ark0f, line 247:
```suggestion
    /// Check if reservation is not used.
```

ark0f (approved):
👌

ark0f (approved):
👌

core-processor/src/ext.rs
```diff
@@ -1043,31 +1233,38 @@ impl Externalities for Ext {
         amount: u64,
         duration: u32,
     ) -> Result<ReservationId, Self::FallibleError> {
-        self.charge_gas_if_enough(self.context.message_context.settings().reservation_fee)?;
+        self.with_changes(|mutator| {
```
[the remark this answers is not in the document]
techraed: > cause in caller code we have references for ext and here are created separate ones

Sorry, didn't catch you. Lock and unlock does not consume any resources that must be returned. The issue we discussed once about gr_read was agreed to be solved in a separate PR

breathx (approved):

PR #3908, opened 2024-04-20
feat(runtime): Account for programs locked balance

ekovalev:
A necessary prerequisite for the staking builtin actor to be added. Being able to participate in staking for programs would mean they'll have part of their balances locked while bonding funds. This hasn't been accounted for so far - all checks of programs' funds availability have been based on their `free_balance` which can produce incorrect estimations and lead to errors in transfers.

### DIFF
--- a/pallets/gear-bank/src/lib.rs
+++ b/pallets/gear-bank/src/lib.rs
@@ -28,7 +28,7 @@ mod tests;
 
 pub use pallet::*;
 
-use frame_support::traits::{Currency, StorageVersion};
+use frame_support::traits::{fungible, tokens::Provenance, Currency, StorageVersion};
 
 #[macro_export]
 macro_rules! impl_config {
@@ -57,14 +57,14 @@ pub mod pallet {
     use frame_support::{
         ensure,
         pallet_prelude::{StorageMap, StorageValue, ValueQuery},
-        sp_runtime::{traits::CheckedSub, Saturating},
-        traits::{ExistenceRequirement, Get, ReservableCurrency, WithdrawReasons},
+        sp_runtime::Saturating,
+        traits::{ExistenceRequirement, Get, ReservableCurrency},
         Identity,
     };
     use pallet_authorship::Pallet as Authorship;
     use parity_scale_codec::{Decode, Encode, MaxEncodedLen};
     use scale_info::TypeInfo;
-    use sp_runtime::traits::Zero;
+    use sp_runtime::{traits::Zero, ArithmeticError, DispatchError, TokenError};
 
     // Funds pallet struct itself.
     #[pallet::pallet]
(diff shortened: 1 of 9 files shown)


ukint-vs (approved):

pallets/gear-bank/src/lib.rs
```diff
@@ -175,11 +179,17 @@ pub mod pallet {
...
+            <CurrencyOf<T> as fungible::Inspect<_>>::can_deposit(
+                &bank_address,
+                value,
+                Provenance::Extant,
+            )
+            .into_result()
+            .map_err(|e| match e {
+                DispatchError::Token(TokenError::BelowMinimum) => Error::<T>::InsufficientDeposit,
+                DispatchError::Arithmetic(ArithmeticError::Overflow) => Error::<T>::Overflow,
+                _ => unreachable!("Unexpected error: {e:?}"),
```
breathx, line 191: what cases could cause this branch?

pallets/gear/src/manager/mod.rs
```diff
@@ -377,7 +383,15 @@ where
         ProgramStorageOf::<T>::remove_program_pages(program_id, memory_infix);
 
         let program_account = program_id.cast();
-        let balance = CurrencyOf::<T>::free_balance(&program_account);
+
+        // Only `reducible_balance` is allowed to be transferred out, even if a part of the
```
breathx, line 387: does reducible balance include ED?

pallets/gear/src/tests.rs
```diff
@@ -6261,6 +6264,83 @@ fn exit_locking_funds() {
...
+            value,
+            false,
+        ));
+
+        run_to_next_block(None);
+
+        // free balance includes the ED
+        assert_eq!(CurrencyOf::<Test>::free_balance(program_account), value);
+
+        // reducible balance for preserved account doesn't include the ED
```
breathx, line 6301: program should be able to operate of the ED.
e.g. 
ed = 10
program has 15, while 2 is frozen
program must be able to send 13

breathx: test it pls

breathx (changes requested):

pallets/gear/src/manager/mod.rs
ekovalev: Depends on `Preservation` requirement. If `Preservation::Expendable` is supplied then the account is allowed to die therefore the ED is included in the `reducible_balance`. Otherwise it's not

pallets/gear/src/tests.rs
ekovalev: if the idea is to always keep a program's account alive then it won't be able to touch the ED (provided we get the `reducible_balance` with `Preservation::Preserve` parameter. Otherwise we can pass the `Preservation::Expandable` as the preservation requirement, thereby allowing the program's account to be wiped as dust in case the balance drops below the ED

pallets/gear-bank/src/lib.rs
ekovalev: In this case only these two are possible. We just need to account for other potential `DispatchError` variants to make it compile.

breathx:
Blocked by #3924

pallets/gear-bank/src/lib.rs
NikVolf: Please no

Other variants of `DispatchError` can be added and this will lead to panic, even if it is impossible now

Please make exhausted match or even better return error and panic in calling code (if absolutely unavoidable)

breathx:
Implemented inside #3924 solution

PR #3909, opened 2024-04-21
feat(gcli): introduce command config

### LINKED ISSUE #3896: Introduce command config for `gcli`
### Problem to Solve

the `--endpoint` flag of `gcli` is hard to use, save the config of gcli globally with static config file

### Possible Solution

```
$ gcli config --set-url dev,mainnet,localnet(9944),<customized-value>
```

Interact with `gring` is considerable as well,

```
$ gcli config --use my-second-keypair # name of the keypair
```

### Notes

_No response_

clearloop:
Resolves #3896 

- [x] remove outdated keystore
- [x] introduce command config
- [x] get/set subcommands for command config

```
Gear command line configuration

Usage: gcli config [OPTIONS] <COMMAND>

Commands:
  set   Set a config setting
  get   Get current config settings
  help  Print this message or the help of the given subcommand(s)

Options:
  -u, --url <URL_OR_MONIKER>  URL for Solana's JSON RPC or moniker [possible values: mainnet, testnet, localhost]
  -h, --help                  Print help

```

@gear-tech/dev


gcli/src/cmd/config.rs
```diff
@@ -0,0 +1,157 @@
...
+const CONFIG_PATH: &str = ".config/vara/config.toml";
+
+/// Gear command line configuration
+#[derive(Clone, Debug, Default, Parser, Serialize, Deserialize)]
+pub struct Config {
+    /// Config actions
+    #[clap(subcommand)]
+    #[serde(skip)]
+    pub action: Action,
+    /// URL for Solana's JSON RPC or moniker
```
StackOverflowExcept1on, line 35:
```suggestion
    /// URL for Vara's JSON RPC or moniker
```

gcli/src/cmd/config.rs
```diff
@@ -0,0 +1,158 @@
...
+            }
+            localhost
+                if localhost == "localhost"
+                    || ignore_case && localhost.to_lowercase() == "localhost" =>
+            {
+                Self::Localhost
+            }
+            _ => Self::Custom(Url::parse(input).map_err(|_| input.to_string())?),
+        })
+    }
```
StackOverflowExcept1on, line 157:
```suggestion
    fn from_str(input: &str, ignore_case: bool) -> Result<Self, String> {
        let input = if ignore_case {
            Cow::Owned(input.to_lowercase())
        } else {
            Cow::Borrowed(input)
        };

        Ok(match input.as_ref() {
            "mainnet" => Self::Mainnet,
            "testnet" => Self::Testnet,
            "localhost" => Self::Localhost,
            input => Self::Custom(Url::parse(input).map_err(|_| input.to_string())?),
        })
    }
```

gcli/src/cmd/mod.rs
```diff
@@ -128,7 +132,17 @@ impl App for Opt {
     }
 
     fn endpoint(&self) -> Option<String> {
```
StackOverflowExcept1on, line 134: there is also same func in `app.rs`

gcli/src/app.rs
```diff
@@ -90,6 +90,16 @@ pub trait App: Parser + Sync {
     /// Exec program from the parsed arguments.
     async fn exec(&self) -> anyhow::Result<()>;
 
+    /// Get gear api without signing in with password.
+    async fn api(&self) -> anyhow::Result<GearApi> {
+        let endpoint = self.endpoint().clone();
+        let timeout = self.timeout();
+        Api::new_with_timeout(endpoint.as_deref(), Some(timeout))
+            .await
+            .map(Into::into)
```
StackOverflowExcept1on, line 99: If I understand correctly, then `.map(Into::into)` here is Alice’s account. Maybe we should use the primary address from gring?

clearloop: Makes sense, this is currently a workaround for adapting that `gclient` force us using a signer but not just a storage api, so the signer discover logic here should never failed

Oh I forgot `gcli info balance` should print the signer's balance

gcli/src/cmd/config.rs
```diff
@@ -16,21 +16,83 @@
...
+    /// just have one config option.
+    pub fn exec(&self) -> Result<()> {
+        if self.action == Action::Set {
+            self.write(None)?
+        }
+
+        println!(
+            "{}: {}",
+            "RPC URL".bold(),
+            self.url.clone().unwrap_or_default().as_ref()
```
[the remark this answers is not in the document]
clearloop: Agree, but atm, this is tricky in `clap` )) 

if the input option is not `bool`, we must follow an argument after it

```
$ gcli config get --url
error: a value is required for '--url <URL_OR_MONIKER>' but none was supplied
  [possible values: mainnet, testnet, localhost]

For more information, try '--help'.
```

Let me try to override the same option in the subcommand `get`

gcli/src/cmd/config.rs
clearloop: interesting, I didn't know we can just `Cow` like this before

gcli/src/app.rs
clearloop: The problem here is that we have to input `password` on `gcli info balance`, need to refactor on this kind of needs that users can run commands without inputing the `password`

gcli/src/cmd/config.rs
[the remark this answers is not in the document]
clearloop: on the other case, `config get` should return the entire config as well

gcli/src/app.rs
clearloop: I'm thinking of introducing a method reading the address of the primary account from `gring`

gcli/src/cmd/config.rs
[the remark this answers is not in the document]
clearloop: ddcaeb8

gcli/src/cmd/config.rs
clearloop: ddcaeb8

gcli/src/app.rs
clearloop: f606e20

StackOverflowExcept1on (approved):

PR #3910, opened 2024-04-21
fix(pallet-gear): don't charge gas in benchmarks

StackOverflowExcept1on:
`CustomConstantCostRules::default()` will add some extra overhead in benchmarks

https://github.com/gear-tech/gear/blob/8ba3c6996be2c4fd060f2d9f16b292c72b201f2c/core/src/code/mod.rs#L274-L280

### DIFF
--- a/pallets/gear/src/lib.rs
+++ b/pallets/gear/src/lib.rs
@@ -550,12 +550,13 @@ pub mod pallet {
             value: BalanceOf<T>,
         ) -> DispatchResultWithPostInfo {
             use gear_core::code::TryNewCodeConfig;
+            use gear_wasm_instrument::gas_metering::CustomConstantCostRules;
 
             let who = ensure_signed(origin)?;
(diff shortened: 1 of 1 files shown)


ukint-vs (approved):
Nice catch

StackOverflowExcept1on:
> Could you provide weights changes info please?

This shouldn't affect the benchmarks much. In the compiled code you will have 1000 syscalls and `call $gas_charge` at the top of `$init` function.

grishasobol:
> > Could you provide weights changes info please?
> 
> This shouldn't affect the benchmarks much. In the compiled code you will have 1000 syscalls and `call $gas_charge` at the top of `$init` function.

I guess in `handle` function. Nevertheless, I hope you will provide an evidence that weights has not been changed.

breathx (approved):
Agree on checking weights just to know

PR #3919, opened 2024-04-24
refactor(workspace): move types to `gprimitives` crate

### LINKED ISSUE #2944: Introduce `gstd-primitives` crate
Details: https://github.com/gear-tech/gear/issues/2944#issuecomment-2079421129

StackOverflowExcept1on:
Resolves #2944

Motivation: Mainly this PR is useful for [`I/O crates`](https://github.com/gear-foundation/dapp-template/blob/master/io/src/lib.rs). If user wants to declare I/O types that depend on `gstd`, `gprimitives = { workspace = true, features = ["codec"] }` should be used instead.

We also encountered linking issue on Windows at hackathon. This is because MSVC linker requires declaration of all imported functions from `gsys`.


gcore-primitives/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+)]
+#[as_ref(forward)]
+#[as_mut(forward)]
+#[cfg_attr(feature = "codec", derive(TypeInfo, Encode, Decode, MaxEncodedLen), codec(crate = scale))]
+pub struct ActorId([u8; 32]);
+
+declare_primitive!(ActorId);
+
+//TODO: find out why this is needed
+impl From<u64> for ActorId {
```
StackOverflowExcept1on, line 128: we keep it for gtest or what?

gstd/src/common/errors.rs
```diff
@@ -115,6 +115,12 @@ impl From<CoreError> for Error {
     }
 }
 
+impl From<TryFromSliceError> for Error {
+    fn from(_: TryFromSliceError) -> Self {
+        Self::Convert(TryFromSliceError::REASON)
```
StackOverflowExcept1on, line 120: for backwards compatibility with `gstd::Error`, but in general we could break it

breathx (changes requested):
Need to sync on impl details since this isn't resolve initial issue (motivation of implementing separate crate)

StackOverflowExcept1on:
Basically this PR is ready and I would like to see review from @breathx and @DennisInSky.

So, the `gprimitives` crate does not require `alloc` and `std` now. The internal implementation is hidden, but the user can mutate the primitives via `as_mut()`. Unfortunately, `cargo` works in such a way that the `gprimitives` features will be combined from `gcore` and `gstd`. This means that we cannot make any exceptions with public fields for `gcore`, i.e. the fields will also be public for `gstd`. The primitives have `fmt::Display` and `fmt::Debug` implementations that output bytes as hex (no longer requires an alloactor for this). Only `ActorId` and `CodeId` can be parsed from `&str` in hex format. If string starts with `0x`, then it is considered as hex. `ActorId::from_str` can also parse SS58 addresses from `&str`. `ActorId` can be converted to vara address via `id.to_ss58check()`. All conversion errors will be of type `gprimitives::ConversionError` and they are compatible with `gstd::errors::Error::Convert(ConversionError)`. I did not add the `mock` feature and all types can be converted from `u64`. This is mainly due to the fact that it seems inconvenient for writing pallet tests, because... `gprimitives = { workspace = true, features = ["mock"] }` does not seem obvious to the developer. The `gsdk` and `gclient` libraries compile successfully and their generated types can be converted to `gprimitives` types via `From` implementations. So the `serde` feature was not added at this stage.

core/src/ids.rs
```diff
@@ -18,270 +18,186 @@
...
+/// Program identifier.
+pub type ProgramId = ActorId;
 
-        impl From<$name> for $crate::ids::Hash {
-            fn from(val: $name) -> $crate::ids::Hash {
-                val.0
-            }
-        }
+pub mod prelude {
+    //! The purpose of this module is to make it easier to import `gprimitives` extensions.
```
ark0f, line 51: Usually `prelude` is used in examples. Also, any IDE provides auto-import feature. So I'm not sure this module really should exist

utils/ss58/src/lib.rs
ark0f, line 1: Should be refactored in another PR

gcore/src/exec.rs
```diff
@@ -115,7 +115,7 @@ pub fn block_timestamp() -> u64 {
 /// ```
 pub fn reply_deposit(message_id: MessageId, amount: u64) -> Result<()> {
     let mut error_code = 0u32;
-    unsafe { gsys::gr_reply_deposit(message_id.as_ptr(), amount, &mut error_code) };
+    unsafe { gsys::gr_reply_deposit(message_id.as_ref().as_ptr() as _, amount, &mut error_code) };
```
ark0f, line 118: Consider to introduce `as_ptr` for `gprimitives` types

gprimitives/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+/// struct. The source `ActorId` for a message being processed can be obtained
+/// using `msg::source()` function. Also, each send function has a target
+/// `ActorId` as one of the arguments.
+#[derive(Clone, Copy, Default, Hash, Ord, PartialEq, PartialOrd, Eq, From, Into, AsRef, AsMut)]
+#[as_ref(forward)]
+#[as_mut(forward)]
+#[cfg_attr(feature = "codec", derive(TypeInfo, Encode, Decode, MaxEncodedLen), codec(crate = scale))]
+pub struct ActorId([u8; 32]);
+
+macros::impl_primitive!(new zero into_bytes from_u64 from_h256 try_from_slice display debug, ActorId);
```
ark0f, line 86: Looks like there is no need to implement different sets of functions for each ID

gprimitives/src/lib.rs
```diff
@@ -0,0 +1,184 @@
...
+    primitive_types::H256,
+    scale_info::{
+        scale::{self, Decode, Encode, MaxEncodedLen},
+        TypeInfo,
+    },
+};
+
+/// The error type returned when conversion fails.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Display)]
+pub enum ConversionError {
```
ark0f, line 47: Too generic name. Consider
```suggestion
pub enum GPrimitiveError {
```

ark0f (changes requested):

core/src/ids.rs
StackOverflowExcept1on: Well, this seems more convenient than `use gear_core::ids::{ActorId, ActorIdExt, CodeId, CodeIdExt}` (also `cargo fmt` will sort alphabetically...)

gcore/src/exec.rs
StackOverflowExcept1on: I see no reason to provide raw ptrs API foruserspace library

StackOverflowExcept1on: We can introduce `MessageIdExt` for `gcore` with `as_ptr` and `as_mut_ptr`.

gprimitives/src/lib.rs
StackOverflowExcept1on: What's the point of `ReservationId::zero()` then? Some methods are useless because you can only get this type using `ReservationId::reserve()`.

utils/ss58/src/lib.rs
StackOverflowExcept1on: We want `gprimitives` to require neither std nor allocator. There was also an idea that `ActorId` could be formatted as Vara/Polkadot address via `debug!("{actor_id:+}");`

breathx (changes requested):
Except comments, awesome work done 👍🏻

utils/ss58/src/lib.rs
breathx: basically, we could impl the formatter (+ these refactoring) in separate pr, that's true
but 1 file with 300 lines looks normal within the PR, not that much

core/src/ids.rs
breathx: im ok having there prelude, but as I've said before to @StackOverflowExcept1on, for me importing in caller code `use gear_core::ids::*` looks ok due to module's basic utility, so exts can be implemented 1 module upper

DennisInSky:
@StackOverflowExcept1on, @breathx what do you guys think about re-exporting `H256`, `U256` from the `primitive-types` crate so I don't need to reference it in sails directly?

core/src/ids.rs
```diff
@@ -18,270 +18,206 @@
...
+            format!("{id:42}"),
+            "5EU7B2s4m2XrgSbUyt8U92fDpSi2EtW3Z3kKwUW4drZ1KAZD"
+        );
+        // Alternate formatter.
+        assert_eq!(
+            format!("{id:#}"),
+            "ActorId(0x6a519a19ffdfd8f45c310b44aecf156b080c713bf841a8cb695b0ea5f765ed3e)"
+        );
+        // Alternate formatter with precision 2.
+        assert_eq!(format!("{id:#.2}"), "ActorId(0x6a51..ed3e)");
```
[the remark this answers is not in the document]
StackOverflowExcept1on: Can we just return `fmt::Error` if more than one formatting is specified?

gcore/src/lib.rs
```diff
@@ -74,9 +74,7 @@ pub mod exec;
 pub mod msg;
 pub mod prog;
 pub use gear_stack_buffer as stack_buffer;
-
-mod general;
-pub use general::*;
+pub use gprimitives::{ActorId, CodeId, MessageHandle, MessageId, ReservationId, Ss58Address};
```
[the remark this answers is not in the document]
StackOverflowExcept1on: because we don't want re-export ConversionError here

breathx:
> @StackOverflowExcept1on, @breathx what do you guys think about re-exporting `H256`, `U256` from the `primitive-types` crate so I don't need to reference it in sails directly?

Exactly 4 comments above yours in this PR: https://github.com/gear-tech/gear/pull/3919#discussion_r1603726637 :D

gprimitives/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+
+use core::{
+    fmt,
+    str::{self, FromStr},
+};
+use derive_more::{AsMut, AsRef, Display, From, Into};
+use gear_ss58::RawSs58Address;
+#[cfg(feature = "codec")]
+use {
+    primitive_types::H256,
```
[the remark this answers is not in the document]
breathx: Just opened an issue https://github.com/gear-tech/gear/issues/3973

gstd/src/msg/basic.rs
```diff
@@ -307,7 +301,7 @@ pub fn signal_code() -> Result<Option<SignalCode>> {
 /// }
 /// ```
 pub fn id() -> MessageId {
-    gcore::msg::id().into()
+    gcore::msg::id()
```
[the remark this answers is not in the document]
StackOverflowExcept1on: 15331905a45887e8ee3e16a79ce339f7744abd11

gstd/src/exec/basic.rs
```diff
@@ -71,7 +71,7 @@ pub fn reply_deposit(message_id: MessageId, amount: u64) -> Result<()> {
 /// }
 /// ```
 pub fn exit(inheritor_id: ActorId) -> ! {
```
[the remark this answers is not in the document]
StackOverflowExcept1on: 15331905a45887e8ee3e16a79ce339f7744abd11

core/src/ids.rs
```diff
@@ -18,270 +18,206 @@
...
+            const SALT: &[u8] = b"reservation";
+            hash_of_array([SALT, msg_id.as_ref(), &nonce.to_le_bytes()]).into()
+        }
     }
 }
 
-#[test]
-fn formatting_test() {
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
StackOverflowExcept1on: 6aa7bb1bd7ef1f494e7cf84310333a4addd93e58

core/src/ids.rs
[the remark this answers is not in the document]
StackOverflowExcept1on: 6aa7bb1bd7ef1f494e7cf84310333a4addd93e58

core/src/program.rs
```diff
@@ -140,12 +140,12 @@ mod tests {
     }
 
     #[test]
-    #[should_panic(expected = "Identifier must be 32 length")]
+    #[should_panic]
     /// Test that ProgramId's `from_slice(...)` constructor causes panic
```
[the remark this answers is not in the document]
StackOverflowExcept1on: 6aa7bb1bd7ef1f494e7cf84310333a4addd93e58

gprimitives/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+
+[dependencies]
+derive_more.workspace = true
+parity-scale-codec = { workspace = true, features = ["derive"], optional = true }
+primitive-types = { workspace = true, features = ["scale-info"], optional = true }
+scale-info = { workspace = true, features = ["derive"], optional = true }
+hex.workspace = true
+gear-ss58.workspace = true
+
+[features]
```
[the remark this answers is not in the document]
StackOverflowExcept1on: why do we need serde?

gprimitives/src/lib.rs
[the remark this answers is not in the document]
StackOverflowExcept1on: f7950d54ab6053bbf3a14efc5fedd5a479eb60e0

gprimitives/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+        Ok(actor_id)
+    }
+}
+
+/// Message identifier.
+///
+/// Gear allows users and program interactions via messages. Each message has
+/// its own unique 256-bit id. This id is represented via the `MessageId`
+/// struct. The message identifier can be obtained for the currently processed
+/// message using the `msg::id()` function. Also, each send and reply functions
```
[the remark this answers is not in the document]
StackOverflowExcept1on: 39a57e1bcbf5539d9a7f0e0663a333f5a34171d5

gprimitives/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq, From, Into)]
+#[cfg_attr(feature = "codec", derive(TypeInfo, Encode, Decode, MaxEncodedLen), codec(crate = scale))]
+pub struct MessageHandle(u32);
+
+/// Program (actor) identifier.
+///
+/// Gear allows user and program interactions via messages. Source and target
+/// program as well as user are represented by 256-bit identifier `ActorId`
+/// struct. The source `ActorId` for a message being processed can be obtained
+/// using `msg::source()` function. Also, each send function has a target
```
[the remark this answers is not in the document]
StackOverflowExcept1on: 39a57e1bcbf5539d9a7f0e0663a333f5a34171d5

gprimitives/src/lib.rs
```diff
@@ -0,0 +1,218 @@
...
+
+macros::impl_primitive!(new zero into_bytes from_u64 from_h256 display debug, MessageId);
+
+/// Code identifier.
+///
+/// This identifier can be obtained as a result of executing the
+/// `gear.uploadCode` extrinsic. Actually, the code identifier is the Blake2
+/// hash of the Wasm binary code blob.
+///
+/// Code identifier is required when creating programs from programs (see `prog`
```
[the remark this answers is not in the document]
StackOverflowExcept1on: 39a57e1bcbf5539d9a7f0e0663a333f5a34171d5

core/src/ids.rs
```diff
@@ -18,270 +18,206 @@
...
 
-        impl From<u64> for $name {
-            fn from(v: u64) -> Self {
-                let mut id = Self(Default::default());
-                id.0[0..8].copy_from_slice(&v.to_le_bytes()[..]);
-                id
-            }
-        }
+    /// Program (actor) identifier extension.
+    pub trait ActorIdExt {
```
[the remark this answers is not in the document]
StackOverflowExcept1on: dde1f8ab89d796637a557526b0793f1ab217c20d

gstd/src/common/primitives_ext.rs
```diff
@@ -0,0 +1,88 @@
...
+///     unsafe { RESERVED = Some(reservation_id) };
+/// }
+///
+/// #[no_mangle]
+/// extern "C" fn handle() {
+///     let reservation_id = unsafe { RESERVED.take().expect("Empty `RESERVED`") };
+///     reservation_id.unreserve().expect("Unable to unreserve");
+/// }
+/// ```
+pub trait ReservationIdExt: Sized {
```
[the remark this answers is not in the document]
StackOverflowExcept1on: dde1f8ab89d796637a557526b0793f1ab217c20d

gprimitives/Cargo.toml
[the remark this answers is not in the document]
StackOverflowExcept1on: d84f1c4fd777b24f598c71f10b7476e55006f0bc

gstd/src/msg/encoded.rs
```diff
@@ -58,8 +59,8 @@ use scale_info::scale::{Decode, Encode};
 ///
 /// - [`load_bytes`](super::load_bytes) function returns a payload as a byte
 ///   vector.
-pub fn load<D: Decode>() -> Result<D> {
-    super::basic::with_read_on_stack(|read_result: Result<&mut [u8]>| -> Result<D> {
+pub fn load<D: Decode>() -> GstdResult<D> {
```
breathx, line 62: crate::Result<D> looks better

gprimitives/src/lib.rs
```diff
@@ -216,3 +227,96 @@ impl FromStr for CodeId {
...
+    use core::str::FromStr;
+
+    fn actor_id() -> ActorId {
+        ActorId::from_str("0x6a519a19ffdfd8f45c310b44aecf156b080c713bf841a8cb695b0ea5f765ed3e")
+            .unwrap()
+    }
+
+    #[test]
+    #[should_panic]
+    fn wrong_format1_test() {
```
breathx, line 293: names are unintuitive, please describe test cases

breathx: moreover, why do we return err on this cases instead of supporting it?
if it's temporary, please open an issue

at least case of +.2 looks valid for me

gprimitives/src/lib.rs
```diff
@@ -24,6 +24,7 @@
 #![doc(html_favicon_url = "https://gear-tech.io/favicons/favicon.ico")]
 
 pub use gear_ss58::Ss58Address;
+pub use primitive_types;
```
breathx, line 28: what if explicitly reexport only dedicated and most-used for us types: H160, H256, U256.
Cause gprimitives::primitive_types looks anyway weird

gprimitives/src/lib.rs
```diff
@@ -24,6 +24,7 @@
 #![doc(html_favicon_url = "https://gear-tech.io/favicons/favicon.ico")]
 
 pub use gear_ss58::Ss58Address;
+#[cfg(feature = "codec")]
```
breathx, line 27: why?

gcore/src/utils.rs
```diff
@@ -16,6 +16,23 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+use gprimitives::{ActorId, CodeId, MessageId, ReservationId};
+
+pub trait AsRawPtr: AsRef<[u8]> + AsMut<[u8]> {
```
breathx, line 21: Sealed as well

gprimitives/src/lib.rs
```diff
@@ -162,15 +162,58 @@ impl FromStr for ActorId {
...
+                .map_err(|_| ConversionError::InvalidSs58Address)?
+                .into();
             actor_id.0[..].copy_from_slice(&raw_address);
         }
 
         Ok(actor_id)
     }
 }
 
+#[cfg(feature = "serde")]
```
breathx, line 176: pls impl these for MessageId and CodeId as well
(or as usual open an issue for faster merging of this)

core/src/ids.rs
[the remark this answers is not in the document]
breathx: Im ok with formatting with two params, to be solved in future, I guess

breathx (approved):

breathx:
Resolve comments above and good to go from my side 👍🏻 

ptal once again @ark0f

gstd/src/msg/encoded.rs
StackOverflowExcept1on: ok, but it will look like `crate::errors::Result`

gprimitives/src/lib.rs
StackOverflowExcept1on: Should it be in root or in module?

breathx: up to you
I'd go with root (like it's primitives for us as well)

gcore/src/utils.rs
StackOverflowExcept1on: Currently these are primitive extensions that add methods `as_ptr()`, `as_mut_ptr()`, which are only available inside `gcore` since `gcore::utils` is private. I think since `gcore` already solves problem of moving from raw pointers to rust types and therefore there is no point in making it public and `: Sealed`.

StackOverflowExcept1on: I'll mark these extensions as `pub(crate)` for better readability

breathx: Adjusting vis is even better

ark0f (approved):
Awesome 👌

gprimitives/src/lib.rs
StackOverflowExcept1on: should we also provide serde for H256, H160...?

breathx: Wasn't it done by default?

breathx: in origin crate

breathx:
Good to merge from my side @StackOverflowExcept1on

PR #3922, opened 2024-04-25
refactor(core): Context-dependent memory

ark0f:
Make it closer to actual executor API which is more flexible


breathx (approved):
imho, one dummy ref-wrap was more comfortable than annoying "undefined" context everywhere

PR #3927, opened 2024-04-29
refactor(core): change prop-test's strategy Value from tuple to `struct`

ByteNacked:
Resolves ##3906 .

Change prop-test's strategy Value from tuple to  `struct MemoryParams`.

@reviewer-or-team


core/src/memory.rs
```diff
@@ -874,6 +874,14 @@ mod tests {
...
                     )
                 })
+                .prop_map(
+                    |(max_pages, mem_size, static_pages, allocations)| MemoryParams {
+                        max_pages,
+                        mem_size,
+                        static_pages,
+                        allocations,
+                    },
+                )
```
ByteNacked, line 884: It's not possible to make this mapping inside the `prop_flat_map` call because it operates with strategies and not values, which these strategies map to.

core/src/memory.rs
```diff
@@ -909,7 +917,7 @@ mod tests {
             ) {
                 let _ = env_logger::try_init();
 
-                let (max_pages, mem_size, static_pages, allocations) = mem_params;
+                let MemoryParams{max_pages, mem_size, static_pages, allocations} = mem_params;
```
grishasobol, line 920: I cannot understand is it ok for `fmt` to be without space `{max_pages`?

grishasobol (approved):

ukint-vs (approved):

core/src/memory.rs
ByteNacked: I guess it's ok, if I understood you correctly.

PR #3933, opened 2024-04-30
fix toolchain matching and remove redundant cargo lock

grishasobol:


utils/wasm-builder/src/cargo_toolchain.rs
```diff
@@ -66,7 +66,7 @@ impl Toolchain {
         static TOOLCHAIN_CHANNEL_RE: Lazy<Regex> = Lazy::new(|| {
             // This regex is borrowed from the rustup code and modified (added non-capturing groups)
```
DennisInSky, line 67: Probably you have to adjust this comment as well as it is no longer a borrowed pattern. Unless they have updated it in rustup as well ;-)

DennisInSky (approved):

clearloop (approved):

utils/wasm-builder/src/cargo_toolchain.rs
grishasobol: done

PR #3944, opened 2024-05-06
refactor(runtime): Dump constants (existential deposit; min voucher duration)

breathx:


runtime/vara/src/constants.rs
```diff
@@ -27,7 +27,7 @@ pub mod currency {
     pub const ECONOMIC_CENTIUNITS: Balance = ECONOMIC_UNITS / 100;
 
     /// The existential deposit.
-    pub const EXISTENTIAL_DEPOSIT: Balance = 10 * UNITS; // 10 Vara
+    pub const EXISTENTIAL_DEPOSIT: Balance = UNITS; // 1 Vara
```
ukint-vs, line 30:
```suggestion
    pub const EXISTENTIAL_DEPOSIT: Balance = 1 * UNITS; // 1 Vara
```
but clippy will warn

ukint-vs (approved):

runtime/vara/src/constants.rs
breathx: yeah, don't like an idea to ignore clippy when its not necessary, moreover there is a comment right in place

PR #3948, opened 2024-05-07
feat(node): Extend bls381 builtin actor

gshep:
- add `aggregate_g1` method for the cases with large number of G1-points to aggregate
- `map_to_g2affine` method for mapping a message to G2Affine-point for signing/verifying
- related test cases

@gear-tech/dev


Cargo.toml
```diff
@@ -175,6 +175,7 @@ ark-serialize = { version = "0.4", default-features = false }
 ark-ec = { version = "0.4.2", default-features = false }
 ark-ff = { version = "0.4.2", default-features = false }
 ark-scale = { version = "0.0.12", default-features = false }
+sha2 = { version = "0.10.8", default-features = false, features = ["oid"] }
```
breathx, line 178: lets avoid any features here mb?

node/service/src/client.rs
```diff
@@ -57,13 +57,15 @@ impl sc_executor::NativeExecutionDispatch for VaraExecutorDispatch {
         gear_ri::gear_ri::HostFunctions,
         gear_ri::sandbox::HostFunctions,
         sp_crypto_ec_utils::bls12_381::host_calls::HostFunctions,
+        gear_ri::gear_bls_12_381::HostFunctions,
```
breathx, line 60: To be consistent in naming with sp_crypto
```suggestion
        gear_ri::gear_bls12_381::HostFunctions,
```

runtime-interface/src/lib.rs
```diff
@@ -301,3 +313,47 @@ pub trait GearDebug {
             .as_nanos()
     }
 }
+
+#[runtime_interface]
+pub trait GearBls12_381 {
+    /// Aggregate provided G1-points. Useful for cases with hundreds or more items.
+    /// Accepts scale-encoded `ArkScale<Vec<G1Projective>>`.
+    /// Result is scale-encoded `ArkScale<G1Projective>`. Returns an empty array on failure.
```
breathx, line 321: Why dont we return Result from host func on error?

breathx (changes requested):

runtime-interface/src/lib.rs
breathx: As its done in sp_crypto

node/service/src/client.rs
gshep: This is an auto-generated name by `#[runtime_interface]` macro

ekovalev (approved):

pallets/gear-builtin/src/benchmarking.rs
```diff
@@ -266,31 +266,28 @@ benchmarks! {
...
 
-        let mut _result = vec![];
+        let mut _result = Err(0);
     }: {
         _result = gear_runtime_interface::gear_bls_12_381::map_to_g2affine(&message);
     } verify {
         assert!(
-            ArkScale::<G2Affine>::decode(&mut &_result[..]).is_ok()
+            matches!(
+                _result, Ok(result) if ArkScale::<G2Affine>::decode(&mut &result[..]).is_ok()
```
breathx, line 289: couldn't be just unwrapped?

breathx (approved):

breathx:
Well done!
I'd like to see @grishasobol's review as well please

pallets/gear-builtin/src/benchmarking.rs
```diff
@@ -254,6 +254,42 @@ benchmarks! {
...
+        );
+    }
+
+    bls12_381_map_to_g2affine {
+        let c in 0 .. MAX_PAYLOAD_SIZE as u32;
+
+        let message = vec![1u8; c as usize];
+
+        let mut _result = Err(0);
+    }: {
```
grishasobol, line 284: Better to use per_kb here like in other places. Also using per kb reduces weight round fluctuation. For example now you have 565 per byte and for kb it can be 565921.

runtime-interface/src/lib.rs
```diff
@@ -47,6 +56,9 @@ pub use gear_sandbox::sandbox;
 
 const _: () = assert!(core::mem::size_of::<HostPointer>() >= core::mem::size_of::<usize>());
 
+// Domain Separation Tag for signatures on G2.
+pub const DST_G2: &[u8] = b"BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_POP_";
```
grishasobol, line 60: Are you sure this value will not be changed in future? I mean may be better to provide it from runtime

grishasobol (approved):

clearloop:
the dispatch label got some issues in the previous version, please merge main or push an empty commit to reset the state

runtime-interface/src/lib.rs
gshep: Nope, this constant is from the milagro_bls library. And if we ever need to change it we will introduce just a new function.

pallets/gear-builtin/src/benchmarking.rs
gshep: create new issue for that - https://github.com/gear-tech/gear/issues/4013

PR #3949, opened 2024-05-07
feat(pallet-gear): Introduce extrinsic to transfer value from terminated/exited program

ark0f:


pallets/gear/src/internal.rs
```diff
@@ -903,22 +906,40 @@ where
...
                 break;
             }
 
-            inheritor = id
+            if holders.len() == max_depth {
+                break;
+            }
+
+            if !holders.insert(next_inheritor) {
+                log::warn!("Cyclic inheritor detected for {next_inheritor}");
```
grishasobol, line 937: It is normal situation for gear runtime - so, why do we need to `warn` about this?
```suggestion
                log::debug!("Cyclic inheritor detected for {next_inheritor}");
```

grishasobol (approved):
LGTM (have not checked gtest)

breathx (changes requested):
Revert option field in JournalNote. Discussed personally

core-processor/src/processing.rs
```diff
@@ -223,6 +223,7 @@ fn process_error(
     // execution, where `gr_wait` was called.
     if dispatch.context().is_none() && value != 0 {
         // Send back value
+        // TODO: handle sending back
```
breathx, line 226: ?

pallets/gear/src/lib.rs
```diff
@@ -900,28 +902,24 @@ pub mod pallet {
...
-            ProgramStorageOf::<T>::get_program(program_id)
-                .map(|program| {
-                    if let Program::Exited(inheritor) = program {
-                        Some(inheritor)
-                    } else {
-                        None
-                    }
-                })
-                .unwrap_or_default()
+            ProgramStorageOf::<T>::get_program(program_id).and_then(|program| {
```
breathx, line 905: you can flatten this two methods, I guess

pallets/gear/src/lib.rs
```diff
@@ -1648,6 +1646,45 @@ pub mod pallet {
...
         }
+
+        /// Transfers value from terminated or exited program to its living inheritor.
+        ///
+        /// Inheritor can be an active program or original user.
+        ///
+        /// `depth` parameter is how far to traverse to inheritor.
+        #[pallet::call_index(8)]
+        #[pallet::weight(<T as Config>::WeightInfo::transfer_value_to_inheritor(depth.get()))]
+        pub fn transfer_value_to_inheritor(
```
breathx, line 1657: I'd rename this method into something like "unlock_stuck_value" and moved this into pallet_gear_program

pallets/gear/src/lib.rs
```diff
@@ -1648,6 +1646,45 @@ pub mod pallet {
...
+            program_id: ProgramId,
+            depth: NonZeroU32,
+        ) -> DispatchResult {
+            ensure_signed(origin)?;
+
+            let depth = depth.try_into().unwrap_or_else(|e| {
+                unreachable!("NonZeroU32 to NoZeroUsize conversion must be infallible: {e}")
+            });
+            let Some((destination, holders)) = Self::inheritor_for(program_id, depth) else {
+                return Err(Error::<T>::InheritorNotFound.into());
```
breathx, line 1668: the error is incorrect, inheritor couldn't be found only for active programs

pallets/gear/src/manager/mod.rs
```diff
@@ -367,27 +366,14 @@ where
...
+        program: &mut ActiveProgram<BlockNumberFor<T>>,
     ) {
-        ProgramStorageOf::<T>::remove_program_pages(program_id, memory_infix);
-
-        let program_account = program_id.cast();
-        let balance = CurrencyOf::<T>::free_balance(&program_account);
-        if !balance.is_zero() {
-            let destination = Pallet::<T>::inheritor_for(value_destination).cast();
-
-            CurrencyOf::<T>::transfer(
```
breathx, line 383: revert this part.
on exit we transfer balance to destination, but only once
any other balance remaining will be transferred through new extrinsic

pallets/gear/src/lib.rs
```diff
@@ -1648,6 +1646,45 @@ pub mod pallet {
...
+        ///
+        /// Inheritor can be an active program or original user.
+        ///
+        /// `depth` parameter is how far to traverse to inheritor.
+        #[pallet::call_index(8)]
+        #[pallet::weight(<T as Config>::WeightInfo::transfer_value_to_inheritor(depth.get()))]
+        pub fn transfer_value_to_inheritor(
+            origin: OriginFor<T>,
+            program_id: ProgramId,
+            depth: NonZeroU32,
```
breathx, line 1660: I'd avoid this not to mislead caller clients like js code, zero depth is just no-op operation

pallets/gear/src/lib.rs
```diff
@@ -1648,6 +1646,45 @@ pub mod pallet {
 
             Ok(())
         }
+
+        /// Transfers value from terminated or exited program to its living inheritor.
+        ///
+        /// Inheritor can be an active program or original user.
```
breathx, line 1652: inheritor could be anything

breathx: if you run into depth limit, for example

pallets/gear/src/internal.rs
```diff
@@ -903,22 +906,40 @@ where
...
+            if get_inheritor(next_inheritor).is_none() {
                 break;
             }
 
-            inheritor = id
+            if holders.len() == max_depth {
+                break;
+            }
+
+            if !holders.insert(next_inheritor) {
```
breathx, line 936: for cyclic inherenting you can just do some specific logic in a call - leave some commented space there: such funds could be transferred into treasury, for example

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -536,6 +536,47 @@ benchmarks! {
...
+        init_block::<T>(None);
+    }: _(RawOrigin::Signed(caller.clone()), program_id, NonZeroU32::MAX)
+    verify {
+        assert_eq!(
+            CurrencyOf::<T>::free_balance(&caller),
+            minimum_balance * d.unique_saturated_into()
+        );
+
+        for program_id in programs {
+            assert_eq!(CurrencyOf::<T>::free_balance(&program_id), BalanceOf::<T>::zero());
```
breathx, line 576:
```suggestion
            assert!(CurrencyOf::<T>::free_balance(&program_id).is_zero());
```

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -536,6 +536,47 @@ benchmarks! {
...
+        let mut programs = vec![];
+        for i in 0..d {
+            let program_id = benchmarking::account::<T::AccountId>("program", i, 100);
+            programs.push(program_id.clone());
+            let _ = CurrencyOf::<T>::deposit_creating(&program_id, minimum_balance);
+            let program_id = program_id.cast();
+            benchmarking::set_program::<ProgramStorageOf::<T>, _>(program_id, vec![], 1.into());
+
+            ProgramStorageOf::<T>::update_program_if_active(program_id, |program, _bn| {
+                // we set program to terminated state and not exited one because
```
breathx, line 556: why?
basically these branches are equal and only depend on your code here

breathx (changes requested):

pallets/gear/src/lib.rs
ark0f: Check new name which can be more suitable

pallets/gear/src/manager/journal.rs
```diff
@@ -155,22 +158,25 @@ where
...
-                    );
-                }
+                Program::Active(program) => Self::clean_inactive_program(id_exited, program),
                 _ => unreachable!("Action executed only for active program"),
             }
 
             *p = Program::Exited(value_destination);
+
+            // `gr_exit()` argument implies value is sent to destination
+            let id_exited = id_exited.cast();
```
breathx, line 168: this logic should be also applied to failing init, according to what we've been discussing

breathx (changes requested):

pallets/gear/src/lib.rs
```diff
@@ -1648,6 +1634,49 @@ pub mod pallet {
...
+        #[pallet::weight(<T as Config>::WeightInfo::claim_value_to_inheritor(depth.get()))]
+        pub fn claim_value_to_inheritor(
+            origin: OriginFor<T>,
+            program_id: ProgramId,
+            depth: NonZeroU32,
+        ) -> DispatchResult {
+            ensure_signed(origin)?;
+
+            let depth = depth.try_into().unwrap_or_else(|e| {
+                unreachable!("NonZeroU32 to NoZeroUsize conversion must be infallible: {e}")
```
breathx, line 1651:
```suggestion
                unreachable!("NonZeroU32 to NonZeroUsize conversion must be infallible: {e}")
```

breathx (changes requested):
Now for me looks great
Requesting changes only to merge it in correct order to master and kindly as @ekovalev for at least quick look on this PR 
Well done 👍🏻

breathx (approved):

pallets/gear/src/lib.rs
```diff
@@ -1673,6 +1659,58 @@ pub mod pallet {
 
             Ok(())
         }
+
+        /// Transfers value from terminated or exited program to its inheritor.
+        ///
+        /// `depth` parameter is how far to traverse to inheritor.
+        #[pallet::call_index(8)]
+        #[pallet::weight(<T as Config>::WeightInfo::claim_value_to_inheritor(depth.get()))]
```
techraed, line 1667: Correct me if I'm wrong - user will be charged by the _depth.get()_ and compensated later for the actually used gas. Is that right?

So the guideline usage for the `depth` is to set it to `MAX`. Right?
I will explain why I'm asking that. Actually, this extrinsic is a great idea. Depth param gives control over funds in 2 ways:
1. You can hack inheritors cycles (read other comments)
2. You can collect funds only from chosen holders.
But to use these great opportunities one should know the correct `depth` param. I can't see any easy way to understand what is the actual depth or what is the desired depth for user. Are there any ways to get this info?

techraed (changes requested):
please introduce some more tests:
1) a separate test on how to get rid of cycles;
2) test showing an invariant that no `Currency` transfers from the extrinsic fail

pallets/gear/src/lib.rs
breathx: In case of MAX the expected weight of extrinsic would exceed maximum allowed for DispatchType::Normal, so it's better to set something reasonable. For most of cases depth will be eq 1

breathx: Moreover, can't see refund here @ark0f 
Please collect amount of visited accounts and return `Ok(WeightInfo::claim_value_to_inheritor(real_depth).into())` instead of `Ok(().into())`

techraed: > Moreover, can't see refund here @ark0f Please collect amount of visited accounts and return `Ok(WeightInfo::claim_value_to_inheritor(real_depth).into())` instead of `Ok(().into())`

Great that you have mentioned this, as all the tests actually use `MAX`. That would be a wrong guideline for the users

techraed (approved):
the invariant check suggested proptest explained in pm, a separate issue will be created for that.

breathx (approved):

breathx:
@ark0f feel free to merge as soon as branch is updated

PR #3961, opened 2024-05-13
refactor(core): Guarantee system accounts existence for programs

### LINKED ISSUE #3924: Proposal: programs' `frame_system::AccountInfo` permanent existence
Motivation of the PR is to increase transparency of the system and isolate user-space from executable environment's restrictions, with upcoming security improvement and bug-safe space within the substrate code.

Invariant: Each program should have ED locked on its account when active.

Implementation details (TODO):
- [ ] Implement charging from balance within `pallet_gear::upload_program` and `pallet_gear::create_program` extrinsics  
    - [ ] 2 tests
- [ ] Implement charging from program balance on backend level in case of calling `gr_create_program` and further charging  in `JournalNote::StoreNewPrograms` processing (track the case when code not exist, so refund should be performed)
    - [ ] 2 tests
- [ ] In case of calling `gr_exit` this funds should be unlocked and transferred to inheritor
    - [ ] 1 test
- [ ] Remove value limits (`check_value`) for sendings in all send-extrinsics and all backend send-syscalls
    - [ ] 2 tests: extrinsics/syscalls (for syscalls could be implemented in `core-processor/ext` units)
- [ ] Replace all transfers in system to `Existence::AllowDeath` version
- [ ] Implement migrations that transfer ED from treasury for each existing program
    - [ ] try-runtime tests that after migrations all program that are in "Active" status has ED locked.
- [ ] Fix existing tests
___
- [ ] Consider (maybe only open an issue or resolving in another PR) to lock the balance for builtin actors and "pallet_gear_staking_rewards" and "pallet_gear_bank"

ekovalev:
Resolves #3924.

The idea is to separate the notion of value ownership/transfer by actors in Actor space from the underlying system accounts management level (e.g. Substrate's frame-system and balances pallets).
Programs (a.k.a. actors) should not care about their "wallets" implementation details and whether or not they require things like existential deposits to stay alive or have to respect limits on transferred value amounts imposed by the system level. Instead, the protocol must guarantee an invariant that for any active program there always is an account able to store any amount of value at any point throughout the program's lifetime.

This change, once implemented, renders useless numerous checks of a value validity in balances transfers - any value should be valid for a transfer and not cause program's account removal.

Specifically, the PR implements:
- [x] Charging the existential deposit to user's balance at the time of a program creation in `Gear::upload_program` and `Gear::create_program` extrinsics
    - [x] 2 tests
- [x] Validating that existential deposit can be charged to a program's balance in case another program being created using the `gr_create_program` syscall (by decrementing the value counter) and further actual transfer of the ED in `JournalNote::StoreNewPrograms` handler
    - [x] Track the case when code does not exist, so that refund must be made
    - [x] 2 tests
- [x] The existential deposit is transferred to the program inheritor upon the `gr_exit` syscall (program deactivation)
    - [x] 1 test
- [x] Removal of value limits (various `check_value`) in all send-like extrinsics and all backend send-like syscalls
    - [x] 2 tests: extrinsics/syscalls (for syscalls could be implemented as `core-processor/ext` unit tests)
- [x] Ensures all transfers in the system are made with `Existence::AllowDeath` requirements (since the platform guarantees the account will not die as long as the program remains in active state)
- [x] Implements migrations to deposit ED from Treasury for each program that already exists in storage
    - [x] try-runtime tests that after migrations all program that are in "Active" status have accounts with at least ED on them
- [x] Fix existing tests


breathx:
Seems ready to be reviewed once fixed benches?

ekovalev:
> Seems ready to be reviewed once fixed benches?

Yes, pretty much.

pallets/gear-program/src/migrations/v7.rs
```diff
@@ -0,0 +1,292 @@
...
+            let mut transferred = 0_u32;
+            let mut active_count = 0_u32;
+            ProgramStorage::<T>::iter().for_each(|(program_id, program)| {
+                weight = weight.saturating_add(T::DbWeight::get().reads_writes(1, 1));
+
+                if let Program::Active(_p) = program {
+                    active_count += 1;
+                    // Transfer ED from treasury account to program's account
+                    // regardless whether the latter is extant or not.
+                    match <T as pallet_treasury::Config>::Paymaster::pay(
```
ukint-vs, line 86: Do we need to topup treasury before the upgrade?

ukint-vs (approved):
lgtm.

pallets/gear-program/src/migrations/v7.rs
ekovalev: You're right. That's why they topped it up manually by 1k Vara a few days ago.

NikVolf:
merge master @ekovalev ?

ekovalev:
> merge master @ekovalev ?

I'd prefer @breathx to take a look first, as well

NikVolf:
@breathx final thoughts?

pallets/gear-program/src/migrations/v7.rs
```diff
@@ -0,0 +1,292 @@
...
+                        &program_id.cast(),
+                        (),
+                        existential_deposit,
+                    ) {
+                        Ok(_res) => {
+                            transferred += 1;
+                        }
+                        Err(e) => {
+                            log::debug!(
```
breathx, line 95:
```suggestion
                            log::error!(
```

pallets/gear/src/lib.rs
```diff
@@ -602,6 +604,15 @@ pub mod pallet {
...
             );
 
+            // Create an account for the program
+            CurrencyOf::<T>::transfer(
+                &who,
+                &program_id.cast(),
+                CurrencyOf::<T>::minimum_balance(),
+                ExistenceRequirement::KeepAlive,
+            )?;
+            System::<T>::inc_consumers(&program_id.cast())?;
```
breathx, line 614: why not locking balance?

pallets/gear/src/lib.rs
```diff
@@ -1242,9 +1243,27 @@ pub mod pallet {
...
+
+            // Before storing the program to `ProgramStorage` we need to make sure that an account
+            // can be created for the program.
+            // Note: making a transfer outside of the `Ext::set_program()` because here a transfer
+            // is allowed to fail (as opposed to creating a program by a program).
+            CurrencyOf::<T>::transfer(
+                &who,
+                &program_id.cast(),
+                CurrencyOf::<T>::minimum_balance(),
+                ExistenceRequirement::KeepAlive,
```
breathx, line 1257: why keep alive?
sender is free to be deleted actually

pallets/gear/src/lib.rs
```diff
@@ -1242,9 +1243,27 @@ pub mod pallet {
...
+            )?;
+
+            // A consumer is added at program's account creation and removed it on termination
+            // to make sure the account is not removed by the runtime while the program exists.
+            // With a consumer, a correct runtime should not remove the account.
+            System::<T>::inc_consumers(&program_id.cast())?;
```
breathx, line 1263: you need to lock this funds, that automatically will inc consumers/providers

pallets/gear/src/lib.rs
```diff
@@ -1690,6 +1716,8 @@ pub mod pallet {
                     Error::<T>::InactiveProgram
                 );
 
+                Self::check_gas_limit(gas_limit)?;
```
breathx, line 1719: why moved below?

pallets/gear/src/manager/journal.rs
```diff
@@ -363,12 +366,30 @@ where
...
+                    // Make sure an account exists for the newly created program.
+                    // Balance validity check has been performed so we don't expect any errors.
+                    CurrencyOf::<T>::transfer(
+                        &program_id.cast(),
+                        &candidate_id.cast(),
+                        CurrencyOf::<T>::minimum_balance(),
+                        ExistenceRequirement::KeepAlive,
+                    )
+                    .unwrap_or_else(|e| unreachable!("Existential deposit transfer error: {e:?}"));
+                    System::<T>::inc_consumers(&candidate_id.cast())
```
breathx, line 390: lock as well

breathx: and unlock on exit journal note with the file

examples/create-program-reentrance/src/wasm.rs
```diff
@@ -0,0 +1,71 @@
+// This file is part of Gear.
+
+// Copyright (C) 2023-2024 Gear Technologies Inc.
```
breathx, line 3:
```suggestion
// Copyright (C) 2024 Gear Technologies Inc.
```

examples/create-program-reentrance/src/wasm.rs
```diff
@@ -0,0 +1,71 @@
...
+        "[CREATE_PROGRAM_REENTRANCE::handle] code_hash: 0x{}, amount: {}",
+        hex::encode(code_hash),
+        amount
+    );
+
+    let code_id: CodeId = code_hash.into();
+    let dest_program_id = unsafe { DST_PROGRAM_ID };
+
+    // Call gr_create_program syscall
+    let (_msg_id, program_id) =
```
breathx, line 49: seems you could just create program, assert balance within the exec
than call it again to just share the info about balance and the balance should be the same if program wasn't created

pallets/gear/src/tests.rs
```diff
@@ -786,13 +780,14 @@ fn calculate_gas_returns_not_block_limit() {
...
         assert_ok!(Gear::upload_code(RuntimeOrigin::signed(USER_1), code));
+        // Generator needs some balance to be able to pay ED for created programs
         assert_ok!(Gear::upload_program(
             RuntimeOrigin::signed(USER_1),
             WASM_BINARY.to_vec(),
             DEFAULT_SALT.to_vec(),
             EMPTY_PAYLOAD.to_vec(),
             BlockGasLimitOf::<Test>::get(),
-            0,
+            1000,
```
breathx, line 790: relay on ED const pls

pallets/gear/src/tests.rs
```diff
@@ -1105,7 +1101,7 @@ fn auto_reply_out_of_rent_mailbox() {
         assert!(utils::is_active(program_id));
 
         let user1_balance = Balances::free_balance(USER_1);
-        assert_balance(program_id, value, 0u128);
+        assert_balance(program_id, value + ed, 0u128);
```
breathx, line 1104: ed shouldn't be free balance, it should be locked (as well for proper informationing on observers like polkajs)

breathx (changes requested):
tests to be re-reviewed once applied suggested changes

otherwise well done

core-processor/src/ext.rs
```diff
@@ -477,16 +477,6 @@ impl BackendExternalities for Ext {
 }
 
 impl Ext {
-    fn check_message_value(&mut self, message_value: u128) -> Result<(), FallibleExtError> {
```
techraed, line 480: Because of this check we had an invariant that any `program_actor -> user_actor` value transfer will be infallible. We still have with these changes this guarantee for `program_actor -> program_actor` interactions, as they always have ED on their balance and it's never touched. However, there are two cases worth testing:

1. program sends reply or just a message to user with a low gas (not enough for mailbox threshold) **with value(!)**.
2. program sends message **with value** to user and it reaches the mailbox.

There's an assumption that without this ED check transfers in `GearBank`/`Currency` from program to user can fail, i.e. cause runtime panic.

@breathx Can you double check it please?

pallets/gear/src/manager/journal.rs
techraed: is there a need if `reducible_balance` is used everywhere?

core-processor/src/ext.rs
techraed: checked the first case - it can potentially panic btw

breathx: Great catch

Transfer in journal handling in case of avoiding mailbox from program to user will cause panic if user doesnt exist and 0 < value < ed, so this case should be handled. Consider on error transfer in into unused bank funds

examples/syscall-error/src/wasm.rs
```diff
@@ -17,19 +17,19 @@
...
 
 #[no_mangle]
 extern "C" fn init() {
-    let res = msg::send(ActorId::default(), "dummy", 250);
+    let res = msg::send(ActorId::default(), "dummy", u128::MAX / 2);
     assert_eq!(
         res,
-        Err(CoreError::Ext(ExtError::Message(
-            MessageError::InsufficientValue
+        Err(CoreError::Ext(ExtError::Execution(
```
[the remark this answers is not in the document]
ekovalev: It doesn't matter actually. This example is not used anywhere as far as I can tell.
 The previous MessageError is no longer thrown, but any error here should fit the bill, should it not?.

gtest/src/manager.rs
```diff
@@ -1122,7 +1114,12 @@ impl JournalHandler for ExtManager {
     }
 
     #[track_caller]
-    fn store_new_programs(&mut self, code_id: CodeId, candidates: Vec<(MessageId, ProgramId)>) {
+    fn store_new_programs(
+        &mut self,
+        _program_id: ProgramId,
```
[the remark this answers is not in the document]
ekovalev: right

pallets/gear/src/lib.rs
ekovalev: > you need to lock this funds, that automatically will inc consumers/providers
Can't see anything like that in Substrate code.

Anyway, having funds locked doesn't necessarily exclude them from reducible amount whereas having non-zero consumers does.
That's how the `Contracts` or `Assets` pallets go about it - by explicitly incrementing consumers for a contract's account to enforce the same invariant of the account to always exist.
Locks serve a different purpose.

pallets/gear/src/lib.rs
ekovalev: It's essentially as good as creating a lock to ensure that an account will always exist. But more efficient as there is no need to iterate through existing locks when adding a new one.

pallets/gear/src/lib.rs
ekovalev: makes sense

pallets/gear-program/src/migrations/allocations.rs
```diff
@@ -118,7 +118,7 @@ impl<T: Config> OnRuntimeUpgrade for MigrateAllocations<T> {
                 "incorrect count of elements old {} != new {}",
             );
             ensure!(
-                Pallet::<T>::current_storage_version() == MIGRATE_TO_VERSION,
+                Pallet::<T>::on_chain_storage_version() == MIGRATE_TO_VERSION,
```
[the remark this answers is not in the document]
ekovalev: It shouldn't. Once the `on_chain_version` has been bumped, the migration inside `on_runtime_upgrade` simply won't run anymore.
On the other hand, comparing the `current_storage_version` to the `MIGRATE_TO_VERSION` would fail on the very first run as long as there is a newer migration for the pallet in the same Runtime. Furthermore, it doesn't indicate if a migration has been successful (because it doesn't depend on the storage at all).

pallets/gear-bank/src/lib.rs
```diff
@@ -258,13 +274,17 @@ pub mod pallet {
...
             Self::ensure_bank_can_transfer(value)?;
 
-            // The check is similar to one that used in transfer implementation.
-            // It allows us define if we cannot transfer funds on early stage
-            // to be able mean any transfer error as insufficient bank
-            // account balance, because other conditions are checked
-            // here and in other caller places.
-            if CurrencyOf::<T>::free_balance(account_id).saturating_add(value)
-                < CurrencyOf::<T>::minimum_balance()
+            // Since funds are not being minted here but transferred, the only error we can
```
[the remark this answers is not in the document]
ekovalev: It's just trying to convey a message that despite the outcome of the `can_deposit` check (when converted into a `Result`) could include different types of errors (like overflows etc.), in this particular case we can only observe the `BelowMinimum` one hence we don't bother to check which error variant it is exactly.

pallets/gear-bank/src/tests.rs
```diff
@@ -737,12 +737,26 @@ fn deposit_value_zero() {
...
 
 #[test]
-fn deposit_value_insufficient_balance() {
+fn deposit_value_overflow() {
     new_test_ext().execute_with(|| {
         const VALUE: Balance = Balance::MAX;
 
         assert!(VALUE > Balances::free_balance(ALICE));
 
+        assert_noop!(
```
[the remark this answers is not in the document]
ekovalev: no. Otherwise the test wouldn't be passing, would it?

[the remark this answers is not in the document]
ekovalev: the original `deposit_value_insufficient_balance` test has been split in two. In the original formulation (when `Balance::MAX` is being deposited) the overflow error would trigger first.
To observe the insufficient balance, I deposit `Balance::MAX/2` - then it works.

pallets/gear/src/lib.rs
ekovalev: We discussed it in DM before. If a message is not for a program then it won't be enqueued hence no point to check gas limit (checking value is being deprecated, as well).

pallets/gear/src/manager/journal.rs
ekovalev: "unlocking" is done in the `clean_inactive_program()` function (called by the exit dispatch handler):
```
// Delete the consumer to allow the account to be reaped.
System::<T>::dec_consumers(&program_account);
```

examples/create-program-reentrance/src/wasm.rs
ekovalev: sending out a message and waiting for reply is crucial to let the `value_counter` to be reset between executions when a message goes through the wait-wake cycle. This should be a part of the test.
Or not?

examples/create-program-reentrance/src/wasm.rs
ekovalev: What are the rules here? Because I see all sorts of years ranges variations all over the codebase.

pallets/gear/src/lib.rs
ekovalev: **Update:** as discussed, replaced the direct consumers increment for a program account with a lock for the existential deposit. In the view of that all the arguments against this further down in this PR can be ignored as obsolete.

pallets/gear/src/lib.rs
ekovalev: Again, the latter argument is no longer valid and can be ignored. Locks are used now.

pallets/gear/src/manager/journal.rs
ekovalev: Now literally "unlocking" of the ED is taking place

pallets/gear/src/tests.rs
ekovalev: Aligned with the request above.

core-processor/src/ext.rs
ekovalev: Fixed

examples/create-program-reentrance/src/wasm.rs
breathx: yeah, that's ok, it was just an option to simplify the test
unnecessary to follow

the aim was to assert prog balance directly inside wasm program

pallets/gear-bank/src/lib.rs
```diff
@@ -224,11 +236,19 @@ pub mod pallet {
...
-            );
+            match <CurrencyOf<T> as fungible::Inspect<_>>::can_deposit(
+                &bank_address,
+                value,
+                Provenance::Extant,
+            ) {
+                DepositConsequence::Success => (), // expected outcome
+                DepositConsequence::BelowMinimum => return Err(Error::<T>::InsufficientDeposit),
+                DepositConsequence::Overflow => return Err(Error::<T>::Overflow),
+                // The rest is unreachable in Gear protocol and can be ignored.
```
breathx, line 247: wouldn't we propagate log:error here for confidence?

pallets/gear/src/internal.rs
```diff
@@ -771,8 +774,23 @@ where
             Some(hold.expected())
         } else {
             // Permanently transferring funds.
+            // Note that we have no guarantees of the user account to exist. Since no minimum
+            // transfer value is enforced, the transfer can fail. Handle it gracefully.
             CurrencyOf::<T>::transfer(&from, &to, value, ExistenceRequirement::AllowDeath)
-                .unwrap_or_else(|e| unreachable!("Failed to transfer value: {:?}", e));
+                .unwrap_or_else(|e| match e {
+                    DispatchError::Token(TokenError::BelowMinimum) => {
+                        log::warn!(
```
breathx, line 782: more likely just debug output, it's totally normal case

pallets/gear/src/internal.rs
```diff
@@ -771,8 +774,23 @@ where
...
+                            "Failed to transfer value: {:?}. User account balance is too low.",
+                            e
+                        );
+                        <CurrencyOf<T> as fungible::Unbalanced<_>>::handle_dust(fungible::Dust(
+                            value,
+                        ));
+                    }
+                    _ => {
```
breathx, line 790: add all of the variants to pattern matching. I think we can handle dust for all case, and for ones that are not BelowMinimum throw log::error instead of panics here?

pallets/gear/src/manager/mod.rs
```diff
@@ -360,7 +365,18 @@ where
...
+
+        // The `reducible_balance` should now include the ED since no consumer is left.
+        // If some part of the program account's `free` balance is still `frozen` for some reason
+        // it will be offset against the `reducible_balance`.
+        let balance = <CurrencyOf<T> as fungible::Inspect<_>>::reducible_balance(
+            &program_account,
+            Preservation::Expendable,
+            Fortitude::Polite,
+        );
         if !balance.is_zero() {
```
breathx, line 380: Add comment that transfer below is infallible since balance >= ed

pallets/gear/src/tests.rs
```diff
@@ -2401,7 +2417,7 @@ fn delayed_program_creation_no_code() {
             DEFAULT_SALT.to_vec(),
             EMPTY_PAYLOAD.to_vec(),
             DEFAULT_GAS_LIMIT * 100,
-            0,
+            1000, // necessary for init function to succeed
```
breathx, line 2420: depend on ed pls

breathx (changes requested):
Im only worried now about match error branches and panics. Otherwise seems awesome 👍🏻 great job

examples/create-program-reentrance/src/wasm.rs
breathx: > What are the rules here? Because I see all sorts of years ranges variations all over the codebase.

*year module/file created* - *actual year*

examples/syscall-error/src/wasm.rs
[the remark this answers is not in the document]
breathx: @ark0f what do you think about it?
Seems it's still "message error", not execution overall

pallets/gear-program/src/migrations/v7.rs
breathx: still unaddressed

breathx:
@techraed kindly asking you to double check the changes and my comments above

pallets/gear-program/src/migrations/v7.rs
ekovalev: > still unaddressed

I'm pretty sure it's not: it's `log::error!` now, as proposed earlier

pallets/gear/src/internal.rs
ekovalev: The problem here is that the error type is `sp_runtime::DispatchError` which has tons of variants, of which the `CurrencyOf::<T>::transfer()` can only return a few. Looks impractical (if at all possible) to list all possible variants.

Regarding the second part (not panicking but logging an error), it depends on the invariants we want to uphold. If there is a value a user hopes to receive back from a program (which can easily be big), but a program can't transfer it despite all the safeguards we have in the code, it's a reason to look into it rather than just letting go, isn't it?

techraed (commented):
Awesome!

I remember we agreed to use locks and to introduce *integrational* test, which tests that underlying `Currency` implementor locks & account existence invariants works as expected.

So, could you, please, introduce the test in this PR?

pallets/gear-bank/src/lib.rs
```diff
@@ -224,11 +236,21 @@ pub mod pallet {
...
+                DepositConsequence::CannotCreate
+                | DepositConsequence::UnknownAsset
+                | DepositConsequence::Blocked => {
+                    log::error!("Unexpected deposit consequence while depositing to bank address");
+                }
```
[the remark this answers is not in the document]
ekovalev: In reality it doesn't matter: this arm should never be `match`ed, and if it does it mandates us to immediately look at the underlying `can_deposit` implementation regardless of the exact `DepositConsequence` variant being matched. I'd suggest keep it simple and not complicate things unnecessarily.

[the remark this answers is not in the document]
techraed: I do not insist on that

techraed (approved):

breathx (approved):

PR #3962, opened 2024-05-13
feat(gtest): Porting gas tree implemenatation into `gtest`

techraed:
This opens set of PRs making `gtest` stable and matching our runtime.

- [x] Use gas tree implementation for `property_tests` as a base gas tree implementation using "in-memory" storage
- [x] Replace basic gas limits logic in `gtest` with a ported gas tree
- [x] Use GAS_ALLOWANCE, instead of `u64::MAX` used everywhere as a gas limit/gas allowance
- [x] Replace todos with issues/tasks
- [x] Make `System` a single thread singleton, which has separate copies on other threads (`OnceCell` within `thread_local` macro). We don't need one instance for all the threads.

### DIFF
--- a/common/src/gas_provider/mod.rs
+++ b/common/src/gas_provider/mod.rs
@@ -21,6 +21,7 @@ use frame_support::traits::tokens::Balance as BalanceTrait;
 use sp_runtime::{traits::Zero, RuntimeDebug};
 use sp_std::marker::PhantomData;
 
+mod auxiliary;
 mod error;
 mod internal;
 mod lockable;
@@ -32,6 +33,7 @@ mod reservable;
 #[cfg(test)]
 mod property_tests;
 
+pub use auxiliary::*;
 pub use error::Error;
 pub use internal::TreeImpl;
 pub use lockable::{LockId, LockableTree};
--- a/common/src/gas_provider/mod.rs
+++ b/common/src/gas_provider/mod.rs
@@ -21,6 +21,7 @@ use frame_support::traits::tokens::Balance as BalanceTrait;
 use sp_runtime::{traits::Zero, RuntimeDebug};
 use sp_std::marker::PhantomData;
 
+#[cfg(feature = "std")]
 mod auxiliary;
 mod error;
 mod internal;
--- a/common/src/gas_provider/mod.rs
+++ b/common/src/gas_provider/mod.rs
@@ -34,6 +34,7 @@ mod reservable;
 #[cfg(test)]
 mod property_tests;
 
+#[cfg(feature = "std")]
 pub use auxiliary::*;
 pub use error::Error;
 pub use internal::TreeImpl;
--- a/common/src/gas_provider/mod.rs
+++ b/common/src/gas_provider/mod.rs
@@ -35,7 +35,9 @@ mod reservable;
 mod property_tests;
 
 #[cfg(feature = "std")]
-pub use auxiliary::*;
+pub use auxiliary::{
+    AuxiliaryGasProvider, ExternalOrigin, GasTreeError, PlainNodeId, ReservationNodeId,
+};
 pub use error::Error;
 pub use internal::TreeImpl;
 pub use lockable::{LockId, LockableTree};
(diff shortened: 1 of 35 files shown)


breathx:
+23k lines 🐳

techraed:
> +23k lines 🐳

accidentally added local copy of the expanded vara substrate runtime module...

breathx (changes requested):
Haven't fully checked it, but please:
1. Separate `gear-common` operational stuff into dedicated PR
2. I don't see any case somebody will assert System::new() and why, so I'd make it panicking

techraed:
> Haven't fully checked it, but please:
> 
> 1. Separate `gear-common` operational stuff into dedicated PR
> 2. I don't see any case somebody will assert System::new() and why, so I'd make it panicking

1. Discussed
2. Will do

common/src/gas_provider/mod.rs
```diff
@@ -32,6 +34,8 @@ mod reservable;
 #[cfg(test)]
 mod property_tests;
 
+#[cfg(feature = "std")]
+pub use auxiliary::*;
```
ark0f, line 38: I think it worth to not re-export because of simple type names like `TotalIssuanceWrap`, `ExternalOrigin`, etc

gtest/Cargo.toml
```diff
@@ -8,6 +8,7 @@ license.workspace = true
 
 [dependencies]
 gear-core.workspace = true
+gear-common = { workspace = true, features = ["std"] }
```
ark0f, line 11: It might be `gtest` is not so lightweight now. Could you please compare compilation time of the crate?

gtest/src/system.rs
```diff
@@ -91,30 +106,37 @@ impl LazyPagesStorage for PagesStorage {
...
 impl System {
     /// Prefix for lazy pages.
     pub(crate) const PAGE_STORAGE_PREFIX: [u8; 32] = *b"gtestgtestgtestgtestgtestgtest00";
 
     /// Create a new testing environment.
+    ///
+    /// # Panics
+    /// Only one instance in the current thread of the `System` is possible to
+    /// create. Instantiation of the other one leads to runtime panic.
+    #[allow(clippy::new_without_default)]
```
ark0f, line 118: I think default can be returned back. Trait implementation can be documented, too

gtest/src/system.rs
```diff
@@ -34,7 +34,22 @@ use gear_core::{
...
+    path::Path,
+    thread,
+};
+
+thread_local! {
+    /// `System` is a singleton with a one instance and no copies returned.
+    ///
+    /// `OnceCell` is used to control one-time instantiation, while `RefCell`
+    /// is needed for interior mutability to uninitialize the global.
+    static SYSTEM_INITIALIZED: RefCell<OnceCell<()>> = const { RefCell::new(OnceCell::new()) };
```
ark0f, line 51: Looks like `RefCell<bool>` will do the same but simpler

gtest/src/manager.rs
```diff
@@ -448,18 +449,22 @@ impl ExtManager {
...
+    pub(crate) fn run_dispatch(&mut self, dispatch: Dispatch, from_task_pool: bool) -> RunResult {
         self.prepare_for(&dispatch);
 
-        self.gas_limits
-            .entry(dispatch.id())
-            .or_insert_with(|| dispatch.gas_limit().unwrap_or(u64::MAX));
-
         if self.is_program(&dispatch.destination()) {
+            // TODO: `gas_limit` being None is an abuse of the `run_dispatch` usage,
+            // as it must be called only for user messages
```
ark0f, line 461: Could you please explain this in details?

ark0f (changes requested):

gtest/src/system.rs
techraed: Well, I disagree here.
Creating more than one instance of the type leads to panic. When you have `Default` trait you can't fully control instantiation, because you have now more than one instantiation points: `new` and `default` functions.

techraed: Also the trait `#[derive]` stuff can cause some problems if someone decides to use `System` as an owned by the other `Default` implementor type.

common/src/gas_provider/mod.rs
techraed: done in [00dec79](https://github.com/gear-tech/gear/pull/3962/commits/00dec791abb2807adb270767548d3fcec83eb50a)

ark0f (approved):

breathx (approved):
🔥

PR #3966, opened 2024-05-16
chore: debug assert reservation states in process_success

playX18:
@breathx


core-processor/src/processing.rs
```diff
@@ -147,12 +147,12 @@ where
                 res.system_reservation_context,
                 ActorExecutionErrorReplyReason::Trap(reason),
             ),
-            DispatchResultKind::Success => process_success(Success, res),
+            DispatchResultKind::Success => process_success(Success, res, Some(system_reservation_ctx)),
```
breathx, line 150: make assertion in place instead of impl new unused argument for pub fn

breathx (changes requested):

core-processor/src/processing.rs
```diff
@@ -147,11 +147,49 @@ where
                 res.system_reservation_context,
                 ActorExecutionErrorReplyReason::Trap(reason),
             ),
-            DispatchResultKind::Success => process_success(Success, res),
+            DispatchResultKind::Success => {
+                if let (Some(current_prev), Some(current_res)) = (
```
breathx, line 151: couldn't matching and extraction be hidden in debug_assert to avoid excess code in prod compile?

core-processor/src/processing.rs
```diff
@@ -147,11 +147,49 @@ where
...
+                    system_reservation_ctx.current_reservation,
+                    res.system_reservation_context.current_reservation,
+                ) {
+                    debug_assert!(current_res <= current_prev);
```
techraed, line 155: I'd add an explanotary comment, as it seems you are asserting the invariant, aren't you?
also, maybe even have a separate match arm for variants, for which reservations must be checked. That could reduce copy-paste I guess

core-processor/src/processing.rs
```diff
@@ -139,25 +139,54 @@ where
...
-                process_success(Wait(duration, waited_type.clone()), res)
+        Ok(res) => {
+            match res.kind {
+                DispatchResultKind::Success
+                | DispatchResultKind::Wait(_, _)
+                | DispatchResultKind::Exit(_) => {
+                    // assert that after processing the initial reservation is less or equal to the current one.
+                    // during execution reservation amount might increase due to `system_reserve_gas` calls
+                    // thus making initial reservation less than current one.
+                    debug_assert!(system_reservation_ctx
```
breathx, line 150: both these checks are aimed to verify system reservations
1. these checks aren't 100% correct
2. you need to check initial task's idea: amount of common reservations (not the amount of gas in system reservation)

breathx (changes requested):

breathx:
@playX18 why is it staling?

breathx (approved):

breathx:
@ark0f please take a look

ark0f (approved):

core-processor/src/processing.rs
```diff
@@ -139,25 +141,60 @@ where
...
+                        .current_reservation
```
techraed, line 153: it's always `None`, why check it?

core-processor/src/processing.rs
```diff
@@ -139,25 +141,60 @@ where
...
+                            .current_reservation
+                            .map(|current| initial <= current))
+                        .unwrap_or(true));
+
+                    debug_assert!(system_reservation_ctx
+                        .previous_reservation
+                        .and_then(|initial| res
+                            .system_reservation_context
+                            .previous_reservation
+                            .map(|current| initial <= current))
```
techraed, line 165: that should be checked for being equal, because of the impementation details

core-processor/src/processing.rs
techraed: still if you want to check that, you must use `unwrap_or_default` for the `system_reservation_ctx.current_reservation` (the one created before execution)

techraed (changes requested):

core-processor/src/processing.rs
```diff
@@ -139,25 +141,60 @@ where
...
-            DispatchResultKind::GasAllowanceExceed => {
-                process_allowance_exceed(dispatch, program_id, res.gas_amount.burned())
+        Ok(res) => {
+            match res.kind {
+                DispatchResultKind::Success
+                | DispatchResultKind::Wait(_, _)
+                | DispatchResultKind::Exit(_) => {
+                    // assert that after processing the initial reservation is less or equal to the current one.
+                    // during execution reservation amount might increase due to `system_reserve_gas` calls
+                    // thus making initial reservation less than current one.
```
techraed, line 151: what you actually wrote here must be checked the different way:
1. take the `system_reservation_ctx.previous_reservation`
2. take the `res.system_reservation_ctx.previous_reservation + res.system_reservation_ctx.current_reservation`
3. check that the result of 2 >= result of 1

techraed: maybe even instead of 2 use `res.context_store.system_reservation` as it already has the sum of 2

core-processor/src/processing.rs
```diff
@@ -139,25 +141,58 @@ where
...
+                    debug_assert!(
+                        res.context_store.system_reservation()
+                            >= system_reservation_ctx.previous_reservation
+                    );
+                    /*debug_assert!(system_reservation_ctx.previous_reservation.and_then(
+                        |initial| res
+                            .system_reservation_context
+                            .previous_reservation
+                            .map(|current| initial == current)
+                    ));*/
```
techraed, line 161: remove that please

techraed (approved):
@ark0f re-check, please

core-processor/src/processing.rs
techraed:
```suggestion
```

ark0f (approved):

breathx (approved):

PR #3982, opened 2024-05-22
refactor(core): remove core program

grishasobol:
append `GenuineProgram` struct to `gtest`
append inner `program` struct for `core-processor`


core-processor/src/executor.rs
```diff
@@ -141,18 +140,18 @@ where
     let execute = || {
         let env = Environment::new(
             ext,
-            program.code_bytes(),
+            program.code.code(),
```
breathx, line 143: to consider renaming here since having `fn code` for `struct Code` looks weird

breathx (approved):

core-processor/src/executor.rs
grishasobol: This would change too many files, so I will make issue to make this in other PR

grishasobol: Because we have the same method in `core::code::Code`

grishasobol: https://github.com/gear-tech/gear/issues/3985

PR #3983, opened 2024-05-23
refactor(common, core): move program from `gear-common` to `gear-core`

grishasobol:


core/src/program.rs
```diff
@@ -18,11 +18,112 @@
...
+        match prog_with_status {
+            Program::Active(p) => Ok(p),
+            _ => Err(InactiveProgramError),
+        }
+    }
+}
+
+/// Active program in storage.
+#[derive(Clone, Debug, Decode, Encode, PartialEq, Eq, TypeInfo)]
+pub struct ActiveProgram<BlockNumber: Copy> {
```
techraed, line 93: what about these?
```rust
#[codec(crate = codec)]
#[scale_info(crate = scale_info)]
```

techraed (commented):
generally looks ok

core/src/program.rs
grishasobol: I'd used `gear-core` style here. Do you this there is any differences? `scale_info` used by default as `scale_info crate`, and in core we use naming `parity-scale-codec` instead of `codec`.

If you think this is must be refactored in core - let's create separate issue for that ?

techraed: discussed

techraed (approved):
I'd recommend to take a look @gshep @ark0f

ark0f (approved):

NikVolf:
rebase master @grishasobol

gshep (approved):
:+1:

PR #3984, opened 2024-05-25
refactor(core-processor): lazy storage access

grishasobol:
Introduce `LazyStorageAccess` in order to move all prepare for processing logic to `gear-core-processor`.
Use it in pallet-gear, pallet-gear benchmarks and gtest.


grishasobol:
This solution must be reconsidered

PR #3987, opened 2024-05-28
Replace cost-* helper macros with functions.

### LINKED ISSUE #3943: Simplify/remove `cost_*` macros
### File Location(s)

pallets/gear/src/schedule.rs

### Proposal

The `cost_*` macros located in `pallets/gear/src/schedule.rs` appear to be overly complicated and difficult to read. It is proposed to simplify them or remove them altogether

ByteNacked:
Resolves #3943.

Replace `cost-*` helper macros with functions.

@reviewer-or-team

### DIFF
--- a/pallets/gear/src/schedule.rs
+++ b/pallets/gear/src/schedule.rs
@@ -19,10 +19,6 @@
 //! This module contains the cost schedule and supporting code that constructs a
 //! sane default schedule from a `WeightInfo` implementation.
 
-#![allow(unused_parens)]
-
-use core::num;
-
 use crate::{weights::WeightInfo, Config, CostsPerBlockOf, DbWeightOf};
 use common::scheduler::SchedulingCostsPerBlock;
 use core_processor::configs::{ExtCosts, ProcessCosts, RentCosts};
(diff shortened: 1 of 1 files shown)


ByteNacked:
Well, it's obviously not a perfect solution, but in my opinion, it significantly simplifies the reading comprehension of these code parts.

### NOTICE:
For a 1:1 function/macros correspondence check, please see the first commit with the tests that verify this.

pallets/gear/src/schedule.rs
```diff
@@ -660,107 +658,72 @@ impl<T: Config> From<MemoryWeights<T>> for LazyPagesCosts {
...
+fn cost_byte_batched(w: fn(u32) -> Weight) -> Weight {
+    Weight::from_parts(cost_batched(w).ref_time() / 1024, 0)
 }
 
-macro_rules! cost {
-    ($name:ident) => {
-        cost_args!($name, 1)
-    };
+#[inline]
+fn cost_byte_batched_args(w: fn(u32, u32) -> Weight, arg1: u32, arg2: u32) -> Weight {
```
ByteNacked, line 682: Currently, Rust does not have functions that can accept a variable number of arguments, and in such cases, it usually makes sense to use macros. However, we have only one such case here, so it seems fine to handle it manually.

pallets/gear/src/schedule.rs
```diff
@@ -660,107 +658,72 @@ impl<T: Config> From<MemoryWeights<T>> for LazyPagesCosts {
...
-
-macro_rules! cost_instr_with_batch_size {
-    ($name:ident, $num_params:expr, $batch_size:expr) => {
-        cost_instr_no_params_with_batch_size!($name, $batch_size).saturating_sub(
-            (cost_instr_no_params_with_batch_size!(instr_i64const, $batch_size) / 2)
-                .saturating_mul($num_params),
-        )
-    };
+#[inline]
+fn cost(w: fn(u32) -> Weight) -> Weight {
```
ByteNacked, line 662: If there are any performance concerns about whether the compiler will be able to reduce these function calls to simple assignments when passing a pointer to a function, it could be replaced with impl Fn(u32) -> Weight. However, this doesn't seem to be the case.

ukint-vs (approved):

PR #3996, opened 2024-06-04
feat(gtest): Introduce gas allowance to `gtest`

### LINKED ISSUE #3977: Introduce a proper gas allowance management in the `gtest`
### Problem to Solve

By default max value is just used and all messages (original and generated during the execution of the original) are being run as they were all packed in one block.

techraed:
Resolves #3977

- Introduce _block_ concept: gas for all generated dispatches are counted not in one single counter, but in the map `block number -> gas used`.
- If processing stopped because of the gas allowance exceed error, the block number is incremented and dispatches execution continues as if it is a new block with updated gas allowance.

### DIFF
--- a/gtest/src/gas_tree.rs
+++ b/gtest/src/gas_tree.rs
@@ -18,7 +18,7 @@
 
 //! Auxiliary (for tests) gas tree management implementation for the crate.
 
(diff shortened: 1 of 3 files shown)


techraed:
merge after #3962 and #3961

NikVolf (approved):

Issue #3997, opened 2024-06-04
Implement new call to send message and collect reply from on-chain state in gsdk and gclient


vobradovich:
### Problem to Solve

Method `gear_calculateReplyForHandle` is not implelemted in gsdk and gclient

### Possible Solution

- Support `gear_calculateReplyForHandle` rpc call according https://github.com/gear-tech/gear/pull/3790

PR #3998, opened 2024-06-04
feat(gsdk,gclient): Implement gsdk api for calling gear_calculateReplyForHandle

vobradovich:
Resolves # https://github.com/gear-tech/gear/issues/3997.

### DIFF
--- a/gsdk/Cargo.toml
+++ b/gsdk/Cargo.toml
@@ -45,6 +45,7 @@ gsdk = { path = ".", features = ["testing"] }
 tokio = { workspace = true, features = [ "full" ] }
 demo-messenger.workspace = true
 demo-new-meta.workspace = true
+demo-vec.workspace = true
 demo-waiter = { workspace = true, features = ["std"] }
 
 [build-dependencies]
--- a/gsdk/Cargo.toml
+++ b/gsdk/Cargo.toml
@@ -45,6 +45,7 @@ gsdk = { path = ".", features = ["testing"] }
 tokio = { workspace = true, features = [ "full" ] }
 demo-messenger.workspace = true
 demo-new-meta.workspace = true
+demo-ping = { workspace = true, features = ["std"] }
 demo-waiter = { workspace = true, features = ["std"] }
 
 [build-dependencies]
--- a/gsdk/Cargo.toml
+++ b/gsdk/Cargo.toml
@@ -45,7 +45,6 @@ gsdk = { path = ".", features = ["testing"] }
 tokio = { workspace = true, features = [ "full" ] }
 demo-messenger.workspace = true
 demo-new-meta.workspace = true
-demo-ping = { workspace = true, features = ["std"] }
 demo-waiter = { workspace = true, features = ["std"] }
 
 [build-dependencies]
--- a/gsdk/src/rpc.rs
+++ b/gsdk/src/rpc.rs
@@ -19,7 +19,10 @@
 //! Gear API RPC methods
 
 use crate::{result::Result, Api, GasInfo};
-use gear_core::ids::{CodeId, MessageId, ProgramId};
+use gear_core::{
+    ids::{CodeId, MessageId, ProgramId},
+    message::ReplyInfo,
+};
 use sp_core::H256;
 use subxt::rpc_params;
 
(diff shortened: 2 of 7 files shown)


gsdk/src/rpc.rs
```diff
@@ -185,4 +188,30 @@ impl Api {
             .await
             .map_err(Into::into)
     }
+
+    /// gear_calculateReplyForHandle
+    pub async fn calculate_reply_for_handle(
```
DennisInSky, line 193: Do we need a param similar to this?
https://github.com/gear-tech/gear/blob/0c429b0a620d1effd9cc0d986748463dd7d00406/gsdk/src/rpc.rs#L86

gsdk/tests/rpc.rs
```diff
@@ -355,6 +359,52 @@ async fn test_program_counters() -> Result<()> {
...
+        .await?;
+
+    assert_eq!(
+        reply_info,
+        ReplyInfo {
+            payload: b"PONG".to_vec(),
+            value: 0,
+            code: ReplyCode::Success(SuccessReplyReason::Manual)
+        }
+    );
```
DennisInSky, line 403: I would assert Alice's account balance hasn't changed, i.e. it was for free.

gsdk/Cargo.toml
```diff
@@ -45,6 +45,7 @@ gsdk = { path = ".", features = ["testing"] }
 tokio = { workspace = true, features = [ "full" ] }
 demo-messenger.workspace = true
 demo-new-meta.workspace = true
+demo-ping = { workspace = true, features = ["std"] }
```
DennisInSky, line 48: Probably I would rather use one of the already referenced demos in order to avoid compilation time increase. The `demo-new-meta` seems having some suitable message handling.

NikVolf:
@clearloop please take a look!

gsdk/src/rpc.rs
vobradovich: No, there is no such parameter

gsdk/tests/rpc.rs
```diff
@@ -355,6 +359,52 @@ async fn test_program_counters() -> Result<()> {
...
+
+    // 1. upload program.
+    let signer = Api::new(Some(&node_uri(&node)))
+        .await?
+        .signer("//Alice", None)?;
+
+    signer
+        .calls
+        .upload_program(
+            demo_ping::WASM_BINARY.to_vec(),
```
clearloop, line 377: use a demo which includes state, you can add a new method in `demo-messenger`, it is just for testing the clients, for example a `counter`

```rust
fn method_change_state_or_other_names() -> i32 {
   *state += 1;
   return state
}
```

1. *simulate the call and get the updated state
2. (advanced: yes, this requires demo-meta ) query the state and it has not changed
3. *run the call check the result is the same

gsdk/Cargo.toml
vobradovich: I looked at `demo-messenger`, but it doesn't contain methods with `msg::reply_bytes`, so I took `ping`

vobradovich: but I'm gonna try it with `demo-new-meta`

gsdk/tests/rpc.rs
vobradovich: After closer look
- there are no methods in `demo-new-meta` that change the state
- exists example `demo-mul-by-const`, but it is not referenced

Just updated test to use `demo-new-meta`, obviously, the state doesn't change

clearloop: sorry my bad, ur right, I just found that both of the two don't have state mutations ))

only `new-meta` has metadata functions among our examples but it relates to too many tests that modify it is not the ideal way.

I'd suggest you to use `demo-mul-by-const` to verify the state changes since the result of the simulation matches the real execution result is the thing to test

btw, like what you said, this RPC calls obviously doesn't change states, and it is how it works since it is just a read call without signature and will never enter the transaction pool, so skipping verify-state-not-changes makes sense as well!

vobradovich: actually, `demo_mul_by_const` does not change state in the handle(

vobradovich: addув test to demo-vec and state change, review pls

gsdk/src/rpc.rs
breathx: @DennisInSky api already exist on node, so this impl is just access-entry so couldn't impl any custom additional logic. Speaking of your idea it was previously decided to keep api as much similar to real tx sending as possible

breathx: I guess, this could be resolved

gsdk/tests/rpc.rs
breathx: I guess it's unnecessary since there's no even signing for the call

breathx: Same for "immutability of the call", but lets keep it cz already coded

breathx (approved):
👍🏻

breathx:
@vobradovich pls push empty commit to trigger CI

clearloop (approved):
lgtm! please resolve conversions

PR #4000, opened 2024-06-05
 fix(benchmarks): update reinstrument bench

### LINKED ISSUE #3992: Take in account different WASM module sections in re-instrumentation bench
### Problem to Solve

Currently, the re-instrumentation benchmark (and the produced weight) measures only the size of the code section, which might not be the heaviest one. Investigate and propose a more robust solution. Implement it.

### Possible Solution

Maybe measure heaviest section (worst case scenario) in re-instrumentation bench

### Notes

See `pallets/gear/src/benchmarking/mod.rs`

ByteNacked:
Resolves #3992 .

Update `reinstrument_per_kb` bench (and corresponding weight).
Now, when generating the input module code for this benchmark, table and element sections are used (previously, only the code section was used) because they are heaviest weight-wise.

@reviewer-or-team

### DIFF
--- a/pallets/gear/src/benchmarking/code.rs
+++ b/pallets/gear/src/benchmarking/code.rs
@@ -104,11 +104,21 @@ pub struct ModuleDefinition {
     pub stack_end: Option<WasmPage>,
 }
 
+#[derive(Default)]
+pub enum InitElements {
+    NoInit,
+    Number(u32),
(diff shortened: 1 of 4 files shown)


gshep (approved):
:+1:

NikVolf (approved):

PR #4017, opened 2024-06-18
feat: Introduce `gear-tasks`

ark0f:
Corresponding Substrate patches: https://github.com/gear-tech/polkadot-sdk/pull/5


utils/gear-tasks/tests/with_client.rs
```diff
@@ -0,0 +1,193 @@
...
+}
+
+fn new_test_ext() -> BackendExternalities {
+    CLIENT.get_or_init(|| {
+        let mut client = TestClientBuilder::new().build();
+        // Substrate's `CodeExecutor::call()` has explicit flag to use native execution,
+        // so it's applicable for `NativeElseWasmExecutor`, too.
+        // The flag is always set to `false` in our case, so
+        // we set it to true
+        client.gear_use_native();
```
clearloop, line 91: I'm wondering if this is just for testing or even production since the current approach has modified 4 substrate packages for passing this `boolean` ...

```
substrate_test_client::Client 
-> sc_client_api::CallExecutor 
-> sc_client::LocalCallExecutor 
-> sp_state_machine::Client
```

I'd suggest to just introduce a `force_native` field in `sc_executor::NativeElseWasmExecutor` with a public mutable method, then we can do this setting stuff in our own repo

- [gear_node_testing::TestClientBuilderExt::use_gear_native](https://github.com/gear-tech/gear/blob/master/node/testing/src/client.rs#L100) for testing
- [gear_service::new_partial](https://github.com/gear-tech/gear/blob/master/node/service/src/lib.rs#L212) for production

ref https://github.com/gear-tech/polkadot-sdk/pull/4

ark0f: > I'm wondering if this is just for testing

It is testing only

ark0f: > I'd suggest to just introduce a force_native field in sc_executor::NativeElseWasmExecutor with a public mutable method

Done. Check again, please

utils/gear-tasks/src/host.rs
```diff
@@ -0,0 +1,215 @@
...
+                            .expect(
+                                "`Externalities::gear_overlayed_changes()` returns invalid type",
+                            );
+                        // Overlayed changes enter runtime mode in `StateMachine::execute()`
+                        // so we definitely want to exit from it before `execute_task()` tries to
+                        // enter it again.
+                        overlayed_changes.exit_runtime().expect(
+                            "Overlayed changes is never called in `sp_state_machine::StateMachine::execute()`",
+                        );
+                        runtime_api.set_overlayed_changes(*overlayed_changes);
```
clearloop, line 135: if we can ensure this logic runs before state machine execution, why we still need to reset the overlayed_changes of state machine (it's holding a mutable reference of OverlayedChanges) but not modify it with `RuntimeApiImpl.changes` directly?

see also: [impl CallExecutor for sc_service::LocalExecutor](https://github.com/gear-tech/polkadot-sdk/blob/master/substrate/client/service/src/client/call_executor.rs#L148)

breathx (changes requested):
Please consider throwing immutable refs to overlay for threads both on gear and substrate side as discussed

PR #4019, opened 2024-06-19
Initial work on Fuzzer randomization

### LINKED ISSUE #3988: Randomization of fuzzer configuration through `arbitrary` crate
### Problem to Solve

At the moment fuzzer config and what we pass to `wasm-smith` for WASM generation is mostly static, this does not allow us to get fully randomized programs that cover wide range of possible usages of Gear API and potentially lowers the coverage percentage we might actually get. 

### Possible Solution

- Randomize `StandardGearWasmConfigsBundle` and `SelectableParams` in `wasm-gen`. Below is the observations based on running fuzzer multiple days with tuning each of configs possible. 
- Increase amount of gas by calculating theorethically needed amount of gas for each program we generate instead of using constant gas which also should help us run more programs without failure after short period of time.

### Notes

# What to not randomize
- `max_funcs`: almost never results in increasing coverage or in rate of successful terminations. It also increases the total runtime for a single generated program. In 5 minutes on my i5 13400f Fuzzer is able to run only about ~50 programs total. 
-  `min_funcs`: very similar to `max_funcs` in terms of what happens to coverage and program amounts. Not really worth it. Lowering the value helps to get more programs to run but overall not a big increase: with all configs set to default (min_funcs: 3, max_funcs: 5) I am getting ~700 programs ran in 5 minutes and 83% of them are successful while other 17% are terminated for various reasons: 80% of terminations come from gas insufficient error. In case of increasing the `max_funcs` we ran out of gas more often and amount of successfully terminated programs shifts to 30% and total amount of programs executed also drops from 700 to 50 in case  of `max_funcs: 20`

# What's worth randomizing

- `max_instructions`: This value highly impacts on how many programs we can run under limited time-frame: the larger it is the less programs we can run as more code == more gas consumption. Plus if we enable `Control` this has even larger impact on gas. My proposal here is to randomize this value from in between 100-800 instructions:
- - `>600 and <900` range is good enough to not significantly lower the amount of programs ran and also does not make programs ran out of gas noticeable faster.
- - `>100 and <500` range is good for running small quick programs with `Control` instructions enabled as it does not use gas so quickly

- Adding or removing `Control` instructions to WASM programs: Control instructions do have a high impact on runtime of a programs and increase coverage significantly: based on pallet/gear I can see decrease of coverage withot control to about ~50-60% while with control it's up to 90% for gear pallet. My suggestion is that it's worth to enable control instructions in case of `<= 500` `max_instructions` which would not result in program runtime increase and also does not change rate of successes/traps in generated programs.
- Random range of `SyscallInjectionTypes` in `StandardGearWasmConfigsBundle`: by default we generate one of each syscalls, with increase of this range to 1..=10 more coverage was seen in pallets plus more branches and more state transitions should happen which should test our runtime more thoroughly. 

# What to increase

I think it's worth to also calculate allowed gas for each program that we generate based on estimation through config and total instructions in the program so we can run the generated programs until completion more often rather than failing because we exceeded our gas limit.

playX18:
Resolves #3988.

# Notes

At the moment this PR contains simple implementation, I am still working on making this randomization better. At the moment some runs show success rates of 80-90% and others show 30%. Coverage is also increased but ever so slightly, I am guessing more testing with syscall injections is needed but for this I need to figure out how to randomize them properly.


playX18:
@techraed request for review now. I've adjusted parameters to match master fuzzer success/failure rate **and** potentially have better success rate. Still need to verify this on dedicated server though

utils/runtime-fuzzer/src/generator/upload_program.rs
```diff
@@ -100,17 +102,45 @@ fn arbitrary_salt(u: &mut Unstructured) -> Result<Vec<u8>> {
...
+        unstructured.int_in_range(80..=400).unwrap()
+    } else {
+        unstructured.int_in_range(500..=1200).unwrap()
+    };
+
+    let (min_funcs, max_funcs) = no_control.then(|| (1, 1)).unwrap_or((3, 4));
+
     let initial_pages = 2;
-    let mut injection_types = SyscallsInjectionTypes::all_with_range(1..=3);
+    // pump up injection rates of syscalls when there's control instructions and lower it when there's no control
```
techraed, line 130: I think of setting "wait*" injection range to `0..=0` in case of `no controls`. Almost all executions of `no-control` then must go to wait list, because you have only 1 functions with < 400 instructions, i.e. chance of being in a waitlist if `gr_wait*` is inserted is ~100%

utils/runtime-fuzzer/src/runtime/account.rs
```diff
@@ -106,10 +106,22 @@ impl<'a> BalanceManager<'a> {
...
         let max_balance = runtime::gas_to_value(runtime::acc_max_balance_gas());
-        let new_balance = self
-            .unstructured
-            .int_in_range(EXISTENTIAL_DEPOSIT..=max_balance)?;
 
+        // In 1/4 cases we're going to get max_balance account which helps us to run code to completion.
+        //
+        // Note that before there was another branch here that also did more calculation on `max_balance` to get into the sweet spot
+        // but it turns out to slightly move the balance of success/failure rate to 50/50 which is not good. With only these two branches
+        // we get around 80/20 success/failure rate. Note that this also depends on number of instructions in the program.
```
techraed, line 114: what eventually affected such adjustment from 50\50 to 80\20?

utils/runtime-fuzzer/src/generator/upload_program.rs
```diff
@@ -100,17 +102,45 @@ fn arbitrary_salt(u: &mut Unstructured) -> Result<Vec<u8>> {
...
+    unstructured: &mut Unstructured,
     programs: Option<&NonEmpty<ProgramId>>,
     codes: Option<&NonEmpty<CodeId>>,
     log_info: Option<String>,
     current_balance: Balance,
-) -> StandardGearWasmConfigsBundle {
+) -> RandomizedGearWasmConfigBundle {
+    // Observation: with balance increased and no control instructions we get *a lot* of programs executed even if we're
+    // running fuzzer for just 5 minutes. Without balance increase disabling control results in gas limit exceeding quite often.
+    let no_control = unstructured.ratio(1, 4).unwrap();
```
techraed, line 113: all `unstructured` usage logic for the `Randomized` config creation must be moved to the type's initializer function in `wasm-gen`.

utils/wasm-gen/src/config.rs
```diff
@@ -217,3 +218,69 @@ impl ConfigsBundle for StandardGearWasmConfigsBundle {
         (gear_wasm_generator_config, selectable_params)
     }
 }
+
+pub struct RandomizedGearWasmConfigBundle {
```
techraed, line 222: after defining `new_artbitrary` for `Randomized` (maybe even implementing `Arbitrary`, not sure) config, please test it the same way as `Standard` config was tested for generating wasms that are always valid for `gear_core::Code` instantiation by `Code::try_new` (see tests)

techraed (changes requested):

utils/runtime-fuzzer/src/runtime/account.rs
playX18: Before it was essentially `max_balance.saturating_div(4).max(EXISTENTIAL_DEPOSIT)` and its variations I tried, this did not result in anything useful so I've just removed that branch. It's much better to either give max balance to program or give it random balance up to max balance from countless run observations.

techraed (approved):
Great!!!

What's with checks on a server - is everything satisfiable?

PR #4034, opened 2024-06-26
refactor(runtime-interface): split sandbox interface out of runtime-interface

clearloop:
Since we hardly can avoid patching `sp-externalities` due to the requirements of https://github.com/gear-tech/gear/pull/4017, gtest would get broken after it.

for the explanation: `gear-sandbox` ( part of gtest's deps ) has `runtime-interface` as dependency, and `runtime-interface` will include gear-tasks after merging #4017 ... after a bench of the dependencies stuffs, we have to upload almost all of the substrate dependencies used in gtest which is on the way to hell

This PR trims the usages of `runtime-interface` in `gear-sandbox` since it only needs `sandbox-interface` but not the entire `gear-runtime-interface`

- [x] split out sandbox interface from runtime-interface
- [x] use sandbox interface instead of the entire runtime interface in gear-sandbox
- [x] patch gear-sandbox-interface in `utils/crates-io`

@gear-tech/dev


ukint-vs (approved):
make sense

PR #4037, opened 2024-07-02
DRAFT: Update wasmer version

ByteNacked:
Resolves # .

-
-
-

@reviewer-or-team

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -164,11 +164,11 @@ thiserror = "1.0.61"
 tokio = { version = "1.38.0" }
 uluru = "3.1.0"
 url = "2.5.0"
-wat = "1.0.83"
+wat = "1.0.71"
 wabt = "0.10.0"
-wasmer = "2.2"
-wasmer-cache = "2.2.1"
-wasmer-types = "2.2"
+wasmer = "4.3"
+wasmer-cache = "4.3"
+wasmer-types = "4.3"
 wasmi = { version = "0.14.0", default-features = false }
 wasmparser = { package = "wasmparser-nostd", version = "0.100.1", default-features = false }
 which = "4.4.2"
@@ -267,8 +267,8 @@ vara-runtime = { path = "runtime/vara", default-features = false }
 wasm-smith = { version = "0.12.21", git = "https://github.com/gear-tech/wasm-tools.git", branch = "gear-stable" }
 
 # Common executors between `sandbox-host` and `calc-stack-height`
-sandbox-wasmer = { package = "wasmer", version = "2.2", features = ["singlepass"] }
-sandbox-wasmer-types = { package = "wasmer-types", version = "2.2" }
+sandbox-wasmer = { package = "wasmer", version = "4.3", features = ["singlepass"] }
+sandbox-wasmer-types = { package = "wasmer-types", version = "4.3" }
 
 # Substrate deps
 frame-benchmarking = { version = "4.0.0-dev", git = "https://github.com/gear-tech/polkadot-sdk.git", branch = "gear-v1.3.0", default-features = false }
--- a/sandbox/host/src/sandbox.rs
+++ b/sandbox/host/src/sandbox.rs
@@ -23,7 +23,7 @@
 mod wasmer_backend;
 mod wasmi_backend;
 
-use std::{collections::HashMap, pin::Pin, rc::Rc};
+use std::{cell::RefCell, collections::HashMap, pin::Pin, rc::Rc};
 
 use codec::Decode;
 use env::Instantiate;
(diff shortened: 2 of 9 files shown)


sandbox/host/src/sandbox.rs
```diff
@@ -243,11 +254,55 @@ impl SandboxInstance {
...
+    /// Returns `Ok(Some(()))` if the global could be modified.
+    ///
+    /// # Safety
+    ///
+    /// Expected to be called only from signal handler.
+    pub unsafe fn signal_handler_set_global_val(
+        &self,
+        name: &str,
+        value: Value,
+    ) -> std::result::Result<Option<()>, error::Error> {
```
breathx, line 296:
```suggestion
    ) -> Result<Option<()>, error::Error> {
```

sandbox/host/src/sandbox/wasmer_backend.rs
```diff
@@ -309,43 +325,64 @@ pub fn instantiate(
...
     }
 
-    let mut import_object = sandbox_wasmer::ImportObject::new();
-    for (module_name, exports) in exports_map.into_iter() {
-        import_object.register(module_name, exports);
-    }
+    let mut import_object = sandbox_wasmer::Imports::new();
+    import_object.register_namespace("env", exports);
+
+    // This is sound since we're only instantiating the module and `start` function is restricted by our code checks.
```
breathx, line 348: Do you mean that there's nothing we actually run so we don't need any context because nothing will access this global value etc?

breathx (approved):

sandbox/host/src/sandbox/wasmer_backend.rs
ByteNacked: Yes, that's correct. No code can access globals or call syscalls, as the start function is not allowed by Code checks. Therefore, this context loads unnecessarily.

It seems both of these can be removed at module instantiation:
```
let instance = SandboxContextStore::using(sandbox_context, || {
        dispatch_function_env::using(&mut dispatch_func_env, || {
```

sandbox/host/src/sandbox/wasmer_backend.rs
```diff
@@ -20,28 +20,59 @@
...
-        BackendInstance, GuestEnvironment, InstantiationError, Memory, SandboxContext,
+        BackendInstanceBundle, GuestEnvironment, InstantiationError, Memory, SandboxContext,
         SandboxInstance, SupervisorFuncIndex,
     },
     util::MemoryTransfer,
 };
 
 environmental::environmental!(SandboxContextStore: trait SandboxContext);
 
+// Hack to allow mutable `environmental!` definition per module
```
ByteNacked, line 44:
```suggestion
// Hack to allow multiple `environmental!` definition per module
```

ByteNacked:
It seems the current approach doesn't really work. During syscalls, second mutable borrow errors happen.

sandbox/host/src/sandbox/wasmer_backend.rs
breathx: Could you extend comment a bit to avoid future misunderstanding please?

NikVolf:
Have you resynced the whole Vara history with it @ByteNacked ?

ByteNacked:
> Have you resynced the whole Vara history with it @ByteNacked ?

Node re-syncing for vara, vara-dev is in progress ATM.

ByteNacked:
> Update2: The problem with random segfaults was that the Module in the LRU cache outlived the Store. I fixed it.

A bit more detail about the segfaults that initially appeared after upgrading Wasmer to version 4.

This looks like an API vulnerability in Wasmer (it appears only in version 4 due to architectural changes in Wasmer) because the segfaults were caused by the Module we stored in the LRU cache containing references to a memory buffer with compiled code stored in the Store. Thus, when the Store was dropped (see [Store::clear](https://github.com/gear-tech/gear/blob/8ff08219798050aa2576b6f1d7eb2f5cd7ab017b/sandbox/host/src/sandbox.rs#L540)), the reference to this buffer outlived the Store and became a dangling pointer.

The problem was resolved by serializing the module before writing it to the LRU cache, similar to what we do when writing to the FS cache. Inside the module, there is an `Artifact` object that can be either `Archived` or `Plain`. If it is `Plain`, it can contain a reference to the compiled code in the `Store`. If it is `Archived`, it is self-contained. During serialization, it transforms into `Archived` and then gets serialized. There is no API to simply make the module `Archived` (there is, but it's low-level and very fragile according to comments in the code).

cc @grishasobol

ByteNacked:
Waiting for resync of the whole Vara history :gear:

sandbox/host/src/sandbox/wasmer_backend.rs
ByteNacked: Extended comment

ByteNacked:
@breathx Please re-review due to substantial changes.

ByteNacked:
:exclamation: :exclamation: :exclamation:  One more problem was found. Both syncs with the testnet and mainnet were successful, but during the testnet sync, the node stucks on a certain block (every sync attempt). After restarting the node, the sync continues as normal.

ByteNacked:
Waiting for one more review/approval @grishasobol  @gshep

ByteNacked:
Update:

* After discussion renamed some entities in sandbox-host crate
* Refactored/simplified code a little bit, move access to `gas_global` into `FunctionEnv`
* Added same store check in `StoreRefCell`

Cargo.toml
```diff
@@ -173,12 +174,12 @@ syn = "2.0.71"
 thiserror = "1.0.62"
 tokio = { version = "1.38.0" }
 uluru = "3.1.0"
-url = "2.5.2"
-wat = "1.0.83"
+url = "2.5.0"
+wat = "1.0.71"
```
gshep, line 178: do we need to downgrade these two?

sandbox/host/src/sandbox.rs
```diff
@@ -243,11 +251,55 @@ impl SandboxInstance {
...
+
+            BackendInstanceBundle::Wasmer { instance, store } => {
+                wasmer_set_global(instance, &mut store.borrow_mut(), name, value)
+            }
+        }
+    }
+
+    /// Get the value from a global with the given `name`. Only for usage in signal handler.
+    ///
+    /// Returns `Some(_)` if the global could be found.
```
gshep, line 266:
```suggestion
    /// Returns `Some(_)` if the global has been found.
```

sandbox/host/src/sandbox.rs
```diff
@@ -243,11 +251,55 @@ impl SandboxInstance {
...
+    ///
+    /// Expected to be called only from signal handler.
+    pub unsafe fn signal_handler_get_global_val(&self, name: &str) -> Option<Value> {
+        match &self.backend_instance {
+            BackendInstanceBundle::Wasmi(wasmi_instance) => wasmi_get_global(wasmi_instance, name),
 
-            BackendInstance::Wasmer(wasmer_instance) => {
-                wasmer_set_global(wasmer_instance, name, value)
+            BackendInstanceBundle::Wasmer { instance, store } => unsafe {
+                // We cannot use `store.borrow_mut()` in single handler context because it's already borrowed during `invoke` call.
```
gshep, line 276:
```suggestion
                // We cannot use `store.borrow_mut()` in signal handler context because it's already borrowed during `invoke` call.
```

gshep: The same in the next func too

sandbox/host/src/sandbox/wasmer_backend/store_refcell.rs
```diff
@@ -0,0 +1,343 @@
...
+pub struct StoreRefCell {
+    store: UnsafeCell<Store>,
+    state: Cell<BorrowState>,
+}
+
+#[derive(Debug)]
+pub struct BorrowScopeError;
+
+impl StoreRefCell {
+    /// Create new `StoreRefCell` with provided `Store
```
gshep, line 116:
```suggestion
    /// Create new `StoreRefCell` with provided `Store`
```

gshep (approved):
Awesome job! :100:

grishasobol (approved):
Better to make full sync once more, in order to be sure that nothing was broken in latest changes. This can be done after merging in master, but before next release.

PR #4039, opened 2024-07-02
feat(gtest): Add send_with_gas

### LINKED ISSUE #4038: Allow specifying gas for external message sendings in `gtest`
### Problem to Solve

Atm the only possible way to send message from user is with block gas limit applied. This should be extended similar to real extrinsics

### Possible Solution

Implement one more function for sending "*_with_gas"

### Notes

_No response_

holykol:
Resolves #4038

### DIFF
--- a/gtest/src/program.rs
+++ b/gtest/src/program.rs
@@ -502,16 +502,16 @@ impl<'a> Program<'a> {
     }
 
     /// Send message to the program with gas limit and value.
-    pub fn send_with_gas<ID, C>(
+    pub fn send_with_gas<ID, P>(
         &self,
         from: ID,
-        payload: C,
+        payload: P,
         gas_limit: u64,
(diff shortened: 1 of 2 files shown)


breathx:
At least single test for demonstration

gtest/src/program.rs
```diff
@@ -501,6 +501,21 @@ impl<'a> Program<'a> {
...
+    pub fn send_with_gas<ID, C>(
+        &self,
+        from: ID,
+        payload: C,
+        gas_limit: u64,
+        value: u128,
+    ) -> RunResult
+    where
+        ID: Into<ProgramIdWrapper>,
+        C: Codec,
```
breathx, line 514: Only Encode is necessary here

breathx (approved):

PR #4044, opened 2024-07-04
fix(gstd): make sure `dbg!` macro always expand to valid expression

### LINKED ISSUE #4043: `gstd::dbg!` causes compilation error when building without debug feature
### Problem

If the program source contains statements like `add(dbg!(2), 2)`, it will expand to `add(, 2)` when the `debug` feature isn't enabled on gstd.

```
  error: macro expansion ends with an incomplete expression: expected expression
    --> /workspaces/gdbg/examples/example-contract/lib.rs:12:9
     |
  12 |     add(dbg!(2), 2);
     |         ^^^^^^^ expected expression
     |
     = note: the macro call doesn't expand to an expression, but it can expand to a statement
  help: add `;` to interpret the expansion as a statement
     |
  12 |     add(dbg!(2);, 2);
     |
```

IMO it should never nuke the wrapped expression. Under release mode, it should work more like an identity operator.

### Steps

1. Cargo.toml

```
[dependencies]
gstd = { version = "1" }
```

2. src/lib.rs

```
fn add(left: u32, right: u32) -> u32 {
    left + right
}

#[no_mangle]
extern "C" fn init() {
    add(gstd::dbg!(2), 2);
}
```

3. cargo build --release

### Possible Solution

#4044

### Notes

_No response_

### Relevant Log Output

_No response_

btwiuse:
Resolves #4043

@reviewer-or-team

### DIFF
--- a/gstd/src/macros/debug.rs
+++ b/gstd/src/macros/debug.rs
@@ -65,7 +65,6 @@ macro_rules! debug {
 ///
 /// Similar to the standard library's
 /// [`dbg!`](https://doc.rust-lang.org/std/macro.dbg.html) macro.
-#[cfg(any(feature = "debug", debug_assertions))]
 #[macro_export]
 macro_rules! dbg {
     () => {
@@ -88,11 +87,4 @@ macro_rules! dbg {
     ($($val:expr),+ $(,)?) => {
         ($($crate::dbg!($val)),+,)
     };
-}
-
-#[cfg(not(any(feature = "debug", debug_assertions)))]
-#[allow(missing_docs)]
-#[macro_export]
-macro_rules! dbg {
-    ($($arg:tt)*) => {};
 }


breathx (changes requested):
For the case it should only return value. Otherwise it will be impossible to use it wout feature, so feature-dependent compilation will be broken

btwiuse:
I think it ok to make `dbg!` depend on `debug!`

cuz when debug is on, `debug!` becomes noop, so `dbg!` is equivalent to identity operation

breathx:
> I think it ok to make `dbg!` depend on `debug!`
> 
> 
> 
> cuz when debug is off, `debug!` becomes noop, so `dbg!` is equivalent to identity operation 

For some reason on first sight thought that you've not removed feature on above impl. Depending on debug! is totally correct approach, yeah

breathx (approved):

breathx:
Please, push an empty commit to trigger CI on labeling

StackOverflowExcept1on (approved):

Issue #4009, opened 2024-06-12
api::new can accept impl Into<Option<T>>


clearloop:
api::new can accept impl Into<Option<T>>

_Originally posted by @breathx in https://github.com/gear-tech/gear/pull/4003#discussion_r1634737895_
            
         
Since String doesn't have `Into<Option<String>>` by default, an `Into` implementation is not easy to go, the solution could be implementing an private trait

```rust
trait MaybeRpc {
  // NOTE: set an output generic type will overkill
  // type Output = T 
  fn as_option(&self) -> Option<&str>; // we need all of the implementations have the same result
}

// impl MaybeRpc for T where T: AsRef<str> {}
impl MaybeRpc for &str {}
impl MaybeRpc for Option<&str> {}

fn new(rpc: impl MaybeRpc) -> Result<Self>;
```

PR #4045, opened 2024-07-05
refactor(gsdk): use `Into<Option<T>>` instead of `AsOption`

clearloop:
Refactor of  #4022.
Ref  #4009

@gear-tech/dev


gsdk/src/api.rs
```diff
@@ -37,8 +37,8 @@ pub struct Api {
 
 impl Api {
     /// Create new API client.
-    pub async fn new(url: impl AsOption<str>) -> Result<Self> {
-        Self::new_with_timeout(url, None).await
+    pub async fn new(url: impl Into<Option<String>>) -> Result<Self> {
```
breathx, line 40: &str as well

breathx (approved):

gsdk/src/api.rs
StackOverflowExcept1on: mb `AsRef<str>`?

clearloop: the problem here is `String` is most frequently used for this interface ...

```
// impl Into<Option<String>>
let api = Api::new(node.ws());

// impl Into<Option<&str>>
let api = Api::new(node.ws().as_str())

// impl Into<Option<T>> where T: AsRef<str>
let api = Api::new::<String>(node.ws())
```

clearloop: there problem here is, `String` is the most frequently used case

```rust
// impl Into<Option<String>>
let api = Api::new(node.ws());
let api = Api::new(Some("".to_string()))
let api = Api::new(None)

// impl Into<Option<&str>>
let api = Api::new(node.ws().as_str())
let api = Api::new(Some(""))
let api = Api::new(None)

// impl<T: AsRef<str>> Into<Option<T>>
let api = Api::new(node.ws())
let api = Api::new::<String>(Some("".to_string()))
let api = Api::new::<String>(None)
```

StackOverflowExcept1on: makes sense if rust cannot infer type

StackOverflowExcept1on (approved):

PR #4046, opened 2024-07-07
feat(gstd): Add handle_reply to MessageFuture

### LINKED ISSUE #4040: Implement option to specify reply-hook for outgoing message
### Problem to Solve

In case of async sendings, if you want any additional logic to run on reply received, you will be forced to manually implement registry of outgoing messages and associated handlers, which is bad in terms of gstd's ux and general developing experience, because most of the things are hidden under the hood

### Possible Solution

Design:
```rust
msg::send_reply_for_bytes(ActorId::zero(), b"PAYLOAD", VALUE, REPLY_DEPOSIT) // Sends message A
    .expect("failed to send message")
    .handle_reply(|| {
        // this closure runs in handle reply, if replied to A
    })
    .await
```

How it works:
While initiating sending message that will be waiting for reply, there's Future implementor type that keeps "origin_msg_id: MessageId", on this type(-s) implement function that will interact with static mut mapping in gstd (something like HandleReplyHooks: HashMap<MessageId, Box<dyn Fn>>): and on each reply received, not only run common gstd async handler, but also check if there's associated handle reply hook specified by user (only one per each allowed).

Implement related gstd logic, propose docs, write demo and provide test coverage.

### Notes

_No response_

holykol:
* Add handle_reply method on any struct using impl_futures!
* Add new async-reply-hook example
* Test basic interaction

Resolves #4040


pallets/gear/Cargo.toml
```diff
@@ -177,27 +178,27 @@ std = [
...
-    "demo-read-big-state/wasm-wrapper",
-    "demo-proxy/wasm-wrapper",
-    "gsys",
-    "demo-reserve-gas/wasm-wrapper",
-    "demo-delayed-sender/wasm-wrapper",
-    "demo-constructor/wasm-wrapper",
-    "demo-waiter/wasm-wrapper",
-    "demo-init-wait/wasm-wrapper",
-    "demo-signal-entry/wasm-wrapper",
+	"frame-benchmarking/runtime-benchmarks",
```
breathx, line 181: why

examples/async-reply-hook/src/wasm.rs
```diff
@@ -0,0 +1,34 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use gstd::{critical, debug, exec, msg, prelude::*, ActorId};
+
+#[gstd::async_main]
+async fn main() {
```
breathx, line 22: I'd patch demo and test to send 2 messages, but apply handle reply hook only for one of them to be able to assert that they're properly called one for each, rather than just always running

gstd/src/async_runtime/mod.rs
```diff
@@ -49,9 +52,20 @@ pub(crate) fn locks() -> &'static mut LocksMap {
...
 }
 
+static mut REPLY_HOOKS: Option<HooksMap> = None;
+
+fn reply_hooks() -> &'static mut HooksMap {
+    unsafe { REPLY_HOOKS.get_or_insert_with(HashMap::new) }
+}
+
 /// Default reply handler.
 pub fn record_reply() {
```
breathx, line 62: I'd rename this now since it's not only record reply, but also calls dedicated hook for it

gstd/src/async_runtime/reply_hooks.rs
```diff
@@ -0,0 +1,42 @@
...
+
+pub(crate) type HooksMap = HashMap<MessageId, Box<dyn FnOnce()>>;
+
+/// Register hook to be executed when a reply for message_id is received.
+pub(crate) fn register_reply_hook<F: FnOnce() + 'static>(mid: MessageId, f: F) {
+    reply_hooks().insert(mid, Box::new(f));
+}
+
+/// Execute hook for message_id (if registered)
+pub(crate) fn execute_reply_hook(mid: MessageId) {
```
breathx, line 33: execute_removing or something to show that it's removed (called just once)

gstd/src/msg/macros.rs
```diff
@@ -64,6 +64,39 @@ macro_rules! impl_futures {
 
                 Ok(self)
             }
+
+            /// Execute a function when the reply is received
```
breathx, line 68: add note that it runs in handle_reply and consumes the reply gas

breathx: as a security note

breathx: you also need to forbid double hooks applied to the same future

breathx (changes requested):

gstd/src/async_runtime/mod.rs
holykol: Wouldn't it break backwards compatibility?
We can't guarantee someone hasn't already written

```
#[no_mangle]
extern "C" fn handle_reply() {
    gstd::record_reply();
}
```

in their program

pallets/gear/Cargo.toml
holykol: Right now this file has both tabs and spaces as indentation. I thought we might wanna make this consistent at least across the file

gstd/src/msg/macros.rs
holykol: added

gstd/src/async_runtime/reply_hooks.rs
holykol: renamed to execute_and_clear_reply_hook

examples/async-reply-hook/src/wasm.rs
holykol: take a look at the test now

gstd/src/msg/macros.rs
```diff
@@ -64,6 +64,44 @@ macro_rules! impl_futures {
 
                 Ok(self)
             }
+
+            /// Execute a function when the reply is received.
+            ///
+            /// Note: This callback will run in reply context and consume reply gas.
```
DennisInSky, line 70: Probably you also want to mention it is called before the continuation if the latter is still awaiting execution

gstd/src/msg/macros.rs
```diff
@@ -64,6 +64,44 @@ macro_rules! impl_futures {
...
+            ///         .await
+            ///         .expect("Received error");
+            /// }
+            /// # fn main() {}
+            /// ```
+            ///
+            /// # Panics
+            ///
+            /// Panics if this is called second time.
+            pub fn handle_reply<F: FnMut() + 'static>(self, f: F) -> Self {
```
DennisInSky, line 101: Seems static lifetime might be a problem if I start using it within Sails service which manages its data as an external reference passed to its constructor:
https://github.com/gear-tech/sails/blob/f8f4d160ef20fa746e6e5c28701463c28e765bf1/examples/demo/app/src/counter/mod.rs#L26

examples/async-reply-hook/src/wasm.rs
```diff
@@ -0,0 +1,59 @@
...
```
DennisInSky, line 22: I don't see the main test case for which this feature was conceived – receiving a reply after the message times out.

gstd/src/msg/macros.rs
DennisInSky: Take my words back. Apparently, if you implement the main idea of these functionality, I won't be able to do anything with non-static data as my service will be dead by the time of this callback is being called

breathx (changes requested):
@holykol 
UPD of initial logic impl request:
- do not clean hook on timeout
- allow setting hook only if non-zero reply-deposit set for sending

gstd/src/common/errors.rs
```diff
@@ -182,6 +186,7 @@ impl fmt::Display for UsageError {
                 write!(f, "System reservation amount can not be zero in config")
             }
             UsageError::ZeroMxLockDuration => write!(f, "Mutex lock duration can not be zero"),
+            UsageError::ZeroReplyDeposit => write!(f, "Reply deposit can not be zero"),
```
breathx, line 189: on setting reply hook

pallets/gear/src/tests.rs
```diff
@@ -14901,6 +14901,10 @@ fn handle_reply_hook() {
 
         run_to_block(4, None);
 
+        let mailbox = MailboxOf::<Test>::iter_key(USER_1)
```
breathx, line 14904: deadcode

examples/async-reply-hook/src/wasm.rs
breathx: @holykol add this test pls when main processing receives timeout and works further, and reply hook is triggered some time soon

breathx (commented):
Looks ok. Waiting for test

examples/async-reply-hook/src/wasm.rs
```diff
@@ -0,0 +1,79 @@
...
+    // Case 3: Message with reply hook and we reply to it
+    let m3 = gstd::msg::send_bytes_for_reply(source, b"for_reply_3", 0, 1_000_000_000)
+        .expect("Failed to send message")
+        .handle_reply(|| {
+            debug!("reply message_id: {:?}", msg::id());
+            debug!("reply payload: {:?}", msg::load_bytes());
+
+            assert_eq!(msg::load_bytes().unwrap(), [3]);
+
+            msg::send_bytes(msg::source(), b"saw_reply_3", 0);
```
DennisInSky, line 48: There is no any assertion this block is executed

examples/async-reply-hook/src/wasm.rs
```diff
@@ -0,0 +1,79 @@
...
+    let m1 = gstd::msg::send_bytes_for_reply(source, b"for_reply_1", 0, 0)
+        .expect("Failed to send message");
+
+    // Case 2: Message with reply hook but we don't reply to it
+    let m2 = gstd::msg::send_bytes_for_reply(source, b"for_reply_2", 0, 1_000_000_000)
+        .expect("Failed to send message")
+        .up_to(Some(5))
+        .expect("Failed to set timeout")
+        .handle_reply(|| {
+            unreachable!("This should not be called");
```
DennisInSky, line 35: @breathx do we have auto-replies for messages sent to user mailbox? Or they work just for app-2-app interactions?

pallets/gear/src/tests.rs
```diff
@@ -14830,6 +14830,112 @@ fn critical_hook_in_handle_signal() {
...
+            1_000_000_000,
+            0,
+            false,
+        ));
+        run_to_block(11, None);
+
+        // Hook should still be executed
+        assert!(
+            MailboxOf::<Test>::iter_key(USER_1).any(|(m, _)| m.payload_bytes() == b"saw_reply_4")
+        );
```
DennisInSky, line 14935: What's the guarantee this message arrived after the one at line 14918

gstd/src/async_runtime/mod.rs
vobradovich: renamed to `handle_reply_with_hook`

pallets/gear/src/tests.rs
vobradovich: added order check

examples/async-reply-hook/src/wasm.rs
DennisInSky: answered in person - we do have ones

examples/async-reply-hook/src/wasm.rs
```diff
@@ -0,0 +1,79 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
```
DennisInSky, line 17: This shouldn't be required as the main licesne resides in the repo root

examples/async-reply-hook/src/wasm.rs
DennisInSky: According to the answer, the message here has to be amended (or comment should be added) to something like - "This should be called in gas / 100 blocks, but the program exits by that time"

examples/async-reply-hook/src/wasm.rs
DennisInSky: Would be nicer to see it here rather than in test :-)

gstd/src/async_runtime/reply_hooks.rs
```diff
@@ -0,0 +1,49 @@
...
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
```
DennisInSky, line 17: Not needed :-)

DennisInSky (approved):

examples/async-reply-hook/src/wasm.rs
vobradovich: deleted it, but there's a lot of files w/ licence

gstd/src/async_runtime/reply_hooks.rs
vobradovich: all the files in the dir have a license. let's remove it in separate PR

examples/async-reply-hook/src/wasm.rs
vobradovich: added check for_reply_3 handle_reply executed, but the test is right place fot this one)

breathx (approved):

PR #4051, opened 2024-07-11
feat(gear-builtin): Separate actor id declaration

### LINKED ISSUE #4049: Builtin actor id declaration
### Problem to Solve

According to gear's design of builtin actors, there are plenty of them could be developed and added to runtime. Moreover, they may differ even between testnet and main net runtimes.

For such cases and ease of development, its much more reasonable to set the id of the builtin not using trait contents, but with some parameters in runtime.

### Possible Solution

Attaching pseudocode how I predict it could be coded:

Some pallet or crate: 
```diff
pub struct Actor<T: Config>(PhantomData<T>);

impl<T: Config> BuiltinActor for Actor<T> {
-    const ID: u64 = 1;

    type Error = BuiltinActorError;

    fn handle(dispatch: &StoredDispatch, gas_limit: u64) -> (Result<Payload, Self::Error>, u64) {
        / .. /
    }
}
```

Runtime:
```diff
pub type BuiltinActors = (
-    pallet_gear_builtin::bls12_381::Actor<Runtime>,
+   BIA<1, bls12_381::Actor<Runtime>>, // Id = 1.

-    example_dummy_bia::Actor<Runtime>,
+    BIA<2, dummy::Actor<Runtime>>, // Id = 2.
);
```

Meanwhile:
```rust
struct BIA<const ID: u64, A: BuiltinActor>(PhantomData<A>);
```

### Notes

_No response_

holykol:
* move ID out of BuiltinActor trait
* introduce ActorWithId struct one can use to assign an id to builtin at compile time
* simplify some tests as a result

Resolves #4049


pallets/gear-builtin/src/lib.rs
```diff
@@ -108,13 +109,28 @@ impl From<BuiltinActorError> for ActorExecutionErrorReplyReason {
...
+    type Error;
+    type Actor: BuiltinActor<Error = Self::Error>;
+}
+
+impl<const ID: u64, E, A: BuiltinActor<Error = E>> BuiltinActorWithId for ActorWithId<ID, A> {
+    const ID: u64 = ID;
+
+    type Error = E;
+    type Actor = A;
+}
```
breathx, line 132:
```suggestion
impl<const ID: u64, A: BuiltinActor> BuiltinActorWithId for ActorWithId<ID, A> {
    const ID: u64 = ID;

    type Error = A::Error;
    type Actor = A;
}
```

breathx (approved):

ekovalev:
Please convince me this is actually a good thing ).
In any case an actor type id is assigned at compile time, and we have a whole `u64` type to make sure every builtin actor we will ever want to introduce has its unique constant ID.
Not to mention that a builtin actor ID determines the 32-byte `ActorID` which is public and made known to contracts developers so that they could rely on the fact it will never change when writing their programs.
Or am I missing the point?

breathx:
> Please convince me this is actually a good thing ).
> 
> In any case an actor type id is assigned at compile time, and we have a whole `u64` type to make sure every builtin actor we will ever want to introduce has its unique ID assigned once for all runtimes it appears in. The latter is important if we recall that this type ID determines the 32-byte `ActorID` address which is then published and used in programs. This way the contracts developers could rely on the fact a builtin's address persists across all runtimes and wouldn't have to write different versions of the same program for different chains.
> 
> Or am I missing the point?

Discussed personally

ekovalev (approved):

PR #4060, opened 2024-07-16
feat(pallet_gear): Introduce invariant: all user non-reply messages go to mailbox

techraed:
Release notes: All instant (no delay) user non-reply ("handle") messages are now sent to mailbox.  

Completely removes `safe_gasful_send` as it introduced some implicit risky invariants. All user messages now go to mailbox, except for reply messages (as it was before)


pallets/gear/src/tests.rs
```diff
@@ -1220,7 +1169,7 @@ fn reply_deposit_to_user_auto_reply() {
             false,
         ));
 
-        run_to_next_block(None);
+        run_to_block(40, None);
```
ark0f, line 1172: It's better to calculate exact block

ark0f (approved):

core-processor/src/ext.rs
```diff
@@ -509,15 +449,21 @@ impl<'a, LP: LazyPagesInterface> ExtMutator<'a, LP> {
     fn charge_expiring_resources<T: Packet>(
         &mut self,
         packet: &T,
-        check_gas_limit: bool,
+        cover_mailbox_threshold: bool,
```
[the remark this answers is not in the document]
techraed: discussed

pallets/gear/src/tests.rs
techraed: done

pallets/gear/src/tests.rs
```diff
@@ -1663,7 +1612,7 @@ fn non_existent_code_id_zero_gas() {
         i32.const 0     ;; salt len
         i32.const 0     ;; payload ptr
         i32.const 0     ;; payload len
-        i64.const 0     ;; gas limit
+        i64.const 3000  ;; gas limit
```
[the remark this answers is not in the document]
techraed: done

Cargo.lock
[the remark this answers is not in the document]
techraed: done

breathx (changes requested):
As far as I remember, it was discussed to impl the invariant in the way that every gasLESS messages sent to user requires MB threshold so further inserted in mailbox in case of user sending. At this point it applies for all messages, which is incorrect with initial idea. Furthermore, isn't the implementation panic when sends with gas = 1 to user? Seems like this.

techraed:
Discussed: the issue is that 0 gas limit for gasful messages is a legit case.

techraed:
@breathx ptal

core-processor/src/ext.rs
```diff
@@ -506,21 +446,47 @@ impl<'a, LP: LazyPagesInterface> ExtMutator<'a, LP> {
...
+                let gas_limit = packet.gas_limit().unwrap_or(mailbox_threshold);
+
+                // Zero gasful message is a special case.
+                if gas_limit != 0 && gas_limit < mailbox_threshold {
+                    return Err(MessageError::InsufficientGasLimit.into());
+                }
+
+                Ok(gas_limit)
+            }
+            DispatchKind::Init | DispatchKind::Reply => {
```
breathx, line 474: what if we create program from code_id and salt and code_id doesnt exist, so `store_new_programs` fails and this init message will be treated as user message, isn't it?

needs double checking how we handle it

breathx (changes requested):

core-processor/src/ext.rs
breathx: discussed

breathx (approved):

PR #4062, opened 2024-07-16
fix(ethexe): run tests on free threads

StackOverflowExcept1on:


.config/nextest.toml
```diff
@@ -7,3 +7,13 @@ path = "junit.xml"
...
+
+[[profile.ci.overrides]]
+filter = 'package(ethexe-cli)'
+retries = 5
+threads-required = "num-test-threads"
+
+[[profile.ci.overrides]]
+filter = 'package(ethexe-observer)'
+retries = 5
+threads-required = "num-test-threads"
```
grishasobol, line 19: Do you think we need this any more ?

StackOverflowExcept1on: well, this may make sense since we have anvil as child process, asynchronous tasks and main thread

grishasobol: Ok

grishasobol (approved):

breathx (approved):

PR #4063, opened 2024-07-16
feat(ethexe): Add latest valid block to db, fetch committed blocks

ukint-vs:
Resolves # .

- add latest valid block to db
- fetch all commited blocks from logs once
- early exit on chain depth

@reviewer-or-team

### DIFF
--- a/ethexe/db/src/database.rs
+++ b/ethexe/db/src/database.rs
@@ -41,6 +41,18 @@ use parity_scale_codec::{Decode, Encode};
 
 const LOG_TARGET: &str = "hyper-db";
 
+enum Key {
+    LatestValidBlock,
+}
+
+impl Key {
+    fn as_bytes(&self) -> &'static [u8] {
+        match self {
+            Key::LatestValidBlock => b"latest_valid_block",
+        }
+    }
+}
+
 #[repr(u64)]
 enum KeyPrefix {
     ProgramToCodeId = 0,
@@ -232,6 +244,17 @@ impl BlockMetaStorage for Database {
     fn set_block_outcome(&self, block_hash: H256, outcome: Vec<StateTransition>) {
         self.kv
             .put(&KeyPrefix::BlockOutcome.one(block_hash), outcome.encode());
+    }
+
+    fn latest_valid_block(&self) -> Option<H256> {
+        self.kv
+            .get(Key::LatestValidBlock.as_bytes())
+            .map(|block_hash| H256::from_slice(&block_hash))
+    }
+
+    fn set_latest_valid_block(&self, block_hash: H256) {
+        self.kv
+            .put(Key::LatestValidBlock.as_bytes(), block_hash.0.to_vec());
     }
 }
 
--- a/ethexe/observer/src/query.rs
+++ b/ethexe/observer/src/query.rs
@@ -8,15 +8,17 @@ use crate::{
     BlobReader,
 };
 use alloy::{
-    primitives::Address as AlloyAddress,
+    eips::BlockNumberOrTag,
+    primitives::{Address as AlloyAddress, B256},
     providers::{Provider, ProviderBuilder},
-    rpc::types::eth::BlockTransactionsKind,
+    rpc::types::{eth::BlockTransactionsKind, Filter},
 };
 use anyhow::{anyhow, Result};
 use ethexe_common::{
     db::{BlockHeader, BlockMetaStorage},
-    events::BlockEvent,
+    events::{BlockCommitted, BlockEvent},
 };
+use ethexe_ethereum::event::{match_log, signature_hash};
 use ethexe_signer::Address;
 use gprimitives::{ActorId, CodeId, H256};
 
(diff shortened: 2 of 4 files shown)


ethexe/observer/src/query.rs
```diff
@@ -70,35 +72,113 @@ impl Query {
...
+                } else {
+                    Ok((block_hash, header))
+                }
+            }
+        } else {
+            log::warn!("Latest valid block not found, sync to genesis.");
+            Ok((
+                self.genesis_block_hash,
+                self.get_block_header_meta(self.genesis_block_hash).await?,
+            ))
```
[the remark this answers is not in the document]
grishasobol: nevertheless this code may stay as is

ethexe/db/src/database.rs
```diff
@@ -41,6 +41,18 @@ use parity_scale_codec::{Decode, Encode};
...
+    fn as_bytes(&self) -> &'static [u8] {
+        match self {
+            Key::LatestValidBlock => b"latest_valid_block",
+        }
+    }
+}
+
 #[repr(u64)]
 enum KeyPrefix {
     ProgramToCodeId = 0,
```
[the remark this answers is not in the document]
ukint-vs: I think it’s better to have static single keys and prefixes separate. I thought about `KeyPrefix::LatestValidBlock::one(&[])` but it look strange. As you wish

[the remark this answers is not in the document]
grishasobol: > I think it’s better to have static single keys and prefixes separate

Static single key does not guarantee that you do not have intersections with other keys. On other hand using only `KeyPrefix` - guaranties that

ethexe/observer/src/query.rs
```diff
@@ -70,35 +72,113 @@ impl Query {
             .collect())
     }
 
+    async fn get_all_committed_blocks(
```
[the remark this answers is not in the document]
ukint-vs: Resolved by DEEP_SYNC const. We will use batches if height diff is more than const

ethexe/observer/src/query.rs
```diff
@@ -39,74 +45,207 @@ impl Query {
...
-    pub async fn get_last_committed_chain(&mut self, block_hash: H256) -> Result<Vec<H256>> {
-        let mut chain = Vec::new();
-        let mut committed_blocks = BTreeSet::new();
+    /// Populate database with blocks using rpc provider.
+    async fn load_chain_batch(
+        &mut self,
+        from_block: u32,
+        to_block: u32,
+    ) -> Result<(Vec<H256>, Vec<H256>)> {
+        let total_blocks = to_block - from_block + 1;
```
grishasobol, line 100: saturating_sub

ethexe/observer/src/query.rs
```diff
@@ -39,74 +45,207 @@ impl Query {
...
+                latest_valid_block_height,
+                self.max_commitment_depth
+            ));
         }
 
+        // Determine if deep sync is needed
+        let is_deep_sync = {
+            // Current block can be lower than latest valid due to reorgs.
+            let block_diff = (current_block.height as i64 - latest_valid_block_height as i64)
+                .unsigned_abs() as u32;
```
grishasobol, line 223: abs is useless. `saturating_sub` should be used. use it everywhere or `checked_sub`.

ethexe/observer/src/query.rs
```diff
@@ -39,74 +45,207 @@ impl Query {
...
+        let (mut chain, committed_blocks) = if is_deep_sync {
+            // Load all blocks from provider by numbers, skip latest valid.
+            self.load_chain_batch(latest_valid_block_height + 1, current_block.height)
+                .await?
+        } else {
+            // Load chain by parent hashes.
+            self.load_chain(block_hash).await?
+        };
+
+        chain.dedup();
```
grishasobol, line 237: What the reason use this ?

ethexe/observer/src/query.rs
```diff
@@ -39,74 +45,207 @@ impl Query {
...
+                block_hash,
+                blocks_events.remove(&block_hash).unwrap_or_default(),
+            );
+            committed_blocks.extend(self.get_committed_blocks(block_hash).await?);
+            block_hashes.push(block_hash);
+        }
+        block_hashes.reverse();
+        log::trace!("{} blocks loaded", block_hashes.len());
+
+        Ok((block_hashes, committed_blocks))
```
grishasobol, line 162: `block_hashes` was collected using block numbers which is not safe. 

More safe solution is to set data in database for all blocks `from_number..to_number` and that __ALL__ where we use numbers - only to access data for blocks.

After that the same logic from load_chain must be used. Because we have already loaded all blocks it will be very fast.

grishasobol (changes requested):

grishasobol (approved):

ethexe/observer/src/query.rs
grishasobol: Solved

PR #4064, opened 2024-07-16
tests(ethexe):  append initial `ethexe-cli` integration testing

grishasobol:
1) Append env for testing
2) Append ping test
3) Some minor fixes

### DIFF
--- a/ethexe/cli/src/tests.rs
+++ b/ethexe/cli/src/tests.rs
@@ -1,0 +1,309 @@
+// This file is part of Gear.
+//
+// Copyright (C) 2024 Gear Technologies Inc.
+// SPDX-License-Identifier: GPL-3.0-or-later WITH Classpath-exception-2.0
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//! Integration tests.
+
+use crate::service::Service;
+use alloy::node_bindings::Anvil;
+use anyhow::{anyhow, Result};
+use ethexe_common::events::BlockEvent;
+use ethexe_db::{Database, MemDb};
+use ethexe_ethereum::{Ethereum, RouterQuery};
+use ethexe_observer::{Event, MockBlobReader, Observer, Query};
+use ethexe_processor::Processor;
+use ethexe_sequencer::Sequencer;
+use ethexe_signer::Signer;
+use ethexe_validator::Validator;
+use futures::StreamExt;
+use gprimitives::{ActorId, CodeId, H256};
+use std::sync::Arc;
+
+struct Listener {
+    receiver: tokio::sync::mpsc::Receiver<Event>,
+    _handle: tokio::task::JoinHandle<()>,
+}
+
+impl Listener {
+    pub fn new(mut observer: Observer) -> Self {
+        let (sender, receiver) = tokio::sync::mpsc::channel::<Event>(8 * 1024 * 1024);
+
+        let _handle = tokio::task::spawn(async move {
+            let observer_events = observer.events();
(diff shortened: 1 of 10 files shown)


ethexe/cli/src/tests.rs
```diff
@@ -0,0 +1,309 @@
...
+            .add_blob_transaction(tx_hash, code.to_vec())
+            .await;
+        Ok((tx_hash, code_id))
+    }
+}
+
+#[tokio::test(flavor = "multi_thread")]
+#[ntest::timeout(60_000)]
+async fn ping() {
+    let _ = env_logger::try_init();
```
StackOverflowExcept1on, line 212: use gear-utils for that

ethexe/cli/src/tests.rs
```diff
@@ -0,0 +1,309 @@
...
+
+    let anvil = Anvil::new().try_spawn().unwrap();
+
+    let mut env = TestEnv::new(anvil.ws_endpoint()).await.unwrap();
+    let mut listener = env.new_listener();
+
+    let service = env.service.take().unwrap();
+    let _ = tokio::task::spawn(service.run());
```
StackOverflowExcept1on, line 220: would be nice to join this gracefully

grishasobol: Add abort presently. May be will make some better solution in future.

ethexe/cli/src/tests.rs
grishasobol: Done

ethexe/cli/src/tests.rs
```diff
@@ -0,0 +1,328 @@
...
+use std::sync::Arc;
+
+struct Listener {
+    receiver: tokio::sync::mpsc::Receiver<Event>,
+    _handle: tokio::task::JoinHandle<()>,
+}
+
+impl Listener {
+    pub fn new(mut observer: Observer) -> Self {
+        let (sender, receiver) = tokio::sync::mpsc::channel::<Event>(8 * 1024 * 1024);
```
StackOverflowExcept1on, line 43: looks like overkill

ethexe/cli/src/tests.rs
```diff
@@ -0,0 +1,328 @@
...
+use ethexe_sequencer::Sequencer;
+use ethexe_signer::Signer;
+use ethexe_validator::Validator;
+use futures::StreamExt;
+use gprimitives::{ActorId, CodeId, H256};
+use std::sync::Arc;
+
+struct Listener {
+    receiver: tokio::sync::mpsc::Receiver<Event>,
+    _handle: tokio::task::JoinHandle<()>,
```
StackOverflowExcept1on, line 38: add `use tokio` instead of full path to types

ethexe/cli/src/tests.rs
```diff
@@ -0,0 +1,328 @@
...
+
+    let (_, code_id) = env.upload_code(demo_ping::WASM_BINARY).await.unwrap();
+
+    log::info!("📗 Waiting for code loaded");
+    listener
+        .apply_until(|event| {
+            if let Event::CodeLoaded(loaded) = event {
+                assert_eq!(loaded.code_id, code_id);
+                assert_eq!(loaded.code.as_slice(), demo_ping::WASM_BINARY);
+                Ok(Some(()))
```
StackOverflowExcept1on, line 230: `Result<Option<R>>` is more flexible, but here `Option<R>` is enough. why can't we just have something like `events.find(predicate)` and then do assertion when event is found?

grishasobol: Because we have `Observer::Event` and `BlockEvent`. One observer event contains maybe several `BlockEvent`, so we need to check them in one `call` like `apply_until_block_event`, or return all of that we need in one `call`.

Anyway, we can make find also. But let's to this in future refactorings

ethexe/cli/src/tests.rs
grishasobol: I think it's not overkill because we have CodeLoaded event which contains code bytes. So, our limits for code size is 512 kBytes,  so 16 CodeLoaded events may overload this buffer.

ethexe/cli/src/tests.rs
grishasobol: Done

ethexe/cli/src/tests.rs
StackOverflowExcept1on: but this is an event channel, not bytes (`channel::<T = Event>`)

grishasobol: You are right. Set it to 1024

StackOverflowExcept1on (approved):

PR #4069, opened 2024-07-19
chore(ethexe): Add bootnodes, update chain-spec

ukint-vs:
Resolves # .

-
-
-

@reviewer-or-team

### DIFF
--- a/ethexe/cli/src/chain_spec.rs
+++ b/ethexe/cli/src/chain_spec.rs
@@ -14,7 +14,7 @@ pub struct ChainSpec {
 
 pub fn testnet_config() -> ChainSpec {
     ChainSpec {
-        ethereum_router_address: "0x05069E9045Ca0D2B72840c6A21C7bE588E02089A".to_string(),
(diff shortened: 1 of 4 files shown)


ethexe/cli/src/chain_spec.rs
```diff
@@ -15,7 +15,10 @@ pub struct ChainSpec {
 pub fn testnet_config() -> ChainSpec {
     ChainSpec {
         ethereum_router_address: "0x05069E9045Ca0D2B72840c6A21C7bE588E02089A".to_string(),
```
StackOverflowExcept1on, line 17: `testnet_config()` is used when the cli option `--chain-spec` is set to `dev`
```suggestion
        ethereum_router_address: "0xf90d2956B6F36194fbac181Fb95b2f67274821da".to_string(),
```

StackOverflowExcept1on: we should also think about `ethexe/chain_spec.toml`. perhaps it should be removed and added to `.gitignore`, because it allows to override the chainspec and make it different from `--chain-spec dev`

StackOverflowExcept1on (approved):

PR #4072, opened 2024-07-21
feat(gstd): stabilize program logs

clearloop:
Resolves # .

Different from `gstd::debug!`, developers may have `log` requirements while building programs in production, e.g. using an indexer to fetch the historical events.

The demand is similar to `log{n}` in EVM's instructions, but since our pallet haven't had an event just for program logging atm (_this is also an option_), our current implementation could be more closed to solana's logs, sending messages to another program, this PR uses `Actor::zero` as a logger program

@gear-tech/dev

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3437,6 +3437,14 @@ dependencies = [
 ]
 
 [[package]]
+name = "demo-log"
+version = "0.1.0"
+dependencies = [
+ "gstd",
+ "gtest",
+]
+
+[[package]]
 name = "demo-messenger"
 version = "0.1.0"
 dependencies = [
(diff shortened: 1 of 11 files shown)


breathx (changes requested):
Similar approach is used in sails for a while.
I'm super against of macros here, so if you really want such functionality to be a part of gstd, I'd go with just gstd::msg::emit/emit_bytes(..) that recalls send with 0 value and 0 destination

@DennisInSky wdyt?

breathx:
Also there's no actually need in testing this)

clearloop:
closing this since these kind of could be added to sails dynamically

PR #4076, opened 2024-07-22
feat(runtime-fuzzer): use previously waited message id as wake syscall param

ByteNacked:
In the runtime fuzzer:

- When generating a syscall from wait-family, it stores the current message ID in the specified Wasm memory storage for later use.
- When `wake` syscall is called, this previously stored/waited message ID is used as a parameter for the wake syscall.

@reviewer-or-team

### DIFF
--- a/utils/wasm-gen/src/config/syscalls/param.rs
+++ b/utils/wasm-gen/src/config/syscalls/param.rs
@@ -127,6 +127,7 @@ impl SyscallsParamsConfig {
             }
             PtrParamAllowedValues::ReservationId => Ptr::Hash(HashType::ReservationId),
             PtrParamAllowedValues::CodeIdsWithValue { .. } => Ptr::HashWithValue(HashType::CodeId),
+            PtrParamAllowedValues::WaitedMessageId => Ptr::Hash(HashType::MessageId),
         };
 
         self.ptr.insert(ptr, allowed_values);
@@ -272,6 +273,8 @@ pub enum PtrParamAllowedValues {
         code_ids: NonEmpty<CodeId>,
         range: RangeInclusive<u128>,
     },
+    /// Variant of `Ptr::Hash` pointer type, where hash is waited message id.
+    WaitedMessageId,
 }
 
 /// Actor kind, which is actually a syscall destination choice.
--- a/utils/wasm-gen/src/config/syscalls/param.rs
+++ b/utils/wasm-gen/src/config/syscalls/param.rs
@@ -97,6 +97,7 @@ impl SyscallsParamsConfig {
                 range,
             })
             .with_ptr_rule(PtrParamAllowedValues::ReservationId)
+            .with_ptr_rule(PtrParamAllowedValues::WaitedMessageId)
     }
 
     /// Set rules for a regular syscall param.
--- a/utils/wasm-gen/src/generator/syscalls/invocator.rs
+++ b/utils/wasm-gen/src/generator/syscalls/invocator.rs
@@ -34,8 +34,8 @@ use gear_utils::NonEmpty;
 use gear_wasm_instrument::{
     parity_wasm::elements::{BlockType, Instruction, Internal, ValueType},
     syscalls::{
-        FallibleSyscallSignature, HashType, ParamType, Ptr, RegularParamType, SyscallName,
-        SyscallSignature, SystemSyscallSignature,
+        FallibleSyscallSignature, ParamType, Ptr, RegularParamType, SyscallName, SyscallSignature,
+        SystemSyscallSignature,
     },
 };
 use gsys::{ErrorCode, Handle, Hash};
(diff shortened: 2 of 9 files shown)


utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -823,6 +834,29 @@ impl<'a, 'b> SyscallsInvocator<'a, 'b> {
         }
     }
 
+    fn store_waited_message_id(&self, instructions: &mut Vec<Instruction>) {
+        let Some(gr_message_id_indexes_handle) = self
+            .syscalls_imports
+            .get(&InvocableSyscall::Loose(SyscallName::MessageId))
+            .map(|&(_, call_indexes_handle)| call_indexes_handle as u32)
+        else {
+            // We automatically enable the `message_id` syscall import if the `wait` syscall is enabled in the config.
```
StackOverflowExcept1on, line 843: I would panic here via `ret.expect("msg")` instead of just `return;`

utils/wasm-gen/src/tests.rs
```diff
@@ -206,6 +213,120 @@ fn test_avoid_waits_works() {
...
+
+        assert!(matches!(
+            termination_reason,
+            TerminationReason::Actor(ActorTerminationReason::Wait(..))
+        ));
+    }
+}
+
+#[test]
+fn test_wake_uses_stored_message_id() {
```
StackOverflowExcept1on, line 282: As I see it, test first writes message id into memory and then calls `gr_wake`. I'm not sure if this is a good approach, because we want to check that syscall sequence `gr_message_id`, `gr_wake` causes `gr_wake` to use the last message id.

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -823,6 +831,29 @@ impl<'a, 'b> SyscallsInvocator<'a, 'b> {
...
```
techraed, line 840: I would panic here, because of not having `gr_message_id` when it's required to be in case of inserted `gr_wait*` functions.

utils/wasm-gen/src/generator/syscalls/invocator.rs
```diff
@@ -823,6 +831,29 @@ impl<'a, 'b> SyscallsInvocator<'a, 'b> {
...
+            // If not, then we don't need to store the message ID.
+            return;
+        };
+
+        let memory_layout = MemoryLayout::from(self.memory_size_bytes());
+        let start_offset = memory_layout.waited_message_id_ptr;
```
techraed, line 846: won't that overwrite other resources?

@StackOverflowExcept1on ptal

techraed (changes requested):
Make sure there's no possibility of having `gr_message_id` not being set in the config while having `gr_wait*`. Currently that's possible.

utils/wasm-gen/src/generator/syscalls/invocator.rs
StackOverflowExcept1on: The macro for memory_layout is designed so that you can easily add new pointers, but this may change the output of wasm-gen.

utils/wasm-gen/src/config/syscalls/param.rs
```diff
@@ -127,6 +128,7 @@ impl SyscallsParamsConfig {
             }
             PtrParamAllowedValues::ReservationId => Ptr::Hash(HashType::ReservationId),
             PtrParamAllowedValues::CodeIdsWithValue { .. } => Ptr::HashWithValue(HashType::CodeId),
+            PtrParamAllowedValues::WaitedMessageId => Ptr::Hash(HashType::MessageId),
```
StackOverflowExcept1on, line 131: ReplyDeposit also uses the Ptr::Hash(HashType::MessageId) signature. Could this lead to conflict?

techraed: I guess no. It's not required to work for loose syscall, only for precise

utils/wasm-gen/src/tests.rs
```diff
@@ -206,6 +213,120 @@ fn test_avoid_waits_works() {
...
+        },
+    );
+
+    let info = ext.into_ext_info(&mut store, &memory).unwrap();
+    let msg_id = info.awakening.first().unwrap().0;
+    let msg_id_bytes = msg_id.into_bytes();
+    assert_eq!(
+        u64::from_le_bytes(msg_id_bytes[..8].try_into().unwrap()),
+        EXPECTED_MSG_ID
+    );
```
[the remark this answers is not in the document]
techraed: Why remove the check for the message id?

utils/wasm-gen/src/config/syscalls/param.rs
StackOverflowExcept1on: Yes, but what if the fuzzer generates config with loose ReplyDeposit syscall?

StackOverflowExcept1on (approved):
LGTM! I ran the tests with `RUST_LOG="syscalls=trace"` and everything seems fine. The only thing we could think about in future refactorings is preserving the execution context. Unfortunately, now we cannot execute `message_id`, `wait`, and then save the context (memory, etc.) and call `wake`.

ByteNacked:
> Make sure there's no possibility of having gr_message_id not being set in the config while having gr_wait*. Currently that's possible.

Fixed

techraed (approved):
nice!

PR #4089, opened 2024-07-26
feat(gtest): Introduce reservations for the `gtest`

techraed:
Release notes: introducing gas reservations creation/deletion and usage for message sending on `gtest`.

TODO:
- [ ] implement `reserve_gas`
- [ ] implement `unreserve_gas`
- [ ] implement message sending from reservations

### DIFF
--- a/gtest/src/manager.rs
+++ b/gtest/src/manager.rs
@@ -110,7 +110,7 @@ impl TestActor {
         matches!(self, TestActor::Uninitialized(..))
     }
 
-    fn genuine_program(&self) -> Option<&GenuineProgram> {
+    pub(crate) fn genuine_program(&self) -> Option<&GenuineProgram> {
         match self {
             TestActor::Initialized(Program::Genuine(program))
             | TestActor::Uninitialized(_, Some(Program::Genuine(program))) => Some(program),
@@ -1003,6 +1003,7 @@ impl JournalHandler for ExtManager {
         message_id: MessageId,
         dispatch: Dispatch,
         bn: u32,
+        // todo add sending from reservation
         _reservation: Option<ReservationId>,
     ) {
         if bn > 0 {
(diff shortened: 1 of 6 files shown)


playX18 (approved):

gtest/src/manager.rs
```diff
@@ -948,6 +948,40 @@ impl ExtManager {
...
+                    None
+                })
+        });
+
+        res.map(|_| {
+            self.gas_tree
+                .consume(reservation)
+                .unwrap_or_else(|e| unreachable!("GasTree corrupted! {:?}", e));
+        })
+    }
```
ByteNacked, line 971: It might be more clear to use a simple `bool` as the return value instead of `Option<()>`. 
Additionally, the map(|_| ..) logic seems a bit complex; perhaps consider using bools for better clarity.

ByteNacked:
LGTM, approved :+1:

ByteNacked (approved):

gtest/src/manager.rs
ark0f: Or consider to merge blocks and use `if let Some(_reservation)`

gtest/src/manager.rs
```diff
@@ -948,6 +948,40 @@ impl ExtManager {
...
+
+    pub(crate) fn adjust_genuine_program<R, F: FnOnce(Option<&mut GenuineProgram>) -> R>(
```
ark0f, line 973:
```suggestion
    pub(crate) fn update_genuine_program<R, F: FnOnce(Option<&mut GenuineProgram>) -> R>(
```
for consistency with program storage

ark0f, line 973:
```suggestion
    pub(crate) fn adjust_genuine_program<R, F: FnOnce(&mut GenuineProgram) -> R>(
```
Every usage unwraps `Option`. Consider to unwrap it in function itself

gtest/src/program.rs
```diff
@@ -1142,12 +1144,12 @@ mod tests {
         assert!(results.iter().any(|result| result.contains(&log)));
     }
 
+    // Possible solution for the issue#3699 worked.
+    // the test must not panic.
```
ark0f, line 1148: Can we just close the issue then?

ark0f (approved):

gtest/src/manager.rs
techraed: > Or consider to merge blocks and use `if let Some(_reservation)`

unfortunately can't do that, as `adjust_genuine_program` actually has a mutable access to the self, while `self.gas_tree` ops - immutable, can't do that.

gtest/src/manager.rs
techraed: well than you have to return from the `adjust_genuine_program` the `Option` then, as the option inside the method won't be unwraped, as it's wrong.

What's more, with such an API you will have to handle 2 results - 1) for the `op` one and for the `adjust_genuine_program`

techraed: done

techraed: done

gtest/src/manager.rs
techraed: > It might be more clear to use a simple `bool` as the return value instead of `Option<()>`. Additionally, the `map(|_| ...)` logic seems a bit complex; perhaps consider using bools for better clarity.

changed in favor of bool

PR #4102, opened 2024-07-30
refactor(pallet_gear, core-processor): Send all user messages from reservation to mailbox

techraed:
Any gasful message sent to user, whether it's sent with a delay or not, reaches mailbox, except for messages sent from reservations, although the reservation minimal amount is a mailbox threshold value. That's because reservation messages with a delay are checked only for being enough for a stash hold, but not for the mailbox. This PR fixes that inconsistency with a basic invariant that all non-zero gasful user messages reach mailbox.

### DIFF
--- a/pallets/gear/src/tests.rs
+++ b/pallets/gear/src/tests.rs
@@ -762,11 +762,7 @@ fn value_counter_set_correctly_for_interruptions() {
         .source("source_store")
         .value("value_store")
         .value_available_as_vec("value_available_store")
-        .send_wgas(
-            "source_store",
-            "value_available_store",
-            <Test as Config>::MailboxThreshold::get(),
-        )
+        .send_wgas("source_store", "value_available_store", 0)
         .send_value(Arg::new([0u8; 32]), Arg::new(vec![]), "value_store")
         .wait_for(1);
 
@@ -795,7 +791,6 @@ fn value_counter_set_correctly_for_interruptions() {
 
         run_to_next_block(None);
         let msg = maybe_last_message(USER_1).expect("Message should be");
-        assert!(MailboxOf::<Test>::contains(&USER_1, &msg.id()));
         let value_available =
             u128::decode(&mut msg.payload_bytes()).expect("Failed to decode value available");
         assert_eq!(value_available, INIT_VALUE + VALUE);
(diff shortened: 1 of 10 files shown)


pallets/gear/src/tests.rs
```diff
@@ -301,7 +301,9 @@ fn calculate_gas_zero_balance() {
 
 #[test]
 fn delayed_send_from_reservation_not_for_mailbox() {
```
ark0f, line 303:
```suggestion
fn delayed_send_from_reservation_not_anywhere() {
```

pallets/gear/src/tests.rs
```diff
@@ -320,13 +322,22 @@ fn delayed_send_from_reservation_not_for_mailbox() {
...
         assert!(Gear::is_initialized(pid));
 
+        let reservation_amount = <Test as Config>::MailboxThreshold::get();
+        let sending_delay = 1u32;
+        let sending_delay_hold_bound: HoldBound<Test> =
+            HoldBoundBuilder::new(StorageType::DispatchStash).duration(sending_delay as u64);
+        let sending_delay_gas = sending_delay_hold_bound.lock_amount();
+        // Sending delayed msg from a reservation checks: reservation_amount - delay_hold - mailbox_therhsold.
+        // Current example sets reseravtion_amount to mailbox_threshold. So, obviously, sending message
+        // from a reservation is impossible - reservation has insufficient gas reserved.
```
ark0f, line 332: Do we have test for exact gas amount of mailbox threshold + dispatch stash holding fee? Would be nice to have

pallets/gear/src/tests.rs
```diff
@@ -320,13 +322,22 @@ fn delayed_send_from_reservation_not_for_mailbox() {
...
+        assert_ne!(sending_delay_gas, 0);
```
ark0f, line 333: Looks like this assertion does nothing

ark0f (approved):

pallets/gear/src/tests.rs
techraed: The test is broken if costs for holding are set wrong. The comment explains that.

pallets/gear/src/tests.rs
techraed: renamed

pallets/gear/src/tests.rs
techraed: Adjusted the `delayed_reservations_to_mailbox` with a calculation for an exact reservation amount

PR #4119, opened 2024-08-07
refactor(gtest): Change block execution model in `gtest` to be similar to the vara-runtime

techraed:
Release notes:
1. Sending message to program simply means adding message to the queue, not executing block immediately as it was done previously. So `program.send(...)` now doesn't trigger block execution.
2. The block execution model is now similar to vara-runtime block execution. Block execution consists of 2 steps: scheduled tasks processing and message processing.
3. To start block execution one of the following methods must be called: `run_next_block`, `run_next_block_with_allowance`, `run_to_block`. Scheduled tasks can be processed separately without processing messages by calling `run_scheduled_tasks`.
4. Instead of `RunResult` a new `BlockRunResult` is introduced, which gives some report data of the block execution.
5. Sending signals from the `Program` API is removed as it violates the rules of the runtime: signals are sent only on panic automatically after failed execution, as it's done on the vara runtime (see https://github.com/gear-tech/gear/commit/bcba7c0bbd08a3de9ed59ffcd5350326f3ca2865).

TODO:
- [x] resolve todos (open corresponding issues)
- [x] fix workspace tests using `gtest`


playX18 (approved):

gtest/src/system.rs
```diff
@@ -174,13 +174,69 @@ impl System {
...
+    /// block tasks:
+    /// - delayed sending
+    /// - waking message
+    /// - removing from the mailbox
+    /// - removing reservations
+    /// - removing stalled wait message.
+    ///
+    /// Messages processing executes messages until either queue becomes empty
+    /// or block gas allowance is fully consumed.
+    pub fn run_next_block(&self) -> BlockRunResult {
```
ByteNacked, line 202: Should it be named run_to_next_block? Or maybe simplified to run_block?

From an API standpoint, this naming seems a bit odd. If a function is named `run_next_block`, what happens to the current one? Where is the `run_current_block function`, etc.?

gtest/src/manager.rs
```diff
@@ -398,9 +396,95 @@ impl ExtManager {
...
+                drop(actors);
+                self.process_dormant(balance, dispatch);
+            } else if let Some((data, code)) = actor.get_executable_actor_data() {
+                drop(actors);
+                self.process_normal(balance, data, code, dispatch);
+            } else if let Some(mock) = actor.take_mock() {
+                drop(actors);
+                self.process_mock(mock, dispatch);
+            } else {
+                unreachable!();
```
ByteNacked, line 481: Manual drops seem very error-prone; maybe it could be improved with a `match`?

ByteNacked (approved):
LGTM :+1:

gtest/src/system.rs
techraed: discussed

gtest/src/manager.rs
techraed: discussed

PR #4145, opened 2024-08-12
feat(common): introduce AuxiliaryStorageWrap to remove boilerplate code

playX18:


common/src/auxiliary/mod.rs
```diff
@@ -136,3 +140,118 @@ impl<K1, K2, V> Default for DoubleBTreeMap<K1, K2, V> {
...
+    fn contains_keys(key1: &Self::Key1, key2: &Self::Key2) -> bool {
+        T::with_storage_mut(|map| map.contains_keys(key1, key2))
+    }
+
+    fn mutate<R, F: FnOnce(&mut Option<Self::Value>) -> R>(
+        _key1: Self::Key1,
+        _key2: Self::Key2,
+        _f: F,
+    ) -> R {
+        unimplemented!()
```
breathx, line 195: shouldn't these be implemented already?

common/src/auxiliary/mod.rs
```diff
@@ -136,3 +140,118 @@ impl<K1, K2, V> Default for DoubleBTreeMap<K1, K2, V> {
         Self::new()
     }
 }
+
+/// An auxiliary storage wrapper type.
+///
+/// Implements DoubleMapStorage and traits like [`IterableByKeyMap`] for such type automatically.
+pub trait AuxiliaryStorageWrap {
+    type Key1: Ord + Clone;
```
breathx, line 148: storage wrap could be just single map

breathx: so assocs here seem incorrect to me

breathx (changes requested):

common/src/auxiliary/mod.rs
breathx: it should be just across T, where T is stored in thread local

breathx (approved):

common/src/auxiliary/mod.rs
playX18: resolved in DM, we'll settle on using two traits for different storages

Issue #3565, opened 2023-12-04
Check publishing with cargo-vendor and cargo-package


clearloop:
### Problem to Solve

> run `cargo package` on everything which verifies the build as-if it were published to crates.io. This requires using an incrementally-built directory registry generated from `cargo vendor` because the versions referenced from `Cargo.toml` may not exist on crates.io.

https://github.com/bytecodealliance/wasmtime/blob/main/scripts/publish.rs#L479

### Possible Solution

see the example in wasmtime https://github.com/bytecodealliance/wasmtime/blob/main/scripts/publish.rs#L509

clearloop:
https://github.com/rust-lang/cargo/issues/9172

clearloop:
This is currently impossible since we have too many local packages

PR #4147, opened 2024-08-12
feat(crates-io): simulate packages publishing

### LINKED ISSUE #4142: feat(crates-io): simulate package publishing
### Problem to Solve

We had some issues publishing packages: https://github.com/gear-tech/gear/actions/workflows/crates-io.yml
Mostly because someone made changes without looking at what should be in `SAFE_DEPENDENCIES`, `STACKED_DEPENDENCIES`. So, it would be nice to be able to run publishing with `--dry-run` flag and have it controlled from some environment variable. It also makes sense to introduce this check on CI.

### Possible Solution

n/a

### Notes

_No response_

StackOverflowExcept1on:
Resolves #4142

This PR is aimed at simplifying debugging of package publishing. Package publishing will be checked on temporary cargo registry, which will allow make additional check on CI.

TODO:
- [x] refactor the code, add rollback of patched files
- [x] use [cargo-http-registry](https://github.com/d-e-s-o/cargo-http-registry), use their api or just write bash script
- [x] write github workflow

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -94,8 +94,8 @@ jobs:
           cargo +stable check --manifest-path utils/wasm-builder/test-program/Cargo.toml
           cargo +stable check --manifest-path utils/cargo-gbuild/test-program/Cargo.toml --workspace --target wasm32-unknown-unknown
 
-      - name: "Check: crates-io packages"
-        run: cargo +stable run --release -p crates-io check
+      - name: "Check: crates-io packages publishing"
+        run: cargo +stable run --release -p crates-io publish --simulate --registry-path /tmp/cargo-http-registry
 
   fuzzer:
     runs-on: [kuberunner, github-runner-01]
--- a/utils/crates-io/src/publisher.rs
+++ b/utils/crates-io/src/publisher.rs
@@ -23,7 +23,7 @@ use crate::{
 };
 use anyhow::{bail, Result};
 use cargo_metadata::{Metadata, MetadataCommand};
-use std::{iter, path::PathBuf};
+use std::path::PathBuf;
 
 /// crates-io packages publisher.
 pub struct Publisher {
@@ -87,10 +87,12 @@ impl Publisher {
 
     /// Restore local files
     pub fn restore(&self) -> Result<()> {
-        if let Some(manifests) = self.manifests() {
-            manifests
-                .map(|manifest| manifest.restore())
-                .collect::<Result<Vec<_>>>()?;
+        self.manifests()
+            .map(|manifest| manifest.restore())
+            .collect::<Result<Vec<_>>>()?;
+
+        if let Some(workspace) = self.workspace.as_ref() {
+            workspace.lock_file().restore()?;
         }
 
         if let Some(simulator) = self.simulator.as_ref() {
@@ -102,11 +104,9 @@ impl Publisher {
 
     /// Patch local files
     fn patch(&self) -> Result<()> {
-        if let Some(manifests) = self.manifests() {
-            manifests
-                .map(|manifest| manifest.patch())
-                .collect::<Result<Vec<_>>>()?;
-        }
+        self.manifests()
+            .map(|manifest| manifest.patch())
+            .collect::<Result<Vec<_>>>()?;
 
         if let Some(simulator) = self.simulator.as_ref() {
             simulator.patch()?;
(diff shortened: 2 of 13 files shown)


utils/crates-io/src/publisher.rs
```diff
@@ -61,53 +61,85 @@ impl Publisher {
...
-        for (name, _) in &index {
+        for name in self.index.iter() {
             let Some(pkg) = self.metadata.packages.iter().find(|pkg| pkg.name == *name) else {
+                println!("Package {name}@{version} not found in cargo metadata!");
                 continue;
             };
 
-            if verify && crate::verify(name, &version)? {
-                println!("Package {}@{} already published .", &name, &version);
+            // TODO: remove this check
```
StackOverflowExcept1on, line 79: ^ try it with one package on local machine: `cargo +stable run --release -p crates-io publish --simulate`

.github/workflows/check.yml
```diff
@@ -94,8 +94,8 @@ jobs:
           cargo +stable check --manifest-path utils/wasm-builder/test-program/Cargo.toml
           cargo +stable check --manifest-path utils/cargo-gbuild/test-program/Cargo.toml --workspace --target wasm32-unknown-unknown
 
-      - name: "Check: crates-io packages"
-        run: cargo +stable run --release -p crates-io check
+      - name: "Check: crates-io packages publishing"
+        run: cargo +stable run --release -p crates-io publish --simulate --registry-path /tmp/cargo-http-registry
```
StackOverflowExcept1on, line 98: This can also be run locally to simulate publishing. If `--registry-path` was passed, then on error it is possible to continue from the last crate that failed to publish.

utils/crates-io/src/simulator.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+impl Simulator {
+    /// Create a new simulator.
+    pub fn new(registry_path: Option<PathBuf>) -> Result<Self> {
+        let path = match registry_path {
+            Some(path) => RegistryPath::Dir(path),
+            None => RegistryPath::TempDir(TempDir::new()?),
+        };
+        let (future, addr) = cargo_http_registry::serve(&path, "127.0.0.1:35503".parse()?)?;
```
clearloop, line 67: pick an available port instead of this fixed one?

.github/workflows/check.yml
StackOverflowExcept1on: @clearloop can we also remove `crates-io check` in favor of running `cargo publish` (which does `cargo build` as well)?

utils/crates-io/src/publisher.rs
```diff
@@ -62,52 +59,82 @@ impl Publisher {
...
+            /*if !["gear-ss58", "gprimitives"].contains(name) {
+                continue;
+            }*/
+
```
clearloop, line 75:
```suggestion

```

.github/workflows/check.yml
clearloop: sure, `crates-io check` is a simple version of the simulation, we df can remove it since we now have a better solution

utils/crates-io/src/version.rs
```diff
@@ -35,14 +34,27 @@ struct Version {
...
 
     let client = reqwest::blocking::Client::builder()
         .user_agent("gear-crates-io-manager")
         .build()?;
 
-    if let Ok(resp) = client
+    if let Some(simulator) = simulator {
+        if client
+            .get(format!(
+                "http://{}/api/v1/crates/{}/{version}/download",
```
clearloop, line 47: smart move 👍

clearloop (approved):
Interesting, if I'm not mistaken, this `crate-io-registry` redirects `api/../download` to crates-io and `api/../new` to the local registry

This could be a better solution than `cargo vendor`, we can close https://github.com/gear-tech/gear/issues/3565 as well in this PR!

utils/crates-io/src/simulator.rs
StackOverflowExcept1on: I wanted to pick a random port, but it seems like that starts creating a lot of folders in `~/.cargo/registry/index` that start with `127.0.0.1-`

utils/crates-io/src/publisher.rs
```diff
@@ -62,52 +59,82 @@ impl Publisher {
...
+        if let Some(simulator) = self.simulator.as_ref() {
+            simulator.restore()?;
+        }
+
+        Ok(())
+    }
+
+    /// Patch local files
+    fn patch(&self) -> Result<()> {
+        if let Some(manifests) = self.manifests() {
```
StackOverflowExcept1on, line 110: It might also make sense to backup Cargo.lock and roll it back here

breathx (approved):

PR #4141, opened 2024-08-12
feat(common, gtest): add AuxiliaryTaskpool

playX18:
Implementation of AuxiliaryTaskpool. At the moment not used in code. Once merged into master will work into merging it into [#4119](https://github.com/gear-tech/gear/pull/4119)

PR #4154, opened 2024-08-14
feat(gtest): Use `AuxiliaryTaskPool` for current scheduled ops on `gtest`

techraed:
Successor of the #4141, that actually uses the `AuxiliaryTaskPool` for the currently available scheduled tasks on the `gtest`, which are: 1) messages wait-wake, 2) delayed sending. 

Note: the implementation of the delayed sending and the messages wait-wake proc is the same, but scheduled to be made more accurate. 

Done:
- actually use aux task pool in `ExtManager::process_tasks`
- use task pool for storing the scheduled tasks for delayed sending and waking messages
- move `JournalHandler` and `TaskHandler` traits impls to corresponding modules.


techraed:
@LouiseMedova this fixes the bug you found for the unreserved system reservation on the message to be consumed

breathx (approved):

PR #4171, opened 2024-08-21
feat(pallett-gear-voucher): Extend voucher with creation of a new program from uploaded code

### LINKED ISSUE #4170: Extend voucher with creation of a new program from uploaded code
### Problem to Solve

Create program with prepaid call

### Possible Solution

Extend voucher issue with allowed CodeIds
Add PrepaidCall::CreateProgram

### Notes

_No response_

vobradovich:
Resolves #4170


DennisInSky (approved):
Not an expert here, but LGTM. Maybe I would think of implementing permissions as a enum and a collection of its values as a set of voucher permissions.

breathx (changes requested):
Will submit proper review soon. pls dont merge

PR #4173, opened 2024-08-21
feat(ethexe): zero program state; handle exec balances topups; handle unrecognised mirrors; refactor

breathx:


ethexe/runtime/common/src/state.rs
```diff
@@ -111,6 +121,41 @@ pub struct ProgramState {
     pub executable_balance: Value,
 }
 
+impl ProgramState {
+    pub const fn zero() -> Self {
+        Self {
```
StackOverflowExcept1on, line 126: mb use derive default?

ethexe/processor/src/lib.rs
```diff
@@ -121,8 +109,7 @@ impl Processor {
         Ok(true)
     }
 
-    // TODO: deal with params on smart contract side.
-    pub fn handle_new_program(&mut self, program_id: ProgramId, code_id: CodeId) -> Result<H256> {
+    pub fn handle_new_program(&mut self, program_id: ProgramId, code_id: CodeId) -> Result<()> {
```
StackOverflowExcept1on, line 112: mb check here that the program does not exist in db?

ethexe/processor/src/lib.rs
```diff
@@ -272,11 +218,9 @@ impl Processor {
     ) -> Result<Vec<LocalOutcome>> {
         log::debug!("Processing upload code {code_id:?}");
 
-        if code_id != CodeId::generate(code) || self.handle_new_code(code)?.is_none() {
-            Ok(vec![LocalOutcome::CodeRejected(code_id)])
-        } else {
-            Ok(vec![LocalOutcome::CodeApproved(code_id)])
-        }
+        let valid = !(code_id != CodeId::generate(code) || self.handle_new_code(code)?.is_none());
```
StackOverflowExcept1on, line 221:
```suggestion
        let valid = code_id == CodeId::generate(code) && self.handle_new_code(code)?.is_some();

```

StackOverflowExcept1on (approved):

ethexe/processor/src/lib.rs
StackOverflowExcept1on: what if `self.handle_new_code(code)?` returns an error? user won't get `LocalOutcome::CodeValidated`

ethexe/runtime/common/src/state.rs
breathx: I think const fns with more intuitive about usage naming for the case looks better

ethexe/processor/src/lib.rs
breathx: program -> code_id matching does exactly the thing

PR #4182, opened 2024-08-22
feat(ethexe): introduce test-only field status for Service

### LINKED ISSUE #4099: ethexe: append possibility to receive info that service is started to listen blocks
### Problem to Solve

Presently we have no ability to understand that service has run blocks listening in tests. That cause the problem that we need to sleep 1 second and hope that it's enough for service to start listening.

### Possible Solution

Maybe it's normal to receive event "StartListening" thru network.

### Notes

_No response_

clearloop:
Resolves #4099

share service status to caller environments, currently test-only features since this may not necessary for production

@gear-tech/dev


ethexe/cli/src/lib.rs
```diff
@@ -15,6 +15,7 @@
 //
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
+#![feature(stmt_expr_attributes)]
```
breathx, line 18: pls void nightly features

breathx (changes requested):
Otherwise 👍🏻

ethexe/cli/src/lib.rs
clearloop: https://github.com/gear-tech/gear/pull/4182/commits/1b4e05a76ddb532a9a8e1c6c450fc2598fc94dc0

breathx (approved):

PR #4185, opened 2024-08-23
refactor(ethexe-observer): simplify events querying

breathx:


ethexe/observer/src/observer.rs
```diff
@@ -267,107 +204,91 @@ pub(crate) async fn read_block_events_batch(
...
+    let (router_and_wvara_logs, mirrors_logs) = future::try_join(
+        provider.get_logs(&router_and_wvara_filter),
+        provider.get_logs(&mirror_filter),
+    )
+    .await?;
+
+    let block_hash_of = |log: &alloy::rpc::types::Log| -> Result<H256> {
+        log.block_hash
+            .map(|v| v.0.into())
+            .ok_or(anyhow!("Block hash is missing"))
```
StackOverflowExcept1on, line 258: mb use ok or else here?

StackOverflowExcept1on (approved):

ethexe/observer/src/observer.rs
breathx: seems that we use value creation everywhere in the code, moreover its still super cheap

PR #4190, opened 2024-08-26
feat(ethexe): Only handle non-informational events

breathx:


ethexe/ethereum/src/mirror/events.rs
```diff
@@ -16,68 +16,65 @@
...
-    pub const ALL: [H256; 8] = [
+    pub const FOR_HANDLING: &[B256] = &[
         EXECUTABLE_BALANCE_TOP_UP_REQUESTED,
         MESSAGE_QUEUEING_REQUESTED,
-        MESSAGE,
         REPLY_QUEUEING_REQUESTED,
-        REPLY,
-        STATE_CHANGED,
-        VALUE_CLAIMED,
         VALUE_CLAIMING_REQUESTED,
```
grishasobol, line 44: MInor: Why do we have naming like "_REQUESTED" in `sol` and naming "_FOR_HANDLING" in `rust`. Why not just use naming REQUESTED or REQUEST anywhere: `BlockEventRequest`, `RouterEventRequest` ?

ethexe/processor/src/lib.rs
```diff
@@ -257,7 +256,7 @@ impl Processor {
         Ok(outcomes)
     }
 
-    fn process_router_event(
+    fn handle_router_event(
```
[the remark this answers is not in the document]
breathx: It seems a little bit more consistent since we have fns "handle_new_code/program" etc. It's more like callbacks for events happening for me

ethexe/ethereum/src/mirror/events.rs
breathx: EventRequest sounds like request for event, but in fact its event that requests for handling (processing), wdyt?

grishasobol: The main idea: we should use in rust the corresponding naming as in .sol

grishasobol (approved):

PR #4192, opened 2024-08-26
fix(benchmarks,runtime-interface): clear store by counter when runnining benchmarks

ByteNacked:
Resolves # .

Fixes the issue of a 10x slowdown in `onetime syscall` benchmarks.

[Original problem](https://github.com/gear-tech/gear/pull/4136/commits/58643b209e890d64945d2f466808745fd52b49f2)

@reviewer-or-team


ByteNacked:
TODO: update weights for onetime syscall benches

runtime-interface/sandbox/Cargo.toml
```diff
@@ -28,3 +28,4 @@ std = [
   "sp-runtime-interface/std",
   "sp-wasm-interface/std",
 ]
+runtime-benchmarks = []
```
breathx, line 31: Newline

runtime-interface/Cargo.toml
```diff
@@ -53,3 +53,6 @@ std = [
 	"ark-scale/std",
 	"sha2/std"
 ]
+runtime-benchmarks = [
+	"gear-sandbox-interface/runtime-benchmarks"
+]
```
breathx, line 58: Newline

breathx (approved):
Doesn't this affect other benches, not only onetime syscalls?

runtime-interface/Cargo.toml
ByteNacked: Fixed

runtime-interface/sandbox/Cargo.toml
ByteNacked: Fixed

ByteNacked:
> Doesn't this affect other benches, not only onetime syscalls?

It's not affect normal benchmarks because they run with :

```
BENCHMARK_STEPS=50
BENCHMARK_REPEAT=20
```

and store cleared normally (by method `Sanboxes::get` or by creating new thread).

gshep:
Another option is to use `Vec::with_capacity` with some good value for collections of `SandboxComponents`. It would allow to avoid redundant reallocations which are the reason of regression.

ByteNacked:
> Another option is to use `Vec::with_capacity` with some good value for collections of `SandboxComponents`. It would allow to avoid redundant reallocations which are the reason of regression.

Will check, but I think the reason is not in big sizes of these `Vecs` (Instance, memory suppose to be just ptr size) but due to internal indexes inside wasmer store object (I think we not supposed to store >= 1000 memory and instances in it).

ByteNacked:
> Another option is to use `Vec::with_capacity` with some good value for collections of `SandboxComponents`. It would allow to avoid redundant reallocations which are the reason of regression.

Checked https://github.com/gear-tech/gear/actions/runs/10576870738/job/29303638506 , it seems doesn't work

ByteNacked:
Updated onetime syscalls' weights

PR #4202, opened 2024-08-28
fix(lazy-pages-fuzzer): remove host address check

ByteNacked:
Resolves # .

Remove lazy pages fuzzer host address check.

@reviewer-or-team


ByteNacked:
I added this assert because I assumed that when pages are allocated in wasmi/wasmer, they would be aligned to the WASM_PAGE_SIZE. However, this turned out to be incorrect, as they are only aligned to the host's page size.

In this case, the assert is pointless and can be removed

PR #4205, opened 2024-08-29
fix(gtest): clean waitlist on exit_dispatch

playX18:
This allows waitlist impl of gtest to match JournalHandler from gear pallet


breathx (approved):

techraed (approved):

gtest/src/manager/journal.rs
```diff
@@ -100,6 +99,11 @@ impl JournalHandler for ExtManager {
     }
 
     fn exit_dispatch(&mut self, id_exited: ProgramId, value_destination: ProgramId) {
+        self.waitlist.drain_key(id_exited).for_each(|entry| {
```
techraed, line 102: since the original rationale for the task was to make the impls of journal handlers on runtime and gtest similar, can you please also introduce the removal of the reservations as it's done on the runtime

techraed (changes requested):

techraed (approved):

PR #4207, opened 2024-08-29
fix(gtest): update `ActorMailbox::claim_value` and `ActorMailbox::reply`

ByteNacked:
Resolves # .

Update `ActorMailbox::claim_value` and `ActorMailbox::reply` implementation in gtest.

@reviewer-or-team


gtest/src/manager.rs
```diff
@@ -294,4 +294,75 @@ impl ExtManager {
 
         Ok(message)
     }
+
+    pub(crate) fn send_reply_impl(
```
techraed, line 298: I'd move that to `state::mailbox::actors`, as it all related to the mailbox logic

techraed (commented):
The implementation is absolutely correct.
However, some things would be very good to implement:
1. `Program::send_*` API works as extrinsics in terms of runtime. The same is about `ActorMailbox::reply*` and `ActorMailbox::claim_value`. All such user initialized messages are sent through `ExtManager::validate_and_route_dispatch`, so you need to use the former inside newly introduced `send_reply_impl` and `claim_value_impl` methods.
2. If you do the **1**, then you have to adjust the `validate_and_route_dispatch` accordingly. Please don't forget to add the check that destination is an active program.

gtest/src/manager.rs
ByteNacked: I think `fn send_reply_impl` should be inside `ExtManager` to maintain a clear separation of abstraction levels.

Moved `send_reply` and `claim_value` to the `manager::exec` module, where `send` extrinsic is located.

ByteNacked:
> * `Program::send_*` API works as extrinsics in terms of runtime. The same is about `ActorMailbox::reply*` and `ActorMailbox::claim_value`. All such user initialized messages are sent through `ExtManager::validate_and_route_dispatch`, so you need to use the former inside newly introduced `send_reply_impl` and `claim_value_impl` methods.
> 
> * If you do the **1**, then you have to adjust the `validate_and_route_dispatch` accordingly. Please don't forget to add the check that destination is an active program.

Discussed, fixed

gtest/src/manager.rs
ByteNacked: Fixed

gtest/src/manager/exec.rs
```diff
@@ -35,6 +35,33 @@ impl ExtManager {
         self.route_dispatch(dispatch)
     }
 
+    pub(crate) fn claim_value_impl(
```
techraed, line 38: by your choice: I'd move it to `mailbox::actor` too

techraed (approved):

gtest/src/manager/exec.rs
ByteNacked: Fixed

PR #4209, opened 2024-08-30
feat(wasm-builder): Successful build returns paths to the wasm and opt wasm files

vobradovich:


utils/wasm-builder/src/lib.rs
```diff
@@ -100,22 +100,27 @@ impl WasmBuilder {
     }
 
     /// Build the program and produce an output WASM binary.
-    pub fn build(self) {
+    ///
+    /// Returns tuple with paths to wasm & opt wasm file
```
DennisInSky, line 104: Do you want to mention when it returns `None`?

DennisInSky (approved):

breathx (approved):

PR #4213, opened 2024-08-31
fix(workflows): add node sync check; refine artefacts append to releases

kvs96:


.github/workflows/check-node-sync.yml
```diff
@@ -0,0 +1,104 @@
...
+  check-testnet:
+    needs: [prepare, build-and-push-image]
+    runs-on: kuberunner
+    timeout-minutes: 7200
+    steps:
+      - name: Run testnet sync
+        run: |
+          docker run --name gear-node --rm ${{ needs.prepare.outputs.image_name }} \
+            gear --chain testnet --sync full 2>&1 | tee -a ./sync.log &
+          while ! grep -q 'Imported' ./sync.log; do sleep 60; done
```
breathx, line 86: What if it will stuck on the same block imported?

breathx (approved):

PR #4221, opened 2024-09-03
fix(ethexe): problem with too fast request to rpc after transaction confirmation

grishasobol:


ethexe/ethereum/src/lib.rs
```diff
@@ -270,3 +271,36 @@ impl SignerSync for Sender {
         self.chain_id
     }
 }
+
+// TODO: Maybe better to append solution like this to alloy.
+/// Works like `tx.get_receipt().await`, but retries a few times if rpc returns a null response.
+async fn get_transaction_receipt<T: transports::Transport + Clone, N: network::Network>(
```
StackOverflowExcept1on, line 277: maybe it would be better to create trait with `try_get_receipt` method
`get_transaction_receipt(builder.send().await?).await?` vs `builder.send().await?.try_get_receipt().await?`

grishasobol: Cannot see the difference, also I think It is a temporary solution...

StackOverflowExcept1on: the code is easier to read and not difficult to implement

grishasobol: Done

ethexe/ethereum/src/lib.rs
```diff
@@ -270,3 +271,40 @@ impl SignerSync for Sender {
...
+    /// Works like `self.get_receipt().await`, but retries a few times if rpc returns a null response.
+    async fn try_get_receipt(self) -> Result<N::ReceiptResponse>;
+}
+
+impl<T: Transport + Clone, N: Network> TryGetReceipt<T, N> for PendingTransactionBuilder<'_, T, N> {
+    async fn try_get_receipt(self) -> Result<N::ReceiptResponse> {
+        let tx_hash = *self.tx_hash();
+        let provider = self.provider().clone();
+
+        let mut err = match self.get_receipt().await {
```
breathx, line 286: Wdyt to put initial query in loop so simplify code

breathx (approved):

ethexe/ethereum/src/lib.rs
grishasobol: It's impossible because initial query is different and borrows the self

PR #4222, opened 2024-09-04
feat(gprimitives): introduce constant builtin actor ids

### LINKED ISSUE #3854: List builtin programs in the RPC of pallet-builtin
### Problem to Solve

https://github.com/gear-tech/gear/blob/master/pallets/gear-builtin/rpc/src/lib.rs

we are currently don't have an active list of builtin programs for querying system programs from users' or docs side

### Possible Solution

1. Expand name and ( usage? including online docs ) in trait `BuiltinActor` https://github.com/gear-tech/gear/blob/master/pallets/gear-builtin/src/lib.rs#L107
2. Add a RPC call for listing all system programs https://github.com/gear-tech/gear/blob/master/pallets/gear-builtin/rpc/src/lib.rs#L73

### Notes

_No response_

clearloop:
Resolves #3854 

1. We have already had the RPC querying actor id of builtin libraries, but it is not convenient for developers that they have to know the `id` of the libraries first
2. Moving constant ids to `gprimitives` benefits programs, that they can get the interfaces from libraries directly but not RPC

@gear-tech/dev


breathx (changes requested):
I still don't agree with hardcoding addresses/indeces, since encoding (creation) of actual program ids for Bias may be changed in future. moreover, some of builtins are able to be replaced with the updates and further address changing, since its just common actors. so it seems better to keep things a bit harder, but pointing out on need of supporting some address-updating logic.

Or this could be just UNSAFE hardcoded options with comments that in only works on Vara at the moment of lib version existence

breathx:
@ekovalev @gshep could you please also elaborate on this discussion

ekovalev:
Can we avoid hardcoding `ProgramId`s? Because it will essentially render the whole concept of `ActorId`s useless - why not identify actors by their `ProgramId`s then? By design tight coupling between `ActorId`s and `ProgramId`s is undesirable, because `ActorId` is supposed to be independent from the Runtime version, account system used in the Runtime... even the underlying blockchain itself which does not necessarily have to always be a Substrate based chain.

Maybe just keep the set of `ActorId`s in `gprimitives` and assign new ones in a more controlled way? And move the derivation of the program id there, so that contract developers could derive it programmatically instead of querying via RPC? 
Although I personally don't see much of a problem here - it's quite common for a blockchain to have some "well-known" addresses anyway.

gshep:
To tell the truth I have nothing to say more.

It seems like we aren't able to implement such convenient library with builtin addresses without violate any other properties.

PR #4223, opened 2024-09-04
refactor(gtest): update user-facing panic messages 

ByteNacked:
In `gtest`:

- remove `#[track_caller]` directive
- update user-facing panic messages

@reviewer-or-team


gtest/src/error.rs
```diff
@@ -88,3 +88,13 @@ pub enum TestError {
     #[display(fmt = "Reading of program state failed: `{_0}`")]
     GbuildArtifactNotFound(String),
 }
+
+macro_rules! _user_panic {
```
breathx, line 92: It can be named without underscore

breathx, line 92: Moreover, it's more like "misuse" of gtest api: so we can even say that we panicked because of users wrong usage of library

gtest/src/error.rs
```diff
@@ -88,3 +88,13 @@ pub enum TestError {
...
+    ($($arg:tt)*) => {{
+        use colored::Colorize;
```
breathx, line 94: as _

gtest/src/manager.rs
```diff
@@ -221,8 +222,9 @@ impl ExtManager {
     #[track_caller]
```
breathx, line 222: Track caller may be also removed here to highlight correct "source" of panic

gtest/src/error.rs
ByteNacked: So what naming do you suggest here?

gtest/src/manager.rs
ByteNacked: Yes, not sure how I missed it

gtest/src/error.rs
ByteNacked: Fixed

gtest/src/manager.rs
ByteNacked: Fixed

gtest/src/error.rs
ByteNacked: Fixed

gtest/src/error.rs
ByteNacked: The idea here is that we use user-facing panics to communicate gtest API usage errors to the user.

Maybe `api_misuse_panic` would be a better name?

ByteNacked: Or `usage_panic`.

ByteNacked: Discussed, fixed

breathx (approved):

PR #4225, opened 2024-09-05
feat(gsdk): introduce api builder

### LINKED ISSUE #4224: Retry feature of `gclient`/`gsdk`
### File Location(s)

`gsdk/src/api.rs`

### Proposal

Seems like retry feature in `gclient`/`gsdk` is not finished. Would be very nice to complete it :hugs:

clearloop:
Resolves #4224

- [x] add retries to rpc client
- [x] api builder for gsdk
- [x] api builder for gclient

@gear-tech/dev


gsdk/src/client.rs
```diff
@@ -190,7 +186,21 @@ impl Rpc {
...
+        let mut retries = 0;
+
+        loop {
+            let r = self
+                .rpc
+                .request(method, params.clone())
+                .await
+                .map_err(Into::into);
+
+            if retries == self.retries || r.is_ok() {
```
gshep, line 198: perhaps it's worth to log an error?

gsdk/src/api.rs
```diff
@@ -23,11 +23,14 @@ use anyhow::Result;
 use core::ops::{Deref, DerefMut};
 use subxt::OnlineClient;
 
+const DEFAULT_GEAR_ENDPOINT: &str = "wss://rpc.vara.network:443";
+const DEFAULT_TIMEOUT: u64 = 60_000;
```
gshep, line 27: If I'm not mistaken the timeout is in milliseconds. It would be nice to mention it in comment or add suffix like `_MILLISECS`

gshep (approved):
Awesome! :1st_place_medal:

gsdk/src/client.rs
clearloop: processed warnings instead of errors since the last try will emit error anyway

gsdk/src/api.rs
clearloop: [b100cd3](https://github.com/gear-tech/gear/pull/4225/commits/b100cd3b7599cfdbb579d13cb11dad823b90271d) updated, btw I just changed the default `retires` to `0` otherwise it could be unexpected from users' end

breathx (approved):

PR #4237, opened 2024-09-16
feat(ethexe): introduce ScaleCodec library for Solidity

osipov-mit:


ethexe/contracts/src/ScaleCodec.sol
```diff
@@ -0,0 +1,479 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.26;
+
+function bytesToUint(bytes memory data, uint256 byte_length) pure returns (uint256) {
```
StackOverflowExcept1on, line 4: plz use camel case
```suggestion
function bytesToUint(bytes memory data, uint256 byteLength) pure returns (uint256) {
```

breathx (changes requested):
To avoid merging

ethexe/contracts/src/ScaleCodec.sol
```diff
@@ -0,0 +1,508 @@
...
+    }
+
+    struct Optional {
+        bool isSome;
+        bytes value;
+    }
+
+    struct Result {
+        bool isOk;
+        bool isErr;
```
breathx, line 36: seems you could use only one field? isErr == !isOk

ethexe/contracts/src/ScaleCodec.sol
```diff
@@ -0,0 +1,508 @@
...
+        uint256 value;
+        uint256 offset;
+    }
+
+    struct DecodedString {
+        string value;
+        uint256 offset;
+    }
+
+    struct Optional {
```
breathx, line 29:
```suggestion
    struct Option {
```

ethexe/contracts/src/ScaleCodec.sol
```diff
@@ -0,0 +1,508 @@
...
+            let byte_value := byte(0, mload(add(src_ptr, i)))
+            result := or(result, shl(mul(i, 8), byte_value))
+        }
+    }
+
+    return result;
+}
+
+library ScaleCodec {
+    struct CompactInt {
```
breathx, line 19:
```suggestion
    struct CompactUint256 {
```

ethexe/contracts/src/ScaleCodec.sol
```diff
@@ -0,0 +1,508 @@
...
+        uint256 value;
+        uint256 offset;
```
breathx, line 21: if offset is amount of zeroes, than you could use less than 256 bits for it?

ethexe/contracts/src/ScaleCodec.sol
```diff
@@ -0,0 +1,508 @@
...
+        bool isOk;
+        bool isErr;
+        bytes value;
+    }
+
+    function sliceBytes(bytes memory data, uint256 start, uint256 end) public pure returns (bytes memory) {
+        bytes memory result = new bytes(end - start);
+
+        for (uint256 i = 0; i < end - start; i++) {
+            result[i] = data[i + start];
```
breathx, line 44: why do you need to fill each slice/bytes etc?

ethexe/contracts/src/ScaleCodec.sol
```diff
@@ -0,0 +1,508 @@
...
+        }
+
+        return result;
+    }
+
+    function concatBytes(bytes[] memory value) public pure returns (bytes memory) {
```
breathx, line 50: gpt suggests to pre-allocate resulting bytes memory here:
```sol
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

contract ByteConcatenation {
    function concat(bytes[] memory data) public pure returns (bytes memory) {
        // Calculate total length
        uint totalLength = 0;
        for (uint i = 0; i < data.length; i++) {
            totalLength += data[i].length;
        }

        // Allocate memory for the result
        bytes memory result = new bytes(totalLength);
        uint offset = 0;

        // Copy each bytes element into the result
        for (uint i = 0; i < data.length; i++) {
            bytes memory currentBytes = data[i];
            uint currentLength = currentBytes.length;

            // Use inline assembly to copy memory
            assembly {
                // Copy currentBytes to result at the offset
                // The `mload` is used to get the length of currentBytes
                let src := add(currentBytes, 0x20) // skip the length
                let dest := add(result, add(0x20, offset)) // skip the length of result
                mcopy(dest, src, currentLength)
            }
            offset += currentLength;
        }

        return result;
    }

    // Helper function for memory copy
    function mcopy(uint dest, uint src, uint length) internal pure {
        for (uint i = 0; i < length; i++) {
            assembly {
                mstore8(add(dest, i), mload8(add(src, i)))
            }
        }
    }
}
```

ethexe/contracts/src/ScaleCodec.sol
breathx: everywhere

ethexe/contracts/src/ScaleCodec.sol
```diff
@@ -0,0 +1,508 @@
...
+    function encodeBoolTo(bool value, bytes memory destination, uint256 offset) internal pure {
+        if (value) {
+            destination[offset] = 0x01;
+        } else {
+            destination[offset] = 0x00;
+        }
+    }
+
+    function decodeBool(bytes memory _bytes, uint256 offset) public pure returns (bool) {
+        return _bytes[offset] == 0x01;
```
breathx, line 154: what if its 0x02?

breathx: you should return result here maybe
for optimised code impl decodeBoolUnchecked

breathx: similar everywhere

ethexe/contracts/src/ScaleCodec.sol
```diff
@@ -0,0 +1,508 @@
...
+
+    function encodeInt16To(int16 value, bytes memory destination, uint256 offset) internal pure {
+        encodeUint16To(uint16(value), destination, offset);
+    }
+
+    function decodeInt16(bytes memory _bytes, uint256 offset) public pure returns (int16) {
+        return int16(decodeUint16(_bytes, offset));
+    }
+
+    function encodeUint32(uint32 value) public pure returns (bytes memory) {
```
breathx, line 216: do you always need `bytes`?
seems that uint32 value has zero cast to bytes4 - so it could be also useful for future usage?

ethexe/contracts/src/ScaleCodec.sol
```diff
@@ -0,0 +1,508 @@
...
+        if (_bytes[offset] == 0x00) {
+            return Optional(false, new bytes(0));
+        } else {
+            return Optional(true, sliceBytes(_bytes, 1 + offset, _bytes.length));
+        }
+    }
+
+    function encodeResult(Result memory value) public pure returns (bytes memory) {
+        if (value.isOk) {
+            return bytes.concat(hex"00", value.value);
```
breathx, line 494: ?
```suggestion
            return bytes.concat(0x00, value.value);
```

breathx (changes requested):

ethexe/contracts/src/ScaleCodec.sol
osipov-mit: doesn't work this way

ethexe/contracts/src/ScaleCodec.sol
osipov-mit: Actually `encodeUint32To` function is supposed to be used most of the time (at least in contracts generated by sails). And here I belief `bytes` is better cuz you probably need to combine those bytes with some others and you can't do it if you have type `bytes4`. And even if you don't need to combine them you need to pass those bytes to the `sendMessage` function which takes `bytes` type

ethexe/contracts/src/ScaleCodec.sol
osipov-mit: Since there is no native Result type in Solidity I think it's not the best idea.
Maybe calling `revert` is better if the bytes are wrong?

ethexe/contracts/src/ScaleCodec.sol
osipov-mit: discussed

breathx:
Won't be merged. Consider moving into separate repo in case of need.

PR #4250, opened 2024-09-23
fix(gtest): Fix timestamp setting on block creation

techraed:
A quick fix for the `gtest` block timestamp calculation.


techraed:
False positive

PR #4252, opened 2024-09-24
Heuristics tests are altered to panic with multiple results

Lazark0x:
Resolves #4246


runtime/vara/src/tests.rs
```diff
@@ -20,15 +20,345 @@ use super::*;
...
+/// Memory pages access costs.
+struct PagesCosts {
+    pub load_page_data: CostOf<GearPagesAmount>,
+    pub upload_page_data: CostOf<GearPagesAmount>,
+    pub mem_grow: CostOf<GearPagesAmount>,
+    pub mem_grow_per_page: CostOf<GearPagesAmount>,
+    pub parachain_read_heuristic: CostOf<GearPagesAmount>,
+}
+
+impl<T: pallet_gear::Config> From<MemoryWeights<T>> for PagesCosts {
```
StackOverflowExcept1on, line 93: btw we already have `gear_core::costs::LazyPagesCosts`. mb reuse it here?
 https://github.com/gear-tech/gear/blob/73929e309d40aceec8da761b8b2313eab7ea41a4/pallets/gear/src/schedule.rs#L1304

StackOverflowExcept1on: we could add the missing fields (from `MemoryWeights<T>`) to `gear_core::costs::LazyPagesCosts` and reuse that for both gtest and these checks.

runtime/vara/src/tests.rs
```diff
@@ -20,15 +20,345 @@ use super::*;
...
+            WeightExpectation::new(
+                weights.$inst_name.into(),
+                expected.$inst_name.into(),
+                INSTRUCTIONS_SPREAD,
+                stringify!($inst_name),
+            )
+        };
+    }
+
+    let expectations = vec![
```
breathx, line 121: It'd be great also to have some check of missing variants

runtime/vara/src/tests.rs
Lazark0x: Done

runtime/vara/src/tests.rs
Lazark0x: Yeah, added checks for the number of fields to be checked

runtime/vara/src/tests.rs
breathx: That way LazyPagesCosts become not actually costs of lazy pages. wdyt @grishasobol ?

Lazark0x: Do you think we need a more generic name for the structure? @breathx

runtime/vara/src/tests.rs
```diff
@@ -18,17 +18,547 @@
...
 use pallet_gear::{InstructionWeights, MemoryWeights, SyscallWeights};
 use pallet_staking::WeightInfo as _;
-use runtime_common::weights::{
-    check_instructions_weights, check_lazy_pages_costs, check_pages_costs, check_syscall_weights,
-    PagesCosts,
-};
 use sp_runtime::AccountId32;
 
+#[cfg(feature = "dev")]
+use frame_support::traits::StorageInstance;
```
breathx, line 30: pls move everything into module. runtime/vara/tests is general testing module. so all "utils" like `WeightExpectation' should be somewhere else. keep here tests only

breathx (approved):
lgtm, Im only aware of naming for lazy pages costs, waiting for additional opinions here

runtime/vara/src/tests.rs
Lazark0x: Done

breathx (approved):

core/src/costs.rs
```diff
@@ -502,6 +502,21 @@ impl SyscallCosts {
     }
 }
 
+/// Memory pages costs.
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct PagesCosts {
```
StackOverflowExcept1on, line 507: also need to update `TYPE_LIST` (add `PagesCosts`, `IoCosts`) and regenerate `core/src/gas_metering/schedule.rs`
https://github.com/gear-tech/gear/blob/b9511183910d6b1ff6a018a5cadb72dc784b4fee/utils/weight-diff/src/main.rs#L305

Lazark0x: Done

StackOverflowExcept1on (approved):

PR #4273, opened 2024-10-01
feat(ethexe): impl scheduled program wakes; restore `many-waits` test

breathx:


ethexe/runtime/common/src/schedule.rs
```diff
@@ -0,0 +1,86 @@
...
+    fn send_user_message(&mut self, _stashed_message_id: MessageId, _to_mailbox: bool) -> u64 {
+        unimplemented!("TODO (breathx)")
+    }
+    // TODO (breathx): consider deprecation of delayed wakes + non-concrete waits.
+    fn wake_message(&mut self, program_id: ProgramId, message_id: MessageId) -> u64 {
+        log::trace!("Running scheduled task wake message {message_id} to {program_id}");
+
+        // TODO (breathx): don't update state if not changed?
+        self.update_state_with_storage(program_id, |storage, state| {
+            let Some(((dispatch, _expiry), new_waitlist_hash)) = storage
```
grishasobol, line 48: Maybe unreachable for case when `_expiry < current`.

grishasobol (approved):

ethexe/runtime/common/src/schedule.rs
breathx: unfortunately, I think it could be reached in some cases. we need to support outdated tasks and handle cases why we've missed something

Issue #938, opened 2022-05-13
Support start section


grishasobol:
### Problem to Solve

Currently we can not guaranty that start section in smart contracts will not emit some problems in executor or smart contracts execution. We should consider this possibilities and may be release features in executor, which will allow to execute start section safely.

### Possible Solution

Currently we disable possibility to execute smart contracts, with start section https://github.com/gear-tech/gear/pull/933

PR #4282, opened 2024-10-08
feat(sandbox-host): update sandbox host wasmi version

### LINKED ISSUE #3017: Update wasmi host executor
### File Location(s)

`sandbox/host`

### Proposal

Update wasmi executor to 0.30.

The main problem is latest versions of main-stream executors (wasmi, wasmer, wasmtime) implement [WASM store](https://webassembly.github.io/spec/core/exec/runtime.html#store) using only Rust borrow semantics and no synchronization primitives. As a result, we cannot simply clone required structures elsewhere.

It is the problem because we have lazy pages concept that requires access to WASM globals during WASM function invocation, when accessing protected memory pages. Pseudocode:
```rust
let store = Store::new(...);
func.call(&mut store, ...);

// inside WASM function call
fn call(...) {
	// ...some WASM instructions

	memory.write 123 at 0xCAFE
	// let's think 0xCAFE address belongs to protected memory page, so:
    // 1. MMU sees protected memory and causes interruption to OS
	// 2. We set signal handler earlier, so OS jumps to `signal_handler()`
	// 3. After handler is done, OS jumps back to `memory.write` and it will be successful now

	// ..execution continues
}

// when interruption occurs, lazy-pages signal handler is in work
fn signal_handler() {
    let store = ???; // how to access mutable reference again, if `func.call` holds it?
	global.set(&mut store, 333);
    memory.unprotect 0xCAFE // unprotect page which address belongs to
}
```

Possible solutions:
- [x] There was the first try in #2931 with https://github.com/gear-tech/wasmi/pull/3. Wrap globals into synchronization primitive. This approach will be required for the future wasmer update, which is harder to implement because wasmer is more complex.
- [x] Suggested by @gshep. Mutable pointer to store. Very dangerous, very unpredictable because it means we have 2 mutable references or multialiasing, which is UB.
- [x] Suggested by @grishasobol. Implement callbacks on every memory access at executor's side. It requires maintaining patches as in the first solution.
- [ ] Perhaps the executor authors know how to deal with the new design?  https://github.com/wasmerio/wasmer/issues/4118

ByteNacked:
Resolves #3017 .

- [x] Update Wasmi host executor version
- [x] Update lazy-pages-fuzzer Wasmi version 
- [x] Add CLI parameters to switch host executor
- [x] Sync mainnet
- [x]  Sync testnet 

@reviewer-or-team


ByteNacked:
Node sync still in progress...

Cargo.toml
```diff
@@ -303,9 +303,7 @@ ethexe-rpc = { path = "ethexe/rpc", default-features = false }
 ethexe-common = { path = "ethexe/common" }
 
 # Common executor between `sandbox-host` and `lazy-pages-fuzzer`
-sandbox-wasmi = { package = "wasmi", git = "https://github.com/gear-tech/wasmi", branch = "v0.13.2-sign-ext", features = [
-    "virtual_memory",
-] }
+wasmi = { package = "wasmi", version = "0.36"}
```
ark0f, line 306: I guess it can be bumped up to `0.38`. It also has a new [`extra-checks`](https://github.com/wasmi-labs/wasmi/pull/1217) feature. 

What do you think, @breathx?

ByteNacked: Should I enable this extra-checks feat?

ByteNacked:
```
# Expected execution overhead is upt to 20%, if enabled.
#
# - Enable if your focus is on safety.
# - Disable if your focus is on execution speed.
extra-checks = []
```

breathx: I think it would be great to have similar feature in our runner, but disable it by default: can’t find any pros at this point, especially with wasmi being such slow anyway

sandbox/host/src/sandbox/wasmi_backend.rs
```diff
@@ -18,393 +18,544 @@
...
+                DEFAULT_MAX_RECURSION_DEPTH,
+            )
+            .expect("infallible"),
+        );
+
+        let engine = Engine::new(&config);
+        let store = Store::new(&engine, None);
+        Backend {
+            store: Rc::new(StoreRefCell::new(store)),
+        }
```
ark0f, line 128: Original code have never set the config. It only used in embedded executor specifically for fuzzer

sandbox/host/src/sandbox/wasmi_backend.rs
```diff
@@ -18,393 +18,544 @@
...
+) -> crate::error::Result<(Memory, Allocation)> {
+    let ty =
+        MemoryType::new(initial, maximum).map_err(|error| Error::Sandbox(error.to_string()))?;
+    let mut alloc = region::alloc(u32::MAX as usize, Protection::READ_WRITE)
+        .unwrap_or_else(|err| unreachable!("Failed to allocate memory: {err}"));
+    // # Safety:
+    //
+    // `wasmi::Memory::new_static()` requires static lifetime so we convert our buffer to it
+    // but actual lifetime of the buffer is lifetime of `Store<T>` itself,
+    // so memory will be deallocated when `Store<T>` is dropped.
```
ark0f, line 150: Update comment according to actual structures hierarchy

sandbox/host/src/sandbox/wasmi_backend.rs
```diff
@@ -18,393 +18,544 @@
...
-            let range = util::checked_range(source_addr.into(), size, source.len())
-                .ok_or_else(|| error::Error::Other("memory read is out of bounds".into()))?;
-
-            Ok(Vec::from(&source[range]))
-        })
+        let mut buffer = Vec::with_capacity(size);
+        let spare_cap = buffer.spare_capacity_mut().len();
+        // # Safety:
+        // `Vec::set_len` is safe to call because we have just allocated enough space and never read from it.
+        unsafe { buffer.set_len(spare_cap) };
```
ark0f, line 188:
```suggestion
        let mut buffer = vec![0; size];
```

sandbox/host/src/sandbox/wasmi_backend.rs
```diff
@@ -18,393 +18,544 @@
...
+pub fn set_global(
+    instance: &wasmi::Instance,
+    store: &mut Store,
+    name: &str,
+    value: Value,
+) -> std::result::Result<Option<()>, error::Error> {
+    let global = match instance.get_global(&*store, name) {
+        Some(e) => e,
+        None => return Ok(None),
+    };
```
ark0f, line 249:
```suggestion
    let Some(global) = match instance.get_global(&*store, name) else {
	return Ok(None);
    };
```

sandbox/host/src/sandbox/wasmi_backend.rs
```diff
@@ -18,393 +18,544 @@
...
-    };
+    let mut store = context.store().borrow_mut();
+
+    let module =
+        Module::new(store.engine(), wasm).map_err(|_| InstantiationError::ModuleDecoding)?;
+    let mut linker = Linker::new(store.engine());
+
+    for import in module.imports() {
+        let module = import.module().to_string();
+        let name = import.name().to_string();
```
ark0f, line 273: `.to_string()` is not actually required

sandbox/host/src/sandbox/wasmi_backend.rs
```diff
@@ -18,393 +18,544 @@
...
+                );
+
+                linker
+                    .define(&module, &name, wasmi_memory.memory)
+                    .map_err(|_| InstantiationError::EnvironmentDefinitionCorrupted)?;
+            }
+            ExternType::Func(func_ty) => {
+                let guest_func_index = guest_env
+                    .imports
+                    .func_by_name(import.module(), import.name());
```
ark0f, line 297:
```suggestion
                    .func_by_name(module, name);
```

sandbox/host/src/sandbox/wasmi_backend.rs
```diff
@@ -18,393 +18,544 @@
...
+
+                let guest_func_index = if let Some(index) = guest_func_index {
+                    index
+                } else {
+                    // Missing import (should we abort here?)
+                    continue;
+                };
```
ark0f, line 304:
```suggestion
                let Some(guest_func_index) = guest_func_index else {
                    // Missing import (should we abort here?)
                    continue;
                };
```

sandbox/host/src/sandbox/wasmi_backend.rs
```diff
@@ -18,393 +18,544 @@
...
+    }
+
+    let instance_pre = linker.instantiate(&mut *store, &module).map_err(|error| {
+        log::trace!("Failed to call wasmi instantiate: {error:?}");
+        InstantiationError::Instantiation
+    })?;
+    let instance = instance_pre.start(&mut *store).map_err(|error| {
+        log::trace!("Failed to call wasmi start: {error:?}");
+        InstantiationError::StartTrapped
     })?;
```
ark0f, line 334: This is where `_supervisor_context: &mut dyn SupervisorContext` must be used

sandbox/host/src/sandbox/wasmi_backend.rs
```diff
@@ -18,393 +18,544 @@
...
+        move |_caller, params, results| -> Result<(), wasmi::Error> {
+            SupervisorContextStore::with(|supervisor_context| {
+                let invoke_args_data = params
+                    .iter()
+                    .map(|value| {
+                        into_value(value).ok_or_else(|| {
+                            host_trap(format!("Unsupported function argument: {:?}", value))
+                        })
+                    })
+                    .collect::<std::result::Result<Vec<_>, _>>()?
```
ark0f, line 361:
```suggestion
                    .collect::<Result<Vec<_>, _>>()?
```
This and other fully-qualified names can be omitted

sandbox/host/src/store_refcell.rs
```diff
@@ -83,38 +83,42 @@
...
-pub struct StoreRefCell {
-    store: UnsafeCell<Store>,
+pub struct StoreRefCell<S> {
+    store: UnsafeCell<S>,
     state: Cell<BorrowState>,
 }
 
+trait GenericAsStoreMut {}
+
+impl<'r, 's> GenericAsStoreMut for &'r mut StoreMut<'s> {}
```
ark0f, line 113: It's better to use full path

utils/lazy-pages-fuzzer/src/wasmi_backend.rs
```diff
@@ -35,127 +37,148 @@ use crate::{
...
-        &self,
-        _field_name: &str,
-        _memory_type: &sandbox_wasmi::MemoryDescriptor,
-    ) -> Result<MemoryRef, sandbox_wasmi::Error> {
-        Ok(self.memory.clone())
+impl InstanceAccessGlobal for InstanceBundle {
+    fn set_global(&self, name: &str, value: i64) -> anyhow::Result<()> {
+        let global = self
+            .instance
+            .get_global(unsafe { &*self.store }, name)
```
ark0f, line 50: Should we ignore these unsafe accesses?

ark0f (changes requested):

runtime-interface/sandbox/src/detail.rs
```diff
@@ -61,13 +61,22 @@ impl Sandboxes {
...
+// Global sandbox backend type selector
+static SANDBOX_BACKEND_TYPE: sandbox_env::AtomicSandboxBackend =
+    sandbox_env::AtomicSandboxBackend::new(sandbox_env::SandboxBackend::Wasmer);
+
 thread_local! {
-    static SANDBOXES: RefCell<Sandboxes> = RefCell::new(Sandboxes::new());
+    static SANDBOXES: RefCell<Sandboxes> = {
+        let sandbox_backend = SANDBOX_BACKEND_TYPE.load(Ordering::SeqCst);
+        RefCell::new(Sandboxes::new(sandbox_backend))
+    }
```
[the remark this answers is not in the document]
ByteNacked: It won't work because we need to initialize the sandbox for every OS thread with the sandbox type. In the suggested approach, it will panic on a new thread.

runtime-interface/sandbox/src/detail.rs
```diff
@@ -61,13 +61,22 @@ impl Sandboxes {
...
 
-pub fn init() {
-    SANDBOXES.with(|sandboxes| {
-        let _store = sandboxes.borrow_mut().get(0);
+pub fn init(sandbox_backend: sandbox_env::SandboxBackend) {
+    SANDBOX_BACKEND_TYPE.store(sandbox_backend, Ordering::SeqCst);
+    // At first access sandbox will be initialized with the provided backend
+    SANDBOXES.with_borrow_mut(|sandboxes| {
+        let _store = sandboxes.get(0);
     })
```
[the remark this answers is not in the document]
ByteNacked: See comment below

sandbox/host/src/sandbox/wasmi_backend.rs
ByteNacked: It won't be used because ATM, we can't have start functions in programs, but I can add it for symmetry with the Wasmer backend implementation.

utils/lazy-pages-fuzzer/src/wasmi_backend.rs
ByteNacked: It similar to wasmer/wasmi backed impl

sandbox/host/src/sandbox/wasmi_backend.rs
ark0f: We can't for now but we want in the future: https://github.com/gear-tech/gear/issues/938

Cargo.toml
ByteNacked: Added disabled by default feat : `wasmi-extra-checks`

sandbox/host/src/lib.rs
```diff
@@ -23,6 +23,7 @@
 
 pub mod error;
 pub mod sandbox;
+pub(crate) mod store_refcell;
```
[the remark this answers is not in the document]
ByteNacked: Fixed

sandbox/host/src/sandbox.rs
```diff
@@ -523,6 +525,8 @@ pub struct SandboxComponents<DT> {
     instances: Vec<Option<(Pin<Rc<SandboxInstance>>, DT)>>,
     /// Memories are `Some` until torn down.
     memories: Vec<Option<Memory>>,
+    /// Allocation used for the Wasmi memory.
+    allocations: Vec<Allocation>,
```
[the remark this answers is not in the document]
ByteNacked: Moved to `Backend` struct

sandbox/host/src/sandbox/wasmi_backend.rs
```diff
@@ -18,393 +18,544 @@
...
-        Ok(sandbox_wasmi::FuncInstance::alloc_host(
-            signature.clone(),
-            idx.0,
-        ))
+fn into_value(value: &wasmi::Val) -> Option<Value> {
+    match value {
+        wasmi::Val::I32(val) => Some(Value::I32(*val)),
+        wasmi::Val::I64(val) => Some(Value::I64(*val)),
+        wasmi::Val::F32(val) => Some(Value::F32((*val).into())),
+        wasmi::Val::F64(val) => Some(Value::F64((*val).into())),
```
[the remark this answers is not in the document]
ByteNacked: Fixed

sandbox/host/src/sandbox/wasmi_backend.rs
ByteNacked: Fixed

sandbox/host/src/sandbox/wasmi_backend.rs
ByteNacked: Fixed

sandbox/host/src/sandbox/wasmi_backend.rs
ByteNacked: Fixed

sandbox/host/src/sandbox/wasmi_backend.rs
ByteNacked: Fixed

sandbox/host/src/sandbox/wasmi_backend.rs
ByteNacked: Fixed

sandbox/host/src/sandbox/wasmi_backend.rs
ByteNacked: Fixed

sandbox/host/src/sandbox/wasmi_backend.rs
ByteNacked: Fixed

sandbox/host/src/sandbox/wasmi_backend.rs
ByteNacked: Fixed

sandbox/host/src/store_refcell.rs
ByteNacked: Fixed

utils/lazy-pages-fuzzer/src/wasmi_backend.rs
ByteNacked: Discussed, added comments

sandbox/host/src/sandbox/wasmi_backend.rs
ByteNacked: Fixed

sandbox/host/src/store_refcell.rs
```diff
@@ -83,38 +83,41 @@
...
     ops::{Deref, DerefMut},
     ptr::NonNull,
 };
 
 use defer::defer;
-use wasmer::{AsStoreMut, AsStoreRef, Store, StoreRef};
 
 #[derive(Debug, Clone, Copy)]
 enum BorrowState {
-    Shared(NonZero<usize>),
```
StackOverflowExcept1on, line 99: why? `type NonZeroUsize = NonZero<usize>`

ByteNacked: I think it makes the code a little bit cleaner.

StackOverflowExcept1on: yes, but NonZero<T> is available since 1.79 why not use generic types?

ByteNacked: I prefer to use non-generic types where possible to avoid overcomplicating things.

utils/lazy-pages-fuzzer/src/wasmer_backend.rs
```diff
@@ -34,6 +34,9 @@ use crate::{
 #[derive(Clone)]
 struct InstanceBundle {
     instance: Instance,
+    // NOTE: Due to the implementation of lazy pages, which need to access the Store to retrieve globals,
+    // we have to use a second mutable reference to the Store in the form of a raw pointer
+    // to use it within the lazy pages' signal handler context.
```
ark0f, line 39: It would be nice to describe why it considered to safe

sandbox/host/src/sandbox/wasmi_backend.rs
```diff
@@ -18,393 +18,534 @@
...
+fn host_trap(msg: impl Into<error::Error>) -> wasmi::Error {
+    wasmi::Error::host(msg.into())
+}
+
+fn into_wasmi_val(value: Value) -> wasmi::Val {
+    match value {
+        Value::I32(val) => wasmi::Val::I32(val),
+        Value::I64(val) => wasmi::Val::I64(val),
+        Value::F32(val) => wasmi::Val::F32(wasmi::core::F32::from_bits(val)),
+        Value::F64(val) => wasmi::Val::F64(wasmi::core::F64::from_bits(val)),
```
ark0f, line 71:
```suggestion
        Value::F64(val) => wasmi::Val::F64(f64::from_bits(val)),
```

Also, please, do the same conversions in `into_value()`

sandbox/host/src/sandbox/wasmi_backend.rs
```diff
@@ -18,393 +18,534 @@
...
+        wasmi::Val::F64(val) => Some(Value::F64(u64::from(*val))),
+        _ => None,
+    }
+}
+
+/// Wasmi specific context
+pub struct Backend {
+    store: Rc<StoreRefCell>,
+    // Allocation should be dropped right after the store is dropped
+    allocations: Vec<Allocation>,
```
ark0f, line 96: Invariant can be violated. Allocations can be dropped before store is dropped because of `Rc` thus lifetime of store can be longer than lifetime of allocations. You need to keep `StoreRefCell` and `Vec<Allocation>` both in the same `Rc`.

ark0f (changes requested):

sandbox/host/src/sandbox/wasmi_backend.rs
ByteNacked: Hmm, your suggested change doesn't even compile. Please note that `f64` and `F64` are different types.

utils/lazy-pages-fuzzer/src/wasmer_backend.rs
ByteNacked: Added

sandbox/host/src/sandbox/wasmi_backend.rs
ByteNacked: Added check in `Drop`

sandbox/host/src/sandbox/wasmi_backend.rs
```diff
@@ -18,393 +18,546 @@
...
-    ) -> std::result::Result<sandbox_wasmi::GlobalRef, sandbox_wasmi::Error> {
-        Err(sandbox_wasmi::Error::Instantiation(format!(
-            "Export {}:{} not found",
-            module_name, field_name
-        )))
+fn into_value(value: &wasmi::Val) -> Option<Value> {
+    match value {
+        wasmi::Val::I32(val) => Some(Value::I32(*val)),
+        wasmi::Val::I64(val) => Some(Value::I64(*val)),
+        wasmi::Val::F32(val) => Some(Value::F32(u32::from(*val))),
```
ark0f, line 86:
```suggestion
        wasmi::Val::F32(val) => Some(Value::F32(val.to_bits())),
```

ark0f (approved):
👍

sandbox/host/src/sandbox/wasmi_backend.rs
ByteNacked: Fixed

ByteNacked:
Seems Wasmi executor unable to sync tests, gonna investigate this problem.

breathx (changes requested):
should we init sandbox for ethexe as it used to be?

ByteNacked:
> should we init sandbox for ethexe as it used to be?

No we're not,
added comment:

```rust
// Sets the global sandbox backend type.
// Buy default, it's set to `Wasmer`, so in case of `Wasmer` it's not necessary to call this function.
pub fn init(sandbox_backend: sandbox_env::SandboxBackend) {
```

breathx (approved):

PR #4288, opened 2024-10-14
refactor(core): Clear ContextStore state created by *_commit syscalls

playX18:
First commit before implementing actual migrations to clear all data creates by `_commit` syscalls family.

### DIFF
--- a/core/src/message/context.rs
+++ b/core/src/message/context.rs
@@ -198,6 +198,13 @@ impl ContextStore {
     pub fn system_reservation(&self) -> Option<u64> {
         self.system_reservation
     }
+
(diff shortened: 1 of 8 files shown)


breathx (changes requested):
Your goal is to fix all the modifying code, not only clear it. We dont need to store/write any related data at all. Moreover, context shouldn't even contain the field

playX18:
Implemented everything that is needed to remove commit state from ContextStore but there's still some tests that are failing and I am unable to figure it out yet.

playX18:
Closing this in favor of opening another PR with much cleaner code and solution to this problem (which does not introduce errors in a fuzzer)

PR #4291, opened 2024-10-16
feat(ethexe): handle incoming transfers: from messaging and from direct WVara erc20.transfer(..)

breathx:


ethexe/processor/src/handling/events.rs
```diff
@@ -151,12 +151,24 @@ impl Processor {
...
         event: WVaraEvent,
     ) -> Result<()> {
         match event {
-            WVaraEvent::Transfer { .. } => {
-                log::debug!("Handler not yet implemented: {event:?}");
+            WVaraEvent::Transfer { from, to, value } => {
+                if let Some(state_hash) = in_block_transitions.state_of(&to) {
+                    if in_block_transitions.state_of(&from).is_none() {
+                        let new_state_hash = self.db.mutate_state(state_hash, |_, state| {
+                            state.balance += value;
```
StackOverflowExcept1on, line 162: it does to.balance += value, but what about from.balance -= value?

breathx: As todo left, there needs additional testing and checking invariants on sendings from external user or local programs. Mostly because we need to also move its balance from router: coming up soon

StackOverflowExcept1on (approved):

PR #4301, opened 2024-10-21
refactor(core-code) Changed type of DispatchKind to bitflag

Lazark0x:
In order to optimize the size of structures, the DispatchKind type was changed from BTreeSet to bitflag(u8)


core-processor/src/ext.rs
```diff
@@ -487,6 +487,7 @@ impl<'a, LP: LazyPagesInterface> ExtMutator<'a, LP> {
                 Ok(packet.gas_limit().unwrap_or(0))
             }
             DispatchKind::Signal => unreachable!("Signals can't be sent as a syscall"),
+            _ => unreachable!("Multiple dispatch kind is impossible here"),
```
ark0f, line 490: Consider to use `enumflags2` create to clearly distinguish `DispatchKind` and set of kinds

ark0f (changes requested):

core-processor/src/ext.rs
Lazark0x: I switched to enumflags2. It turned out not very nice because it is impossible to apply TypeInfo trait on BitFlags<T>.

breathx (approved):

ark0f (approved):

breathx:
Discussion about this PR lead to a decision not to implement it in order to keep codes/programs human-readable from observers. Optimisation doesn't have that much profit to cover expenses

PR #4304, opened 2024-10-22
refactor(core): commit family of syscalls does not keep state between executions

playX18:
New PR based on #4288 but with cleaner code and solution that does not break a fuzzer. Still WIP but all that is left is write a migration and fix tests. 

@breathx


pallets/gear/src/tests.rs
```diff
@@ -12636,9 +12636,12 @@ fn incomplete_async_payloads_kept() {
             None,
             Some("OK PING"),
             Some("OK REPLY"),
-            None,
-            Some("STORED COMMON"),
-            Some("STORED REPLY"),
+            // payload is not kept between executions so we get a panic here
```
breathx, line 12639: tests for dedicated functionality should be removed

core/src/message/context.rs
```diff
@@ -814,18 +814,13 @@ mod tests {
             Error::OutgoingMessagesBytesLimitExceeded,
         );
     }
-
-    #[test]
+    /*     #[test]
```
breathx, line 817: why?

core/src/message/context.rs
```diff
@@ -583,7 +582,8 @@ impl MessageContext {
     }
 
     /// Destructs context after execution and returns provided outcome and store.
-    pub fn drain(self) -> (ContextOutcome, ContextStore) {
+    pub fn drain(mut self) -> (ContextOutcome, ContextStore) {
+        self.store.local_nonce += self.tmp_store.outgoing.len() as u32;
```
breathx, line 586: that's kind of dangerous, since we could adjust local nonce on each access, so I'd go this way

core/src/message/context.rs
```diff
@@ -374,10 +370,13 @@ impl MessageContext {
 
     /// Pushes payload into stored payload by handle.
     pub fn send_push(&mut self, handle: u32, buffer: &[u8]) -> Result<(), Error> {
-        let data = match self.store.outgoing.get_mut(&handle) {
+        let data = match self.tmp_store.outgoing.get_mut(&handle) {
             Some(Some(data)) => data,
             Some(None) => return Err(Error::LateAccess),
-            None => return Err(Error::OutOfBounds),
+            None => {
+                //            debug!("OOB: {:#?}", backtrace::Backtrace::new().frames());
```
breathx, line 377: deadcode

core/src/message/context.rs
```diff
@@ -141,33 +141,39 @@ impl ContextOutcome {
         }
     }
 }
+/// Store of current temporary message execution context.
+#[derive(Clone, Default, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Decode, Encode, TypeInfo)]
+pub struct TmpStore {
```
breathx, line 146: I'd consider other naming options. Could be "Payloads" or something like that

breathx (changes requested):
Also need to fix CI

breathx:
Otherwise, lgtm

core/src/message/context.rs
playX18: Since we got rid of payload storage in Dispatch MessageContext can't be created with payload bytes exceeding limit. This test is removed now

core/src/message/context.rs
```diff
@@ -222,29 +229,16 @@ impl MessageContext {
...
-
         Some(Self {
             kind,
             outcome: ContextOutcome::new(program_id, message.source(), message.id()),
             current: message,
             store: store.unwrap_or_default(),
+            outgoing_payloads: OutgoingPayloads::default(),
             settings,
-            outgoing_bytes_counter,
+            // message context *always* starts with zero bytes outgoing. Before it could be non-zero and eventually
```
breathx, line 239: counter is better to move to OutgoingPayloads itself since in facts it's the size of OutgoingPayloads

core/src/message/context.rs
```diff
@@ -141,34 +141,40 @@ impl ContextOutcome {
         }
     }
 }
+/// Store of current temporary message execution context.
+#[derive(Clone, Default, Debug, Eq, Hash, Ord, PartialEq, PartialOrd, Decode, Encode, TypeInfo)]
+pub struct OutgoingPayloads {
+    outgoing: BTreeMap<u32, Option<Payload>>,
```
breathx, line 147: I'd rename for clearance 
```suggestion
    handle: BTreeMap<u32, Option<Payload>>,
```

pallets/gear-messenger/src/migrations/context_store.rs
```diff
@@ -0,0 +1,195 @@
...
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+use core::marker::PhantomData;
+
+use crate::{Config, Pallet, Waitlist};
```
breathx, line 20: keep imports together and formatted pls
```suggestion
// along with this program. If not, see <https://www.gnu.org/licenses/>.

use core::marker::PhantomData;
use crate::{Config, Pallet, Waitlist};
```

pallets/gear-messenger/src/migrations/context_store.rs
```diff
@@ -0,0 +1,195 @@
...
+
+#[cfg(test)]
+#[cfg(feature = "try-runtime")]
+mod test {
+    use super::*;
+    use crate::mock::{new_test_ext, *};
+    use frame_support::traits::StorageVersion;
+
+    #[test]
+    fn context_store_migration_works() {
```
breathx, line 185: if you impl test, so its better to have some data translated + asserted than`

core/src/message/context.rs
breathx: so it will be named just bytes_counter

breathx (approved):

PR #4311, opened 2024-10-24
refactor(ethexe): Subscribe blocks when `Observer` is created

### LINKED ISSUE #4099: ethexe: append possibility to receive info that service is started to listen blocks

techraed:
Resolves #4099 

TODO:
- [x] use macro to avoid copy-paste in `EventsStreamProducer`

**UPD**:
Still partially resolves #4099


techraed:
I'd say that tests are really fragile: removing timers results in underlying spawned tasks being not synchronized, which creates many obstacles in testing. The approach to test impl should be changed in future, so we can write them easily without thinking too much about underlying details of how `Service` and other entities are implemented.

 [6e450c4](https://github.com/gear-tech/gear/pull/4311/commits/6e450c46487c1b2a372002ad7b667cb973a38186) - this works perfectly fine, but requires some timeout for sync.

[a465d72](https://github.com/gear-tech/gear/pull/4311/commits/a465d72aa3ceae68f3f5a8705bf56829f96ca095) - this avoids sync problems when channgels try to read data, when there's no actual data sent from the `anvil`, because it works in a non-block-production mode. However, it requires increasing a number of attempts to get the receipt for the sent tx (could be the `anvil` problem).

ethexe/ethereum/src/lib.rs
```diff
@@ -297,7 +297,9 @@ impl<T: Transport + Clone, N: Network> TryGetReceipt<T, N> for PendingTransactio
             Err(err) => err,
         };
 
-        for _ in 0..3 {
+        log::trace!("Failed to get transaction receipt for {tx_hash}. Retrying...");
+        for n in 0..25 {
```
StackOverflowExcept1on, line 301: it's ok for tests, but bad for production

StackOverflowExcept1on (approved):

ethexe/cli/src/service.rs
```diff
@@ -418,30 +420,46 @@ impl Service {
...
             mut query,
             mut router_query,
             mut processor,
-            mut sequencer,
-            signer: _signer,
             mut validator,
-            metrics_service,
             rpc,
             block_time,
+            signer: _signer,
```
StackOverflowExcept1on, line 458:
```suggestion
```

ethexe/cli/src/service.rs
```diff
@@ -844,6 +855,62 @@ impl Service {
...
+        let Self {
+            service,
+            events_stream_producer,
+        } = self;
+
+        service
+            .run_inner(events_stream_producer)
+            .await
+            .map_err(|err| {
+                log::error!("Service finished work with error: {:?}", err);
```
StackOverflowExcept1on, line 908:
```suggestion
                log::error!("Service finished work with error: {err:?}");
```

techraed:
After some discussions we came to the conclusion that having blocks automatically produced is not a good solution, so dividing the PR to the 2:
- subscribing to blocks during `Observer` instantiation
- actually solving `sleep` problem

ethexe/cli/src/service.rs
```diff
@@ -418,30 +420,46 @@ impl Service {
...
+        if let Some(metrics_service) = self.metrics_service.take() {
+            tokio::spawn(metrics_service.run(
+                observer.get_status_receiver(),
+                self.sequencer.as_mut().map(|s| s.get_status_receiver()),
+            ));
+        }
+        let events_stream_producer = observer.events_stream_producer();
+
+        self.run_inner(events_stream_producer).await.map_err(|err| {
+            log::error!("Service finished work with error: {:?}", err);
```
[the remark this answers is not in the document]
techraed: not actual

ethexe/ethereum/src/lib.rs
techraed: not actual

ethexe/cli/src/service.rs
techraed: done

StackOverflowExcept1on (approved):

grishasobol (approved):

PR #4320, opened 2024-10-31
refactor(gtest): removed use of read_state_bytes_using_wasm

Lazark0x:
Removed from everywhere except rpc and runtime api


gclient/src/api/rpc.rs
```diff
@@ -353,7 +354,8 @@ impl GearApi {
         .await
     }
 
-    /// Same as [`read_state_using_wasm_by_path`](Self::read_state_using_wasm_by_path), but reads the program's state at the block identified by its hash.
+    /// Same as [`read_state_using_wasm_by_path`](Self::read_state_using_wasm_by_path),
+    /// but reads the program's state at the block identified by its hash.
     pub async fn read_state_bytes_using_wasm_by_path_at(
```
breathx, line 359: gclient is not rpc or runtime api (it's client that call them), so as we agreed, rpc hole itself won't be removed atm, but each new user space libs shouldn't be able to call it, so feel free to remove it from here as well

breathx (changes requested):

gclient/src/api/rpc.rs
Lazark0x: Right, additionally removed `*_using_wasm` functions from `gclient` and `gsdk`

breathx (approved):

PR #4325, opened 2024-11-04
fix(sandbox): Concurrent file operations in embedded executor cache

ark0f:


ByteNacked (approved):
LGFM 👍

utils/gear-wasmer-cache/src/lib.rs
```diff
@@ -167,7 +167,12 @@ impl FileSystemCache {
     fn load(&self, key: Hash) -> Result<Vec<u8>, io::Error> {
         let path = self.path.join(key.to_string());
 
-        fs::read(path)
+        let mut file = File::open(path)?;
+        file.lock_exclusive()?;
```
ByteNacked, line 171: Is  `unlock` called  on drop?

StackOverflowExcept1on: > Furthermore, the lock is released either by an explicit LOCK_UN operation on any of these duplicate descriptors, or when all such descriptors have been closed.

https://linux.die.net/man/2/flock

Cargo.lock
```diff
@@ -6635,6 +6635,7 @@ dependencies = [
 name = "gear-wasmer-cache"
 version = "1.6.2"
 dependencies = [
+ "fs2",
```
StackOverflowExcept1on, line 6638: mb switch to fork of fs2: https://github.com/al8n/fs4-rs? since it is not maintained for 7 years

StackOverflowExcept1on (approved):

utils/gear-wasmer-cache/src/lib.rs
StackOverflowExcept1on: Explicit unlock mb make sense for windows
> If a process terminates with a portion of a file locked or closes a file that has outstanding locks, the locks are unlocked by the operating system. However, the time it takes for the operating system to unlock these locks depends upon available system resources.

https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-lockfileex

ark0f:
To be reimplemented to mitigate concurrent read/write operations for multi-threaded process (PR was done for a few single-threaded processes case)

breathx (approved):

PR #4328, opened 2024-11-08
feat(ethexe): extend ethexe cli

breathx:
Rpc endpoint now starts by default


ethexe/cli/src/config.rs
```diff
@@ -116,14 +120,14 @@ pub struct Config {
     /// Sender address to send Ethereum transaction.
     pub sender_address: Option<String>,
 
-    // Network configuration
+    /// Network configuration.
     pub net_config: Option<NetworkEventLoopConfig>,
 
-    // Prometheus configuration
+    /// Prometheus configuration/
```
grishasobol, line 126: .

ethexe/cli/src/params/rpc.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+impl RpcParams {
+    /// Creates [`RpcConfig`].
+    pub fn as_config(&self) -> Option<RpcConfig> {
+        if self.no_rpc {
+            return None;
+        };
+
+        let ip = if self.rpc_external {
```
StackOverflowExcept1on, line 47: looks like over-engineering. mb rename `rpc_external` to `rpc_addr`

StackOverflowExcept1on (approved):

ethexe/cli/src/params/rpc.rs
breathx: default for substrate (the same approach used for prometheus in cli)

ethexe/cli/src/args.rs
```diff
@@ -89,6 +86,10 @@ pub struct Args {
     #[arg(long, default_value = "12")]
     pub block_time: u64,
 
+    /// Amount of threads to be used.
+    #[arg(long, default_value = "16")]
```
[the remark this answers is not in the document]
breathx: it's not physical threads + it should be able to configure for tests etc

ethexe/processor/src/handling/run.rs
```diff
@@ -34,31 +34,32 @@ enum Task {
 }
 
 pub fn run(
-    threads_amount: usize,
+    num_workers: usize,
```
grishasobol, line 37: I do not recommend to do this renaming. Because this is real threads amount in tokio runtime and we have also workers amount which is amount of virtual threads. Of course we can set real threads amount to the amount of workers but better to specify that it is different things.

ethexe/cli/src/args.rs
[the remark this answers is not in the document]
grishasobol: I would recommend to use here term `virtual threads to be used in processor` instead of just `threads`

PR #4331, opened 2024-11-11
fix(ethexe): Disable jemalloc in RocksDB for MSVC

ark0f:


ethexe/db/src/database.rs
```diff
@@ -101,7 +101,8 @@ impl Clone for Database {
     }
 }
 
-#[derive(Debug, Clone, Default, Encode, Decode, serde::Serialize)]
+#[cfg_attr(test, derive(serde::Serialize))]
```
breathx, line 104: we need it not only for tests. why?

Issue #3752, opened 2024-02-26
Check gas limit for builtin actors before calling their respective `handle` method


ekovalev:
In the current implementation of the builtin actor framework (#3624) the `builtin_dispatcher::run()` doesn't check whether the gas provided is sufficient for the actor to execute a message - this can require decoding the message itself and, hence, is done inside the actor's `handle` function.
We might want to change this flow to let the `builtin_dispatcher` decide upfront if the gas limit is sufficient. One of the ways to do that is to make builtins report their maximum possible incurred gas cost and require the message gas limit to surpass that value. If the actually burned gas turns out to be less that that amount the unused gas will naturally be returned to the gas tree.

breathx:
And skip builtin message processing until next block in case that max of builtin weights is bigger than current gas allowance

PR #4345, opened 2024-11-18
feat(runtime): Track block gas allowance in builtins

ekovalev:
Addresses #3752.

The initial implementation of builtin actors overlooked the need of keeping track of the block gas allowance. So far only individual message gas limit has been taken in consideration.

To address this, the way how gas usage is tracked in builtin actors has been modified. Specifically:
- `BuiltinActor::handle` method signature has been change to return just a `Result<Payload, BuilintActorError` and not a tuple with actual gas burned; instead, gas usage is captured by the `GasCounter` (as a part of mutable `context` passed as an input to `handle`);
- `BuiltinActor::max_gas()` method has been added to the trait to allow builtin actors report the maximum possible amount of gas that can be burned during a message processing. The `handle` would only be called if it fits in the current block in the worst possible case based on the `max_gas` output (similarly to how it is decided whether an extrinsic is added to a block based on its weight). It doesn't have to be 100% accurate, I guess - but better be not too far from real values.

The refactoring has rendered the codebase more readable.

TODO (#4396):

- Fill the `max_gas()` for every builtin actor with actual values - either through benchmarking or theoretically.


pallets/gear-builtin/src/lib.rs
```diff
@@ -93,6 +95,9 @@ pub enum BuiltinActorError {
     /// Actor's inner error encoded as a String.
     #[display(fmt = "Builtin execution resulted in error: {_0}")]
     Custom(LimitedStr<'static>),
+    /// Occurs if a builtin actor execution does not fit in the current block.
+    #[display(fmt = "Block gas allowance exceeded")]
```
breathx, line 99: Im wondering If we could abstract it since it more about how system implemented rather than each particular builtin

pallets/gear-builtin/src/bls12_381.rs
```diff
@@ -143,29 +144,46 @@ fn multi_miller_loop<T: Config>(
...
+    }
+    if GasAllowanceOf::<T>::get() < to_spend {
+        return (
+            Err(BuiltinActorError::GasAllowanceExceeded),
+            gas_limit.saturating_sub(gas_left),
+        );
     }
 
-    gas_spent += to_spend;
+    gas_left -= to_spend;
```
breathx, line 175: saturating?

pallets/gear-builtin/src/bls12_381.rs
```diff
@@ -174,27 +192,38 @@ fn final_exponentiation<T: Config>(
...
```
breathx, line 215: here and in other places

breathx (changes requested):
Generally looks good. Regarding "potential todo" mentioned in the description: nobody should under no circumstances be charged for anything in case of gas allowance exceeded case.

gshep (approved):
Good job! :1st_place_medal:

pallets/gear-builtin/src/bls12_381.rs
ekovalev: Safe due to a check above - if `gas_left` is less than `to_spend` we abort earlier

ekovalev: But for the sake of universality we can use saturated arithmetics - no problem

breathx:
discussed

pallets/gear/src/builtin.rs
```diff
@@ -23,33 +23,48 @@ use gear_core::{
     message::{Payload, StoredDispatch},
 };
 
-pub type HandleFn<E> = dyn Fn(&StoredDispatch, u64) -> (Result<Payload, E>, u64);
+/// Builtin actor `handle` function signature.
+pub type HandleFn<C, E> = dyn Fn(&StoredDispatch, &mut C) -> Result<Payload, E>;
+
+/// Builtin actor `max_gas` function signature.
+// TODO: let the weight function take a complexity argument similar to extrinsics weight functions
```
gshep, line 30: Sounds like a new epic story so definitely needs an issue =)

pallets/gear-builtin/src/lib.rs
```diff
@@ -108,7 +113,49 @@ impl From<BuiltinActorError> for ActorExecutionErrorReplyReason {
...
                 ActorExecutionErrorReplyReason::Trap(TrapExplanation::Panic(e))
             }
+            BuiltinActorError::GasAllowanceExceeded => {
+                unreachable!("Never supposed to be converted to error reply reason")
+            }
+        }
+    }
+}
+
+/// TODO:
```
gshep, line 123: It would be nice to have some description

pallets/gear-builtin/src/lib.rs
```diff
@@ -118,8 +165,11 @@ pub trait BuiltinActor {
     /// Handles a message and returns a result and the actual gas spent.
     fn handle(
         dispatch: &StoredDispatch,
-        gas_limit: u64,
-    ) -> (Result<Payload, BuiltinActorError>, u64);
+        context: &mut BuiltinContext,
+    ) -> Result<Payload, BuiltinActorError>;
+
+    /// Returns the maximum gas that can be spent by the actor.
```
gshep, line 171: Does  zero value get interpreted in special way?

gshep (approved):

pallets/gear-builtin/src/lib.rs
ekovalev: > Does zero value get interpreted in special way?

Not really: it'll just always pass the "initial screening" in the `BuiltinDispatcher`, but checks inside the `handle()` still remain (every time the gas counter is charged). However, if the `max_gas()` reports an accurate estimation, the `handle` would be skipped entirely and message re-queued. In the latter case we are certain no gas is going to be wasted if we still have to requeue the message after all, but only know it half way.

pallets/gear-builtin/src/lib.rs
```diff
@@ -220,45 +273,41 @@ pub mod pallet {
...
-                        ))),
-                        actual_gas,
-                    )
-                }
-            }
+            // Now actually charge the gas
+            context.try_charge_gas(actual_gas)?;
+
+            res.map(|_| {
+                log::debug!(
```
breathx, line 295: you can use inspect and inspect_err with such purpose across the code base

breathx (approved):

pallets/gear-builtin/src/staking.rs
```diff
@@ -122,18 +122,19 @@ where
...
-        let Ok(request) = Request::decode(&mut payload) else {
-            return (Err(BuiltinActorError::DecodingError), 0);
-        };
+        let request =
+            Request::decode(&mut payload).map_err(|_| BuiltinActorError::DecodingError)?;
 
         // Handle staking requests
         let call = Self::cast(request);
-        let (result, gas_spent) = Pallet::<T>::dispatch_call(origin, call, gas_limit);
+        Pallet::<T>::dispatch_call(origin, call, context).map(|_| Default::default())
```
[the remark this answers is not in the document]
ekovalev: > why is it always returns default also?

The `handle` signature assumes it always returns some payload in case of success, even if no meaningful return message is expected by the sender. This is exactly the case for the `staking` builtin who just relays messages converting them into dispatchable calls for the `pallet_staking`.

pallets/gear-builtin/src/bls12_381.rs
```diff
@@ -32,250 +32,189 @@ pub struct Actor<T: Config>(PhantomData<T>);
...
-    }
-
-    gas_spent += to_spend;
+    context.try_charge_gas(to_spend)?;
 
     let mut items = vec![0u8; len as usize];
     let bytes_slice = items.as_mut_slice();
-    let result = input.read(bytes_slice).map(|_| items).map_err(|_| {
+
+    input.read(bytes_slice).map(|_| items).map_err(|_| {
```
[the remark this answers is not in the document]
ekovalev: Not really. I guess this is a bit beyond the scope of the current PR ).

pallets/gear-builtin/src/lib.rs
```diff
@@ -318,32 +375,34 @@ impl<T: Config> BuiltinDispatcher for BuiltinRegistry<T> {
             );
         }
 
-        // Creating a gas counter to track gas usage (because core processor needs it).
-        let mut gas_counter = GasCounter::new(gas_limit);
+        // We only allow a message to even start processing if it can fit into the current block
+        // (where the gas estimation is as accurate as the actor's `max_gas()` method).
+        let current_gas_allowance = GasAllowanceOf::<T>::get();
+        if max_gas() > current_gas_allowance {
```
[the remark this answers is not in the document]
ekovalev: This is a long-standing discussion. To find out which of the potential branches will be taken inside the `handle` we must enter the function and at least decode a part of the payload. The `bls12_381` actor does this but we can't expect or enforce every builtin actor to do that. The whole purpose of introducing the `max_gas()` method is to be able to predict form the outside how much gas the `handle` would require in the worst case. Very similar to how we assess the extrinsic max weight and the actual post-dispatch weight.
It's quite a rare case anyway to find a call to a builtin actor at the very end of a block.

pallets/gear-builtin/src/proxy.rs
```diff
@@ -77,20 +77,18 @@ where
...
-                (result.map(|_| Default::default()), actual_gas)
-            }
-            Err(e) => (Err(e), gas_limit),
-        }
+        let call = Self::cast(request)?;
+        Pallet::<T>::dispatch_call(origin, call, context).map(|_| Default::default())
+    }
+
+    fn max_gas() -> u64 {
+        Default::default()
```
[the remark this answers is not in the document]
ekovalev: Interesting idea. Instead of the complexity parameter (i.e. the payload size) we could let `max_gas` take some small fraction of the payload. Actually, each `max_gas` could interpret it differently depending whether we gauge the gas cost though benchmarking or theoretical considerations.

PR #4348, opened 2024-11-21
fix(lazy-pages-fuzzer): Increase Wasmi's max stack size

ByteNacked:
Increase Wasmi's maximum stack size in the lazy-pages fuzzer.
This should fix failures caused by stack overflows.

@reviewer-or-team


breathx (approved):
But I can't actually remember when we've faced fuzzer stack overflows

ByteNacked:
> But I can't actually remember when we've faced fuzzer stack overflows

We're facing them right now, check lazy-page fuzzer alerts in tg.

PR #4351, opened 2024-11-23
feat(gcore): Stack-based logging

ark0f:


gcore/src/utils.rs
```diff
@@ -113,3 +149,40 @@ pub mod ext {
         unsafe { gsys::gr_oom_panic() }
     }
 }
+
+/// Add a debug message to the log.
+///
+/// Same as `gstd::debug!()` but uses a stack-allocated buffer.
```
StackOverflowExcept1on, line 155: it can use stack, but not always:
https://github.com/gear-tech/gear/blob/b9511183910d6b1ff6a018a5cadb72dc784b4fee/stack-buffer/src/lib.rs#L100-L104

ark0f: https://github.com/gear-tech/gear/pull/4351/files#diff-2203746c3be74ee3be9d80b919b9ed85f5fe98f45a953f96f7cb92eae852d6aeR92

StackOverflowExcept1on: ok, add same note for macro. not only for `fn stack_debug()`

gcore/src/utils.rs
```diff
@@ -38,8 +38,9 @@ impl AsRawPtr for ReservationId {}
 
 /// Extensions for additional features.
 pub mod ext {
-    #[cfg(any(feature = "debug", debug_assertions))]
-    use crate::errors::{Error, Result};
+    use crate::stack_buffer;
+    use core::{fmt, fmt::Write as _, mem::MaybeUninit};
```
StackOverflowExcept1on, line 42:
```suggestion
    use core::{fmt::{self, Write as _}, mem::MaybeUninit};
```

gcore/src/utils.rs
```diff
@@ -50,16 +51,51 @@ pub mod ext {
...
+            buf: &'a mut [MaybeUninit<u8>],
+            pos: usize,
+        }
+
+        impl fmt::Write for StackFmtWriter<'_> {
+            fn write_str(&mut self, s: &str) -> fmt::Result {
+                let upper_bound = (self.pos + s.len()).min(MAX_BUFFER_SIZE);
+                if let Some(buf) = self.buf.get_mut(self.pos..upper_bound) {
+                    let buf = buf as *mut [MaybeUninit<u8>] as *mut [u8];
+                    let s = &s.as_bytes()[..buf.len()];
```
StackOverflowExcept1on, line 78: we could avoid panicking here and return `fmt::Error`. instead of `[..buf.len()]` use `s.as_bytes().get(..buf.len())`.

ark0f: It's not panicking. It's requirement of `copy_from_slice` to both slices to be equal length and trimming in case of message exceeds buffer

StackOverflowExcept1on: `ops::Index` (`[..buf.len()]`) is panicking operation

breathx (approved):

PR #4362, opened 2024-11-27
feat(Gear.exe): impl initial election logic

grishasobol:
update libraries


ethexe/contracts/foundry.toml
```diff
@@ -1,4 +1,5 @@
 [profile.default]
+solc_version = "0.8.26"
```
StackOverflowExcept1on, line 2: pin it to latest version
```suggestion
solc_version = "0.8.28"
```

StackOverflowExcept1on:
It is better to update dependencies to specific version rather than master (applies to all dependencies except symbiotic). Because openzeppelin-contracts probably has an audit for 5.1.0 tag, but does not for master.

ethexe/contracts/src/Middleware.sol
```diff
@@ -267,12 +276,50 @@ contract Middleware {
...
+    function makeElectionAt(uint48 ts, uint256 maxValidators) public view returns (address[] memory) {
+        require(maxValidators > 0, "Max validators must be greater than zero");
+
+        (address[] memory activeOperators, uint256[] memory stakes) = getActiveOperatorsStakeAt(ts);
+
+        if (activeOperators.length <= maxValidators) {
+            return activeOperators;
+        }
+
+        // Bubble sort descending
```
StackOverflowExcept1on, line 288: openzeppelin has quicksort:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/Arrays.sol

ethexe/contracts/foundry.toml
grishasobol: done

grishasobol:
> It is better to update dependencies to specific version rather than master (applies to all dependencies except symbiotic). Because openzeppelin-contracts probably has an audit for 5.1.0 tag, but does not for master.

Specify last release tag for openzeppelin libs

.gitmodules
```diff
@@ -4,12 +4,15 @@
 [submodule "ethexe/contracts/lib/openzeppelin-contracts"]
```
StackOverflowExcept1on, line 4: plz also add tag for forge-std

StackOverflowExcept1on: > Specify last release tag for openzeppelin libs

it's ok, 5.1.0 is latest version of openzeppelin: https://github.com/OpenZeppelin/openzeppelin-contracts/releases

ethexe/contracts/src/Middleware.sol
grishasobol: I do not see how it can be usable here, because stake and operators must be sort together. Of course we can sort indexes but this will cause additional elements cloning after sorting. Also this is supposed to be off-chain call only

ethexe/contracts/test/Middleware.t.sol
```diff
@@ -78,6 +81,79 @@ contract MiddlewareTest is Test {
...
+            assertEq(res[1], operators[1]);
+            assert(res[2] == operators[0] || res[2] == operators[2] || res[2] == operators[4]);
+        }
+
+        {
+            address[] memory res = middleware.makeElectionAt(uint48(vm.getBlockTimestamp() - 1), 4);
+            assertEq(res.length, 4);
+            assertEq(res[0], operators[3]);
+            assertEq(res[1], operators[1]);
+            assert(res[2] == operators[0] || res[2] == operators[2] || res[2] == operators[4]);
```
StackOverflowExcept1on, line 134: assert - solidity built-in, assertTrue - foundry test-specific
```suggestion
            assertTrue(res[2] == operators[0] || res[2] == operators[2] || res[2] == operators[4]);
```

StackOverflowExcept1on (approved):

ethexe/contracts/test/Middleware.t.sol
grishasobol: Done

PR #4375, opened 2024-12-05
feat(ethexe): optimize inheritor field passed to eth

breathx:


breathx:
This optimization isn't needed rn

PR #4379, opened 2024-12-06
feat: Initial `multivalue` WASM proposal support

ark0f:


Cargo.toml
```diff
@@ -592,9 +592,10 @@ debug = true
 [patch.crates-io]
 # sp-io = { version = "38.0.0", git = "https://github.com/gear-tech/polkadot-sdk.git", branch = "gear-polkadot-stable2409" }
 # sp-application-crypto = { version = "38.0.0", git = "https://github.com/gear-tech/polkadot-sdk.git", branch = "gear-polkadot-stable2409" }
+
+# these patched dependecies force their `sign_ext` feature to be used by Substrate dependencies
```
StackOverflowExcept1on, line 596: Are these patches necessary? Substrate (runtime) does not use `sign-ext`: https://github.com/paritytech/polkadot-sdk/blob/906fa9e51306635245a22e03160d1c761fae6cc3/substrate/utils/wasm-builder/src/wasm_project.rs#L850. Same goes for outdated version wasmi 0.13.2 in our Cargo.lock.

utils/wasm-instrument/Cargo.toml
```diff
@@ -17,6 +17,10 @@ gwasm-instrument = { workspace = true, features = ["sign_ext"] }
 derive_more.workspace = true
 enum-iterator.workspace = true
 
+# TODO: remove when tooling is updated
```
StackOverflowExcept1on, line 20: Before updating tooling, it makes sense to replace `parity-wasm` with `gear-wasm` in all our crates (except wasmi 0.13.2 and other substrate-related dependencies).

Also try applying existing patches to our tooling as git dependency:
- https://github.com/gear-tech/parity-wasm/pull/4
- https://github.com/gear-tech/wasm-instrument/pull/8

StackOverflowExcept1on (approved):

utils/wasm-instrument/Cargo.toml
StackOverflowExcept1on: It should also be under `[target.'cfg(force_wasmer_cranelift_i_know_what_i_do)'.dependencies]`.

playX18 (approved):

StackOverflowExcept1on:
One more note: the `multivalue` function refers not only to `fn f() -> (i32, i32)`, but also to the fact that `loop`, `if`, `block` instructions can input/output multiple parameters. This might be worth checking and putting restrictions on it (as we do for `call_indirect`). Here's an example wat: https://github.com/gear-tech/wasm-instrument/blob/multi-value/tests/fixtures/stack-height/multi_value_if.wat.

core/src/gas_metering/schedule.rs
```diff
@@ -117,6 +117,8 @@ pub struct Limits {
     #[doc = " is not weight metered its costs must be static (via this limit) and included in"]
     #[doc = " the costs of the instructions that cause them (call, call_indirect)."]
     pub parameters: u32,
+    #[doc = " Maximum numbers of parameters a function can have."]
```
grishasobol, line 120: Parameters ?

pallets/gear/src/benchmarking/mod.rs
```diff
@@ -1840,6 +1840,36 @@ benchmarks! {
...
+                    num_elements,
+                    function_index: OFFSET_AUX,
+                    init_elements: Default::default(),
+                }),
+                .. Default::default()
+            }))
+        };
+    }: {
+        #[cfg(force_wasmer_cranelift_i_know_what_i_do)]
+        sbox.invoke();
```
grishasobol, line 1870: I think this leads to the fact that automatic benchmarks run would zeroes this weight in weights.rs

ark0f:
Close as non-substantial

PR #4381, opened 2024-12-07
feat(code-backend) implemented extractors in backend to avoid accidental memory registry creation

### LINKED ISSUE #3891: Use extractors in backend to avoid accidental memory registry creation
We can avoid manager creation at all via actix-like extractors in syscall signature. But this is for the future PR or there will be too many changes

_Originally posted by @ark0f in https://github.com/gear-tech/gear/pull/3846#discussion_r1560673990_

Lazark0x:
Resolves #3891 

Implemented extractors as proposed in issue #3891 
Extractors are used in methods with a variable number of arguments, while having a variable number of constructor arguments themselves.
After all extractors have been created, the MemoryAccessRegistry::pre_process() method is called
If the method did not need any extractors, the MemoryAccessRegistry::pre_process() method will not be called either


core-backend/src/funcs.rs
```diff
@@ -592,54 +584,55 @@ where
...
-                    };
-                    let res = f();
-                    let unlock_bound = ctx.ext_mut().unlock_payload(payload_access.into_lock());
-
-                    DropPayloadLockBound::from((unlock_bound, res))
-                })
-                .into_inner()
-                .map_err(Into::into)
-        })
+    pub fn read(at: u32, buffer: Write) -> impl Syscall<Caller> {
```
breathx, line 587: isn't it a bug, changing order of args?

breathx (changes requested):

core-backend/src/funcs.rs
Lazark0x: Yeah, but no. The arguments are in reverse order, but there is no other way but to change the syscall signature, which is obviously not worth it.

```
let ptr = SyscallValue(args[1]).try_into()?;
let size = SyscallValue(args[0]).try_into()?;
```

core-backend/src/funcs.rs
```diff
@@ -128,14 +87,20 @@ impl From<u32> for SyscallReturnValue {
...
+
+    fn gas(&self) -> Option<Gas>;
 }
 
 impl SyscallContext for () {
     fn from_args(args: &[Value]) -> Result<(Self, &[Value]), HostError> {
         Ok(((), args))
     }
+
+    fn gas(&self) -> Option<Gas> {
```
ark0f, line 101: I actually do not see the reason of such changes (for `run_fallible` and `run_any`, too). Could you clarify, please?

core-backend/src/accessors.rs
```diff
@@ -0,0 +1,486 @@
...
+        Ok(registry.as_mut().unwrap().register_write_as(ptr))
+    }
+
+    fn post_process<Caller, Ext>(output: Self::Output, _ctx: &CallerWrap<Caller>) -> Self {
+        Self { write: output }
+    }
+}
+
+impl Read {
+    pub fn value(self) -> Result<Vec<u8>, MemoryAccessError> {
```
ark0f, line 419:
```suggestion
    pub fn into_inner(self) -> Result<Vec<u8>, MemoryAccessError> {
```

core-backend/src/runtime.rs
```diff
@@ -71,6 +77,23 @@ where
...
+        self.io = Some(io);
+    }
+
+    pub fn io_ref(
+        &self,
+    ) -> Result<&MemoryAccessIo<Caller, BackendMemory<ExecutorMemory>>, MemoryAccessError> {
+        match &self.io {
+            Some(Ok(io)) => Ok(io),
+            Some(Err(err)) => Err(err.clone()),
+            None => Err(MemoryAccessError::ManagerError),
```
ark0f, line 94:
```suggestion
            None => unreachable!(),
```
Syscall didn't set its requirement for the manager and thus implementation is invalid and should be never met in runtime

core-backend/src/funcs.rs
```diff
@@ -183,26 +151,55 @@ where
...
+                    index += args_count;
                 )+
-                Ok((self)($($generic),+))
+
+                if(index != ARGS_AMOUNT) {
+                    return Err(HostError);
+                }
+
+                if registry.is_some() {
+                    let io = registry.unwrap().pre_process(caller);
```
ark0f, line 191: Error can be set like in `CallerWrap::set_termination_reason()` and `HostError` returned, so you don't need to handle late return

ark0f (changes requested):

core-backend/src/funcs.rs
```diff
@@ -183,26 +151,55 @@ where
...
+                    let args_count = $generic::REQUIRED_ARGS;
+                    let args_slice = &args[index..index + args_count];
+                    let $generic = $generic::pre_process(&mut registry, args_slice).map_err(|_| HostError)?;
+                    index += args_count;
                 )+
-                Ok((self)($($generic),+))
+
+                if(index != ARGS_AMOUNT) {
+                    return Err(HostError);
+                }
```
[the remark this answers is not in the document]
Lazark0x: Done

core-backend/src/funcs.rs
```diff
@@ -183,26 +151,55 @@ where
...
+
+                if registry.is_some() {
+                    let io = registry.unwrap().pre_process(caller);
+                    caller.set_io(io);
+                }
```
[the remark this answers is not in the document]
Lazark0x: I agree, thanks

core-backend/src/funcs.rs
```diff
@@ -372,8 +380,18 @@ where
...
 
         let (ctx, args) = Call::Context::from_args(args)?;
-        let syscall = builder.build(args)?;
+
+        if ctx.gas().is_some() {
+            caller
+                .state_mut()
+                .ext
+                .decrease_current_counter_to(ctx.gas().unwrap());
+        }
```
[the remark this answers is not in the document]
Lazark0x: Done

core/src/memory.rs
```diff
@@ -208,7 +208,7 @@ pub trait Memory<Context> {
     fn size(&self, ctx: &Context) -> WasmPagesAmount;
 
     /// Set memory region at specific pointer.
-    fn write(&self, ctx: &mut Context, offset: u32, buffer: &[u8]) -> Result<(), MemoryError>;
+    fn write(&self, ctx: &Context, offset: u32, buffer: &[u8]) -> Result<(), MemoryError>;
```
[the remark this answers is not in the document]
Lazark0x: Okay, that's possible

core-backend/src/accessors.rs
```diff
@@ -0,0 +1,486 @@
...
+        registry: &mut Option<MemoryAccessRegistry<Caller>>,
+        args: &[Value],
+    ) -> Result<Self::Output, HostError>
+    where
+        Caller: AsContextExt<State = HostState<Ext, BackendMemory<ExecutorMemory>>>,
+        Ext: BackendExternalities + 'static,
+    {
+        if args.len() != Self::REQUIRED_ARGS {
+            return Err(HostError);
+        }
```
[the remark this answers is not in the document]
Lazark0x: Done

core-backend/src/accessors.rs
```diff
@@ -0,0 +1,486 @@
...
+
+        let ptr = SyscallValue(args[0]).try_into()?;
+        let size = SyscallValue(args[1]).try_into()?;
+
+        if size as usize > N {
+            Ok(None)
+        } else {
+            Ok(Some(registry.as_mut().unwrap().register_read(ptr, size)))
```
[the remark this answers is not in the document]
Lazark0x: Yeah, you're right, thanks

core-backend/src/accessors.rs
```diff
@@ -0,0 +1,486 @@
...
+    {
+        Self {
+            size: output.size,
+            result: ctx.io_ref().and_then(|io| io.read(ctx, output)),
+        }
+    }
+}
+
+impl<const N: usize> SyscallArg for ReadLimited<N> {
+    type Output = Option<WasmMemoryRead>;
```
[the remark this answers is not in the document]
Lazark0x: I rewrote it with your version

core-backend/src/funcs.rs
Lazark0x: Getting gas from SyscallContext is necessary to set the amount of gas in Ext::decrease_current_counter_to() before calling Builder::build() and Syscall::execute(), because MemoryAccessRegistry::pre_process() consumes gas that we would not otherwise include

core-backend/src/runtime.rs
Lazark0x: Agree

core-backend/src/funcs.rs
Lazark0x: This would be a great idea, we could get rid of a lot of code, except for the fact that when an error occurs in FallibleSyscall, the error content is written to memory in different formats depending on the type of syscall

ark0f (changes requested):

core-backend/src/funcs.rs
```diff
@@ -183,26 +151,45 @@ where
...
             Call: Syscall<Caller, Res>,
-            $( $generic: TryFrom<SyscallValue, Error = HostError>,)+
+            Caller: AsContextExt<State = HostState<Ext, BackendMemory<ExecutorMemory>>>,
+            Ext: BackendExternalities + 'static,
+            $( $generic: SyscallArg, )+
         {
-            fn build(self, args: &[Value]) -> Result<Call, HostError> {
-                const ARGS_AMOUNT: usize = impl_syscall_builder!(@count $($generic),+);
+            fn build(self, ctx: &mut MemoryCallerContext<Caller>, args: &[Value]) -> Result<Call, HostError> {
+                let ARGS_AMOUNT: usize = 0 $( + $generic::REQUIRED_ARGS )+;
```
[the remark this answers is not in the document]
Lazark0x: I can't do this because of the fact that these generics are declared at the impl level, not at the function level [E0401](https://doc.rust-lang.org/error_codes/E0401.html)
And rust's compiler «hoist» for inner consts

core-backend/src/accessors.rs
```diff
@@ -0,0 +1,513 @@
...
+        Ext: BackendExternalities + 'static;
+}
+
+pub(crate) struct Read {
+    result: Result<Vec<u8>, MemoryAccessError>,
+    size: u32,
+}
+
+pub(crate) struct ReadLimited<const N: usize = { Payload::MAX_LEN }> {
+    result: Result<Result<Vec<u8>, MemoryAccessError>, PayloadSizeError>,
```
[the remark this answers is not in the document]
Lazark0x: Done

core-backend/src/funcs.rs
```diff
@@ -183,26 +151,45 @@ where
...
+
+                debug_assert_eq!(args.len(), ARGS_AMOUNT);
```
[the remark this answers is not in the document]
Lazark0x: Thanks, my mistake

core-backend/src/funcs.rs
Lazark0x: All in all I don't know what would be the best way to do it, let it be your variant =)

core-backend/src/runtime.rs
```diff
@@ -23,14 +23,64 @@ use crate::{
...
 };
 use gear_core::{costs::CostToken, pages::WasmPage};
 use gear_sandbox::{AsContextExt, HostError};
 use gear_wasm_instrument::SyscallName;
 
+pub(crate) type MemoryAccessIoOption<Caller> =
+    Option<Result<MemoryAccessIo<Caller, BackendMemory<ExecutorMemory>>, MemoryAccessError>>;
+pub(crate) struct MemoryCallerContext<'a, Caller> {
+    pub caller_wrap: CallerWrap<'a, Caller>,
+    pub memory_wrap: MemoryAccessIoWrap<Caller>,
```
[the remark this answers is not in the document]
breathx: > This is done specifically to be able to get mut refs to both of these structure fields for cases like: `io.write(&mut ctx.caller_wrap, ...)` where `io` is `&mut MemoryAccessIo`

for cases like gr_read?

core-backend/src/accessors.rs
```diff
@@ -0,0 +1,525 @@
...
+    {
+        match output {
+            Some(output) => Self {
+                result: ctx
+                    .memory_wrap
+                    .io_mut_ref()
+                    .and_then(|io| io.read_decoded(&mut ctx.caller_wrap, output)),
+            },
+            None => Self {
+                result: Ok(Default::default()),
```
breathx, line 392: should we return option for the case?

core-backend/src/funcs.rs
```diff
@@ -128,13 +85,13 @@ impl From<u32> for SyscallReturnValue {
     }
 }
 
-pub(crate) trait SyscallContext: Sized {
-    fn from_args(args: &[Value]) -> Result<(Self, &[Value]), HostError>;
+pub(crate) trait SyscallContext: Sized + Copy {
+    fn from_args(args: &[Value]) -> Result<(Self, Option<Gas>, &[Value]), HostError>;
```
breathx, line 89: why is gas optional?

core-backend/src/funcs.rs
```diff
@@ -286,31 +278,31 @@ where
...
     fn execute(
         self,
-        caller: &mut CallerWrap<Caller>,
+        caller: &mut MemoryCallerContext<Caller>,
         context: Self::Context,
         syscall_name: SyscallName,
     ) -> Result<(Gas, ()), HostError> {
         let Self { token, f, .. } = self;
-        let FallibleSyscallContext { gas, res_ptr } = context;
+        let FallibleSyscallContext { res_ptr, .. } = context;
```
breathx, line 286: remove this trampoline for future errors from everywhere please
```suggestion
        let FallibleSyscallContext { res_ptr } = context;
```

core-backend/src/funcs.rs
```diff
@@ -453,50 +416,46 @@ where
     }
 
     pub fn send(
-        pid_value_ptr: u32,
-        payload_ptr: u32,
-        len: u32,
+        pid_value: ReadAs<HashWithValue>,
+        payload: ReadPayloadLimited,
```
breathx, line 420: can we maybe do it somehow like ReadAs< PayloadLimited >?

core-backend/src/funcs.rs
```diff
@@ -616,55 +574,60 @@ where
...
+                    .into_inner()
+                    .map_err(Into::into)
+            },
+        )
     }
 
-    pub fn size(size_ptr: u32) -> impl Syscall<Caller> {
-        InfallibleSyscall::new(CostToken::Size, move |ctx: &mut CallerWrap<Caller>| {
-            let size = ctx.ext_mut().size()? as u32;
+    pub fn size(size_write: WriteAs<[u8; 4]>) -> impl Syscall<Caller> {
```
breathx, line 602: WriteAS< u32 > doesnt work? u32: Encode resulting in le bytes

core-backend/src/funcs.rs
```diff
@@ -1183,160 +1162,167 @@ where
...
+                gas.write(ctx, gas_available.to_le_bytes())
                     .map_err(Into::into)
             },
         )
     }
 
-    pub fn message_id(message_id_ptr: u32) -> impl Syscall<Caller> {
-        InfallibleSyscall::new(CostToken::MsgId, move |ctx: &mut CallerWrap<Caller>| {
-            let message_id = ctx.ext_mut().message_id()?;
+    pub fn message_id(message_id_write: WriteAs<[u8; 32]>) -> impl Syscall<Caller> {
```
breathx, line 1186: etc 
```suggestion
    pub fn message_id(message_id_write: WriteAs<MessageId>) -> impl Syscall<Caller> {
```

breathx (approved):
Lgtm

@ark0f pls take another look

core-backend/src/runtime.rs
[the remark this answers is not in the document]
Lazark0x: Yep, and for any write call, where we need `&mut ctx`

core-backend/src/funcs.rs
Lazark0x: Certainly not as `ReadAs`, because reading a payload requires a length argument as well, `ReadAs` only works for structures with known size
The need for `ReadPayloadLimited` comes from the encapsulation of early length checking and the conversion of bytes into `Payload`

core-backend/src/funcs.rs
Lazark0x: Yeah, you're right, converting to bytes was redundant here

core-backend/src/funcs.rs
Lazark0x: Yep

core-backend/src/funcs.rs
Lazark0x: Ok

core-backend/src/funcs.rs
Lazark0x: Since we have an `impl SyscallContext for ()` that doesn't have gas

core-backend/src/accessors.rs
Lazark0x: The original code from `fn register_and_read_value` returned 0 if the address matched `PTR_SPECIAL`, I think this is the desired behavior...

core-backend/src/funcs.rs
breathx: seems we could remove it? @ark0f ?

core-backend/src/accessors.rs
breathx: just from coding perspective 
meet PTR_SPECIAL => return None => unwrap_or_default 
but not a big deal actually

breathx (approved):
once replied by @ark0f, maybe better to remove option for gas - otherwise looks good

core-backend/src/funcs.rs
```diff
@@ -183,26 +147,53 @@ where
...
+                    let args_slice = &args[index..index + args_count];
+                    let $generic = $generic::pre_process(&mut registry, args_slice)?;
+                    index += args_count;
+                )+
+
+                debug_assert_eq!(index, ARGS_AMOUNT);
+
+                if index != ARGS_AMOUNT {
+                    return Err(HostError);
+                }
```
ark0f, line 182:
```suggestion
                debug_assert_eq!(index, ARGS_AMOUNT);
```

ark0f (approved):
🎉

core-backend/src/funcs.rs
Lazark0x: `RawSyscall` replaced with `InfallibleSyscall` with `Null` gas cost

PR #4397, opened 2024-12-12
fix(ethexe): allocation failure on huge journal returned from runtime 

breathx:


ethexe/runtime/src/wasm/api/mod.rs
```diff
@@ -46,13 +46,18 @@ fn _run(arg_ptr: i32, arg_len: i32) -> i64 {
...
-    let res = run::run(
+    let journal = run::run(
         program_id,
         original_code_id,
         state_root,
         maybe_instrumented_code,
     );
 
+    let chunks = journal.encoded_size() / 32 * 1024 * 1024 + 1; // never zero
+    let chunk_size = (journal.len() / chunks).max(1); // never zero
```
grishasobol, line 57: Better to panic in case it is zero because in other case we would panic in allocator

grishasobol (commented):
Would be greate if you add test or TODO to make test for big memory

PR #4405, opened 2024-12-17
refactor(ethexe): use fixed size array instead of BTreeMap in MemoryPages

playX18:
@breathx

### DIFF
--- a/ethexe/runtime/common/src/state.rs
+++ b/ethexe/runtime/common/src/state.rs
@@ -25,7 +25,12 @@ use alloc::{
     vec::Vec,
 };
 use anyhow::{anyhow, Result};
-use core::{any::Any, marker::PhantomData, mem};
+use core::{
+    any::Any,
+    marker::PhantomData,
+    mem,
+    ops::{Index, IndexMut},
+};
 use ethexe_common::gear::Message;
 use gear_core::{
     ids::{prelude::MessageIdExt as _, ProgramId},
@@ -40,6 +45,8 @@ use gear_core_errors::{ReplyCode, SuccessReplyReason};
 use gprimitives::{ActorId, MessageId, H256};
 use parity_scale_codec::{Decode, Encode};
 use private::Sealed;
+#[cfg(feature = "std")]
+use serde::{de::Visitor, ser::SerializeTuple};
 
 pub use gear_core::program::ProgramState as InitStatus;
 
@@ -239,6 +246,14 @@ impl<S: Sealed> MaybeHashOf<S> {
         if let Some(other) = other {
             *self = other;
         }
+    }
+
+    pub fn map<T>(&self, f: impl FnOnce(HashOf<S>) -> T) -> Option<T> {
+        self.0.map(f)
+    }
+
+    pub fn take(&mut self) -> Self {
+        Self(self.0.take())
     }
 }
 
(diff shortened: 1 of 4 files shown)


ethexe/runtime/common/src/state.rs
```diff
@@ -812,12 +827,81 @@ impl Mailbox {
...
+
+impl IndexMut<RegionIdx> for MemoryPages {
+    fn index_mut(&mut self, index: RegionIdx) -> &mut Self::Output {
+        &mut self.0[index.0 as usize]
+    }
+}
+
+/// An inner structure for [`MemoryPages`]. Has at most 256 entries because
+/// [`RegionIdx`] is `u8` internally.
+pub type MemoryPagesInner = [MaybeHashOf<MemoryPagesRegion>; 256];
```
grishasobol, line 904: Why we should support all 256 ? We use only 16 regions , so what the reason to make it 256 size which increase encode/decode serialize/deserialize time significally

grishasobol (changes requested):

ethexe/runtime/common/src/state.rs
```diff
@@ -240,6 +247,14 @@ impl<S: Sealed> MaybeHashOf<S> {
             *self = other;
         }
     }
+
+    pub fn map<T>(&self, f: impl FnOnce(HashOf<S>) -> T) -> Option<T> {
```
breathx, line 251: `map` is a copy of `with_hash`. `take` seems unnecessary

ethexe/runtime/common/src/state.rs
breathx: Agree. Please use constant from below

breathx (changes requested):

ethexe/runtime/common/src/state.rs
```diff
@@ -812,12 +819,83 @@ impl Mailbox {
...
 pub struct MemoryPages(MemoryPagesInner);
 
-// TODO (breathx): replace BTreeMap here with fixed size array.
-pub type MemoryPagesInner = BTreeMap<RegionIdx, HashOf<MemoryPagesRegion>>;
+impl Default for MemoryPages {
+    fn default() -> Self {
+        Self([MaybeHashOf::empty(); MemoryPages::REGIONS_AMOUNT])
+    }
+}
+// implemented manually because serde_derive does not work with arrays larger than 32 elements
```
breathx, line 830: now its less than 32

ethexe/runtime/common/src/state.rs
```diff
@@ -812,12 +819,83 @@ impl Mailbox {
...
+    fn index(&self, idx: RegionIdx) -> &Self::Output {
+        &self.0[idx.0 as usize]
+    }
+}
+
+impl IndexMut<RegionIdx> for MemoryPages {
+    fn index_mut(&mut self, index: RegionIdx) -> &mut Self::Output {
+        &mut self.0[index.0 as usize]
+    }
+}
```
breathx, line 894: I think small sugar doesnt bring more beauty rather than adds up option for unhandled implicit panic in threads.rs. More remove in favor of just `.get(..)?`

playX18: discussed in DM, leaving Index implementation in

breathx (approved):

grishasobol (approved):

PR #4407, opened 2024-12-18
feat(wasm-builder): switch to new `wasm32v1-none` target

### LINKED ISSUE #4235: wasm-builder uses hacks to build WASMs
### Problem

Starting with Rust 1.82 WASMs contain [`multi-value`](https://github.com/WebAssembly/spec/blob/main/proposals/multi-value/Overview.md) and [`reference-types`](https://github.com/WebAssembly/spec/blob/main/proposals/reference-types/Overview.md) features: https://blog.rust-lang.org/2024/09/24/webassembly-targets-change-in-default-target-features.html. However, Rust does not currently even provide support for `multi-value` and `reference-types` in language.

We currently use hack #4247 to solve this issue. This hack results in end user having to:
- install `rust-src` component for building smart contracts
- `-Zbuild-std` increases build time (while wasm-builder itself is slow)
- removing `-Clinker-plugin-lto` increases binary sizes (it was removed because I found some bug in Rust/LLVM: rust-lang/rust#130604)), compiler also cannot delete panic locations
- `RUSTC_BOOTSTRAP=1` reduces stability since the user may be using nightly features in WASMs

Other optimization/parsing/executions issues:
- `wasmer` signlepass does not support `multi-value` (https://docs.wasmer.io/runtime/features#support-of-features-by-backend)
- `wasm-opt` does not support `multi-value` (WebAssembly/binaryen#6407)
- `parity-wasm` fork does not support `reference-types`
- `parity-wasm-instrument` fork does not support `reference-types` (stack height instrumentation and gas charge for new instructions)

### Steps

_No response_

### Possible Solution

- [x] Update `wasmi` and `wasmer` executor
- [ ] Support `multi-value` and `reference-types` parsing in our `parity-wasm` fork
  - [x] https://github.com/gear-tech/parity-wasm/pull/4
  - [ ] TODO: `reference-types`
- Support `multi-value` and `reference-types` in `parity-wasm-instrument`
  - [x] https://github.com/gear-tech/wasm-instrument/pull/8
  - [ ] TODO: `reference-types`
- [ ] Benchmarks of new instructions, possibly charging gas for tables, etc.

### Notes

_No response_

### Relevant Log Output

_No response_

StackOverflowExcept1on:
Resolves #4235 by switching to new `wasm32v1-none` target (for runtime and smart-contracts).
- requires Rust 1.84 (will be stable on January 9, 2025).
- `-C target-feature=+sign-ext` also works for `core` and `alloc` without `-Zbuild-std` and `rust-src` component.
- `-C linker-plugin-lto` still doesn't work and we can't remove panic locations.

Output WASM files are now in different location `./target/<target>/<profile>/demo_vec.wasm`:
```diff
- <target> = wasm32-unknown-unknown
+ <target> = gear-wasm
```


.github/workflows/PR.yml
```diff
@@ -2,7 +2,7 @@ name: PR
 
 on:
   pull_request:
-    branches: [master]
+    #branches: [master]
```
StackOverflowExcept1on, line 5: should be removed

utils/cargo-gbuild/src/cli.rs
```diff
@@ -112,10 +112,12 @@ impl GBuild {
             kargo.set_profile(profile);
         }
 
+        // TODO: not sure how to set target dir here
```
StackOverflowExcept1on, line 115: should be removed

Cargo.toml
```diff
@@ -314,133 +314,132 @@ ethexe-common = { path = "ethexe/common" }
...
+sp-wasm-interface-common = { version = "7.0.0", git = "https://github.com/gear-tech/polkadot-sdk.git", branch = "gear-polkadot-stable2409-wasm32v1-none", default-features = false }
+substrate-build-script-utils = { version = "11.0.0", git = "https://github.com/gear-tech/polkadot-sdk.git", branch = "gear-polkadot-stable2409-wasm32v1-none" }
+substrate-frame-rpc-system = { version = "39.0.0", git = "https://github.com/gear-tech/polkadot-sdk.git", branch = "gear-polkadot-stable2409-wasm32v1-none" }
+substrate-rpc-client = { version = "0.44.0", git = "https://github.com/gear-tech/polkadot-sdk.git", branch = "gear-polkadot-stable2409-wasm32v1-none" }
+substrate-state-trie-migration-rpc = { version = "38.0.0", git = "https://github.com/gear-tech/polkadot-sdk.git", branch = "gear-polkadot-stable2409-wasm32v1-none" }
+substrate-test-client = { version = "2.0.0", git = "https://github.com/gear-tech/polkadot-sdk.git", branch = "gear-polkadot-stable2409-wasm32v1-none" }
+substrate-wasm-builder = { version = "24.0.1", git = "https://github.com/gear-tech/polkadot-sdk.git", branch = "gear-polkadot-stable2409-wasm32v1-none" }
```
StackOverflowExcept1on, line 441: now `substrate-wasm-builder` is used from the `gear-polkadot-stable2409-wasm32v1-none` branch: 
https://github.com/gear-tech/polkadot-sdk/commit/761ea7010f91f195c3d3b9d8b22c75f59ff23dce

utils/wasm-optimizer/src/cargo_toolchain.rs
```diff
@@ -94,16 +94,11 @@ impl Toolchain {
             self.raw_toolchain_str() == toolchain,
             anyhow!(
                 "recommended toolchain `{x}` not found, install it using the command:\n\
-        rustup toolchain install {x} --component llvm-tools --target wasm32-unknown-unknown\n\n\
+        rustup toolchain install {x} --component llvm-tools --target wasm32v1-none\n\n\
```
StackOverflowExcept1on, line 97: there is no point in advising to install the llvm-tools component, which is only used for runtime fuzzer to ecosystem developer

.github/workflows/build.yml
```diff
@@ -165,7 +165,7 @@ jobs:
         if: ${{ inputs.production && matrix.profiles.name == 'release' }}
         run: |
           mkdir -p artifact
-          tar czvf artifact/examples.tar.gz target/wasm32-unknown-unknown/release/*.wasm
+          tar czvf artifact/examples.tar.gz target/release/wasms/*.wasm
```
ark0f, line 168: It would be nice to have path like this to be consistent how `cargo` creates folders for specified target (e.g. `gear-wasm` can be imagined as virtual target)

```suggestion
          tar czvf artifact/examples.tar.gz target/gear-wasm/release/*.wasm
```

ark0f (changes requested):

Cargo.toml
ark0f: Could you explain the point to patch Substrate to use new target? I think it's up to Substrate what it uses

StackOverflowExcept1on: To remove the `rust-src` component and `wasm32-unknown-unknown` target from `rust-toolchain.toml` (less binaries need to be downloaded). Also see paritytech/polkadot-sdk#7008 where I have already received 2 approves for this change.

.github/workflows/build.yml
StackOverflowExcept1on: Good idea, but I don't like naming. Maybe `wasm32-gear` would be more suitable?

utils/wasm-builder/src/wasm_project.rs
```diff
@@ -127,7 +127,7 @@ impl WasmProject {
             .expect("Could not find target directory");
 
         let mut wasm_target_dir = target_dir.clone();
-        wasm_target_dir.push("wasm32-unknown-unknown");
+        wasm_target_dir.push("gear-wasm");
```
breathx, line 130:
```suggestion
        wasm_target_dir.push("wasm32-gear");
```

breathx (changes requested):

utils/wasm-builder/src/wasm_project.rs
breathx: how did you pass CI? 🤔

StackOverflowExcept1on: ok, let's run wasm builder smoke tests on CI: https://github.com/gear-tech/gear/actions/runs/12823207597

breathx (approved):

ark0f (approved):
Don't forget to merge updates from `gear-polkadot-stable2409` or merge your changes into that branch

PR #4409, opened 2024-12-19
refactor(backend): refine approach of funcs forbidding

### LINKED ISSUE #4399: refactor(backend): refine approach of funds forbidding
Currently blacklisting funds is implemented on env definition process: https://github.com/gear-tech/gear/blob/20ac111a84252b88bdcff15e38d18c35fbcd2d17/core-backend/src/env.rs#L136-L141

This should be replaced with checking by each sys call on backend isn't it forbidden (by calling some context funcs like self.ext.forbidden_funcs.contains(my_name)). If forbidden, return trap.

It's required to implement dynamic context changing for Environment as wasm module processor.

Lazark0x:
Resolves #4399 .

Moved syscalls forbiddenes check to backend side


core/src/env.rs
```diff
@@ -391,4 +391,7 @@ pub trait Externalities {
 
     /// Return the set of functions that are forbidden to be called.
     fn forbidden_funcs(&self) -> &BTreeSet<SyscallName>;
+
+    /// Return the set of functions that are forbidden to be called in this endpoint.
```
breathx, line 395: I'd remove this. On backend side for each execution better to accept some "forbidden_funcs". Endpoint one's to be taken on backend from kind itself

breathx (changes requested):
Except the comment, looks good 👍🏻

core-backend/src/funcs.rs
```diff
@@ -234,6 +235,7 @@ where
         self,
         caller: &mut CallerWrap<Caller>,
         (): Self::Context,
+        _syscall_name: SyscallName,
```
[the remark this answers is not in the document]
Lazark0x: Yes, added processing to RawSyscall as well

core-backend/src/runtime.rs
```diff
@@ -184,4 +164,19 @@ where
             },
         }
     }
+
+    pub fn check_func_forbiddenness(&mut self, syscall_name: SyscallName) -> Result<(), HostError> {
+        let endpoint_forbidden_funcs = self.ext_mut().endpoint_dispatch_kind().forbidden_funcs();
```
breathx, line 169: maybe inverse this check into ?
entry.forbids(call) -> bool

core-processor/src/executor.rs
```diff
@@ -130,13 +128,13 @@ where
         program_id: program.id,
         program_candidates_data: Default::default(),
         forbidden_funcs: settings.forbidden_funcs,
+        endpoint_dispatch_kind: kind,
```
breathx, line 131: better to do as a part of message ctx

core-backend/src/runtime.rs
Lazark0x: Yeah, that seems like a more direct solution

core-processor/src/executor.rs
Lazark0x: I'd agree

core-backend/src/runtime.rs
```diff
@@ -160,4 +164,20 @@ where
             },
         }
     }
+
+    pub fn check_func_forbiddenness(&mut self, syscall_name: SyscallName) -> Result<(), HostError> {
+        if self.ext_mut().forbidden_funcs().contains(&syscall_name)
+            || self
+                .ext_mut()
+                .endpoint_dispatch_kind()
```
breathx, line 172: doesn't it make more sense from where we take this value?
```suggestion
                .ext_mut()
                .msg_ctx()
                .dispatch_kind()
                .forbids(x)
```

breathx: it's not like it belongs to endpoint

breathx (approved):

core-backend/src/runtime.rs
Lazark0x: Agree

breathx (approved):

PR #4410, opened 2024-12-19
feat(ethexe): Support slash commitments for validators

ukint-vs:
Resolves # .

- update abi
- add slash commitments to router
- 
@reviewer-or-team


breathx:
To be reimplemented soon

PR #4411, opened 2024-12-19
feat(ethexe/contracts): improve interaction with MirrorProxy via etherscan

StackOverflowExcept1on:
"Read/Write Contract as Proxy" tab on Etherscan will now be more useful


breathx:
Only not sure if we need events here ?

ethexe/contracts/src/MirrorProxy.sol
```diff
@@ -12,12 +12,55 @@ import {IRouter} from "./IRouter.sol";
...
+    bytes32 public stateHash;
+    uint256 public nonce;
+
     constructor(address _router) {
         router = _router;
     }
 
+    /* Primary Gear logic */
+
+    function sendMessage(bytes calldata payload, uint128 value) external returns (bytes32) {
```
StackOverflowExcept1on, line 38: not sure about `returns` here
```suggestion
    function sendMessage(bytes calldata payload, uint128 value) external /*returns (bytes32)*/ {
```

PR #4419, opened 2024-12-23
refactor(ethexe): remove to_vec and mem::transmute from ethexe

playX18:
@breathx


ethexe/processor/src/host/api/mod.rs
```diff
@@ -58,7 +59,7 @@ impl MemoryWrap {
         store: impl Into<StoreContext<'a, T>>,
         ptr_len: i64,
     ) -> &'a [u8] {
-        let [ptr, len]: [i32; 2] = unsafe { mem::transmute(ptr_len) };
+        let (ptr, len) = unpack_i64(ptr_len);
```
Byt3nack3d, line 62: Not sure if this is intended, but it seems `unpack_i64` and `mem::transmute::<i64, [i32; 2]>` do not produce the same byte transformation on little-endian architectures.

ByteNacked, line 62: Not sure if this is intended, but it seems `unpack_i64` and `mem::transmute::<i64, [i32; 2]>` do not produce the same byte transformation on little-endian architectures.

ByteNacked: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=81b61a8ba2e5803a88b68ac6188dbcba

grishasobol: First - better to operate with unsigned and only in the end before returning convert them to sign type.
Second - transmute return depends on architecture: on little-endian -> [low, high] , big-endian -> [high, low]

breathx (commented):
impl the same packers/unpackers for i64 across (u32, u32) or even (usize, usize) with according checks. they could be placed in exe/runtime/common, since both exe/runtime and exe/processor(/host) use it

sandbox/host/src/sandbox/wasmer_backend.rs
```diff
@@ -345,8 +345,8 @@ fn dispatch_common(
     let (serialized_result_val_ptr, serialized_result_val_len) = {
         // Cast to u64 to use zero-extension.
         let v = serialized_result as u64;
-        let ptr = (v >> 32) as u32;
-        let len = (v & 0xFFFFFFFF) as u32;
+        let len = (v >> 32) as u32;
```
breathx, line 348: we were working here like with big endian previously? 🤔

sandbox/sandbox/src/host_executor.rs
```diff
@@ -306,8 +306,10 @@ extern "C" fn dispatch_thunk<T>(
         let result_ptr = result.as_ptr() as u64;
         let result_len = result.len() as u64;
         mem::forget(result);
-
-        (result_ptr << 32) | result_len
```
breathx, line 310: previous impl was a bit more obvious

sandbox/host/src/sandbox/wasmer_backend.rs
breathx: why there wasn't any bugs related to it?

breathx (approved):

sandbox/host/src/sandbox/wasmer_backend.rs
playX18: We were using `std::mem::transmute` and it's ordering afaik is not guaranteed. According to @ByteNacked comment it seems that transmute did produce result in reverse order which was hard-coded everywhere.

breathx: > We were using `std::mem::transmute` and it's ordering afaik is not guaranteed. According to @ByteNacked comment it seems that transmute did produce result in reverse order which was hard-coded everywhere.

its guaranteed on the same arch

grishasobol: > its guaranteed on the same arch

yes but we need but better to guarantee that it will have the same behaviour for each arch

ByteNacked: > According to @ByteNacked comment it seems that transmute did produce result in reverse order which was hard-coded everywhere.

Byte shifting and `std::mem::transmute` both depend on the endianness of the architecture. The best practice is to use methods like `::from_le_bytes` and `::to_le_bytes` (or be) to handle byte-order conversions explicitly.

However, in our case, this isn't a significant concern since it is unlikely we will ever work with big-endian architectures

runtime-interface/src/lib.rs
```diff
@@ -234,20 +236,14 @@ pub mod lazy_pages_detail {
         let mut writes_intervals = Vec::with_capacity(writes_len / mem_interval_size);
         deserialize_mem_intervals(writes, &mut writes_intervals);
 
-        let mut gas_counter = LittleEndian::read_u64(gas_bytes);
-
-        let res = match gear_lazy_pages::pre_process_memory_accesses(
+        match gear_lazy_pages::pre_process_memory_accesses(
```
breathx, line 239: could be simplified

breathx (approved):

ark0f (approved):

ByteNacked (approved):
:+1:

PR #4420, opened 2024-12-23
fix(sandbox): clear sandbox store inside a block

ByteNacked:
Resolves # .

Relates #4192

This hack fixes the error:

```
 tokio-runtime-worker gear::authorship: ⌛️ Pseudo-inherent is taking too long and will not be included in the block.    
````

HACK: It was discovered that starting with version 4.0, Wasmer experiences a slowdown
 when creating a large number of memory/instances beyond a certain threshold.
The usual method to clear the store doesn't work for benchmarks (see `Sandboxes::get`)
or when too many instances/memories are created **within a single block**, as the store
is only cleared at the start of a new block.

This is a temporary solution to reset the store after reaching a certain limit
(see `SANDBOX_STORE_CLEAR_COUNTER_LIMIT`) for memory/instances.
Otherwise, the store grows too large, leading to performance degradation during
normal node execution and benchmarks.

@reviewer-or-team


ByteNacked:
Synced:
 - [x] testnet
 - [ ] mainnet

grishasobol (approved):

gshep (approved):

breathx (approved):

PR #4421, opened 2024-12-23
refactor(ethexe/cli): lots of changes; improved redesign

breathx:


ethexe/service/src/tests.rs
```diff
@@ -1090,6 +1151,7 @@ mod utils {
         }
     }
 
+    // TODO (breathx): remove me in favor of crate::config::NodeConfig.
```
grishasobol, line 1154: I do not think this should be done. This config is special for testing purposes. It must be possible to simply expanded by developer for testing.

techraed (approved):

Makefile
```diff
@@ -6,7 +6,7 @@ ethexe-pre-commit: ethexe-contracts-pre-commit ethexe-pre-commit-no-contracts
 ethexe-pre-commit-no-contracts:
 	@ echo " > Formatting ethexe" && cargo +nightly fmt --all -- --config imports_granularity=Crate,edition=2021
 	@ echo " >> Clippy checking ethexe" && cargo clippy -p "ethexe-*" --all-targets --all-features -- --no-deps -D warnings
-	@ echo " >>> Testing ethexe" && cargo nextest run -p "ethexe-*"
+	@ echo " >>> Testing ethexe" && cargo nextest run -p "ethexe-*" --no-fail-fast
```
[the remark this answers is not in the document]
breathx: meaningless to fail it for local dev. personally I prefer to check all of the errors occurred to be able to identify its source: test or global logic error

ethexe/service/src/tests.rs
breathx: it will keep the same interface for tests, but will remove duplicated structures

ethexe/cli/src/commands/key.rs
```diff
@@ -0,0 +1,180 @@
...
+            }
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Subcommand)]
+pub enum KeySubcommand {
+    Clear,
```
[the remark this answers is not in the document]
breathx: its already there, update gh page to watch all changes

ethexe/service/src/tests.rs
grishasobol: The tests can be changed in future and we may want to add some additional fields which is needed only for tests. So, still - it's bad idea to use the same struct in tests and code

PR #4426, opened 2024-12-25
feat(ethexe): support late commitments validation

grishasobol:
Append param validationDelay which allows for commitments from previous era to commit blocks from previous era even if new era is started.


ethexe/contracts/script/Deployment.s.sol
```diff
@@ -48,6 +48,7 @@ contract DeploymentScript is Script {
                         address(wrappedVara),
                         1 days,
                         2 hours,
+                        1 minutes,
```
breathx, line 51: isn't it too small? even normal track of tx may exceed 1min

breathx (approved):

ethexe/contracts/script/Deployment.s.sol
grishasobol: Ok let's make it 5 minutes. Anyway after this period of time new validators will be able to sign their commitment if old validators do not committed yet.

PR #4366, opened 2024-11-28
feat(ethexe-tx-pool): Introduce basic tx-pool

techraed:
This PR introduces a basic logic for the ethexe tx pool which works as following:
- Tx pool itself has a service part and a core part. The service part handles communications from outside to the core and from inside (core) to the outer "world". The core part basically performs main tx pool ops (adding a tx for now).
- The `InputTask` is a task that is requested to be executed by the tx pool. The task is sent via `TxPoolInputTaskSender` sending end of the channel between the sender and the tx pool service. The sendind end is supplied to the main service and the rpc. 
- Basic flow:
    1. Rpc receives a request for sending a tx. It sends the tx via the channel to the tx pool service.
    2. Tx pool service receives the `InputTask` and calls corresponding tx pool core method to handle the task.
    3. `InputTask::AddTransaction` task triggers tx pool core to 1) validate the transaction, 2) store it in db
    4. When the `InputTask::AddTransaction` is executed successfully the tx pool services sends `OutputTask::PropogateTransaction` to the main service. 
    5. The main service handles the `OutputTask::PropogateTransaction` by gossiping it to other nodes.
    6. When a node receives a transaction, it creates `InputTask::AddTransaction` task and sends it to the tx pool. If tx exists, it's not added to the storage.

Here you can also find a tiny refactoring of the `ethexe-signer` which is all about clarifying comments, simplified API and moving types to their own modules.

**UPD**
Changes in #4427 related to this PR
- `TxPoolTrait`, `TxPoolCore`
- `Transaction` trait is heavily changed
- `EthexeTransaction` is changed, as all transactions.

**UPD**
#4427 merged into this PR

PR #4427, opened 2024-12-26
feat(ethexe-tx-pool): introduce proper tx pool integration with rpc and service

techraed:
This PR is to be review/merged after https://github.com/gear-tech/gear/pull/4366.

It brings several updates:
- RPC now accepts proper data for sending message to the program
- Complex tx validation with different validators (mortality, signature, uniqueness, executable) is delivered. Validators are tested.
- Output task for requesting the execution
- Input task for requesting pre-execution validation
- Tests for the tx pool `AddTransaction` and `CheckPreExecutionTransactionValidity` tasks are updated. Tests there mostly focus on the correctness of business logic and implementation (channels not panicking and etc.) stuff. The gossip test in `ethexe-cli` is now a base for an integration layer testing for the tx pool service.
- `Transaction` trait is now an extension over several smaller traits defining behaviour for getting tx signature, hash, reference block hash and etc.
- DB now has recent block hashes window (30 blocks for now). That's a window of latest valid blocks that is used for checking transaction mortality. Whenever a valid block is set, the window is updated.
- `TxPool*` related traits and types were removed.

Some security considerations must be taken into account like the 1) window state while syncing, 2) not accepting txs when syncing, 3) proper (deterministic, same for all the nodes) order of execution of the transactions in the pool and etc. Everything is stated in todos marked with issue number in gh.


techraed:
Actual execution is scheduled to be done by @breathx

PR #4429, opened 2025-01-11
fix(crates-io): publish of core-processor with new dependency

breathx:


breathx:
IMPORTANT: revert c97768176182588d3cb6d5f7547cf975ffe6112a before merge

Issue #4438, opened 2025-01-13
Reduce the size/amount of data sections in `.opt.wasm`


ecol-master:
### Problem

There is an issue with the .opt.wasm files, where an excessive number of data segments are being generated. Specifically, in the example below (see ref) the WebAssembly optimizer creates over 11,000 data segments when building the program. This behavior is problematic and exceeds the current network limits with 1024 segments. 

Reference: https://github.com/gear-foundation/gear.exe-dapps/tree/lm-cnn-data-section/cnn-cats-dogs

### Steps

1.
2.
3.

### Relevant Log Output

<details><summary>Click to expand/collapse</summary>
<p>

```
<Paste your log here>
```

</p>
</details>

PR #4444, opened 2025-01-16
feat(wasm-builder): reduce the amount of data sections in opt-wasm

ecol-master:
Resolves issue #4438  Reduce the size/amount of data sections in .opt.wasm .

- bump `binaryen` to 123 version
- add example `big-data-section` which tests the correctness of reducing the amount of data sections in output `wasm`

@grishasobol


examples/big-data-section/tests/test_big_data_section.rs
```diff
@@ -0,0 +1,56 @@
...
+            &sys,
+            "../../target/wasm32-unknown-unknown/debug/demo_big_data_section.opt.wasm",
+        );
+        sys.mint_to(gtest::constants::DEFAULT_USER_ALICE, 100000000000);
+
+        // Skipping program initialization
+        let _ = prog.send_bytes(USER_ID, b"");
+        let _ = sys.run_next_block();
+
+        let random_payload: Vec<Vec<u8>> = vec![
```
breathx, line 24: could you pls use here proptests to make sure its not coincidence? also run a few times locally with huge amount of repetitions

ecol-master:
Now,  utils from `binaryen` package will be install directed in docker images.  
Current `binaryen` version - **123**.

utils/wasm-builder/Cargo.toml
```diff
@@ -26,7 +26,4 @@ gear-core.workspace = true
 gear-wasm-instrument.workspace = true
 gear-wasm-optimizer.workspace = true
 rustc_version.workspace = true
-itertools.workspace = true
-
-[features]
-wasm-opt = ["gear-wasm-optimizer/wasm-opt"]
+itertools.workspace = true
```
grishasobol, line 29: end of file must be \n

utils/wasm-proc/Cargo.toml
```diff
@@ -17,8 +17,4 @@ clap = { workspace = true, features = ["derive"] }
 log.workspace = true
 tracing-subscriber.workspace = true
 gear-wasm-builder.workspace = true
-gear-wasm-instrument = { workspace = true, features = ["std"] }
-
-[features]
-default = ["wasm-opt"]
-wasm-opt = ["gear-wasm-builder/wasm-opt"]
+gear-wasm-instrument = { workspace = true, features = ["std"] }
```
grishasobol, line 20: end of file must be \n

examples/big-data-section/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+gtest.workspace = true
+proptest = { workspace = true }
+
+[build-dependencies]
+gear-wasm-builder.workspace = true
+
+[features]
+debug = ["gstd/debug"]
+default = ["std"]
+std = []
```
grishasobol, line 28: ==//==

examples/big-data-section/proptest-regressions/lib.txt
```diff
@@ -0,0 +1,7 @@
+# Seeds for failure cases proptest has generated in the past. It is
+# automatically read and these particular cases re-run before any
+# novel cases are generated.
```
grishasobol, line 3: why do we need this file ?

grishasobol:
> Resolves issue #4438 Reduce the size/amount of data sections in .opt.wasm .
> 
> * added methods for `Optimizer` struct `utils/wasm-optimizer/optimize.rs`
> * added program example with large data section
> 
> @grishasobol

Fix the comment

.github/workflows/build.yml
```diff
@@ -37,7 +37,6 @@ env:
   RUST_BACKTRACE: short
   RUST_LIB_BACKTRACE: 1
   TERM: xterm-256color
-  BINARYEN_VERSION: version_111
```
grishasobol, line 40: IMPORTANT: What about all other places where binaryen version_111 is using? See in other yaml/yml files

grishasobol (changes requested):

utils/wasm-optimizer/src/optimize.rs
```diff
@@ -188,6 +182,8 @@ pub fn do_optimization<P: AsRef<Path>>(
         .arg("-mvp")
         .arg("--enable-sign-ext")
         .arg("--enable-mutable-globals")
+        .arg("--limit-segments")
+        .arg("--pass-arg=limit-segments@1024")
```
StackOverflowExcept1on, line 186: mb add this flags only if binaryen version >= 123? e.g. ubuntu packages can be very outdated.

ecol-master: yeah, it make sense, but I tested locally and found that `wasm-opt` ignores unknown flags. So we don't need to check the correctness the installed binaryen version

.github/workflows/build.yml
ecol-master: I remove the `binaryen` installation from places where we use linux image on runners. In windows I didn't do this

StackOverflowExcept1on (approved):

examples/big-data-section/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+#[cfg(test)]
+mod tests {
+    use crate::data_access::DataAccess;
+    use gtest::{Log, Program, System};
+
+    const USER_ID: u64 = gtest::constants::DEFAULT_USER_ALICE;
+
+    use proptest::{
+        arbitrary::any, collection::vec, proptest, test_runner::Config as ProptestConfig,
+    };
```
ark0f, line 44:
```suggestion
    use gtest::{Log, Program, System};
    use proptest::{
        arbitrary::any, collection::vec, proptest, test_runner::Config as ProptestConfig,
    };
    
    const USER_ID: u64 = gtest::constants::DEFAULT_USER_ALICE;
```

examples/big-data-section/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+    // Testing random access to data section
+    proptest! {
+        #![proptest_config(ProptestConfig {
+                    cases: 200, // Set the number of test cases to run
+                    .. ProptestConfig::default()
+                })]
```
ark0f, line 50:
```suggestion
        #![proptest_config(ProptestConfig {
            cases: 200, // Set the number of test cases to run
            .. ProptestConfig::default()
        })]
```

utils/wasm-optimizer/src/optimize.rs
ark0f, line 186: Consider to move `DATA_SEGMENTS_AMOUNT_LIMIT` into `gear-core` and use it between crates so constant is always the same

ark0f (commented):
👍

PR #4447, opened 2025-01-17
feat(ethexe-contracts): add support for FROST signatures

StackOverflowExcept1on:


ethexe/contracts/test/Base.t.sol
```diff
@@ -121,6 +121,10 @@ contract Base is POCBaseTest {
                             uint256(eraDuration),
                             uint256(electionDuration),
                             uint256(validationDelay),
+                            Gear.AggregatedPublicKey(
+                                0x0000000000000000000000000000000000000000000000000000000000000001,
+                                0x4218F20AE6C646B363DB68605822FB14264CA8D2587FDD6FBC750D587E76A7EE
+                            ),
```
StackOverflowExcept1on, line 127: this public key on the secp256k1 curve, which is not used now because we don't have the FROSTOffchain library to sign messages from Solidity

ethexe/contracts/.env.example
```diff
@@ -3,6 +3,11 @@ SEPOLIA_RPC_URL="https://ethereum-sepolia-rpc.publicnode.com"
 HOLESKY_RPC_URL="https://ethereum-holesky-rpc.publicnode.com"
 PRIVATE_KEY="0x0000000000000000000000000000000000000000000000000000000000000000"
 ETHERSCAN_API_KEY="AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"
+# example of valid ROUTER_AGGREGATED_PUBLIC_KEY
```
breathx, line 6: uncomment it?

ethexe/contracts/test/Base.t.sol
```diff
@@ -258,19 +260,16 @@ contract Base is POCBaseTest {
         );
     }
 
+    // TODO: add SignatureType as param here
```
breathx, line 263: should be added now for tests?

ethexe/contracts/test/POC.t.sol
```diff
@@ -98,9 +100,22 @@ contract POCTest is Base {
         depositInto(vaults[1], 10_000);
         depositInto(vaults[2], 10_000);
         rollBlocks((eraDuration - electionDuration) / blockDuration);
+
+        SigningKey _signingKey = FROSTOffchain.newSigningKey();
+        Vm.Wallet memory _publicKey = vm.createWallet(_signingKey.asScalar());
+
+        // TODO: makeElectionAt should also return Gear.AggregatedPublicKey
```
breathx, line 107: put now some placeholder

breathx (approved):

ethexe/contracts/.env.example
StackOverflowExcept1on: I don't want to make it default parameters because whoever is doing deploy should be responsible for public key that can control commits.

ethexe/contracts/test/Base.t.sol
StackOverflowExcept1on: We may be able to remove the ECDSA in the future. Right now we're testing this in Rust code anyway.

ethexe/contracts/test/POC.t.sol
StackOverflowExcept1on: `SigningKey _signingKey = FROSTOffchain.newSigningKey();` and there is this placeholder. Instead of doing DKG we just use single `SigningKey`.

PR #4455, opened 2025-01-20
refactor(ethexe): simplify program creation and message sending in tests

playX18:
- Updated test cases to remove unnecessary payload and value parameters in `create_program` calls.
- Replaced direct assertions on initialization message properties with assertions on reply messages.


ethexe/service/src/tests.rs
```diff
@@ -1202,18 +1267,10 @@ mod utils {
         }
 
         // TODO (breathx): split it into different functions WITHIN THE PR.
```
breathx, line 1269: rm todo

ethexe/service/src/tests.rs
```diff
@@ -1202,18 +1267,10 @@ mod utils {
...
-        pub async fn create_program(
-            &self,
-            code_id: CodeId,
-            payload: &[u8],
-            value: u128,
-        ) -> Result<WaitForProgramCreation> {
+        pub async fn create_program(&self, code_id: CodeId) -> Result<WaitForProgramCreation> {
```
breathx, line 1270: rm EXECUTABLE BALANCE constant in favor of func argument / another func that will create program and topup balance

breathx: so in case of absence of exec balance there's no need in approve

breathx (changes requested):

ethexe/service/src/tests.rs
```diff
@@ -155,22 +155,28 @@ async fn ping() {
...
+        .unwrap()
+        .wait_for()
+        .await
+        .unwrap();
+
     assert_eq!(
         res.reply_code,
         ReplyCode::Success(SuccessReplyReason::Manual)
     );
+    assert_eq!(res.reply_payload, b"PONG");
```
breathx, line 178: overall refactor would be good as well 
```suggestion
    assert_eq!(reply.payload, b"PONG"); // etc
```

ethexe/service/src/tests.rs
playX18: added it as a parameter to `create_program` as I think it's easier to pass approved balance for program at its creation without a need to call another function.

ethexe/service/src/tests.rs
```diff
@@ -1216,43 +1260,34 @@ mod utils {
             Ok(WaitForUploadCode { listener, code_id })
         }
 
-        // TODO (breathx): split it into different functions WITHIN THE PR.
         pub async fn create_program(
             &self,
             code_id: CodeId,
-            payload: &[u8],
-            value: u128,
+            balance_to_approve: u128,
```
breathx, line 1266:
```suggestion
            executable_balance: u128,
```

breathx (approved):

PR #4458, opened 2025-01-21
feat(ethexe): dev mode, set blob via rpc

osipov-mit:


ethexe/cli/src/params/ethereum.rs
```diff
@@ -41,6 +41,11 @@ pub struct EthereumParams {
     #[arg(long, alias = "eth-router")]
     #[serde(rename = "router")]
     pub ethereum_router: Option<String>,
+
+    /// Ethereum block time in seconds.
+    #[arg(long, alias = "eth-block-time")]
+    #[serde(rename = "block-time")]
+    pub block_time: Option<usize>,
```
breathx, line 48:
```suggestion
    pub ethereum_block_time: Option<u64>,
```

ethexe/rpc/src/apis/dev.rs
```diff
@@ -0,0 +1,57 @@
...
+use jsonrpsee::{
+    core::{async_trait, RpcResult},
+    proc_macros::rpc,
+};
+use sp_core::Bytes;
+
+#[rpc(server)]
+pub trait Dev {
+    #[method(name = "blob_set")]
+    async fn blob_set(&self, blob: Bytes) -> RpcResult<(H256, CodeId)>;
```
breathx, line 33: Otherwise function call looks like it's already happened 
```suggestion
    async fn set_blob(&self, blob: Bytes) -> RpcResult<(H256, CodeId)>;
```

ethexe/service/src/lib.rs
```diff
@@ -84,15 +84,25 @@ pub enum NetworkMessage {
...
+                    &config.ethereum.beacon_rpc,
+                    config.ethereum.block_time,
+                )
+                .await
+                .with_context(|| "failed to create blob reader")?,
             )
-            .await
-            .with_context(|| "failed to create blob reader")?,
-        );
+        };
```
breathx, line 105: could be flattened into single if block 
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=62a69e942fc810a88c47654c68df0503

breathx (approved):

PR #4463, opened 2025-01-22
feat(ethexe): Add more context to returned errors

techraed:
Adds `anyhow::Context::context` calls to errors returned from the most important crates and fns (`ethexe-cli` and 2-3 nested levels)


breathx:
To be reimplemented later

PR #4442, opened 2025-01-15
feat(core): Plonky2 proofs onchain verification (part 1)

ekovalev:
## Overview
This is the first part of a bigger change (#4393 ) in order to break down the original PR into several more or less independent and tractable parts. 
The ultimate goal is to provide a framework for zk-related apps on top of (currently) Vara blockchain. Specifically, it supports on-chain verification of Plonky2 proofs for arbitrary circuits. This would require speeding up of computationally greedy tasks in order to ensure any `verify` operation fits in a single block. 
The big picture looks like so:
```
                            +-------------------------------------------------------------------------+
                            |                                                          WASM           |
+------------+              |   +---------------------------+                ( Gear programs realm )  |
|            |     proof    |   |   Application Contract    |                                         |
|   Prover   | -------------|-> |                           |                                         |
|            |              |   |  (verifier_circuit_data)  |                                         |
+------------+              |   +---------------------------+                                         |
                            |                 |                                                       |
                            |                 | concatenate( verifier_circuit_data, proof )           |
                            |                 v                                                       |
                            |   +---------------------------+                                         |
                            |   |     Verifier Contract     |                                         |
                            |   |                           |         +---------------------------+   |
                            |   |     plonky2::verify()     |         |   GoldilocksField  impl   |   |
                            |   |       |                   |         |                           |   |
                            |   |       +- GF::poseidon() --+-------> |    Poseidon::permute()    |   |
                            |   +---------------------------+         +-------------|-------------+   |
                            +-------------------------------------------------------|-----------------+
                                                                                    | syscall
                                                                                    v
                                                                      +---------------------------+
                                                                      |        Vara runtime       |
                                                                      |                           |
                                                                      |   gr_poseidon_permute()   |
                                                                      +-------------|-------------+
                                                                                    | host call
                                                                                    v
                                                                      +---------------------------+
                                                                      |           Client          |
                                                                      |                           |
                                                                      |  permute primitives impl  |
                                                                      +---------------------------+
```
Most of the workflow happens in the user-space. We define a set of contracts with (potentially) well-known addresses:
- an `Application contract` contains the main app logic, and, therefore, stores the circuit-specific data; every proof sent to it for verification is prepended with the encoded circuit data;
- the `Verifier contract` (can be a single instance for the entire chain) wraps the original Plonky2 code for proofs verification; however, the underlying primitives (`GoldilocksField`) are custom-implemented in such a way that heavy operations are offloaded to the host (through a chain of a syscall->host call).

[...]

PR #4465, opened 2025-01-23
feat(node): Plonky2 proofs onchain verification (part 2)

ekovalev:
### Overview
Introduces client side support for onchain Plonky2 proofs verification (#4442).

### Changes
Changes are mostly ported from the original Plonky2 repository with maybe some slight modifications to ensure we only bring over necessary stuff. Specifically:
- GoldilocksField arithmetic to support computations on the host;
- Poseidon hash operations.

#### Note
This PR intentionally omits some architecture-specific code from the original Plonky2 repository (again, for the sake of a smaller PR size). Whatever is lacking here will be added in the next PR in the series.


breathx:
To be reimplemented differently as agreed with @ekovalev

PR #4466, opened 2025-01-24
refactor(core): Implement LimitedIntervalsTree and MaxEncodedLen for JournalNote

playX18:
# Changes

- Added new type to core: `LimitedIntervalsTree` which is based on `IntervalsTree` but checks limits before inserting intervals. It implements `MaxEncodedLen`
- Added type-alias `WasmPagesIntervalsTree` which is `LimitedIntervalsTree<WasmPage, PayloadSizeError, u16::MAX/2+1>`.
- Changed most of the code to use `WasmPagesIntervalsTree` instead of unlimited `IntervalsTree`
- Implemented MaxEncodedLen for `JournalNote`
- Test that `JournalNote`, when encoded is less than 32MiB in size
- Changed LimitedVec formatting implementation to include two new traits for limited formatting instead of requiring `T` to be `where [T]: AsRef<[u8]>`: this allows us to debug/display LimitedVec with any `T` that implements corresponding trait. 

# Notes

- Program storage was not updated, migration to migrate from `IntervalsTree` can be a separate PR. Only public APIs were updated to convert IntervalsTree to LimitedIntervalTree. 

@breathx


breathx:
conflicts to resolve

breathx (changes requested):
Discussed:
- Newtype for tree is to be removed
- Considering removal of pallet configuration of outgoing messages in favor of crate level (protocol) constant

playX18:
> I think this PR better to wait for constants separation (you're on it, right? @playX18)

Yes, I am on it right now

breathx:
To be reopened at a more suitable time: after journal notes refactor

PR #4475, opened 2025-01-27
feat(ethexe): compute service initial impl with non-blocking codes and blocks processing

grishasobol:


Cargo.toml
```diff
@@ -311,6 +311,7 @@ ethexe-prometheus = { path = "ethexe/prometheus", default-features = false }
 ethexe-validator = { path = "ethexe/validator", default-features = false }
 ethexe-rpc = { path = "ethexe/rpc", default-features = false }
 ethexe-common = { path = "ethexe/common", default-features = false }
+ethexe-compute = { path = "ethexe/compute" }
```
breathx, line 314: default features false (despite the fact that it may not have any atm)

ethexe/common/src/lib.rs
```diff
@@ -34,3 +34,26 @@ pub const fn u64_into_uint48_be_bytes_lossy(val: u64) -> [u8; 6] {
 
     [b1, b2, b3, b4, b5, b6]
 }
+
+// TODO: move to submodule.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct BlockData {
+    pub hash: gprimitives::H256,
+    pub header: db::BlockHeader,
+    pub events: alloc::vec::Vec<events::BlockEvent>,
```
breathx, line 43:
```suggestion
    pub header: BlockHeader,
    pub events: Vec<BlockEvent>,
```

ethexe/compute/src/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+// TODO (gsobol): add state monitoring in prometheus
+// TODO (gsobol): append off-chain transactions handling
+// TODO (gsobol) asap: add tests for compute service
+pub struct ComputeService {
+    db: Database,
+    processor: Processor,
+    query: Query,
+    blocks_queue: VecDeque<BlockData>,
+    process_block: Option<BoxFuture<'static, Result<BlockProcessed>>>,
+    process_codes: JoinSet<Result<CodeCommitment>>,
```
breathx, line 62: pls use FuturesUnordered

ethexe/compute/src/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+                let context = ChainHeadProcessContext {
+                    db: self.db.clone(),
+                    processor: self.processor.clone(),
+                    query: self.query.clone(),
+                };
+
+                self.process_block = Some(Box::pin(context.process(block)));
+            } else {
+                self.process_block = None;
+            }
```
breathx, line 80:
```suggestion
            self.process_block = self.blocks_queue.pop_front().map(|block| {
                let context = ChainHeadProcessContext {
                        db: self.db.clone(),
                        processor: self.processor.clone(),
                        query: self.query.clone(),
                };

                Box::pin(context.process(block))
            });
```

ethexe/compute/src/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+}
+
+impl FusedStream for ComputeService {
+    fn is_terminated(&self) -> bool {
+        false
+    }
+}
+
+impl ComputeService {
+    pub fn new(db: Database, processor: Processor, query: Query) -> Self {
```
breathx, line 104: I'd say that this service should be responsible for runtime instantiation, so only it creates processor (not any external essense)

breathx (approved):

breathx (changes requested):

ethexe/compute/src/lib.rs
grishasobol: Where?

ethexe/compute/src/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+
+            return Poll::Ready(Some(res.map(ComputeEvent::BlockProcessed)));
+        }
+
+        if let Poll::Ready(Some(res)) = self.process_codes.poll_join_next(cx) {
+            let res = match res {
+                Ok(res) => res.map(ComputeEvent::CodeProcessed),
+                Err(err) => Err(err.into()),
+            };
+            return Poll::Ready(Some(res));
```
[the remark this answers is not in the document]
grishasobol: no

ethexe/compute/src/lib.rs
```diff
@@ -0,0 +1,299 @@
...
+    query: Query,
+    blocks_queue: VecDeque<BlockData>,
+    process_block: Option<BoxFuture<'static, Result<BlockProcessed>>>,
+    process_codes: JoinSet<Result<CodeCommitment>>,
+}
+
+impl Stream for ComputeService {
+    type Item = Result<ComputeEvent>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
```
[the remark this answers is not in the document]
grishasobol: > otherwise poll will not be called
Yes and this is correct behaviour

ethexe/compute/src/lib.rs
grishasobol: I would better to add TODO in order to think about. I don't think compute should be used in rpc calls and processor should be used there directly.

ethexe/compute/src/lib.rs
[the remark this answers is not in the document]
grishasobol: flatten can be used here but unfortunately it is unstable...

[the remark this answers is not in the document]
breathx: oh, that's res of res. In that case - ".and_then()" https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=2dfaa8599bb9e1f68e6dbd2e90f2bd2a

ethexe/compute/src/lib.rs
breathx: Nice idea about todo. But for production actions processor shouldn't be used in rpc. For case of read states etc there's no problem

Cargo.toml
grishasobol: done

ethexe/compute/src/lib.rs
grishasobol: done

ethexe/compute/src/lib.rs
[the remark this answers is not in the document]
grishasobol: done

ethexe/compute/src/lib.rs
grishasobol: So, I think changing process_codes: JoinSet<Result<CodeCommitment>> to FuturesUnordered is useless here.

ethexe/compute/src/lib.rs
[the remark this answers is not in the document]
grishasobol: So, what about this?

breathx (approved):

PR #4484, opened 2025-01-30
feat(ethexe): add more accurate gas estimation for commitments

StackOverflowExcept1on:
Previously we could spend sequencer funds to send failed transaction. Now gas will be evaluated for everything except `Gear.blockIsPredecessor()`.


ethexe/ethereum/src/router/mod.rs
```diff
@@ -198,8 +203,29 @@ impl Router {
                 .map(|signature| Bytes::copy_from_slice(signature.as_ref()))
                 .collect(),
         );
+
+        let mut state_diff = HashMap::default();
```
StackOverflowExcept1on, line 207: code duplication because the `let builder` type is quite complex and writing generic function would require lot of generics

breathx (approved):

PR #4495, opened 2025-02-07
chore(scripts/CI): fuzzer expand

breathx:


.github/workflows/check.yml
```diff
@@ -142,7 +142,7 @@ jobs:
         run: ./scripts/gear.sh test fuzzer-tests
 
       - name: "Check fuzzer competence with mutation test"
-        run: ./scripts/check-fuzzer.sh
+        run: ./scripts/check-fuzzer.sh || cat fuzz_run
```
ark0f, line 145: Job fails only if `cat` fails

ark0f (changes requested):

ark0f (approved):

PR #4503, opened 2025-02-10
feat(ethexe): Introduce centralized event broadcasting system for services

### LINKED ISSUE #4483: refactor(ethexe): Add events from all services and send them to desired output channel
### File Location(s)

_No response_

### Proposal

At the moment testing code uses timeouts. If we have a way to receive events from all services we can simply wait for event to be received instead of waiting. 

For example instead of `sleep(Duration::from_secs(1))` in `start_service` ([ethexe/service/src/tests.rs#L1648](https://github.com/gear-tech/gear/blob/aefe5e3258b638abdf6957410f387ec6853e6cf2/ethexe/service/src/tests.rs#L1648)) we can simply wait for `Event::Started` to be produced. 

# How to achieve

- Collect all possible events from all services into single enum
- Introduce an output channel for producing events into it.

playX18:
Resolves #4483 

Status: W.I.P, this code works but needs refactoring to make public API better to use. In particular I am unsure how to get rid of `from_broadcaster` function in tests.


breathx (approved):
all review fixes are done in https://github.com/gear-tech/gear/pull/4517

PR #4509, opened 2025-02-12
refactor(core): store bool flag instead of full init allocations

### LINKED ISSUE #4497: refactor(core): remove `init_allocations` from `AllocationsContext`
### File Location(s)

core/src/memory.rs

### Proposal

Storing init allocations is almost useless now and uses only to identify whether allocations was changed during execution or not. So, there can be two more effective solutions:
1) `init_allocations` ->  `allocations_changed: bool` which is set by default to false. if `alloc` or `free` or `free_range` appears during execution the set it to true.
2) `init_allocations` -> `new_allocations: IntervalsTree` which is updated by `alloc` and `free` and `free_range`. In the end of execution it can be used to indent whether allocations has been changed. This more precise solution, because it can detect cases where for example one pages has been allocated and then freed in the same execution.

ecol-master:
Resolves #4497 

- store `allocations_changed` in `AllocationsContext` instead of `init_allocations`

@grishasobol


core/src/memory.rs
```diff
@@ -478,6 +480,7 @@ impl AllocationsContext {
         if let Some(heap) = self.heap {
             if interval.start() >= heap.start() && interval.end() <= heap.end() {
                 self.allocations.remove(interval);
```
breathx, line 482: off top, but: why do we ignore result of removal here, but sensitive to this in common free? cc @grishasobol

breathx (approved):

core/src/memory.rs
ecol-master: yep, u are absolutely right, I missed that line

breathx: It seems that it was the same before your changes

ecol-master: yes, it was added in previous pr (https://github.com/gear-tech/gear/commit/f860337d6e7852512fd2e38f605dee346f8e2d5e)

core-processor/src/ext.rs
```diff
@@ -617,7 +617,7 @@ impl<LP: LazyPagesInterface> ProcessorExternalities for Ext<LP> {
             gas_amount: gas_counter.to_amount(),
             gas_reserver,
             system_reservation_context,
-            allocations: (allocations != initial_allocations).then_some(allocations),
+            allocations: allocations_changed.then_some(allocations),
```
grishasobol, line 620: Please append a comment here that allocations_changed can be some times true event if final state of allocations is the same as before execution. `allocations_changed` allocations has been changed at least once per execution.

grishasobol (approved):

core/src/memory.rs
breathx: If allocations wasn't modified (depending on returning result) than what's the point to set changed = true?

breathx (changes requested):

breathx (approved):

PR #4512, opened 2025-02-14
feat(ethexe): observer on-chain data sync 

grishasobol:
Now:
- fully download the state from the ethereum via the observer.
- added a new BlockSynced  observer event, which signals that all data for the block has been loaded and the block is ready for further processing.
- Compute service now do not looks on commitments and process uncomputed blocks only.
- Validator is engaged in the aggregation of block commitments for publication.


ethexe/observer/src/lib.rs
```diff
@@ -196,88 +261,495 @@ impl ObserverService {
...
+    ) -> Result<Vec<BlockData>> {
+        log::trace!("Querying blocks from {from_block} to {to_block}");
+
+        let mut batch = BatchRequest::new(provider.client());
+
+        let headers_request: FuturesUnordered<_> = (from_block..=to_block)
+            .map(|bn| {
+                batch
+                    .add_call::<_, Option<<Ethereum as Network>::BlockResponse>>(
+                        "eth_getBlockByNumber",
```
StackOverflowExcept1on, line 456: https://docs.rs/alloy-provider/0.11.1/src/alloy_provider/provider/trait.rs.html#392

ethexe/observer/src/lib.rs
```diff
@@ -75,20 +74,28 @@ pub enum ObserverEvent {
...
 pub struct ObserverService {
-    blobs: Arc<dyn BlobReader>,
     provider: Provider,
+    database: Box<dyn OnChainStorage>,
+    // TODO (gsobol): consider to make clone boxed for BlobRead, in order to avoid Arc usage.
+    blobs_reader: Arc<dyn BlobReader>,
     subscription: Subscription<Header>,
 
-    router: Address,
+    router_address: Address,
```
breathx, line 89: mb flatten them into some RuntimeConfig?

ethexe/observer/src/lib.rs
```diff
@@ -97,32 +104,59 @@ impl Stream for ObserverService {
...
+            };
+
+            log::trace!("Received a new block: {data:?}");
+
+            self.block_sync_queue.push_back(header);
+
+            return Poll::Ready(Some(Ok(ObserverEvent::Block(data))));
+        }
+
+        if self.sync_future.is_none() {
```
breathx, line 132: shouldn't sync be prioritized?

ethexe/observer/src/lib.rs
```diff
@@ -147,22 +181,37 @@ impl FusedStream for ObserverService {
...
-                .context("failed to create blob reader")?,
-        );
-
-        Self::new_with_blobs(config, blobs).await
-    }
+    pub async fn new<DB>(
+        eth_cfg: &EthereumConfig,
+        max_sync_depth: u32,
+        db: &DB,
+        blobs_reader: Option<Arc<dyn BlobReader>>,
```
breathx, line 188: such options always look strange, especially if in most of places we use Some. I'd keep it splitted into two fns

ethexe/cli/src/params/node.rs
```diff
@@ -98,7 +98,7 @@ impl NodeParams {
                 .with_context(|| "invalid `validator` key")?,
             validator_session: ConfigPublicKey::new(&self.validator_session)
                 .with_context(|| "invalid `validator-session` key")?,
-            max_commitment_depth: self.max_depth.unwrap_or(Self::DEFAULT_MAX_DEPTH).get(),
+            max_eth_sync_depth: self.max_depth.unwrap_or(Self::DEFAULT_MAX_DEPTH).get(),
```
breathx, line 101: I'd rename cli param for parsing to also match the naming: so it will be max_sync_depth or --eth-max-sync-depth

ethexe/service/src/lib.rs
```diff
@@ -392,22 +375,44 @@ impl Service {
 
             match event {
                 Event::ServiceStarted => unreachable!("never handled here"),
+                Event::Observer(event) => match event {
```
breathx, line 378: wdyt about writing guarantees comments here: e.g. before calling `compute.receive_synced_head` will be `// Observer service already wrote block data and events in storage` like highlighting invariants while developing. Some of this codes new parts unlikely will be changed soon

ethexe/service/src/tests.rs
```diff
@@ -72,7 +72,7 @@ async fn basics() {
         sequencer: Default::default(),
         validator: Default::default(),
         validator_session: Default::default(),
-        max_commitment_depth: 1_000,
+        max_eth_sync_depth: 1_000,
```
breathx, line 75: not in this place, but tbh for prod I'd put much bigger amount. something like 10k blocks from eth

breathx: or even maybe 50k

ethexe/service/src/tests.rs
```diff
@@ -1589,20 +1584,29 @@ mod utils {
             self.apply_until_block_event_with_header(|e, _h| f(e)).await
         }
 
+        // TODO: consider to iterates also thru skipped blocks
```
breathx, line 1587: whats skipped blocks in this context? you mean for blocks that alloy didn't share headers of due to brut force bump of bn?

ethexe/db/src/database.rs
```diff
@@ -467,7 +461,8 @@ impl Database {
         else {
             bail!("No latest valid block found");
         };
-        let Some(reference_block_header) = self.block_header(reference_block_hash) else {
+        let Some(reference_block_header) = OnChainStorage::block_header(self, reference_block_hash)
```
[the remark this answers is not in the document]
grishasobol: It cannot called by self. because we have two traits definitions for block_header() in OnChainStorage and BlocksMetaStorage

ethexe/db/src/database.rs
```diff
@@ -325,21 +311,23 @@ impl CodesStorage for Database {
...
     fn set_program_code_id(&self, program_id: ProgramId, code_id: CodeId) {
         self.kv.put(
-            &KeyPrefix::ProgramToCodeId.one(program_id),
+            &KeyPrefix::ProgramToCodeId.two(self.router_address, program_id),
             code_id.into_bytes().to_vec(),
         );
     }
 
+    // TODO (gsobol): consider to move to another place
+    // TODO (gsobol): test this method
```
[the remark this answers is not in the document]
grishasobol: Cannot find them

ethexe/observer/src/lib.rs
```diff
@@ -75,20 +74,28 @@ pub enum ObserverEvent {
...
+    Block(SimpleBlockData),
+    BlockSynced(H256),
 }
 
+// TODO (gsobol): make tests for observer service
 pub struct ObserverService {
-    blobs: Arc<dyn BlobReader>,
     provider: Provider,
+    database: Box<dyn OnChainStorage>,
+    // TODO (gsobol): consider to make clone boxed for BlobRead, in order to avoid Arc usage.
```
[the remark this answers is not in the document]
grishasobol: Because we already use arc inside blob reader and it's useless . 
inside Provider - Arc
inside http_client - Arc
insider MockBlobReader - Arc as well

ethexe/observer/src/lib.rs
grishasobol: You mean this two addresses only?

ethexe/observer/src/lib.rs
grishasobol: Yes, but I think we should create BlobReader outside always. And provide it to observer after.

ethexe/cli/src/params/node.rs
grishasobol: agree

ethexe/service/src/tests.rs
grishasobol: it's already 100k for production see DEFAULT_MAX_DEPTH , you set it))

ethexe/service/src/tests.rs
grishasobol: Yes, I mean blocks which observer skips because they were skipped by alloy herder stream. This is possible on reorgs for example. Currently it's useless for us to look on these blocks, but maybe in future we would have a test where skipped blocks contains sensetive for test events.

ethexe/db/src/database.rs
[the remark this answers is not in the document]
grishasobol: This is fixed in next PR https://github.com/gear-tech/gear/pull/4521

ethexe/observer/src/lib.rs
```diff
@@ -38,27 +36,28 @@ use futures::{
...
 #[cfg(test)]
 mod tests;
 
 pub use blobs::*;
-pub use observer::*;
-pub use query::*;
 
 type BlobDownloadFuture = BoxFuture<'static, Result<(CodeId, u64, Vec<u8>)>>;
-type BlockFuture = BoxFuture<'static, Result<(H256, BlockHeader, Vec<BlockEvent>)>>;
+type BlockFuture = BoxFuture<'static, Result<(H256, Vec<(CodeId, CodeInfo)>)>>;
```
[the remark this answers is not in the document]
grishasobol: Agree, changed

ethexe/service/src/lib.rs
grishasobol: Agree, append

ethexe/db/src/database.rs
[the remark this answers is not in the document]
breathx: I meant this https://github.com/gear-tech/gear/blob/7935c43dee8a9951518d6b126e1128b0403dbd45/ethexe/db/src/lib.rs#L111
that transitively tests this

ethexe/observer/src/lib.rs
breathx: addresses + depth params

ethexe/observer/src/lib.rs
breathx: even if we don't send it to network? if we're unsynced we can't do anything anyway, but actually for most of cases it doesn't matter

ethexe/observer/src/lib.rs
breathx: Why? There're some default configurations we support (as ConsensusBlobReader)

ethexe/service/src/tests.rs
breathx: )))

breathx (approved):
Good to go, merge as soon as you want

ethexe/db/src/database.rs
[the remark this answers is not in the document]
grishasobol: This is good test, but better to have direct testing also

ethexe/observer/src/lib.rs
grishasobol: > if we're unsynced we can't do anything anyway

For example we can start to process off-chain transactions. Also BP could identify itself and starting to prepare for signature generation.

ethexe/observer/src/lib.rs
grishasobol: 1) You can see in the code - each time we creates observer we provides blob reader argument 
2) Maybe in future blob reader can support not only receive blobs using RPC requests, but also request blobs from vara or from p2p.

ethexe/observer/src/lib.rs
breathx: we cannot start it if we're unsynced since we have another states rather than most of validators (if changed, but anyway)

ethexe/observer/src/lib.rs
breathx: ok, lets always pass it

ethexe/observer/src/lib.rs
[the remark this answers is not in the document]
breathx: whats the difference? why clone boxed is better than arc?

ethexe/observer/src/lib.rs
grishasobol: Done

ethexe/observer/src/lib.rs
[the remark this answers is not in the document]
grishasobol: It's not better, but because of internal implementations we use here: Arc<Arc<>>. This is redundant. Actually I think better to use generic T: Clone, but because currently we prefer dyn, then clone_boxed is needed.

ethexe/observer/src/lib.rs
grishasobol: We can start in future when there will be some kind of off-chain consensus between validators. We must to do it in order to provide off-chain tx guaranties.

ethexe/observer/src/lib.rs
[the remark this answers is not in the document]
grishasobol: Anyway in comment we have: `consider to do`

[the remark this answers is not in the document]
grishasobol: let's discuss this with @ark0f

ethexe/observer/src/lib.rs
grishasobol: Ok, todo is added and hope @ecol-master will do it

PR #4515, opened 2025-02-17
feat(ethexe): add support for FROST signatures to sequencer & validator

StackOverflowExcept1on:
This PR adds support for FROST threshold signatures, which require only $\approx 10500$ gas to validate on Ethereum, allowing us to spend 30-40x less gas on large validator set.


ethexe/contracts/src/Router.sol
```diff
@@ -272,8 +272,9 @@ contract Router is IRouter, OwnableUpgradeable, ReentrancyGuardTransient {
 
     function commitBatch(
         Gear.BatchCommitment calldata _batchCommitment,
-        Gear.SignatureType _signatureType,
-        bytes[] calldata _signatures
+        uint256 _signatureRX,
+        uint256 _signatureRY,
+        uint256 _signatureZ
```
grishasobol, line 277: Could we concat this in one structure?

grishasobol (commented):
do we have test which tests this for incorrect validators?

ethexe/contracts/src/Router.sol
StackOverflowExcept1on: make sense

StackOverflowExcept1on: wdym but incorrect validatos? you mean malicious validators?

StackOverflowExcept1on: https://github.com/gear-tech/gear/blob/80e008b2d0240a5eda0e440305672bc15ca7b373/ethexe/sequencer/src/lib.rs#L280-L295

StackOverflowExcept1on: We don't listen to unauthenticated messages

ethexe/service/src/lib.rs
```diff
@@ -154,12 +169,36 @@ impl Service {
...
+            .validators_verifiable_secret_sharing_commitment()
+            .await
+            .with_context(|| "failed to query validators verifiable secret sharing commitment")?;
+        let verifiable_secret_sharing_commitment =
+            VerifiableSecretSharingCommitment::deserialize_whole(
+                &validators_verifiable_secret_sharing_commitment,
+            )
+            .with_context(|| "failed to decode VerifiableSecretSharingCommitment")?;
+
+        let public_key_package =
```
StackOverflowExcept1on, line 189: also makes sense to check that `public_key_package.verifying_key() == Router.validatorsAggregatedPublicKey()`. if condition is not met, then we don't even run the sequencer/validator and consider that whole network has been compromised

breathx:
Closing to clean up — will be reimplemented once the approach is finalized.

PR #4521, opened 2025-02-23
ethexe(refactor): db refactoring

grishasobol:


ethexe/common/src/db.rs
```diff
@@ -75,46 +74,32 @@ pub struct CodeInfo {
...
-    fn set_block_start_program_states(&self, block_hash: H256, map: BTreeMap<ActorId, H256>);
+    fn previous_not_empty_block(&self, block_hash: H256) -> Option<H256>;
+    fn set_previous_not_empty_block(&self, block_hash: H256, prev_commitment: H256);
 
-    fn block_end_program_states(&self, block_hash: H256) -> Option<BTreeMap<ActorId, H256>>;
-    fn set_block_end_program_states(&self, block_hash: H256, map: BTreeMap<ActorId, H256>);
+    fn block_program_states(&self, block_hash: H256) -> Option<BTreeMap<ActorId, H256>>;
+    fn set_block_program_states(&self, block_hash: H256, map: BTreeMap<ActorId, H256>);
 
+    fn block_outcome_is_empty(&self, block_hash: H256) -> Option<bool>;
```
breathx, line 89: could have default impl?

ethexe/db/src/database.rs
```diff
@@ -106,132 +104,195 @@ impl Clone for Database {
...
     }
+}
 
-    fn block_commitment_queue(&self, block_hash: H256) -> Option<VecDeque<H256>> {
-        self.block_small_meta(block_hash)
-            .and_then(|meta| meta.commitment_queue)
+#[cfg_attr(test, derive(serde::Serialize))]
+#[derive(Debug, Clone, Default, Encode, Decode, PartialEq, Eq)]
+struct BlockSmallData {
+    block_header: Option<BlockHeader>,
```
breathx, line 237: why is it optional?

breathx: if we have some hash that we must have header; event in sync process first of all we download all headers

ethexe/db/src/database.rs
```diff
@@ -43,26 +43,24 @@ use gprimitives::H256;
 use parity_scale_codec::{Decode, Encode};
 use std::collections::{BTreeMap, BTreeSet, VecDeque};
 
-const LOG_TARGET: &str = "ethexe-db";
-
 #[repr(u64)]
 enum KeyPrefix {
```
breathx, line 47: what if we somehow highlight usage of these prefixes: on types (preferrable) or at least with comments which keys must be used with/without router address and some additional keys ?

ethexe/db/src/mem.rs
```diff
@@ -32,13 +32,13 @@ impl CASDatabase for MemDb {
...
-    fn read(&self, hash: &H256) -> Option<Vec<u8>> {
+    fn read(&self, hash: H256) -> Option<Vec<u8>> {
         let key = hash.as_bytes().to_vec();
         self.inner.get(&key).map(|v| v.value().clone())
     }
 
-    fn write_by_hash(&self, hash: &H256, data: &[u8]) {
-        debug_assert_eq!(*hash, crate::hash(data));
+    fn write_by_hash(&self, hash: H256, data: &[u8]) {
+        debug_assert_eq!(hash, crate::hash(data));
```
breathx, line 41: btw I'd keep this assertion even in prod. it's critical for security to maintain db hashes invariants

ethexe/observer/src/lib.rs
```diff
@@ -133,7 +133,7 @@ impl Stream for ObserverService {
             if let Some(header) = self.block_sync_queue.pop_front() {
                 let sync = ChainSync {
                     provider: self.provider.clone(),
-                    database: self.database.clone_boxed(),
+                    database: self.database.clone(),
```
breathx, line 136: is clone_boxed about to be deprecated soon?

ethexe/observer/src/lib.rs
```diff
@@ -81,7 +81,7 @@ pub enum ObserverEvent {
 // TODO (gsobol): make tests for observer service
 pub struct ObserverService {
     provider: Provider,
-    database: Box<dyn OnChainStorage>,
+    database: Database,
```
breathx, line 84: mb just db: Database for code consistency

ethexe/observer/src/sync.rs
```diff
@@ -40,17 +40,17 @@ use std::{
 };
 
 // TODO (gsobol): make tests for ChainSync
-pub(crate) struct ChainSync {
+pub(crate) struct ChainSync<DB: OnChainStorage + CodesStorage> {
```
breathx, line 43: if im not mistaken we were discussion about this, what dont we just put Database itself?

breathx (approved):

ethexe/db/src/database.rs
grishasobol: BlockSmallData is an internal structure which is just to optimise database implementation. For database BlockSmallData contains several in-depended fields each one can be set or not, so each field is Option, except bools which can be set by default.

ethexe/db/src/mem.rs
grishasobol: Keeping this in produciton would make this method useless. Because the only one difference it can provide for us is: whether we need to calculate hash or not. Actually we do not uses it anyway))

ethexe/observer/src/lib.rs
grishasobol: If we would decide to do this) actually we use it inside db:
```
    pub fn from_one<DB: CASDatabase + KVDatabase>(db: &DB, router_address: [u8; 20]) -> Self {
        Self {
            cas: CASDatabase::clone_boxed(db),
            kv: KVDatabase::clone_boxed_kv(db),
            router_address,
        }
    }
```

ethexe/observer/src/sync.rs
grishasobol: Here we have no problems to define as generic because:
1) it's not public
2) here we specify what chain sync could do : just fill codes and on-chain storages

ethexe/common/src/db.rs
grishasobol: donw

ethexe/db/src/database.rs
grishasobol: Sorry, but I cannot understand what do you mean here

ethexe/observer/src/lib.rs
grishasobol: done

ethexe/observer/src/lib.rs
grishasobol: Renamed clone_boxed_kv -> clone_boxed

ethexe/db/src/database.rs
grishasobol: Discussed

ethexe/db/src/database.rs
grishasobol: Waiting

ethexe/observer/src/lib.rs
grishasobol: Clone_boxed currently cannot be removed so leave it as is

ethexe/observer/src/sync.rs
grishasobol: Discussed - ok

ethexe/db/src/database.rs
grishasobol: Discussed

PR #4524, opened 2025-03-06
feat(pallet-gear-eth-bridge): Add message fee

ByteNacked:
Resolves # .

Adds a fee for messages when the queue fee threshold is exceeded.

If the number of messages in the queue exceeds a certain set threshold, a predefined fee is charged for message inclusion in queue.

@reviewer-or-team


pallets/gear-eth-bridge/src/lib.rs
```diff
@@ -295,16 +341,113 @@ pub mod pallet {
...
+                // Return error if fee is greater than value
+                let refund = value.checked_sub(&fee).ok_or(Error::<T>::InsufficientFee)?;
+
+                AccountsFee::<T>::mutate(account_id.clone(), |details| {
+                    let details = details.get_or_insert_with(Default::default);
+                    details.fee = details.fee.saturating_add(fee);
+                    details.refund = details.refund.saturating_add(refund);
+                });
+            }
+            // Queue threshold not exceeded, no fee charged, we should make full value refund
```
gshep, line 378: Sorry for this insufficient remark. However I left another one so should be ok ).

I would place that comment and the previous one within the corresponding block.

pallets/gear-eth-bridge/src/lib.rs
```diff
@@ -405,6 +551,9 @@ pub mod pallet {
                     // Setting zero queue root, keeping invariant of this key existence.
                     QueueMerkleRoot::<T>::put(H256::zero());
 
+                    // Transfer fees to treasury and refund to users.
+                    Self::transfer_fees();
```
gshep, line 555: It may be too time consuming so I suggest to benchmark it and take the weight into consideration

gshep (approved):
Looks good! :1st_place_medal: 

Nevertheless please take into account the remark about weight.

breathx (changes requested):
don't merge pls, will submit review soon

pallets/gear-eth-bridge/src/lib.rs
ByteNacked: These comments describe what happens inside the corresponding block. In this case, I prefer to place them just before the block rather than inside it.

pallets/gear-eth-bridge/src/lib.rs
ByteNacked: Added benchmark

ekovalev:
Is there a reference to some issue or task? I'm having a bit of a hard time understanding what it is all about and how does this fit into the bridging flow.
A short description in the PR itself would also help )

ByteNacked:
> Is there a reference to some issue or task? I'm having a bit of a hard time understanding what it is all about and how does this fit into the bridging flow. A short description in the PR itself would also help )

Updated PR description

breathx (changes requested):
reconsidering fee extraction approach

ekovalev:
My take: from implementation standpoint not many questions - everything looks fine.
From conceptual point of view I think making fees payable as value (that is, in native tokens) might be a bit compelling: it is possible that the user sending a bridging request doesn't have Vara tokens (let's say, he wants to completely unwrap some external assets previously wrapped). Also, unlike gas, value is not propagated across messages tree (if the original sender is not the one submitting bridging request). Is there a way to charge extra fees via gas somehow?

ByteNacked:
PR superseded by #4590.

PR #4535, opened 2025-03-13
fix(ethexe): Receiver can miss events in tests

ark0f:


breathx (approved):

breathx:
receiver still may remain clonable?

playX18:
How does removal of mutex + arc actually prevents receiver from missing events? I am not quite understanding it

breathx:
> How does removal of mutex + arc actually prevents receiver from missing events? I am not quite understanding it

the issue @ark0f faced is that receivers on resubscribe don't have historical events, in his branch it leads to troubles. generally, cant see any cons using it with ref

playX18 (approved):

PR #4537, opened 2025-03-13
feat(ethexe): append service identity (block-identity) 

grishasobol:
Additionally remove Arc usage from ethexe-ethereum (alloy already use Arc as inner field inside everywhere)

### DIFF
--- a/ethexe/ethereum/src/mirror/mod.rs
+++ b/ethexe/ethereum/src/mirror/mod.rs
@@ -25,14 +25,13 @@ use anyhow::{anyhow, Result};
 use ethexe_signer::Address as LocalAddress;
 use events::signatures;
 use gprimitives::{MessageId, H256};
-use std::sync::Arc;
 
 pub mod events;
 
 type InstanceProvider = AlloyProvider;
 type Instance = IMirror::IMirrorInstance<(), InstanceProvider>;
 
-type QueryInstance = IMirror::IMirrorInstance<(), Arc<RootProvider>>;
+type QueryInstance = IMirror::IMirrorInstance<(), RootProvider>;
 
 pub struct Mirror(Instance);
 
@@ -48,7 +47,7 @@ impl Mirror {
     pub fn query(&self) -> MirrorQuery {
         MirrorQuery(QueryInstance::new(
             *self.0.address(),
-            Arc::new(self.0.provider().root().clone()),
+            self.0.provider().root().clone(),
         ))
     }
 
@@ -114,7 +113,7 @@ pub struct MirrorQuery(QueryInstance);
 
 impl MirrorQuery {
     pub async fn new(rpc_url: &str, router_address: LocalAddress) -> Result<Self> {
-        let provider = Arc::new(ProviderBuilder::default().connect(rpc_url).await?);
+        let provider = ProviderBuilder::default().connect(rpc_url).await?;
 
         Ok(Self(QueryInstance::new(
             Address::new(router_address.0),
--- a/ethexe/ethereum/src/mirror/mod.rs
+++ b/ethexe/ethereum/src/mirror/mod.rs
@@ -29,7 +29,7 @@ use std::sync::Arc;
 
 pub mod events;
 
-type InstanceProvider = Arc<AlloyProvider>;
+type InstanceProvider = AlloyProvider;
 type Instance = IMirror::IMirrorInstance<(), InstanceProvider>;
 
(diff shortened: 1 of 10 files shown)


ethexe/ethereum/src/mirror/mod.rs
```diff
@@ -29,7 +29,7 @@ use std::sync::Arc;
 
 pub mod events;
 
-type InstanceProvider = Arc<AlloyProvider>;
+type InstanceProvider = AlloyProvider;
```
breathx, line 31: rm alias ?

ethexe/observer/src/lib.rs
```diff
@@ -64,15 +64,24 @@ pub struct EthereumConfig {
     pub block_time: Duration,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct BlockSyncedData {
```
breathx, line 68: more likely BlockAuthorities, wdyt?

breathx: also please place todo as we agreed to cache and operate with validators more accurate

ethexe/observer/src/lib.rs
```diff
@@ -299,6 +304,10 @@ impl ObserverService {
         }
     }
 
+    pub fn block_time(&self) -> u64 {
```
breathx, line 307:
```suggestion
    pub fn block_time_secs(&self) -> u64 {
```

ethexe/service/src/lib.rs
```diff
@@ -723,3 +741,61 @@ impl Service {
         Ok(tx_hash)
     }
 }
+
+mod identity {
+    use super::*;
+
+    pub struct ServiceIdentity {
```
breathx, line 748: It doesnt look like service rn. maybe just IdentityStore or smth?

ethexe/service/src/lib.rs
```diff
@@ -723,3 +741,61 @@ impl Service {
...
+                let slot = block.header.timestamp / self.block_duration;
+                let index = Self::block_producer_index(data.validators.len(), slot);
+                self.producer = data.validators.get(index).cloned();
+                Ok(true)
+            }
+        }
+
+        pub fn block_producer(&self) -> Result<Address> {
+            self.producer
+                .ok_or_else(|| anyhow!("block producer is not set"))
```
breathx, line 789:
```suggestion
                .ok_or_else(|| anyhow!("block producer is not set: receive synced block first"))
```

ethexe/service/src/lib.rs
```diff
@@ -723,3 +741,61 @@ impl Service {
...
+            }
+        }
+
+        pub fn receive_new_chain_head(&mut self, block: SimpleBlockData) {
+            self.block = Some(block);
+            self.producer = None;
+        }
+
+        /// Returns whether synced block is previously received chain head.
+        pub fn receive_synced_block(&mut self, data: &BlockSyncedData) -> Result<bool> {
```
breathx, line 771: shouldn't this be performed automatically on receive new head?

breathx (changes requested):

ethexe/observer/src/lib.rs
grishasobol: I do not think BlockAuthorities is a good name, because this is abstract data which is attached to block synced events. something else also could be added here or removed

ethexe/service/src/lib.rs
grishasobol: Actually here I mean Identity of Service. For services we have other structure: Something**Service**, where Service is in the end.

ethexe/service/src/lib.rs
grishasobol: how to do this without validators information? we can make optimisation after where new validators info could be propagated from previous block validators in most cases

ethexe/ethereum/src/mirror/mod.rs
grishasobol: done

ethexe/observer/src/lib.rs
grishasobol: > also please place todo as we agreed to cache and operate with validators more accurate

added in service/lib.rs

ethexe/observer/src/lib.rs
grishasobol: done

ethexe/service/src/lib.rs
grishasobol: done

breathx (approved):

PR #4542, opened 2025-03-14
feat(ethexe): append codes queue propagation (does not fix the problem with lost codes)

grishasobol:
Lost codes could be fixed much simpler in block producer model

### DIFF
--- a/ethexe/db/src/database.rs
+++ b/ethexe/db/src/database.rs
@@ -20,7 +20,7 @@
 
 use crate::{
     overlay::{CASOverlay, KVOverlay},
-    CASDatabase, KVDatabase,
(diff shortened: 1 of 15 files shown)


ethexe/db/src/database.rs
```diff
@@ -139,6 +139,14 @@ impl Database {
         }
     }
 
+    pub fn new_in_memory(router_address: [u8; 20]) -> Self {
```
breathx, line 142: also use `from_one` to keep it the same object?
```suggestion
    pub fn memory(router_address: [u8; 20]) -> Self {
```

breathx (approved):

PR #4541, opened 2025-03-14
refactor(ethexe): create separate DB folders for routers, do not use router address as key prefix

playX18:
@breathx

PR #4543, opened 2025-03-14
feat(ethexe): remove router address from Database

ecol-master:
Remove router address from Database and modify the appropriate methods and tests. 

@reviewer-or-team


playX18:
This PR is duplicate of another PR https://github.com/gear-tech/gear/pull/4541, @breathx and me also discussed on simplification of KeyPrefix type, so it is more like enum variants rather than a simple enum.

PR #4576, opened 2025-03-26
feat(ethexe/contracts): add parsing of ethereum events and replies from sails framework

### LINKED ISSUE #4338: ethexe: manually create decoders; impl callbacks

StackOverflowExcept1on:
Resolves #4338

Also see:
- gear-tech/sails#844
- gear-tech/sails#859


ethexe/contracts/src/Mirror.sol
```diff
@@ -172,14 +176,104 @@ contract Mirror is IMirror {
...
+            uint8 replyCodeDiscriminant = uint8(bytes1(replyCode));
+
+            bytes memory payload;
+
+            if (replyCodeDiscriminant == 0) {
+                /* gear_core::message::ReplyCode::Success = 0 */
+                payload = _message.payload;
+            } else {
+                /* gear_core::message::ReplyCode::{Error = 1, Unsupported = 255} */
+                payload = abi.encodeWithSelector(ICallbacks.errorReply.selector, _message.id, replyCode);
```
StackOverflowExcept1on, line 266: should we pass other message details to `errorReply()` such as `_message.payload` (e.g. for emit event with panic string)?

ethexe/contracts/src/Mirror.sol
```diff
@@ -172,14 +176,104 @@ contract Mirror is IMirror {
...
+            }
+
+            (bool success,) = _message.destination.call{gas: 500_000}(_message.payload);
+
+            if (success) {
+                return;
+            }
+        } else {
+            emit Reply(_message.payload, _message.value, _message.replyDetails.to, _message.replyDetails.code);
```
StackOverflowExcept1on, line 275: 1. there was question about not emitting auto replies.
2. should we emit Reply event always (to preserve the protocol logic)? now we only emit it if destination is user.
3. should we process result `(bool success,) = _message.destination.call{gas: 500_000}(_message.payload)` and emit Reply event in case of error?

ethexe/contracts/src/Mirror.sol
```diff
@@ -172,14 +176,104 @@ contract Mirror is IMirror {
...
+                            assembly ("memory-safe") {
+                                log3(memPtr, size, topic1, topic2, topic3)
+                            }
+                        } else if (topicsLength == 4) {
+                            assembly ("memory-safe") {
+                                log4(memPtr, size, topic1, topic2, topic3, topic4)
+                            }
+                        }
+
+                        return;
```
StackOverflowExcept1on, line 242: here we also do not emit Message event, which also breaks core protocol logic just to save gas.

ethexe/contracts/src/Mirror.sol
StackOverflowExcept1on: maybe it's also worth renaming method to something like `onErrorReply`?

ethexe/contracts/src/Mirror.sol
```diff
@@ -172,14 +176,104 @@ contract Mirror is IMirror {
...
+
+            if (replyCodeDiscriminant == 0) {
+                /* gear_core::message::ReplyCode::Success = 0 */
+                payload = _message.payload;
+            } else {
+                /* gear_core::message::ReplyCode::{Error = 1, Unsupported = 255} */
+                payload = abi.encodeWithSelector(ICallbacks.errorReply.selector, _message.id, replyCode);
+            }
+
+            (bool success,) = _message.destination.call{gas: 500_000}(_message.payload);
```
StackOverflowExcept1on, line 269: when we commit blocks, we use gas limit `10_000_000`, and here gas limit is `500_000` per call. it is unlikely that we can occupy the entire ethereum block with gas limit `36_000_000`. this means that we can only make only `10_000_000 / 500_000 = 20` calls to `replyOn`.

StackOverflowExcept1on: maybe we should reconsider that `500_000` limit

ethexe/contracts/src/Mirror.sol
```diff
@@ -172,14 +176,105 @@ contract Mirror is IMirror {
...
+
         emit Message(_message.id, _message.destination, _message.payload, _message.value);
     }
 
     /// @dev Non-zero value always sent since never goes to mailbox.
     function _sendReplyMessage(Gear.Message calldata _message) private {
         _transferValue(_message.destination, _message.value);
 
-        emit Reply(_message.payload, _message.value, _message.replyDetails.to, _message.replyDetails.code);
+        if (_message.destination.code.length > 0) {
```
StackOverflowExcept1on, line 255: it may no longer work in next hardfork

ethexe/contracts/src/Mirror.sol
```diff
@@ -1,14 +1,18 @@
...
 import {ERC1967Utils} from "@openzeppelin/contracts/proxy/ERC1967/ERC1967Utils.sol";
 import {StorageSlot} from "@openzeppelin/contracts/utils/StorageSlot.sol";
+import {ICallbacks} from "./ICallbacks.sol";
 import {IMirror} from "./IMirror.sol";
 import {IRouter} from "./IRouter.sol";
 import {IWrappedVara} from "./IWrappedVara.sol";
 import {Gear} from "./libraries/Gear.sol";
 
 contract Mirror is IMirror {
+    address internal constant ETH_EVENT_ADDR = 0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF;
```
grishasobol, line 14: Whould be very nice if you explain strange constants here and in other places (comments).

ethexe/contracts/src/Mirror.sol
```diff
@@ -172,14 +176,105 @@ contract Mirror is IMirror {
 
     /// @dev Value never sent since goes to mailbox.
     function _sendMailboxedMessage(Gear.Message calldata _message) private {
+        bytes calldata payload = _message.payload;
+
+        if (_message.destination == ETH_EVENT_ADDR && _message.value == 0 && payload.length > 0) {
+            uint256 topicsLength;
+            assembly ("memory-safe") {
+                topicsLength := shr(248, calldataload(payload.offset))
```
grishasobol, line 184: an here: why 248 ?

ethexe/contracts/src/Mirror.sol
```diff
@@ -172,14 +176,105 @@ contract Mirror is IMirror {
...
+                        uint256 memPtr = Memory.allocate(size);
+                        assembly ("memory-safe") {
+                            calldatacopy(memPtr, add(payload.offset, topicsLengthInBytes), size)
+                        }
+
+                        bytes32 topic2;
+                        bytes32 topic3;
+                        bytes32 topic4;
+                        assembly ("memory-safe") {
+                            topic2 := calldataload(add(payload.offset, 33))
```
grishasobol, line 219: And here? What the magic numbers?))

ethexe/contracts/src/Mirror.sol
grishasobol: This could be an init param for mirror and router

ethexe/contracts/src/Mirror.sol
StackOverflowExcept1on: this is internal sails address
https://github.com/gear-tech/sails/blob/master/rs/src/solidity.rs

ethexe/contracts/src/Mirror.sol
StackOverflowExcept1on: we need to load `uint8` to stack, but `calldataload` returns `uint256`, so this shift is needed to remove the extra bits.

ethexe/contracts/src/Mirror.sol
grishasobol: Could you specify this in comments please?

ethexe/contracts/src/Mirror.sol
StackOverflowExcept1on: since we use `topicsLength` to get type of log: `log1`, `log2`, `log3`, `log4` we need to add `+ 1` to already existing offset (`+ 32`, `+ 64`, ...)

ethexe/contracts/src/MirrorProxySmall.sol
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.28;
+
+/// @dev This is custom implementation of ERC1167 that supports upgrades.
```
breathx, line 4: some comments how and why etc

breathx (changes requested):
For similar changes in future, please split into different isolated PRs

ethexe/contracts/src/Mirror.sol
```diff
@@ -1,19 +1,26 @@
...
+    /// @dev Special address to which Sails contract sends messages so that Mirror can decode events:
+    ///      https://github.com/gear-tech/sails/blob/master/rs/src/solidity.rs
+    address internal constant ETH_EVENT_ADDR = 0xFFfFfFffFFfffFFfFFfFFFFFffFFFffffFfFFFfF;
+
     address public immutable router;
 
     address public inheritor;
     /// @dev This nonce is the source for message ids unique generations. Must be bumped on each send. Zeroed nonce is always represent init message by eligible account.
     address public initializer;
+    bool isSmall;
```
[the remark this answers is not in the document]
StackOverflowExcept1on: they are located nearby for optimization (address and bool can fit into 1 slot)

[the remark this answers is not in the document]
breathx: damn sol ...

ethexe/contracts/src/Mirror.sol
StackOverflowExcept1on: > why is it?

https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7702.md
see `extcodesize` changes. it's no longer possible to know is address user or contract.
`extcodesize` has also been removed in `EOFv1` smart contracts.

> shouldn't this check be == 4?

don't get it

ethexe/contracts/src/Mirror.sol
StackOverflowExcept1on: 3. this makes sense, but how do we do it? since we have either data for `call` or data for `Reply` event.

ethexe/contracts/src/Mirror.sol
breathx: oh, for some reason on the first sight though that its about reply code ...

breathx (approved):

ethexe/contracts/src/Mirror.sol
StackOverflowExcept1on: what is the point of this event if calldata is encoded differently?

PR #4585, opened 2025-04-01
fix(runtime): adjust economical constants, bump gas price

breathx:


ukint-vs (approved):

ekovalev (approved):
Just an observation: seems that for the `GasMultiplier` it should always hold that strictly one of the two fields of the struct is a non-zero integer, while the other must always be set to 0 (with a single exception when they are both set to 1).
I assume this invariant is somehow enforced across the codebase, is it?

PR #4588, opened 2025-04-02
refactor(gear-bank): Derive gear bank account id from the pallet id

ekovalev:
Transfers to gear bank can occur multiple times at each block (with every message sent).
Currently for an external observer these transfers look as transfers to an unknown address (which might look weird if not suspicious).
This change links the gear bank account id to the pallet id the same way it is done for other pallets (treasury, staking rewards pool etc.).
The bank account id is stored in the pallet storage which spares the address recalculation each time it is needed (potentially, many times in a single block); on the contrary, it should be fetched from storage only once per block and be kept in overlay across all extrinsics.

Changes:

- [x] Update the derivation of the gear bank account id (involves modification of the `Config` trait)
- [x] Apply a migration to move all funds from the old account id to the new one.
- [x] Update the `gsdk` and `gclient` API.

### DIFF
--- a/pallets/gear-bank/src/lib.rs
+++ b/pallets/gear-bank/src/lib.rs
@@ -33,10 +33,13 @@ use alloc::format;
 
 pub use pallet::*;
 
-use frame_support::traits::{
-    fungible,
-    tokens::{Fortitude, Preservation, Provenance},
-    Currency, StorageVersion,
+use frame_support::{
+    traits::{
+        fungible,
+        tokens::{Fortitude, Preservation, Provenance},
+        Currency, StorageVersion,
+    },
+    PalletId,
 };
 
 #[macro_export]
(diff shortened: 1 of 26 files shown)


ukint-vs (approved):
Good catch

pallets/gear-bank/src/lib.rs
```diff
@@ -228,6 +233,18 @@ pub mod pallet {
...
+    pub struct DefaulBankAddress<T: Config>(PhantomData<T>);
+    impl<T: Config> Get<AccountIdOf<T>> for DefaulBankAddress<T> {
+        fn get() -> AccountIdOf<T> {
+            Pallet::<T>::bank_address()
+        }
+    }
+
+    // Private storage to hold the GearBank's AccountId.
+    #[pallet::storage]
+    pub type BankAddress<T> = StorageValue<_, AccountIdOf<T>, ValueQuery, DefaulBankAddress<T>>;
```
breathx, line 246: This default works only as rust types shortcut, it doesnt insert value into storage on first access, so every caller will do compute of the address: set it somewhere explicitly please

pallets/gear-bank/src/migrations.rs
```diff
@@ -0,0 +1,248 @@
...
+                new_balances.1 == old_balances.0,
+                "Balance at destination is different from the source"
+            );
+            ensure!(
+                Pallet::<T>::on_chain_storage_version() == 1,
+                "v1 not applied"
+            );
+
+            // Ensure the new bank address has been written to storage
+            let prefix = BankAddress::<T>::storage_value_final_key();
```
breathx, line 129: without default this check will be simpler

breathx (changes requested):

breathx (approved):

PR #4598, opened 2025-04-12
feat(ci): Instrument whole network

ark0f:


breathx (changes requested):
I'd consider usage of try runtime snapshots for this, because there gonna be a lot of problems fetching big storage: try-runtime cli solved this problems for us in two ways: smart retrying logic of fetching (still can be failed since time may be longer than pruning on rpc), and cached operations across dumped storage keys

PR #4599, opened 2025-04-13
feat(ethexe-ethereum): add ability to wait for code via http rpc

StackOverflowExcept1on:


ethexe/ethereum/src/router/mod.rs
```diff
@@ -117,28 +117,46 @@ impl Router {
...
-                if event.codeId == code_id {
-                    return Ok(event.valid);
+                while let Some(log) = router_events.next().await {
+                    if Self::is_validated_code(log, code_id)? {
+                        return Ok(true);
+                    }
                 }
             }
-        }
+            None => loop {
```
StackOverflowExcept1on, line 132: because http rpc's can be unreliable and  connection

ethexe/ethereum/src/router/mod.rs
```diff
@@ -117,28 +117,46 @@ impl Router {
...
+                        }
+                    }
+                }
+            },
+        };
 
         Err(anyhow!("Failed to define if code is validated"))
     }
 
+    fn is_validated_code(log: Log, code_id: CodeId) -> Result<bool> {
```
grishasobol, line 148: I think you have 3 statuses here instead of 2:
1) Some(true) - it's code id and is valid
2) Some(false) - it's code id and is not valid
3) None - it's not got validated event or for another code

breathx: actually None case is unreachable since it checks the first topic inside

breathx (changes requested):
seems that now it never returns that code is failed validation

StackOverflowExcept1on:
I had some problems with http rpc. In some cases they would not send response for long time.

breathx:
> I had some problems with http rpc. In some cases they would not send response for long time.

Im ok with a solution itself, but it will never resolve loop if code wasn't approved and you wait for it. previously and in case of rpc it returns event.valid as an Ok clause

breathx:
Not actual

PR #4630, opened 2025-04-30
refactor(core-processor) got rid of the payload cloning

### LINKED ISSUE #3773: Remove unnecessary cloning in core-processor
whoa, I was missing it. I'll open an issue from this comment to remove cloning here because it could be pretty harmful

_Originally posted by @breathx in https://github.com/gear-tech/gear/pull/3743#discussion_r1507915346_

Lazark0x:
Resolves #3773 .


breathx (changes requested):
Discussing.

General concerns: caller mutability issues

breathx (approved):

core-processor/src/executor.rs
```diff
@@ -229,7 +229,7 @@ where
     // Output
     Ok(DispatchResult {
         kind,
-        dispatch,
+        dispatch: info.dispatch,
```
breathx, line 232: it's unchanged in current design, right?

core-processor/src/ext.rs
```diff
@@ -1751,17 +1778,14 @@ mod tests {
     fn test_send_push_input() {
         let mut ext = Ext::new(
             ProcessorContextBuilder::new()
-                .with_message_context(MessageContextBuilder::new().build())
+                .with_message_context(
+                    MessageContextBuilder::new()
+                        .with_payload(vec![1, 2, 3, 4, 5, 6].try_into().unwrap())
```
breathx, line 1783: hide inside 
```suggestion
                        .with_payload(vec![1, 2, 3, 4, 5, 6])
```

core-processor/src/ext.rs
```diff
@@ -1154,17 +1157,24 @@ impl<LP: LazyPagesInterface> Externalities for Ext<LP> {
...
             )?;
-            PayloadSliceLock::try_new((at, end), &mut mutator.ext.context.message_context)
-                .ok_or_else(|| FallibleExecutionError::ReadWrongRange.into())
-        })
-    }
 
-    fn unlock_payload(&mut self, payload_holder: &mut PayloadSliceLock) -> UnlockPayloadBound {
-        UnlockPayloadBound::from((&mut self.context.message_context, payload_holder))
+            PayloadSlice::try_new(
+                at,
```
breathx, line 1162: in this place we do have invariant for end >= at, but it'd be nice to have also inside PayloadSlice::try_new as well just in case

core-processor/src/processing.rs
```diff
@@ -130,7 +130,8 @@ where
 
     let exec_result = executor::execute_wasm::<Ext>(
         balance,
-        dispatch.clone(),
+        // dispatch clone is cheap because is uses `Arc` for payload
```
breathx, line 133: mv comment inside?

breathx: next to the actual .clone

core/src/message/context.rs
```diff
@@ -211,8 +210,7 @@ impl ContextStore {
 /// Context of currently processing incoming message.
 #[derive(Debug)]
 pub struct MessageContext {
-    kind: DispatchKind,
-    current: IncomingMessage,
+    dispatch: IncomingDispatch,
```
breathx, line 213: actually it seems bad to use not de-constructed IncomingDispatch since it provides option to misuse which store: MessageContext::store or MessageContext::dispatch::context

breathx: worth checking if we could remove one of them

core/src/message/incoming.rs
```diff
@@ -23,29 +23,26 @@ use crate::{
...
     source: ProgramId,
-    /// Message payload.
-    payload: Payload,
     /// Message gas limit. Required here.
     gas_limit: GasLimit,
     /// Message value.
     value: Value,
     /// Message details like reply message ID, status code, etc.
     details: Option<MessageDetails>,
+    /// Message payload.
```
breathx, line 44: I'd keep it on the same place above gas limit

core/src/message/incoming.rs
```diff
@@ -74,12 +71,17 @@ impl IncomingMessage {
             self.id,
             self.source,
             destination,
-            self.payload,
+            Arc::unwrap_or_clone(self.payload),
```
breathx, line 74: shouldn't we make it just unwrap? try_into_strored -> Option<StoredMessage>

pallets/gear-builtin/src/lib.rs
```diff
@@ -71,12 +71,11 @@ use gear_core::{
     utils::hash,
 };
 use impl_trait_for_tuples::impl_for_tuples;
+pub use pallet::*;
```
breathx, line 74: separate pub uses from just uses

breathx (changes requested):
generally, lgtm

breathx:
and move payload slice to core/buffer

core-processor/src/ext.rs
Lazark0x: Yeah, that would be better

core-processor/src/ext.rs
Lazark0x: Added this check

core-processor/src/processing.rs
Lazark0x: Oops, fixed

core/src/message/context.rs
Lazark0x: Yeah, I think with the cloned dispatch it's better to put it back the way it was

core/src/message/incoming.rs
Lazark0x: I think we could get `Option`, but we would have to handle the error situation somehow, here we get what we want anyway even if the `payload` has many hard refs

pallets/gear-builtin/src/lib.rs
Lazark0x: Oops, my IDE changed it, fixed

core/src/message/incoming.rs
breathx: can we at least add here log::warn so we could check if we miss refs somewhere ?

breathx (approved):
@ark0f please review

core/src/message/incoming.rs
```diff
@@ -75,12 +72,22 @@ impl IncomingMessage {
             self.id,
             self.source,
             destination,
-            self.payload,
+            Arc::try_unwrap(self.payload).unwrap_or_else(|payload| {
+                log::error!(
+                    "IncomingMessage payload has multiple references, this is unexpected behavior"
```
techraed, line 77: tests show multiple references exist

Lazark0x: The original dispatch from DispatchResult has been removed, fixing the issue.

core-processor/src/executor.rs
```diff
@@ -103,6 +100,14 @@ where
     });
     let value_counter = ValueCounter::new(value_available);
 
+    // Creating message context.
+    let message_context = MessageContext::new(
+        // dispatch clone is cheap because is uses `Arc` for payload
+        dispatch.clone(),
```
breathx: useless clone

breathx: mv comment to caller side (processing.rs)

pallets/gear-builtin/src/lib.rs
```diff
@@ -423,13 +422,12 @@ impl<T: Config> BuiltinDispatcher for BuiltinRegistry<T> {
...
 
-                let mut dispatch_result =
-                    DispatchResult::success(dispatch.clone(), actor_id, gas_amount);
+                let mut dispatch_result = DispatchResult::success(&dispatch, actor_id, gas_amount);
 
                 // Create an artificial `MessageContext` object that will help us to generate
                 // a reply from the builtin actor.
                 let mut message_context =
-                    MessageContext::new(dispatch, actor_id, Default::default());
+                    MessageContext::new(dispatch.clone(), actor_id, Default::default());
```
breathx, line 431: comment on clone

PR #4633, opened 2025-05-02
fix(builtins): mint and lock ED for builtins as for programs

breathx:
- impl storage version for builtin pallet
- impl 0 -> 1 migration that topups builtins balances and locks ed as for programs
- top up basic endowment amount for development genesises

### DIFF
--- a/runtime/vara/src/migrations.rs
+++ b/runtime/vara/src/migrations.rs
@@ -22,11 +22,10 @@ use crate::*;
 pub type Migrations = (
     BagsListMigrate<Runtime>,
     pallet_gear_bank::migrations::MigrateToV1<Runtime>,
+    pallet_gear_builtin::migration::MigrateToV1<Runtime, TreasuryAccount>,
 );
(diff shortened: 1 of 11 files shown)


runtime/vara/src/migrations.rs
```diff
@@ -22,11 +22,10 @@ use crate::*;
 pub type Migrations = (
     BagsListMigrate<Runtime>,
     pallet_gear_bank::migrations::MigrateToV1<Runtime>,
+    pallet_gear_builtin::migration::MigrateToV1<Runtime, TreasuryAccount>,
```
ukint-vs, line 25: We will top up Treasury, right?

ukint-vs (approved):

runtime/vara/src/migrations.rs
breathx: Treasury has enough funds on both networks to payout from 3 to 4 tokens

PR #4634, opened 2025-05-02
feat!(gear): invariant for value passing only with messages; error replies inherit value as well

breathx:
Previously, on error occurred, error reply had value=0 and initially applied one was immediately silently returned on the balance of the source. Now it's attached to err reply and unlocks once err reply handled.

UPD:
Short summary:
1. Value is always transferred on message processed (for the first time, as usual).
2. In case of error occurred, value refund will be done by attaching it to error reply, if some, otherwise - kept.
3. In case of error occurred, if there will be error reply, value is transferred (1.) already locked inside gear-bank, so it won't require GearBank::deposit_value as well as won't require any pallet-balances interactions and ED manipulations and this (error-reply) message sending


pallets/gear/src/tests.rs
```diff
@@ -81,6 +82,158 @@ use utils::*;
...
+
+        assert_balance(USER_1, user_balance - VALUE, VALUE);
+
+        run_to_next_block(None);
+
+        let pid_balance = Balances::free_balance(pid.cast::<AccountId>());
+        assert_eq!(pid_balance, get_ed() + VALUE);
+
+        assert_balance(USER_1, user_balance - VALUE, 0u8);
+        assert_balance(pid, pid_balance, 0u8);
```
techraed, line 161: At first I thought that it was implementation error and a bug. But from tests I see that it was done intentionally.
So in case non-reply-able message with value fails you still transfer the value.
What benefits are intended to gain with that?

That seems to be inconsistent with a general behavior of the system - in case of error all important state transitions must be reverted.

techraed (changes requested):

pallets/gear/src/tests.rs
techraed: discussed

I'm still against these changes, but it doesn't bring that much changes to the development of smart contracts, so I'm ok with introducing a new invariant

ark0f (approved):

PR #4649, opened 2025-05-07
feat(ethexe/mirror): partially support whether to call message or just emit event

breathx:
in addition, fix for `exit(address(0))` was added (previously never exited to 0 addr) as well as some beautifications made


ethexe/contracts/src/Mirror.sol
```diff
@@ -287,36 +316,47 @@ contract Mirror is IMirror {
...
             (bool success,) = _message.destination.call{gas: 500_000}(_message.payload);
 
             if (success) {
                 return;
             }
-        } else {
-            emit Reply(_message.payload, _message.value, _message.replyDetails.to, _message.replyDetails.code);
         }
+
+        emit Reply(_message.payload, _message.value, _message.replyDetails.to, _message.replyDetails.code);
```
StackOverflowExcept1on, line 356: in this case it would also make sense to pass `_message.call` to `Reply`

StackOverflowExcept1on: since service that handles replies may not expect payload to be different

ethexe/contracts/src/IMirror.sol
```diff
@@ -87,7 +103,7 @@ interface IMirror {
 
     /* Primary Gear logic */
 
-    function sendMessage(bytes calldata payload, uint128 value) external returns (bytes32);
```
StackOverflowExcept1on, line 90: in that case shouldn't we update MirrorProxy?

StackOverflowExcept1on (approved):

ethexe/contracts/src/Mirror.sol
```diff
@@ -15,107 +15,126 @@ contract Mirror is IMirror {
...
+    /// @dev Address of the router contract, which is the sole authority.
     address public immutable router;
 
+    /// @dev Program's current state hash.
+    bytes32 public stateHash;
+
+    /// @dev Source for message ids unique generation.
+    ///      In-fact represents amount of messages received from Ethereum.
+    ///      Zeroed nonce is always represent init message.
+    uint256 public nonce;
```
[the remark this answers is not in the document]
breathx: exited packed with inheritor
isSmall is packed with initializer, seems it won't have more optimization

PR #4663, opened 2025-05-19
refactor(core): [p.3] remove ids extension traits [logic unchanged]

breathx:
This PR only moves hashing into gprimitives, removes gear-core/ids mod in favor of raw funcs in core, as well as implements ids querying everywhere where possible instead of pre-calculations - it will minimize PR with refactoring of the inner logic [refining of ids generation]

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6163,7 +6163,6 @@ dependencies = [
  "numerated",
  "parity-scale-codec",
  "paste",
- "primitive-types",
  "proptest",
  "rand 0.8.5",
  "scale-info",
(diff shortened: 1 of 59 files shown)


utils/runtime-fuzzer/src/generator/upload_program.rs
```diff
@@ -88,9 +88,6 @@ pub(crate) fn generate(
     log::trace!("Sending value (upload_program) - {value}");
     log::trace!("Current balance (upload_program - {current_balance}");
 
-    let program_id = ActorId::generate_from_user(CodeId::generate(&code), &salt);
-    log::trace!("Generated code for program id - {program_id}");
```
ark0f, line 92: I guess trace should be added back

ark0f (approved):

utils/runtime-fuzzer/src/generator/upload_program.rs
breathx: program id won't be precalculatable

PR #4666, opened 2025-05-25
refactor(pallet-gear) Removed all rent constants and types

Lazark0x:


pallets/gear-scheduler/src/migrations/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+pub mod v3_remove_program_pause_tasks;
```
breathx: this can be moved into lib.rs/{ mod migrations }

pallets/gear-scheduler/src/migrations/v3_remove_program_pause_tasks.rs
```diff
@@ -0,0 +1,251 @@
...
+        TryRuntimeError,
+    },
+    sp_std::vec::Vec,
+};
+
+const MIGRATE_FROM_VERSION: u16 = 2;
+const MIGRATE_TO_VERSION: u16 = 3;
+const ALLOWED_CURRENT_STORAGE_VERSION: u16 = 3;
+
+pub struct MigrateRemoveProgramPauseTasks<T: Config>(PhantomData<T>);
```
breathx, line 42: actually if we still iterate all over scheduler storage, it may make sense to migrate other idxs so we won't need to support legacy codec indexes for kept task variants

breathx (changes requested):
otherwise ltgm

PR #4667, opened 2025-05-25
refactor(pallet-gear-debug): remove legacy pallet-gear-debug 

Lazark0x:


breathx (changes requested):
* test that are not related to only pallet testing itself should be ported somewhere, more likely into pallet-gear
* pallet-gear-debug should be removed from contruct_runtime macro in runtime/vara/src/lib.rs and put a comment not to use this idx anymore
* migrations should only be applied with dev feature in runtime/vara/src/migrations.rs

pallets/gear/src/lib.rs
```diff
@@ -1763,6 +1743,29 @@ pub mod pallet {
...
             .into())
         }
+
+        /// A dummy extrinsic with programmatically set weight.
+        ///
+        /// Used in tests to exhaust block resources.
+        ///
+        /// Parameters:
+        /// - `fraction`: the fraction of the `max_extrinsic` the extrinsic will use.
+        #[pallet::call_index(9)]
```
breathx, line 1753:
```suggestion
        #[cfg(feature = "dev")]
        #[pallet::call_index(255)]
```

breathx (changes requested):
GearDebug may already be deleted without any migrations since the storages in prod deployments are empty

runtime/vara/src/lib.rs
```diff
@@ -23,7 +23,9 @@
 #![allow(clippy::legacy_numeric_constants)]
 #![allow(non_local_definitions)]
 
-// Make the WASM binary available.
+#[cfg(feature = "runtime-benchmarks")]
+#[macro_use]
+extern crate frame_benchmarking; // Make the WASM binary available.
```
breathx: newline

pallets/gear/src/tests.rs
```diff
@@ -16730,4 +17332,96 @@ pub(crate) mod utils {
             })
             .collect()
     }
+
```
breathx, line 17342: most of the code of this "snapshot" seems useless within gear pallet, so please rework the test logic - you only need to check pages as far as I see, any checks for program active status and queue may be done directly from gear pallet or even removed as redundant checks

breathx (changes requested):
Except tests porting - looks good 👍🏻

PR #4669, opened 2025-05-27
feat(ethexe-service): Reduced event usage in fast sync

ark0f:


ethexe/ethereum/src/router/mod.rs
```diff
@@ -370,6 +370,30 @@ impl RouterQuery {
             .map(|res| res.into_iter().map(|c| CodeId::new(c.0)).collect())
             .map_err(Into::into)
     }
+
+    pub async fn programs_count_at(&self, block: H256) -> Result<u64> {
```
breathx, line 374: Use usize here?

ethexe/service/src/fast_sync.rs
```diff
@@ -535,45 +578,26 @@ async fn sync_from_network(
...
+        let original_code = db
+            .original_code(code_id)
+            .expect("`sync_from_network` must fulfill database");
+        compute.process_code(CodeAndIdUnchecked {
+            code_id,
+            code: original_code,
+        });
     }
 
     while let Some(event) = compute.next().await {
```
breathx, line 614: What if compute events another event? It's gonna be skipped

ethexe/service/src/fast_sync.rs
```diff
@@ -98,81 +107,155 @@ impl EventData {
...
+        network,
+        db_sync::Request::program_ids(latest_committed_block, programs_count),
+    )
+    .await?;
+
+    let program_code_ids = response.unwrap_program_ids();
+    Ok(program_code_ids)
+}
+
+async fn collect_code_ids(
```
breathx, line 210: Some comments on each fn would be good. As far as I see this returns not just codes, but codes to be instrumented?

ethexe/ethereum/src/router/mod.rs
ark0f: It's always preferable to use specific integer type for counters rather than platform-dependent integer type

ethexe/service/src/fast_sync.rs
```diff
@@ -98,81 +107,155 @@ impl EventData {
...
+        let Some(latest_committed_batch) = latest_committed_batch else {
             return Ok(None);
         };
 
-        let latest_committed_batch = latest_committed_batch.ok_or_else(|| {
-            log::error!("Inconsistent block events: block commitment without batch commitment");
-            anyhow!("latest committed batch not found")
-        })?;
+        let Some(latest_committed_block) = latest_committed_block else {
+            anyhow::bail!("Inconsistent block events: batch commitment without block commitment");
```
grishasobol: Batch commitment can be without any block commitments

ethexe/service/src/fast_sync.rs
```diff
@@ -98,81 +107,155 @@ impl EventData {
...
+        .filter(|&code_id| db.code_valid(code_id).is_none())
+        .collect();
+
+    Ok(code_ids)
+}
+
+async fn collect_program_states(
+    observer: &mut ObserverService,
+    at: H256,
+    program_code_ids: &BTreeMap<ActorId, CodeId>,
```
grishasobol, line 240:
```suggestion
    program_code_ids: impl Iterator<Item = ActorId>,
```

ethexe/service/src/fast_sync.rs
```diff
@@ -98,81 +107,155 @@ impl EventData {
...
+    let response = net_fetch(
+        network,
+        db_sync::Request::valid_codes(latest_committed_block, codes_count),
+    )
+    .await?;
+
+    let code_ids = response
+        .unwrap_valid_codes()
+        .into_iter()
+        .filter(|&code_id| db.code_valid(code_id).is_none())
```
grishasobol, line 230: What if validation status exists but different from received?

ethexe/service/src/fast_sync.rs
```diff
@@ -98,81 +107,155 @@ impl EventData {
...
+        .await?;
+
+    let response = net_fetch(
+        network,
+        db_sync::Request::program_ids(latest_committed_block, programs_count),
+    )
+    .await?;
+
+    let program_code_ids = response.unwrap_program_ids();
+    Ok(program_code_ids)
```
grishasobol, line 206: Could you specify please where is a check that for each program we receive right corresponding code id?

grishasobol (changes requested):

ethexe/service/src/fast_sync.rs
ark0f: It means processor is not deterministic

ethexe/service/src/fast_sync.rs
ark0f: What's the point?

ethexe/service/src/fast_sync.rs
ark0f: `db_sync::ResponseHandler::handle_program_ids()`

ethexe/service/src/fast_sync.rs
grishasobol: not important

ethexe/service/src/fast_sync.rs
grishasobol: But you receives them from the network. That means that somebody share with use not valid codes also. At least a warning should be printed in that case

grishasobol (approved):

ethexe/service/src/fast_sync.rs
ark0f: They are validated via `db_sync::ResponseHandler::handle_valid_codes()`

breathx (approved):

PR #4670, opened 2025-05-27
refactor(ethexe-service): Implement `TestableEvent`

ark0f:


ethexe/service/src/tests/utils/events.rs
```diff
@@ -0,0 +1,247 @@
...
+            } => Self::OffchainTransaction {
+                transaction: transaction.clone(),
+            },
+        }
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+#[allow(dead_code)]
+pub(crate) enum TestableEvent {
```
breathx, line 98:
```suggestion
pub(crate) enum TestingEvent {
```

breathx: and everywhere

ethexe/service/src/lib.rs
```diff
@@ -352,18 +346,12 @@ impl Service {
...
             log::trace!("Primary service produced event, start handling: {event:?}");
 
-            // Broadcast event.
-            // Never supposed to be Some in production.
-            if let Some(sender) = sender.as_ref() {
-                sender
-                    .send(event.clone())
-                    .expect("failed to broadcast service event");
-            }
+            #[cfg(test)]
```
breathx, line 349: it wont work for integrational testing, please replace with some custom feature (passed as well for dev-deps)

ethexe/service/src/lib.rs
```diff
@@ -235,6 +229,7 @@ impl Service {
 
         let fast_sync = config.node.fast_sync;
 
+        #[allow(unreachable_code)]
```
breathx, line 232: directly on use place

breathx (approved):
approving to not force you to wait for re-review, but following comments are must-to-apply

ethexe/service/src/lib.rs
ark0f: We don't have such case yet, so why not leave it as is?

ethexe/service/src/lib.rs
ark0f: `unreachable!()` applies to the whole expression (e.g. `Ok(...)`), so it's impossible to move it elsewhere

ethexe/service/src/lib.rs
breathx: You moved tests from unit to integration ones

when you run `cargo test -p ethexe-service`, the code under `#[cfg(test)` isn't compiled for tests in ethexe/service/tests/mod.rs etc

PR #4677, opened 2025-05-30
refactor(ci): Separated node build and gsdk & gcli

Lazark0x:


breathx (changes requested):
it seems incorrect to hold all the linux CI for node build - these should run in parallel
gcli etc build may be grouped in one job, since provisioning runners for them are expensive operation. 

don't forget to label your PRs "pleasereview" to actually run CI so you'd see the difference in time

breathx:
#4709

PR #4678, opened 2025-05-31
feat(vara-runtime): Report offline validators to offence

ukint-vs:
Validators now can be punished for being offline while in the active set.

@gear-tech/dev


breathx:
Currently won't be implemented

PR #4705, opened 2025-06-03
refactor(ethexe): split db interfaces to read and write

grishasobol:


ethexe/common/src/primitives.rs
```diff
@@ -25,6 +25,8 @@ use gprimitives::{ActorId, MessageId, H256};
 use parity_scale_codec::{Decode, Encode};
 use sha3::Digest as _;
 
+pub type ProgramStatesMap = BTreeMap<ActorId, StateHashWithQueueSize>;
```
ark0f: What about just `ProgramStates`?

grishasobol: renamed

PR #4706, opened 2025-06-04
feat(`ethexe`): switch from anyhow to thiserror (blob-loader, compute, processor)

ecol-master:
Resolves # .

-
-
-

@reviewer-or-team


ethexe/processor/src/handling/events.rs
```diff
@@ -31,15 +31,17 @@ impl ProcessingHandler {
...
-                    self.db.original_code(code_id).is_some(),
-                    "db corrupted: missing code [OR] code existence wasn't checked on Eth"
-                );
-
-                ensure!(
-                    self.db.program_code_id(actor_id).is_none(),
-                    "db corrupted: unrecognized program [OR] program duplicates wasn't checked on Eth"
-                );
+                if self.db.original_code(code_id).is_none() {
+                    return Err(ProcessorError::DbCorrupted(String::from(
```
ark0f: Each `DbCorrupted` usage can be a new variant itself

ethexe/rpc/src/errors.rs
```diff
@@ -24,8 +24,8 @@ pub fn db(err: &'static str) -> ErrorObject<'static> {
     ErrorObject::owned(8000, "Database error", Some(err))
 }
 
-pub fn runtime(err: anyhow::Error) -> ErrorObject<'static> {
-    ErrorObject::owned(8000, "Runtime error", Some(format!("{err}")))
+pub fn runtime(err: String) -> ErrorObject<'static> {
```
ark0f:
```suggestion
pub fn runtime(err: impl ToString) -> ErrorObject<'static> {
```
To avoid manual conversion like in `program.rs`

ethexe/compute/src/lib.rs
```diff
@@ -47,6 +46,36 @@ pub enum ComputeEvent {
...
+    #[error("code({0}) marked as validated, but not found in db")]
+    ValidatedCodeNotFound(CodeId),
+    #[error("codes queue nоt found for computed block({0})")]
+    CodesQueueNotFound(H256),
+    #[error("commitment queue not found for computed block({0})")]
+    CommitmentQueueNotFound(H256),
+    #[error("previous commitment not found for computed block ({0})")]
+    PreviousCommitmentNotFound(H256),
+
+    // `Processor` errors
```
ark0f:
```suggestion
```
Processor already has its own error type

ethexe/processor/src/lib.rs
```diff
@@ -43,6 +42,68 @@ mod handling;
...
+    #[error("wasmtime error: {0}")]
+    Wasm(#[from] wasmtime::Error),
+
+    // parity-scale-codes
+    #[error("")]
+    ParityScaleCodes(#[from] parity_scale_codec::Error),
+
+    // sp-allocator
+    #[error("")]
+    SpAllocator(#[from] sp_allocator::Error),
```
ark0f: These comments can be removed because there will be only 1 variant for each crate

ethexe/processor/src/lib.rs
```diff
@@ -43,6 +42,68 @@ mod handling;
...
+    #[error("failed to write call input: {0}")]
+    CallInputWrite(String),
+    #[error("host state should be set before call and reset after")]
+    HostStateNotSet,
+    #[error("allocator should be set after `set_host_state`")]
+    AllocatorNotSet,
+
+    // `ProcessingHandler`
+    #[error("db corrupted: {0}")]
+    DbCorrupted(String),
```
ark0f: Consider to introduce error type for each mentioned structure

ethexe/blob-loader/src/local.rs
```diff
@@ -54,14 +53,12 @@ impl LocalBlobStorage {
     pub async fn get_code(self, code_id: CodeId) -> Result<BlobData> {
         let storage = self.inner.read().await;
 
-        let Some(code) = storage.get(&code_id).cloned() else {
-            return Err(anyhow!("code {code_id} not found in db"));
-        };
+        let code = storage.get(&code_id).cloned().unwrap();
```
grishasobol: why unwrap?

ethexe/compute/src/lib.rs
```diff
@@ -312,11 +340,11 @@ impl<DB: OnChainStorage + BlockMetaStorage> ChainHeadProcessContext<DB> {
         // Propagate prev commitment (prev not empty block hash or zero for genesis).
         if db
             .block_outcome_is_empty(parent)
-            .ok_or_else(|| anyhow!("emptiness not found for computed block {parent}"))?
+            .ok_or(ComputeError::BlockOutcomeNotFound(block))?
```
grishasobol: parent not found

ethexe/processor/src/handling/events.rs
grishasobol: What the reason to use String inside DbCorrupted instead of 'static str

ark0f (approved):

ethexe/blob-loader/src/lib.rs
```diff
@@ -125,23 +157,23 @@ impl ConsensusLayerBlobReader {
...
         let block = self
             .provider
             .get_block_by_hash(block_hash)
             .await?
-            .ok_or_else(|| anyhow!("failed to get block"))?;
+            .ok_or(BlobLoaderError::BlockNotFound(H256(block_hash.0)))?;
         let slot = (block.header.timestamp - BEACON_GENESIS_BLOCK_TIME)
             / self.config.beacon_block_time.as_secs();
-        let blob_bundle_result = match attempts {
+        let blob_bundle = match attempts {
```
grishasobol: Consider to change this match to `let attempts = attempts.unwrap_or(0);`

ethexe/compute/src/lib.rs
```diff
@@ -82,16 +111,17 @@ impl Stream for ComputeService {
...
-                    {
-                        waiting_codes.remove(&commitment.id);
+                Ok(res) => {
+                    if let Ok(commitment) = &res {
+                        if let BlockPreparationState::WaitForCodes { waiting_codes, .. } =
+                            &mut self.state
+                        {
+                            waiting_codes.remove(&commitment.id);
+                        }
                     }
```
grishasobol:
```suggestion
if let (Ok(commitment), BlockPreparationState::WaitForCodes { waiting_codes, .. }) = (&res, &mut self.state) {
    waiting_codes.remove(&commitment.id);
}
```

grishasobol (commented):
why do we use 1.0.62 thiserror version? https://crates.io/crates/thiserror

ethexe/compute/src/lib.rs
kuzminxyz: replace with attribute macro `#[error(transparent)]` for processor error

PR #4707, opened 2025-06-05
fix(ethexe): blob loader and producer codes processing

grishasobol:


ethexe/blob-loader/src/lib.rs
```diff
@@ -254,25 +259,33 @@ impl BlobLoaderService for BlobLoader {
...
                 continue;
             }
 
-            let code_info = self
+            let CodeBlobInfo { timestamp, tx_hash } = self
                 .db
                 .code_blob_info(code_id)
                 .ok_or(anyhow!("not found code info for {code_id} in db"))?;
 
+            if let Some(code) = self.db.original_code(code_id) {
```
grishasobol, line 271: Optimisation - if original code in db - takes it from db, but still generates output event because maybe it's needed for outer services

ethexe/compute/src/lib.rs
```diff
@@ -42,7 +41,7 @@ pub struct BlockProcessed {
 #[derive(Debug, Clone, Eq, PartialEq, derive_more::Unwrap)]
 pub enum ComputeEvent {
     RequestLoadCodes(HashSet<CodeId>),
-    CodeProcessed(CodeCommitment),
+    CodeProcessed(CodeId),
```
grishasobol, line 44: CodeCommitment is useless to return. Just return information what code has been processed

ethexe/compute/src/lib.rs
```diff
@@ -172,15 +171,30 @@ impl ComputeService {
         }
     }
 
-    pub fn process_code(&mut self, code_id: CodeId, timestamp: u64, code: Vec<u8>) {
+    pub fn process_code(&mut self, code_id: CodeId, _timestamp: u64, code: Vec<u8>) {
+        if let Some(valid) = self.db.code_valid(code_id) {
```
grishasobol, line 175: Optimisation - if code is already processed, then skip processing and return CodeProcessed event

ethexe/service/src/tests/utils/env.rs
```diff
@@ -342,7 +342,6 @@ impl TestEnv {
             })
             .unzip();
 
-        self.blobs_storage.change_db(db.clone());
```
grishasobol, line 345: I do not know why we ever need this

ethexe/blob-loader/src/lib.rs
kuzminxyz: I think this optimization is useless, because of if original code exists in db Compute service won't request it to load

grishasobol: As discussed in dm: this case could appear if reorg appear on ethereum

ecol-master (approved):

PR #4738, opened 2025-06-19
chore: Use `cargo-shear` to detect unused deps

ark0f:


examples/async-signal-entry/Cargo.toml
```diff
@@ -15,7 +15,6 @@ parity-scale-codec.workspace = true
 gear-wasm-builder.workspace = true
 
 [dev-dependencies]
```
breathx: empty [dep] to be removed

PR #4745, opened 2025-06-21
feat(ethexe-observer): Store validator set in db

ukint-vs:
Avoid redundant rpc calls to the Router each block and cache validator set to DB.

@reviewer-or-team

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4716,7 +4716,6 @@ dependencies = [
  "ethexe-common",
  "ethexe-db",
  "ethexe-ethereum",
- "ethexe-observer",
  "ethexe-service-utils",
  "ethexe-signer",
  "futures",
--- a/ethexe/consensus/Cargo.toml
+++ b/ethexe/consensus/Cargo.toml
@@ -8,7 +8,6 @@ homepage.workspace = true
 repository.workspace = true
 
 [dependencies]
-ethexe-observer.workspace = true
 ethexe-db.workspace = true
 ethexe-service-utils.workspace = true
 ethexe-signer.workspace = true


ethexe/observer/src/lib.rs
```diff
@@ -79,13 +79,28 @@ impl fmt::Debug for ObserverEvent {
     }
 }
 
+#[allow(dead_code)]
+#[derive(Clone, Debug)]
+struct Timelines {
+    era_duration: u64,
+    election_duration: u64,
+    validation_delay: u64,
+}
```
kuzminxyz: maybe should move to `ethexe_common::gear`? we already define some structure from Gear.sol lib here

ethexe/observer/src/lib.rs
```diff
@@ -79,13 +79,28 @@ impl fmt::Debug for ObserverEvent {
...
+#[derive(Clone, Debug)]
+struct RouterConfig {
+    wvara_address: Address,
+    genesis_block: BlockHeader,
+    timelines: Timelines,
+}
+
 #[derive(Clone, Debug)]
 struct RuntimeConfig {
     router_address: Address,
```
kuzminxyz: maybe should also move to `RouterConfig` ?

ethexe/observer/src/lib.rs
```diff
@@ -183,33 +198,47 @@ impl ObserverService {
...
+                era_duration: timelines.era.to::<u64>(),
+                election_duration: timelines.election.to::<u64>(),
+                validation_delay: timelines.validationDelay.to::<u64>(),
+            },
+        };
+
         let config = RuntimeConfig {
             router_address: *router_address,
-            wvara_address,
+            // wvara_address,
```
kuzminxyz: deadcode

ethexe/common/src/db.rs
```diff
@@ -77,6 +78,7 @@ pub trait OnChainStorageRead {
     fn code_blob_info(&self, code_id: CodeId) -> Option<CodeBlobInfo>;
     fn block_is_synced(&self, block_hash: H256) -> bool;
     fn latest_synced_block_height(&self) -> Option<u32>;
+    fn validator_set_at(&self, next_era_start: u64) -> Option<Vec<Address>>;
```
kuzminxyz: we can not store validators by the era index because of can be reorg in eth

ecol-master:
Add:
1. validator set propagation from parent to child block
2. appropriate methods for `OnchainStorageRead` and `OnchainStorageWrite` traits

@grishasobol  what do you think about removing `BlockSyncedData` struct. Now it stored synced block hash and validator set:
```
pub struct BlockSyncedData {
    pub block_hash: H256,
    pub validators: Vec<Address>,
}
```

But now we don't need to sync validator set every block because of it propagates from parent to children and update if node receive appropriate event. So i think this struct can be removed, because of in any place we can get validator set for any block.

ethexe/compute/src/prepare.rs
```diff
@@ -148,6 +167,7 @@ mod tests {
         let initial_digest = Digest([42; 32]);
         db.set_last_committed_batch(parent_hash, initial_digest);
         db.set_block_codes_queue(parent_hash, VecDeque::new());
+        db.set_validator_set(parent_hash, vec![]);
```
breathx: shouldn't we forbid empty validator set? it's incorrect data

ethexe/observer/src/sync.rs
```diff
@@ -65,16 +65,23 @@ impl<DB: SyncDB> ChainSync<DB> {
 
         self.mark_chain_as_synced(chain.into_iter().rev());
 
-        let validators =
-            RouterQuery::from_provider(self.config.router_address.0.into(), self.provider.clone())
-                .validators_at(block)
-                .await?;
+        let validators = if let Some(validators) = self.db.validator_set(block) {
```
breathx: if we set default in the very beginning - this won't ever work, since there're always key in db?

ethexe/observer/src/sync.rs
```diff
@@ -65,16 +65,23 @@ impl<DB: SyncDB> ChainSync<DB> {
...
+            validators
+        } else {
+            let validators = RouterQuery::from_provider(
+                self.config.router_address.0.into(),
+                self.provider.clone(),
+            )
+            .validators_at(block)
+            .await?;
+            self.db.set_validator_set(block, validators.clone());
```
breathx: I'd rename getter/setter
```suggestion
            self.db.set_validators(block, validators.clone()); // self.db.validators(block)
```

ethexe/observer/src/sync.rs
```diff
@@ -118,6 +125,20 @@ impl<DB: SyncDB> ChainSync<DB> {
...
+                    log::trace!(
+                        "NextEraValidatorsCommitted event detected. Setting a new validator set."
+                    );
+                    let validators = RouterQuery::from_provider(
+                        self.config.router_address.0.into(),
+                        self.provider.clone(),
+                    )
+                    .validators()
+                    .await?;
+                    self.db.set_validator_set(block, validators);
```
breathx: that's incorrect. you shoud set it for future block, not for now?

ethexe/contracts/src/Router.sol
```diff
@@ -224,6 +224,10 @@ contract Router is IRouter, OwnableUpgradeable, ReentrancyGuardTransientUpgradea
         return _router().protocolData.validatedCodesCount;
     }
 
+    function timelines() public view returns (Gear.Timelines memory) {
```
breathx, line 227: this api is unused?

kuzminxyz: Timelines will be used later. Info about era duration will be needed in rewards implementation.

ethexe/observer/src/lib.rs
```diff
@@ -274,9 +283,12 @@ impl ObserverService {
         db.set_block_program_states(genesis_block_hash, Default::default());
         db.set_block_schedule(genesis_block_hash, Default::default());
         db.set_block_outcome(genesis_block_hash, Default::default());
-        db.set_latest_computed_block(genesis_block_hash, genesis_header);
+        db.set_latest_computed_block(genesis_block_hash, genesis_header.clone());
+        // TODO: think about validators set for genesis block
+        // Maybe should be fill with the value from config
+        db.set_validators(genesis_block_hash, Default::default());
```
grishasobol: genesis block is a block of router creation, so validators can be already accessed from rpc at genesis block

ethexe/observer/src/sync.rs
```diff
@@ -186,4 +195,20 @@ impl<DB: SyncDB> ChainSync<DB> {
             self.db.set_latest_synced_block_height(block_header.height);
         }
     }
+
+    fn first_era_block(&self, chain_head: &BlockHeader) -> Result<bool> {
```
grishasobol: first_era_block sounds like - "block of the first era"
better to rename to "era_first_block"

ethexe/observer/src/sync.rs
```diff
@@ -186,4 +195,20 @@ impl<DB: SyncDB> ChainSync<DB> {
...
+        let chain_head_era = self.block_era_index(chain_head.timestamp);
+
+        let parent = self.db.block_header(chain_head.parent_hash).ok_or(anyhow!(
+            "header not found for block({:?})",
+            chain_head.parent_hash
+        ))?;
+
+        let parent_era_index = self.block_era_index(parent.timestamp);
```
grishasobol, line 218: maybe this is a little twisted, but this would not work for genesis era because era index is negative for parent of genesis block

anyway maybe better to use signed types for logic here

ethexe/observer/src/lib.rs
kuzminxyz: done in the next PR

ethexe/observer/src/sync.rs
kuzminxyz: yes, now i am setting new validators in the beginning of the new era

Cargo.toml
```diff
@@ -504,7 +504,7 @@ keyring = "1.2.1"                                            # gcli
 libp2p = "=0.51.4"                                           # gcli (same version as sc-consensus)
 mimalloc = { version = "0.1.46", default-features = false }  # node/cli
 nacl = "0.5.3"                                               # gcli
-nonempty = "0.8.1"                                           # utils/utils
+nonempty = {version = "0.12.0", default-features = false}     # utils/utils
```
breathx:
```suggestion
nonempty = { version = "0.12.0", default-features = false }     # utils/utils
```

ethexe/common/src/gear.rs
```diff
@@ -236,6 +236,13 @@ impl ToDigest for BatchCommitment {
     }
 }
 
+#[derive(Clone, Debug, Default, Encode, Decode, PartialEq, Eq)]
```
breathx, line 239: impl ethereum/abi/gear.rs From

ethexe/ethereum/src/router/mod.rs
```diff
@@ -371,6 +371,19 @@ impl RouterQuery {
             .map_err(Into::into)
     }
 
+    pub async fn timelines(&self) -> Result<Timelines> {
+        self.instance
+            .timelines()
+            .call()
+            .await
+            .map(|res| Timelines {
+                era: res.era.to::<u64>(),
```
breathx: here from will be used

ethexe/observer/src/sync.rs
```diff
@@ -186,4 +200,26 @@ impl<DB: SyncDB> ChainSync<DB> {
...
+            return Ok(false);
+        }
+
+        let parent = self.db.block_header(chain_head.parent_hash).ok_or(anyhow!(
+            "header not found for block({:?})",
+            chain_head.parent_hash
+        ))?;
+
+        let parent_era_index = self.block_era_index(parent.timestamp);
+        Ok(chain_head_era == parent_era_index + 1)
```
breathx: chain head may be more than in 1 era if eth blocks were not built, so I'd put only ">" check here

breathx (changes requested):

ethexe/compute/src/prepare.rs
```diff
@@ -114,6 +125,14 @@ fn propagate_data_from_parent<
     // Propagate last committed batch
     db.set_last_committed_batch(block, last_committed_batch);
 
+    // Propagate last validator set if it is not set in the `ObserverService`
```
grishasobol: this should be done right in observer, because it's bad idea to use OnChainStorageWrite outside observer. It's like double mutable borowinge

grishasobol: Can be case when observer do something like this
```
let mut something = db.get_something();
something.change();
db.set_something(something);
```

right in the same time other service (compute in your case) do
```
db.set_something(some);
```

This could cause unexpectable errors.

I will make db accesses more secure later, to protect data in db from cases like that.

ethexe/observer/src/sync.rs
grishasobol: Good point

ethexe/consensus/src/validator/coordinator.rs
```diff
@@ -77,7 +77,7 @@ impl StateHandler for Coordinator {
 impl Coordinator {
     pub fn create(
         mut ctx: ValidatorContext,
-        validators: Vec<Address>,
+        validators: nonempty::NonEmpty<Address>,
```
grishasobol:
```suggestion
        validators: NonEmpty<Address>,
```

grishasobol: and in all other places

ethexe/consensus/src/validator/coordinator.rs
```diff
@@ -165,7 +171,9 @@ mod tests {
     fn process_validation_reply() {
         let (mut ctx, keys) = mock_validator_context();
         ctx.signatures_threshold = 3;
-        let validators: Vec<_> = keys.iter().take(3).map(|k| k.to_address()).collect();
+        let validators =
+            nonempty::NonEmpty::from_vec(keys.iter().take(3).map(|k| k.to_address()).collect())
```
grishasobol: NonEmpty does not implements FromIterator ?

ethexe/consensus/src/validator/producer.rs
```diff
@@ -324,7 +323,7 @@ mod tests {
     #[tokio::test]
     async fn code_commitments_only() {
         let (ctx, keys) = mock_validator_context();
-        let validators = vec![ctx.pub_key.to_address(), keys[0].to_address()];
+        let validators = nonempty::nonempty![ctx.pub_key.to_address(), keys[0].to_address()];
```
grishasobol:
```suggestion
        let validators = nonempty![ctx.pub_key.to_address(), keys[0].to_address()];
```
and in all other places

grishasobol (approved):

ethexe/consensus/src/validator/coordinator.rs
kuzminxyz: yes, it is can not be impl because of empty iterator

ecol-master (approved):

breathx (approved):

PR #4728, opened 2025-06-18
chore(gear-core): Use `thiserror` to derive `core::error::Error` trait

vobradovich:
Resolves #4724 

fixed
- gear-core-errors
- gcore
- gprimitives
- gstd
- gtest

PR #4750, opened 2025-06-22
refactor: remove useless trait derives

breathx:
after #4728

### DIFF
--- a/core-processor/src/executor.rs
+++ b/core-processor/src/executor.rs
@@ -294,10 +294,23 @@ where
         Default::default(),
     );
 
+    let incoming_dispatch = IncomingDispatch::new(
+        Default::default(),
+        IncomingMessage::new(
+            Default::default(),
+            Default::default(),
+            Default::default(),
+            Default::default(),
+            Default::default(),
+            Default::default(),
+        ),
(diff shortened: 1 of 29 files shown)


core-processor/src/executor.rs
```diff
@@ -294,10 +294,23 @@ where
...
+        IncomingMessage::new(
+            Default::default(),
+            Default::default(),
+            Default::default(),
+            Default::default(),
+            Default::default(),
+            Default::default(),
+        ),
+        Default::default(),
+    );
```
vobradovich, line 292: looks like `IncomingDispatch` should impl `Default`

breathx: nope - it's conceptually incorrect setting some defaults for addresses and message ids
the thing is that these tests will soon migrate to gear-core, where incoming dispatch implements default under test cfg

vobradovich: redundand code
if not Default and for test create new ctor `IncomingMessage::mock()`

breathx: done

vobradovich (approved):

PR #4769, opened 2025-07-14
refactoring(`blob-loader`): use ready macro

ecol-master:
.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ log/
 weight-dumps/
 .binpath
 /.vscode
+/.cursor
 .DS_Store
 .idea
 .log


ethexe/blob-loader/src/lib.rs
```diff
@@ -212,17 +212,12 @@ impl<DB: Database> Stream for BlobLoader<DB> {
...
-            Poll::Ready(Some(result)) => match result {
-                Ok(code_and_id) => {
-                    self.codes_loading.remove(&code_and_id.code_id);
-                    Poll::Ready(Some(Ok(BlobLoaderEvent::BlobLoaded(code_and_id))))
-                }
-                Err(e) => Poll::Ready(Some(Err(e))),
-            },
-            _ => Poll::Pending,
-        }
+        let res = futures::ready!(self.futures.poll_next_unpin(cx)).map(|result| {
```
breathx, line 215: new approach seems missing case while self.futures returned `Poll::Ready(None)`. Previously it was returning Pending, while now returns while BlobLoader stream returns `Poll::Ready(None)` meaning that there'll be no further elements of stream afterword => terminating BlobLoader

grishasobol: Looks like it's the same like was before:
```
macro_rules! ready {
    ($e:expr $(,)?) => {
        match $e {
            $crate::task::Poll::Ready(t) => t,
            $crate::task::Poll::Pending => return $crate::task::Poll::Pending,
        }
    };
}
```

grishasobol (approved):

ethexe/blob-loader/src/lib.rs
grishasobol: Ah see now. Yes looks like ready(none) is not handled properly now

kuzminxyz: yeah, sorry, i forget that tokio::select! will think that service end it's work. I return the handling of None case

PR #4772, opened 2025-07-17
feat(ethexe-consensus): optimize squash_chain_commitments

### LINKED ISSUE #4744: ethexe: improve squashing - removing redundant state transitions
In file ethexe/consensus/src/utils.rs
https://github.com/gear-tech/gear/blob/3efb2265b2725b7059ff18255d8d699e000f522a/ethexe/consensus/src/utils.rs#L288

ukint-vs:
Resolves #4744.

Implemented de-duplication logic that keeps only the latest transition for each actor.
Preserved original order of transitions
 
@gear-tech/dev

### DIFF
--- a/ethexe/consensus/src/utils.rs
+++ b/ethexe/consensus/src/utils.rs
@@ -285,7 +285,6 @@ pub fn create_batch_commitment<DB: BlockMetaStorageRead>(
     }))
 }
 
-// TODO #4744: improve squashing - removing redundant state transitions
 pub fn squash_chain_commitments(
     chain_commitments: Vec<ChainCommitment>,
 ) -> Option<ChainCommitment> {
@@ -293,16 +292,42 @@ pub fn squash_chain_commitments(
         return None;
     }
 
-    let mut transitions = Vec::new();
     let mut gear_blocks = Vec::new();
 
-    for commitment in chain_commitments {
-        transitions.extend(commitment.transitions);
-        gear_blocks.extend(commitment.gear_blocks);
+    // Use a map to track the latest transition for each actor
+    let mut latest_transitions = std::collections::HashMap::new();
+
+    // Process transitions in order to preserve the final state for each actor
+    for commitment in &chain_commitments {
+        for transition in &commitment.transitions {
+            // Always keep the latest transition for each actor_id
+            latest_transitions.insert(transition.actor_id, transition);
+        }
+        gear_blocks.extend(commitment.gear_blocks.clone());
     }
 
+    // Collect the latest transitions for each actor, preserving original order
+    // by processing the flattened transitions and keeping only the latest for each actor
+    let mut seen_actors = std::collections::HashSet::new();
+    let mut deduped_transitions = Vec::new();
+
+    // Process in reverse to get latest occurrences first, then reverse to maintain order
+    let all_transitions: Vec<_> = chain_commitments
+        .into_iter()
+        .flat_map(|c| c.transitions)
+        .collect();
+
+    for transition in all_transitions.into_iter().rev() {
+        if seen_actors.insert(transition.actor_id) {
+            deduped_transitions.push(transition);
+        }
+    }
+
+    // Reverse to restore original order (latest for each actor)
+    deduped_transitions.reverse();
+
     Some(ChainCommitment {
-        transitions,
+        transitions: deduped_transitions,
         gear_blocks,
     })
 }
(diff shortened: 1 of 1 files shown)


grishasobol (commented):
do you think equal transitions is not super rare case? I thought that optimisation is something like 
program transitions: H1 -> H2, H2 -> H3, H3 -> H4
optimized: H1 -> H4

ethexe/consensus/src/validator/participant.rs
```diff
@@ -250,10 +251,35 @@ mod tests {
...
+        let chain_commitment =
+            utils::aggregate_chain_commitment(&ctx.core.db, batch.block_hash, false, None)
+                .unwrap()
+                .map(|(commitment, _)| commitment);
+        let code_commitments = utils::aggregate_code_commitments(
+            &ctx.core.db,
+            batch.code_commitments.clone().into_iter().map(|c| c.id),
+            false,
+        )
+        .unwrap();
```
breathx: remove all unwraps

ethexe/consensus/src/mock.rs
```diff
@@ -234,12 +234,15 @@ pub fn prepared_mock_batch_commitment(db: &Database) -> BatchCommitment {
         From::from([code_commitment1.id, code_commitment2.id]),
     );
 
+    let mut transitions = [cc1.transitions, cc2.transitions].concat();
+    transitions.sort_by_key(|t| t.actor_id);
```
breathx: why sort?

ethexe/consensus/src/utils.rs
```diff
@@ -263,9 +262,65 @@ pub fn aggregate_chain_commitment<DB: BlockMetaStorageRead + OnChainStorageRead>
             .parent_hash;
     }
 
+    // Collect all transitions in true chronological order (oldest -> newest)
+    let all_transitions: Vec<StateTransition> = transitions.into_iter().rev().flatten().collect();
```
breathx: why aren't they from the very beginning in correct oldest -> newest order?

ethexe/consensus/src/utils.rs
```diff
@@ -263,9 +262,65 @@ pub fn aggregate_chain_commitment<DB: BlockMetaStorageRead + OnChainStorageRead>
...
+
+    // Group transitions by actor_id and squash consecutive transitions for the same actor
+    let mut actor_transitions: std::collections::BTreeMap<ActorId, Vec<StateTransition>> =
+        std::collections::BTreeMap::new();
+    for transition in all_transitions {
```
breathx: you may avoid at least one allocation for vec of all_transitions var if you will be inserting the values in map in the same time

breathx: or even mutate in place (instead of collecting vec of transitions belonging to specific actor), so you will only have vec -> map -> into_vec()

ethexe/consensus/src/utils.rs
```diff
@@ -263,9 +262,65 @@ pub fn aggregate_chain_commitment<DB: BlockMetaStorageRead + OnChainStorageRead>
...
+    }
+
+    let mut squashed_transitions = Vec::new();
+    for transitions_for_actor in actor_transitions.into_values() {
+        if transitions_for_actor.is_empty() {
+            continue;
+        }
+
+        // Pick the newest transition
+        let mut squashed = transitions_for_actor.last().unwrap().clone();
```
breathx: I'd add in debug assert check for choronological order:

A -> B -> C -> D etc

ethexe/consensus/src/utils.rs
```diff
@@ -263,9 +262,65 @@ pub fn aggregate_chain_commitment<DB: BlockMetaStorageRead + OnChainStorageRead>
...
+                }
+            }
+
+            if squashed.exited
+                && let Some(inheritor) = exit_inheritor
+            {
+                squashed.inheritor = inheritor;
+            }
+
+            squashed.messages = all_messages;
```
breathx: you may just create a new value instead of some mutations where it's possible to make a mistake in future extensions

breathx: e.g. forget about new field in transition

ethexe/consensus/src/utils.rs
```diff
@@ -497,6 +552,107 @@ mod tests {
...
+
+        let squashed = &commitment.transitions[0];
+        assert_eq!(squashed.actor_id, actor_a);
+
+        // Final state should be the newest state from the head-side transition: a2
+        assert_eq!(squashed.new_state_hash, H256::from([2; 32]));
+
+        // Messages should be accumulated: [msg1, msg2]
+        assert_eq!(squashed.messages.len(), 2);
+        assert!(squashed.messages.contains(&msg1));
```
breathx: you actually don't have any order assertions here.

I'd suggest to go with 1) easier naming: why block2 sends message1 etc; 2) provide full exact assertions

breathx: so you may have just 1, but much more useful test instead of 4-5 of them

breathx:
I'm also wondering why this action is performed only by participants, but not by block producer? he may want to specify if they're (validators) gonna have squashed commitment for this time etc??

PR #4773, opened 2025-07-18
fix(ci): install Rust stable toolchain as default

breathx:


.github/actions/install-rust/action.yml
```diff
@@ -10,4 +10,13 @@ runs:
...
+
     - uses: dsherret/rust-toolchain-file@v1
+
+    - run: rustup default stable
+      if: ${{ runner.os != 'Windows' }}
+      shell: bash
+
+    - run: rustup default stable
+      if: ${{ runner.os == 'Windows' }}
+      shell: powershell
```
ark0f:
```suggestion
    - run: rustup default stable
      shell: bash
```

ark0f (approved):

PR #4774, opened 2025-07-18
feat(`ethexe`); restriction for empty validator set

ecol-master:
Resolves # .

@grishasobol  @StackOverflowExcept1on

### DIFF
--- a/ethexe/db/src/database.rs
+++ b/ethexe/db/src/database.rs
@@ -703,6 +703,7 @@ impl OnChainStorageWrite for Database {
     }
 
     fn set_validators(&self, block_hash: H256, validator_set: Vec<Address>) {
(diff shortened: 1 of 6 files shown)


ethexe/db/src/database.rs
```diff
@@ -703,6 +703,7 @@ impl OnChainStorageWrite for Database {
     }
 
     fn set_validators(&self, block_hash: H256, validator_set: Vec<Address>) {
+        assert!(!validator_set.is_empty(), "validator set can not be empty");
```
grishasobol, line 706: consider to use https://docs.rs/vec1/latest/vec1/

StackOverflowExcept1on: or https://docs.rs/nonempty/latest/nonempty (since we're already using it in the code base)

PR #4784, opened 2025-07-23
feat(gear-ss58): make crate `#![no_std]`

StackOverflowExcept1on:


utils/ss58/src/lib.rs
```diff
@@ -167,8 +167,7 @@ impl fmt::Display for Error {
     }
 }
 
-#[cfg(feature = "std")]
-impl std::error::Error for Error {}
+impl core::error::Error for Error {}
```
breathx, line 170: could be derived?

utils/ss58/src/lib.rs
```diff
@@ -167,8 +167,7 @@ impl fmt::Display for Error {
     }
```
breathx, line 167: mb import derive_more here?

utils/ss58/src/lib.rs
StackOverflowExcept1on: why do we need derive_more/thiserror dependency just to implement Error?

PR #4787, opened 2025-07-23
feat(ethexe): Implement database visitor & verifier

### LINKED ISSUE #4624: ethexe: Implement DB integrity verifier
By @breathx:

A utility could be developed, similar to a `ScheduleRestorer` (implemented in #4539) which, with a single function call, performs the following tasks:

- Iterates over the storage.
- Verifies that all keys are present.
- Ensures that no hashes are missing.

The purpose of this verifier would be to check the validity of the database, specifically focusing on the data's integrity.

It would simplify and enhance the network test by making it more comprehensive. For example, after synchronizing, the `.check_db_validity()` method could be called, or alternatively, `.check_db_validity_at(eth_block_hash)` could be used to verify that the data is complete.

Additionally, checks for other elements, such as `schedule_set` or `program_states_set`, should be incorporated.

---

From me:

Such verifier can be implemented via visitor pattern to traverse structures and their hashes

ark0f:
Closes #4624


ethexe/db/src/verifier.rs
```diff
@@ -0,0 +1,723 @@
...
+            self.errors
+                .push(IntegrityVerifierError::InvalidCodeLenInMetadata {
+                    metadata_len: code_metadata.original_code_len(),
+                    original_len: original_code.len() as u32,
+                });
+        }
+    }
+
+    fn visit_code_valid(&mut self, _code_id: CodeId, code_valid: bool) {
+        if !code_valid {
```
grishasobol: Why code status can not be invalid? In tests we do not check this case but this does not mean that it's impossible

ark0f: Traversal starts with chain so it should be visited only for correct codes

ecol-master:
u can link issue #4624

ethexe/db/src/iterator.rs
```diff
@@ -0,0 +1,1357 @@
...
+    Error(DatabaseIteratorError),
+}
+
+impl From<MemoryPagesNode> for Node {
+    fn from(value: MemoryPagesNode) -> Self {
+        Self::MemoryPages(Box::new(value))
+    }
+}
+
+impl Node {
```
grishasobol: is not possible to use enum-as-inner ?

grishasobol (approved):

ethexe/db/src/verifier.rs
```diff
@@ -0,0 +1,706 @@
...
+    }
+
+    fn visit_block_meta(&mut self, _block: H256, meta: BlockMeta) {
+        if !meta.synced {
+            self.errors.push(IntegrityVerifierError::BlockIsNotSynced);
+        }
+        if !meta.prepared {
+            self.errors.push(IntegrityVerifierError::BlockIsNotPrepared);
+        }
+        if !meta.computed {
```
breathx, line 146: latest block may not be computed etc

PR #4788, opened 2025-07-23
feat(ethexe): commit only head of blocks

grishasobol:
instead of committing a vector of GearBlock, we commit now only hash of the last block in chain.
This should be considered as temporary solution and head have to be changed from block hash to announce hash, I will do in the next PR.


ethexe/common/Cargo.toml
```diff
@@ -28,4 +29,9 @@ std = [
     "parity-scale-codec/std",
     "sha3/std",
     "k256/std",
+    "primitive-types",
+    "primitive-types/std",
```
breathx: it enables dependency by default 
```suggestion
    "primitive-types/std",
```

ethexe/common/Cargo.toml
```diff
@@ -28,4 +29,9 @@ std = [
...
 ]
+
+[package.metadata.cargo-shear]
```
breathx: why

ethexe/common/src/events/router.rs
```diff
@@ -25,7 +25,7 @@ pub enum Event {
     BatchCommitted {
         digest: Digest,
     },
-    GearBlockCommitted(GearBlock),
+    HeadCommitted(H256),
```
breathx, line 28: I'd consider renaming. By the moment of its submission on eth it won't be head in any manner: locally and on eth

ethexe/common/src/primitives.rs
```diff
@@ -76,6 +76,8 @@ pub struct BlockMeta {
     pub synced: bool,
     pub prepared: bool,
     pub computed: bool,
+    pub last_committed_batch: Option<Digest>,
```
breathx, line 79: mb impl some struct Commitment { digest, block_hash } 
it may be useful here and in testing

breathx (approved):

ethexe/common/Cargo.toml
grishasobol: I noticed that "cargo test -p ethexe-common" does not work on my branch (as well as on master). The problem is that we haven't "std" feature in "gprimitives", so feature "std" cannot be forwarded to "primitive-types" thru "gprimitives". So, I use this hack to set "std" feature and be able to use "H256::random()" when std

grishasobol: but because primive-types is not used in rust code, cargo sheap consider it as useless and throw error

ethexe/common/src/events/router.rs
grishasobol: Actually is a head of committed blocks. In next PR it will be head of committed announces. Do you mean it's not head of ethereum block-chain ?

ethexe/common/src/primitives.rs
grishasobol: Do you mean join last_committed_batch and last_committed_head in one? Because if so - it's not correct, since batch commitment can be without "chain-commitment".

If you mean that block_hash is hash of corresponding to batch ethereum block, than yes is possible, but we can also store in DB full commitment by Digest (digest is uniq, since each commitment references to previous)

ethexe/common/Cargo.toml
breathx:
```toml
[dev-dependencies]
gprimitives = { workspace = true, features = ["std"] }
```
fixes it

ethexe/common/src/events/router.rs
breathx: yeah, it misleads to meaning eth head

ethexe/common/Cargo.toml
grishasobol: package `ethexe-common` depends on `gprimitives` with feature `std` but `gprimitives` does not have that feature.

ethexe/common/Cargo.toml
grishasobol: removed

ethexe/common/src/events/router.rs
grishasobol: It will be renamed in next PR to "head_announce"

ethexe/common/Cargo.toml
grishasobol: std feature added

PR #4796, opened 2025-07-28
fix(benchmarks): Fix `instantiate_module_type_section_per_kb` benchmark

ByteNacked:
Resolves # .

- Fix `instantiate_module_type_section_per_kb` benchmark

### DIFF
--- a/pallets/gear/src/benchmarking/code.rs
+++ b/pallets/gear/src/benchmarking/code.rs
@@ -283,8 +283,8 @@ where
 
         // Add dummy type section
         if let Some(types) = def.types {
-            for _ in 0..types.num_elements {
-                program.push_type(FuncType::new([ValType::I64; 6], [ValType::I64; 1]));
+            for proto in generate_uniq_prototypes(types.num_elements as usize).into_iter() {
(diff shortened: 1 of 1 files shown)


pallets/gear/src/benchmarking/code.rs
```diff
@@ -784,3 +784,51 @@ impl BitWidth {
...
 }
+
+// Generate `number` unique WASM function prototypes
+fn generate_uniq_prototypes(number: usize) -> Vec<FuncType> {
+    const ALPHABET: [ValType; 5] = [
+        ValType::I32,
+        ValType::I64,
+        ValType::F32,
+        ValType::F64,
+        ValType::V128,
```
StackOverflowExcept1on, line 797: f32, f64, v128 are not supported in our contracts

ByteNacked: You right, but this is done to increase number of possible shortest permutation when generating unique types.
So I think it's OK because it's only used for generating dummy type sections.
 
 Gonna add a comment though.

breathx (approved):

PR #4824, opened 2025-08-20
feat(ethexe-contracts): update dependencies, fix CI and lints

StackOverflowExcept1on:


ethexe/contracts/test/Router.t.sol
```diff
@@ -11,16 +11,16 @@ contract RouterTest is Base {
     using MessageHashUtils for address;
     using FROSTOffchain for SigningKey;
 
-    address immutable deployer = 0x116B4369a90d2E9DA6BD7a924A23B164E10f6FE9;
+    address public deployer = 0x116B4369a90d2E9DA6BD7a924A23B164E10f6FE9;
```
grishasobol, line 14: Could you explain please why not immutable now or it does not matter cause it's tests?

grishasobol (approved):

ethexe/contracts/test/Router.t.sol
StackOverflowExcept1on: I don't see any point in saving gas for storage in tests

ethexe/contracts/src/Mirror.sol
```diff
@@ -417,6 +444,9 @@ contract Mirror is IMirror {
             callReply := calldataload(0x24)
         }
 
+        // TODO: maybe we should reject value > type(uint128).max?
```
breathx: that's unreachable case since total wvara supply ever expected fits in uint128

breathx (approved):

PR #4826, opened 2025-08-24
feat(`ethexe-ethereum`): support middleware in test deployment

ecol-master:
Implement: 
- `EthereumDeployer`
- Convertions between `alloy::primitives::Address` and `ethexe_common::Address` types
- Update github workflow to use the json-abi files from symbiotic-core and symbiotic-rewards libraries

@grishasobol

### DIFF
--- a/ethexe/ethereum/Cargo.toml
+++ b/ethexe/ethereum/Cargo.toml
@@ -50,6 +50,3 @@ alloy = { workspace = true, features = [
     "node-bindings",
 ] }
 gear-utils.workspace = true
-
-[features]
-local-deploy = []
--- a/ethexe/ethereum/src/lib.rs
+++ b/ethexe/ethereum/src/lib.rs
@@ -51,15 +51,11 @@ use std::time::Duration;
 mod abi;
 mod eip1167;
 
+pub mod deploy;
 pub mod middleware;
 pub mod mirror;
 pub mod router;
 pub mod wvara;
-
-// By design, the deployment module is only needed for testing purposes.
-// Real deployments are done via scripts in the `contracts/scripts/` folder.
-#[cfg(feature = "local-deploy")]
-pub mod deploy;
 
 pub mod primitives {
     pub use alloy::primitives::*;
--- a/ethexe/observer/Cargo.toml
+++ b/ethexe/observer/Cargo.toml
@@ -34,7 +34,6 @@ log.workspace = true
 nonempty.workspace = true
 
 [dev-dependencies]
-ethexe-ethereum = { workspace = true, features = ["local-deploy"] }
 gear-utils.workspace = true
 wasmparser.workspace = true
 wat.workspace = true


ethexe/service/src/tests/utils/env.rs
```diff
@@ -179,17 +179,17 @@ impl TestEnv {
...
-                signer.clone(),
-                sender_address,
-                verifiable_secret_sharing_commitment,
-            )
-            .await?
+            let validators_addresses = validators
+                .iter()
+                .map(|k| k.public_key.to_address())
+                .collect();
+            EthereumDeployer::new(&rpc_url, signer.clone(), sender_address) // verifiable_secret_sharing_commitment,)
```
grishasobol, line 186: ?

ethexe/contracts/script/Deployment.s.sol
```diff
@@ -68,13 +68,17 @@ contract DeploymentScript is Script {
         if (!(vm.envExists("DEV_MODE") && vm.envBool("DEV_MODE"))) {
```
grishasobol, line 68: Please remind me why don't deploy middleware in dev mode? In Dev mode we still needs middleware.
I think we must have here two cases of deployment:
1) FULL - with symbiotic deployment
2) DEV - symbiotic already deployed and we have to provide contracts addresses

ethexe/ethereum/src/lib.rs
```diff
@@ -79,162 +75,71 @@ pub(crate) type ExeFiller =
...
-            self.wvara_address,
-            self.provider.clone(),
-        )
+        Router::new(self.router, self.wvara, self.provider())
     }
+}
 
-    pub fn mirror(&self, address: LocalAddress) -> Mirror {
-        Mirror::new(address.0.into(), self.provider.clone())
+impl EthereumWithMiddleware {
```
grishasobol: What the profit of using different structure here? Even more I can see disadvantage to make like this. For example in ethexe/service/src/tests/utils/env.rs we have field eth: Ethereum. But in some tests we may wanna to use deployment with middleware - we should we do in that case?
In case Ethereum would contain middleware as option it would be much simplier to operate with that in code
```
struct Ethereum {
...
middleware: Optino<Address>,
}

impl Ethereum {
     pub fn middleware() -> Option<Middleware> {
            ...
     }
}
```

ethexe/observer/Cargo.toml
```diff
@@ -10,7 +10,7 @@ repository.workspace = true
 [dependencies]
 ethexe-common = { workspace = true, features = ["std"] }
 ethexe-signer.workspace = true
-ethexe-ethereum.workspace = true
+ethexe-ethereum = { workspace = true, features = ["local-deploy"] }
```
grishasobol: If you use local-deploy for tests only as you write in the comment above, then why you use it here in non-dev deps?

ethexe/ethereum/Cargo.toml
```diff
@@ -48,3 +49,7 @@ alloy = { workspace = true, features = [
     "kzg",
     "node-bindings",
 ] }
+gear-utils.workspace = true
+
+[features]
+local-deploy = []
```
grishasobol: Why naming is "local-deploy" ?

kuzminxyz: `local-deploy` - deploy to local rpc node. How it would be better to name?

ethexe/observer/Cargo.toml
kuzminxyz: yes, it could be moved to dev deps

ethexe/ethereum/src/lib.rs
kuzminxyz: It would operate easier, but interface is not clear. For example you deploy without middleware, but trying to use middleware for some purposes.

ethexe/contracts/script/Deployment.s.sol
kuzminxyz: Dev mode was added to decrease the number of deployed contracts to test the correctness of gear exe program.  For testing purposes you only need, Router, Wrapped VARA and Mirror. Middleware logic is not required.

ethexe/ethereum/Cargo.toml
grishasobol: this can be used not only for local rpc node deploying. for remote rpc also

PR #4827, opened 2025-08-25
chore(ci): Add CI for `linux-aarch64` 

ByteNacked:
Resolves # .

- Add CI for `linux-aarch64` target

@reviewer-or-team


scripts/src/init.sh
```diff
@@ -33,10 +33,17 @@ cargo_init() {
   if [ -z $CI ] ; then
     cargo install cargo-hack
```
StackOverflowExcept1on, line 34: https://github.com/cargo-bins/cargo-binstall

ByteNacked: Hmm, good idea, but  we still need  to:
*  `curl -L https://github.com/cargo-bins/cargo-binstall/releases/download/v1.15.1/cargo-binstall-x86_64-unknown-linux-gnu.tgz` 
* `curl -L https://github.com/cargo-bins/cargo-binstall/releases/download/v1.15.1/cargo-binstall-aarch64-unknown-linux-gnu.full.tgz` 

so it's kinda misses the point

StackOverflowExcept1on: IIRC cargo binstall installs into `~/.cargo/bin`, which should exist on Rust developers machines, so we could silently install cargo binstall if it wasn't there.

.github/workflows/build-linux.yml
```diff
@@ -134,6 +134,45 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+  lazy-pages-linux-aarch64-tests:
```
ark0f, line 122: Should in separate file like `build-windows.yml` and `build-macos.yml`

.github/workflows/check.yml
```diff
@@ -103,6 +103,9 @@ jobs:
           - name: linux-x86_64
             target: x86_64-unknown-linux-gnu
             runs-on: ubuntu-24.04-arm
+          - name: linux-aarch64
+            target: aarch64-unknown-linux-gnu
+            runs-on: ubuntu-24.04-arm
```
ark0f, line 114: `stack-height` job should be adjusted too

.github/workflows/build-linux.yml
ByteNacked: Discussed

scripts/src/init.sh
ByteNacked: Let's discuss it again and implement in separate PR if needed

.github/workflows/check.yml
ByteNacked: Fixed

.github/workflows/build-linux.yml
```diff
@@ -119,6 +119,45 @@ jobs:
...
+    runs-on: ubuntu-24.04-arm
+    strategy:
+      matrix:
+        profiles: ${{ fromJSON(inputs.profiles) }}
+      fail-fast: false
+    name: lazy pages linux-aarch64 tests (${{ matrix.profiles.name }})
```
ark0f:
```suggestion
    runs-on: ubuntu-24.04-arm
    needs: workspace
    strategy:
      matrix:
        profiles: ${{ fromJSON(inputs.profiles) }}
      fail-fast: false
    name: linux-aarch64 (${{ matrix.profiles.name }})
```

PR #4835, opened 2025-09-02
feat(gtest): Introduce builtins to `gtest`

techraed:
1. Adds `eth-bridge` and `bls12_381` builtins with corresponding tests. The introduced codebase almost identically copies builtins from pallet, but doesn't adapt it fully to the `gtest`, leaving some "obtain payload length" functionality for future gas charge feature.
2. Refactors a little bit `WasmProgram` trait for mock programs

The gas charging for builtins is missing and is intended to be introduced in following PRs

### DIFF
--- a/gtest/src/builtins/mod.rs
+++ b/gtest/src/builtins/mod.rs
@@ -1,12 +1,9 @@
 mod bls12_381;
 mod eth_bridge;
 
-pub use bls12_381::{
-    BLS12_381_ID, Request as Bls12_381Request, Response as Bls12_381Response,
-    process_bls12_381_dispatch,
-};
+pub use bls12_381::{BLS12_381_ID, Bls12Request, Bls12Response, process_bls12_381_dispatch};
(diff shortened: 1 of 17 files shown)


gtest/src/builtins/mod.rs
```diff
@@ -0,0 +1,74 @@
...
+pub(crate) use bls12_381::process_bls12_381_dispatch;
+pub(crate) use eth_bridge::process_eth_bridge_dispatch;
+
+use core_processor::common::{ActorExecutionErrorReplyReason, TrapExplanation};
+use gear_core::str::LimitedStr;
+use parity_scale_codec::{Decode, Encode};
+
+/// Builtin actor errors.
+#[derive(Encode, Decode, Clone, Debug, PartialEq, Eq)]
+pub enum BuiltinActorError {
```
breathx: this may be put in somewhere common to be reused here and on pallet 
consider just gbuiltins crate

breathx: same as impls for bls cryptography maybe, abstracted out of host calls

techraed: done

PR #4836, opened 2025-09-02
chore: remove regular docs publishing

breathx:


.github/workflows/build-linux.yml
```diff
@@ -52,29 +52,6 @@ jobs:
       - name: Test docs
         run: ./scripts/gear.sh test docs
 
-      - name: Build docs
```
ark0f, line 55: `cargo test --doc` might not catch all of documentation problems so I'd revert this step

ark0f (changes requested):

PR #4838, opened 2025-09-03
test(ethexe): improve mock environment for inner services testing

grishasobol:
additionally
1) change LatestDataStorage interface
2) change block announces from Vec to BTreeSet
3) rename set_validators -> set_block_validators
4) traits for db write access now subtrait of read access

### DIFF
--- a/ethexe/common/src/mock.rs
+++ b/ethexe/common/src/mock.rs
@@ -32,7 +32,7 @@ use itertools::Itertools;
 use nonempty::{NonEmpty, nonempty};
 use std::collections::{BTreeSet, VecDeque};
 
-// TODO +_+_+: use `proptest::Arbitrary` instead
+// TODO #4881: use `proptest::Arbitrary` instead
 pub trait Mock<Args> {
     fn mock(args: Args) -> Self;
 }


ethexe/common/src/mock.rs
```diff
@@ -0,0 +1,497 @@
...
+    gear::{BatchCommitment, ChainCommitment, CodeCommitment, Message, StateTransition},
+};
+use alloc::{collections::BTreeMap, vec};
+use gear_core::code::{CodeMetadata, InstrumentedCode};
+use gprimitives::{CodeId, H256};
+use itertools::Itertools;
+use nonempty::{NonEmpty, nonempty};
+use std::collections::{BTreeSet, VecDeque};
+
+pub trait Mock<Args> {
```
kuzminxyz, line 36: Maybe we should introduce a test-utils crate and move it there?
In future we can move all the stuff connected to integration testing in this crate

ethexe/common/src/db.rs
```diff
@@ -160,17 +160,16 @@ pub trait LatestDataStorageRead {
...
-            if let Some(data) = data.as_mut() {
-                f(data);
-                return_value = Some(());
-            }
-        });
-        return_value
+pub trait LatestDataStorageWrite: LatestDataStorageRead {
+    fn set_latest_data(&self, data: LatestData);
+    fn mutate_latest_data(&self, f: impl FnOnce(&mut LatestData)) -> Option<()> {
+        if let Some(mut latest_data) = self.latest_data() {
```
kuzminxyz, line 173: Maybe we should move the `LatestData` fields under and Option<>. It would be smarter because of fields are initialize in different places and shouldn't be filled by default values.

```rust
#[derive(Debug, Clone, Default, Encode, Decode, PartialEq, Eq)]
pub struct LatestData {
    /// Latest synced block height
    pub synced_block_height: u32,
    /// Latest prepared block hash
    pub prepared_block_hash: H256,
    /// Latest computed announce hash
    pub computed_announce_hash: AnnounceHash,
    /// Genesis block hash
    pub genesis_block_hash: H256,
    /// Genesis announce hash
    pub genesis_announce_hash: AnnounceHash,
    /// Start block hash: genesis or defined by fast-sync
    pub start_block_hash: H256,
    /// Start announce hash: genesis or defined by fast-sync
    pub start_announce_hash: AnnounceHash,
}
```

The  `Default` impl must return struct with all `None` fields and you will able to  use here `unwrap_or_default` in mutating latest_data

ethexe/observer/src/sync.rs
```diff
@@ -203,7 +189,7 @@ impl<DB: SyncDB> ChainSync<DB> {
 
             let _ = self
                 .db
-                .mutate_latest_data_if_some(|data| data.synced_block_height = block_header.height)
+                .mutate_latest_data(|data| data.synced_block_height = block_header.height)
```
kuzminxyz, line 192: can be changed after refactor latest data

ethexe/tx-pool/src/tests.rs
```diff
@@ -80,16 +80,14 @@ impl BlocksManager {
...
-                        prepared_block_hash: block_hash,
-                        synced_block_height: header.height,
-                        computed_announce_hash: AnnounceHash::zero(),
-                        genesis_block_hash: block_hash,
-                        genesis_announce_hash: AnnounceHash::zero(),
-                        start_block_hash: block_hash,
-                        start_announce_hash: AnnounceHash::zero(),
-                    });
+                self.db.set_latest_data(LatestData {
+                    prepared_block_hash: block_hash,
```
kuzminxyz, line 84: and refactor here

.envrc
```diff
@@ -1 +1 @@
-use flake
+use nix
```
grishasobol, line 1: It's github bug actually gsobol/aanounce has "use nix" also

ethexe/common/src/mock.rs
grishasobol: Actually I cannot see reasons to do this. Also having separate crate like "test-utils" may lead to cycle dependencies problem ("test-utils" needs to import crate "A" to implement something for structures from crate "A", but crate "A" needs to import "test-utils" to use it in crate "A" tests)

ethexe/common/src/db.rs
grishasobol: Actually it's initialised in one place always: when we set genesis in db

.envrc
grishasobol: outdated

ethexe/common/Cargo.toml
```diff
@@ -35,3 +39,4 @@ std = [
     "k256/std",
     "alloy-primitives/std",
 ]
+mock = ["std", "itertools", "itertools/use_std", "tap"]
```
breathx: it's enabled as a dep automatically 
```suggestion
mock = ["std", "itertools/use_std", "tap"]
```

ethexe/common/src/mock.rs
```diff
@@ -0,0 +1,497 @@
...
+    fn mock(args: Args) -> Self;
+}
```
ark0f, line 38: I cannot see actual reason for trait. `mock` constructors can be used inside usual impls right in this file

grishasobol: https://github.com/gear-tech/gear/blob/149819e654024214dd925f52555a54d08b1a97c1/ethexe/consensus/src/mock.rs#L99-L104

grishasobol: actually is better to use Arbitrary from proptest crate instead of Mock, but I'm not ready to doing so big work now. Maybe in future.

ethexe/common/src/mock.rs
grishasobol: create new issue to consider your idea https://github.com/gear-tech/gear/issues/4880

ethexe/common/src/mock.rs
ark0f: It would be nice to have an issue for this

grishasobol: https://github.com/gear-tech/gear/issues/4881

ethexe/common/src/db.rs
grishasobol: I think this can be resolved for now

PR #4841, opened 2025-09-05
feat(contracts): store lock files; impl CI check for ABIs

breathx:


ethexe/contracts/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "devDependencies": {
+    "@openzeppelin/upgrades-core": "^1.44.1"
```
StackOverflowExcept1on: https://github.com/OpenZeppelin/openzeppelin-foundry-upgrades/blob/update-gitignore/src/internal/Versions.sol
`^1.37.0`

StackOverflowExcept1on (approved):

PR #4842, opened 2025-09-05
refactor(ethexe): Make `TxPoolService` an actual service

ark0f:


ethexe/tx-pool/src/lib.rs
```diff
@@ -54,6 +69,46 @@ impl TxPoolService {
...
+
+impl Stream for TxPoolService {
+    type Item = TxPoolEvent;
+
+    fn poll_next(mut self: Pin<&mut Self>, _cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        if let Some(event) = self.events.pop_front() {
+            return Poll::Ready(Some(event));
+        }
+
+        Poll::Pending
```
techraed, line 114: shouldn't you use a waker to tell the runtime to wake when a new event has come?

techraed: I guess you can find an impl of that in commits of the PR which introduced the tx pool

ethexe/tx-pool/src/lib.rs
```diff
@@ -54,6 +77,48 @@ impl TxPoolService {
...
+            .context("Failed to validate offchain transaction")?;
+        let tx_hash = validated_tx.tx_hash();
+
+        // Set valid transaction
+        self.db.set_offchain_transaction(validated_tx.clone());
+
+        // Propagate transaction
+        self.events
+            .push_back(TxPoolEvent::PublishOffchainTransaction(validated_tx));
+        self.wake();
```
grishasobol, line 96: why do we need wake here? process_offchain_transaction is calling when new service event is processing, this means that it's guarantied that main service will poll all services after event processing

ark0f: The only reason is future tests

techraed: @grishasobol correct me If I'm wrong, but `tokio::select!` macro will poll all the futures/streams in each iteration of the loop of the service, but it only polls them when the tokio runtime thinks there might be a new data available.
If so, wake is necessary, as it tells to the runtime that the future is ready to progress (be polled).

techraed (approved):

ethexe/tx-pool/src/lib.rs
ark0f: > but it only polls them when the tokio runtime thinks there might be a new data available.

You're almost right. `tokio::select!()` checks the current task budget before it randomly polls the provided futures

PR #4846, opened 2025-09-08
refactor(core): Rework `Limited{Vec, Str}`

liferooter:


core/src/buffer.rs
```diff
@@ -76,7 +78,32 @@ where
...
+    }
+}
+
+impl From<LimitedVecError> for &'static str {
+    fn from(err: LimitedVecError) -> Self {
+        err.message()
+    }
+}
+
+impl Display for LimitedVecError {
```
ark0f: Can be replaced with `derive_more::Display`

core/src/buffer.rs
```diff
@@ -76,7 +78,32 @@ where
...
+pub struct LimitedVecError;
+
+impl LimitedVecError {
+    /// Returns a static error message.
+    pub const fn message(&self) -> &'static str {
+        "vector length limit is exceeded"
+    }
+}
+
+impl From<LimitedVecError> for &'static str {
```
ark0f: It's better to make it `into_static_str()` because such conversion is not obvious

core/src/buffer.rs
```diff
@@ -76,7 +78,32 @@ where
...
-impl<T: Clone + Default, E: Default, const N: usize> Debug for LimitedVec<T, E, N>
+/// Error type for limited vector overflowing.
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq)]
+pub struct LimitedVecError;
+
+impl LimitedVecError {
+    /// Returns a static error message.
+    pub const fn message(&self) -> &'static str {
+        "vector length limit is exceeded"
+    }
```
ark0f:
```suggestion
    const ERROR_MSG: &str = "vector length limit is exceeded";
```

core/src/buffer.rs
```diff
@@ -85,110 +112,188 @@ where
...
+}
+
+impl<T, const N: usize> TryFrom<Vec<T>> for LimitedVec<T, N> {
+    type Error = LimitedVecError;
+    fn try_from(vec: Vec<T>) -> Result<Self, Self::Error> {
+        Self::validate_len(vec.len()).map(|_| Self { inner: vec })
+    }
+}
+
+impl<T, const N: usize> AsRef<[T]> for LimitedVec<T, N> {
```
ark0f: `AsRef`, `Deref` and others can be replaced with corresponding `derive_more` macros too

ark0f (changes requested):

core/src/buffer.rs
liferooter: Adding associated constant is a good idea.

Having `message` method is convenient for usage in code where the error is mapped into its message. We have such places, e.g. https://github.com/gear-tech/gear/blob/8bde2a7182a1f73be7a055f60d508a9deb086dc1/pallets/gear/src/lib.rs#L578

core/src/limited/str.rs
```diff
@@ -1,6 +1,6 @@
 // This file is part of Gear.
 
-// Copyright (C) 2023-2025 Gear Technologies Inc.
+// Copyright (C) 2021-2025 Gear Technologies Inc.
```
ark0f: Start date should not be updated

core/src/limited/str.rs
```diff
@@ -16,58 +16,123 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
 
-        s.truncate(last_byte);
-    }
-}
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//! This module provides type for string with limited length.
```
ark0f: Duplicated license header. Should be removed

core/src/limited/str.rs
```diff
@@ -16,58 +16,123 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
-//! String with limited length implementation
+//! This module provides type for string with statically limited length.
```
ark0f:
```suggestion
//! This module provides type for string with limited length.
```

Because it's limited in runtime too

core/src/limited/str.rs
```diff
@@ -16,58 +16,123 @@
...
+    pub fn truncated(s: &'a str) -> Self {
+        Self(s[..Self::cut_index(s, Self::MAX_LEN)].into())
+    }
+
+    /// Constructs a limited string from a [`String`]
+    /// truncating it if it's too long.
+    pub fn owned_truncated(mut s: String) -> Self {
+        s.truncate(Self::cut_index(&s, Self::MAX_LEN));
+        Self(s.into())
+    }
```
ark0f:
```suggestion
    pub fn truncated(s: impl Into<Cow<'a, str>>) -> Self {
        match s.into() {
        	// ...
        }
    }
```

Can be just one constructor

core/src/limited/str.rs
```diff
@@ -16,58 +16,123 @@
...
+    pub const MAX_LEN: usize = 1024;
+
+    /// Calculates safe truncation position to truncate
+    /// a string down to [`Self::MAX_LEN`] bytes or less.
+    fn cut_index(s: &str, pos: usize) -> usize {
+        (0..=pos.min(s.len()))
+            .rev()
+            .find(|&pos| s.is_char_boundary(pos))
+            .unwrap_or(0)
+    }
```
ark0f:
```suggestion
    /// Returns the nearest UTF-8 character boundary at or before `pos`
    fn nearest_char_boundary(s: &str, pos: usize) -> usize {
        (0..=pos.min(s.len()))
            .rev()
            .find(|&pos| s.is_char_boundary(pos))
            .unwrap_or(0)
    }
```

core/src/limited/str.rs
```diff
@@ -16,58 +16,123 @@
...
+    /// string literal small enough to fit the limit.
+    ///
+    /// Should be used only with static string literals.
+    /// In that case it can check the string length
+    /// in compile time.
+    ///
+    /// # Panics
+    ///
+    /// Can panic in runtime if the passed string is
+    /// not a static string literal and is too long.
```
ark0f, line 105:
```suggestion
    /// Constructs a limited string from a static
    /// string literal small enough to fit within the limit.
    ///
    /// # Panics
    ///
    /// Can panic in runtime if the passed string is too long.
```
String is static by signature itself

core/src/limited/str.rs
ark0f: Also `pos` can be removed because the parameter is always `Self::MAX_LEN`

core/src/limited/vec.rs
```diff
@@ -0,0 +1,363 @@
...
+    pub fn repeat(value: T) -> Self
+    where
+        T: Clone,
+    {
+        Self(vec![value; N])
+    }
+
+    /// Creates a new full limited vector by
+    /// repeatedly cloning `T::default()`.
+    pub fn repeat_default() -> Self
```
ark0f: `_default()` constructors can be removed as they are just wrappers about pretty simple expressions

core/src/limited/vec.rs
```diff
@@ -0,0 +1,363 @@
...
+
+        let p1 = hex::encode(&self[..e1]);
+        let p2 = hex::encode(&self[s2..]);
+        let sep = if e1 != s2 { ".." } else { "" };
+
+        if f.alternate() {
+            write!(f, "LimitedVec(0x{p1}{sep}{p2})")
+        } else {
+            write!(f, "0x{p1}{sep}{p2}")
+        }
```
ark0f: Can be replaced with `gprimitives::ByteSliceFormatter`

core/src/limited/str.rs
liferooter: That's how it was in `str.rs`, I didn't update anything.

core/src/limited/str.rs
liferooter: Fixed

core/src/limited/str.rs
liferooter: Done

core/src/limited/str.rs
liferooter: Static lifetime doesn't guarantee availability of the value in the constant context. The value still can be dynamic in many ways.

```rust
const fn func(s: &'static str) {
    todo!()
}

func(["a", "b", "c"].choose(&mut rng));

let s = String::leak(std::fs::read_file(FILE_PATH).unwrap());
func(s);
```

It would be nice to have lifetimes that can work backward, not only forward. But currently we don't.

core/src/limited/str.rs
liferooter: Done

core/src/limited/vec.rs
liferooter: Done

core/src/limited/vec.rs
liferooter: `gprimitives::ByteSliceFormatter` handles precesion in a different way. Its default precision is full instead of 8. Should I unify all the code that formats `LimitedVec` with `ByteSliceFormatter` or implement the previous behaviour on top of `ByteSliceFormatter`?

core/src/limited/str.rs
liferooter: Done

core/src/limited/str.rs
ark0f: It was exactly `2023-2025`. The problem was duplicate header

core/src/limited/vec.rs
ark0f: > or implement the previous behaviour on top of ByteSliceFormatter?

This, please. Semantics of `ByteSliceFormatter` are pretty straightforward - just prints HEX representation of slice. `LimitedVec` is expected to contain a lot of data and thus precision is chosen to be 8

core/src/limited/str.rs
liferooter: Ok, so it's fixed now.

core/src/limited/vec.rs
liferooter: Done

core/src/limited/vec.rs
```diff
@@ -0,0 +1,338 @@
...
+            |f, bytes| write!(f, "{bytes}")
+        };
+
+        if let Some(precision) = precision {
+            fmt_bytes(f, format_args!("{bytes:.precision$}"))
+        } else if sign_plus {
+            fmt_bytes(f, format_args!("{bytes}"))
+        } else {
+            fmt_bytes(f, format_args!("{bytes:.8}"))
+        }
```
ark0f: Can be simplified like this
```suggestion
        let bytes = ByteSliceFormatter::Dynamic(self.0.as_slice().as_ref());

        let precision = f.precision();
        let sign_plus = f.sign_plus();

        let bytes = if let Some(precision) = precision {
            format_args!("{bytes:.precision$}")
        } else if sign_plus {
            format_args!("{bytes}")
        } else {
            format_args!("{bytes:.8}")
        };
        
        if f.alternate() {
            write!(f, "LimitedVec({bytes})")
        } else {
            write!(f, "{bytes}")
        }
```

ark0f (approved):
Congrats 🎉

core/src/limited/vec.rs
liferooter: No, it can not.

https://doc.rust-lang.org/std/macro.format_args.html#lifetime-limitation

ark0f: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=5c4e74ebb93aa581e1b59e32b2899318

PR #4848, opened 2025-09-09
build(nix): rework Nix devenv

liferooter:
- Nix flake is replaced with classic `shell.nix` with inputs managed using [npins](https://github.com/andir/npins).
- Nix shell's `stdenv` is replaced with an LLVM-based one.
- Rust in Nix devenv is now managed with `rustup`, because the code requires it.
- Some unneeded dependencies are removed from Nix devenv
  - `rocksdb`, as it's vendored by crates that use it and always compiled from source.
  - `jemalloc` for the same reason.
  - `git`, because it's the way you get the repository and usually already presented on the computer of the developer.
  - `llvmPackages.clang` and `llvmPackages.libclang`, as they're provided by `stdenv`.
- Some unneeded environment variables are removed from Nix devenv
  - `IN_NIX_SHELL`, as it's always set by Nix.
  - `CC`, as it's always the default with LLVM-based `stdenv`.
- Some missing packages are added to Nix devenv
- `make typos` now prints hint instead of automatic installation of `typos` with `cargo install`, which is more agnostic to package manager. 
- Nix-specific ad-hoc code is removed, since Nix devenv now uses `rustup`


playX18 (approved):
LGTM

PR #4850, opened 2025-09-11
refactor(gtest): improve API and docs

techraed:
- Rename `ActorMailbox` to `UserMailbox`, because there's no mailbox for programs
- Rework logs concept:
   `Log`s (`Log` and `CoreLog`, `DecodedCoreLog`) were introduced in the beginning of the `gtest`. Being useful, however, the naming was wrong. Also several types and API being 1) dependent on `Into<Log>`, 2) using `Log` to compare it to `CoreLog`, which itself is derived from user stored message, 3) having vague builder impl for `Log`, 4) having no utility for users of `CoreLog` made the API ambiguous. Now the concept is straightforward and clean. You have just `UserMessageEvent` derived from `StoredMessage`s stored as events (not in mailbox) and `EventBuilder` to build those events and perform ops with them in `BlockRunResult` or `UserMailbox`
- Actualize crate docs, making it deeper, clarifying the concepts and usage


grishasobol:
outdated

PR #4862, opened 2025-09-23
chore: begin replacement of Make and `./scripts/gear.sh` with Just

liferooter:
- Introduces `justfile` as an alternative to existing `Makefile` and `./scripts/gear.sh`.
- Implements some commonly used Make targets as Just recipes.
- Replaces usage of `./scripts/gear.sh` with usage of Just in some parts of CI.
- Inline some usages of `./scrpts/gear.sh` in CI where there's no need for abstraction into a script.

### DIFF
--- a/.github/actions/install-cargo-extensions/action.yml
+++ b/.github/actions/install-cargo-extensions/action.yml
@@ -4,4 +4,5 @@ runs:
   using: composite
   steps:
     - uses: cargo-bins/cargo-binstall@main
-    - cargo binstall cargo-hack cargo-nextest
+    - run: cargo binstall cargo-hack cargo-nextest
+      shell: bash


.github/workflows/check.yml
```diff
@@ -145,7 +151,7 @@ jobs:
           key: "${{ matrix.name }}"
 
       - name: "Check clippy"
-        run: ./scripts/gear.sh clippy gear --all-targets --all-features --locked
+        run: just clippy native
```
breathx, line 154: its more likely 
just clippy :: { workspace ; examples } -> not the point that its wasms
but moreover these steps could be merged into one

.github/workflows/test-measurements.yaml
```diff
@@ -26,15 +26,17 @@ jobs:
       - name: "Install: Rust toolchain"
         uses: dsherret/rust-toolchain-file@v1
 
+      - name: "Install: Cargo extensions"
+        uses: ./.github/actions/install-cargo-extensions
+
       - name: "Show: Versioning"
```
breathx, line 34: such simple commands could be nameless in CI

liferooter: Other such commands, like "Install: Rust toolchain" or "Install: Foundry" are not nameless.

.github/actions/install-cargo-extensions/action.yml
ark0f, line 1: Consider to use https://github.com/marketplace/actions/install-cargo-binstall

justfile
```diff
@@ -0,0 +1,95 @@
...
+@default:
+    echo 'To list available recipes:'
+    echo '> just --list'
+    echo
+    echo 'To list available recipes from namespace:'
+    echo '> just --list <namespace>'
+
+# Remove untracked files and build caches
+[group('actions')]
+[confirm('Remove all untracked files and build caches? (y/n)')]
```
ark0f, line 21: Build caches (target dir) are not cleaned

just/lib.just
ark0f, line 1: `require()` may simplify script

https://just.systems/man/en/functions.html#executables

justfile
```diff
@@ -0,0 +1,95 @@
...
+
+# Format code via `rustfmt`
+[group('actions')]
+fmt:
+    cargo fmt --all
+
+# Check formatting with `rustfmt`
+[group('checks')]
+fmt-check:
+    cargo fmt --all --check
```
ark0f, line 37:
```suggestion
# Format code via `rustfmt`
[group('actions')]
fmt *ARGS:
    cargo fmt --all {{ARGS}}
```

justfile
```diff
@@ -0,0 +1,95 @@
...
+
+# Check code for typos via `typos-cli`
+[group('checks')]
+typos: (ensure-binary "typos" "cargo install typos-cli")
+    # Checking the repository for typos
+    typos
+
+# Check code with `cargo check`
+[group('checks')]
+check:
```
ark0f: Can be removed. We always use clippy

justfile
```diff
@@ -0,0 +1,95 @@
...
+        --exclude runtime-fuzzer \
+        --exclude runtime-fuzzer-fuzz
+
+# Run documentation tests
+[group('checks')]
+test-doc:
+    # Running documentation tests
+    __GEAR_WASM_BUILDER_NO_BUILD=1 \
+    SKIP_WASM_BUILD=1 \
+        cargo test --doc --workspace --no-fail-fast
```
ark0f, line 62:
```suggestion
[group('checks')]
test-doc $__GEAR_WASM_BUILDER_NO_BUILD="1" $SKIP_WASM_BUILD="1":
    cargo test --doc --workspace --no-fail-fast
```

justfile
```diff
@@ -0,0 +1,95 @@
...
+        --exclude gclient \
+        --exclude gcli \
+        --exclude gsdk \
+        --exclude gear-authorship \
+        --exclude pallet-gear-staking-rewards \
+        --exclude gear-wasm-gen \
+        --exclude demo-stack-allocations \
+        --exclude gring \
+        --exclude runtime-fuzzer \
+        --exclude runtime-fuzzer-fuzz
```
ark0f:
```suggestion
test *ARGS: (ensure-cargo "hack") (ensure-cargo "nextest")
    # Running workspace tests
    cargo nextest run --workspace --all-targets --all-features -E 'not package(/fuzz/)' {{ARGS}}
```

justfile
```diff
@@ -0,0 +1,95 @@
...
+
+# Check runtime imports
+[group('checks')]
+check-runtime-imports:
+    # Checking runtime imports
+    cargo build -p wasm-proc
+    cargo build -p vara-runtime
+    ./target/debug/wasm-proc \
+        --check-runtime-imports \
+        target/debug/wbuild/vara-runtime/vara_runtime.wasm
```
ark0f, line 72:
```suggestion
[group('checks')]
check-runtime-imports:
    # Checking runtime imports
    cargo build -p vara-runtime
    cargo run -p wasm-proc -- \
        --check-runtime-imports \
        target/debug/wbuild/vara-runtime/vara_runtime.wasm
```

just/clippy.just
```diff
@@ -0,0 +1,52 @@
...
+    # Checking all native code with Clippy
+    __GEAR_WASM_BUILDER_NO_BUILD=1 SKIP_WASM_BUILD=1 \
+      cargo clippy \
+        --locked --workspace --all-targets --all-features \
+        -- --no-deps -D warnings
+
+# Check all WASM code
+wasm:
+    # Checking all WASM code with Clippy
+    cargo metadata --no-deps --format-version=1 \
```
ark0f, line 15: Consider to use shebang scripts and paste lines directly from `./scripts/src/clippy.sh`

justfile
liferooter: Target directory is a part of `.gitignore` so is fully removed by `git clean -fdx`.

justfile
liferooter: Fixed

just/lib.just
liferooter: My custom `ensure-*` repices are more powerful than Just's `require` function.
- It shows hint that describes how to install the program.
- It is recipe-specific, not global.
- It is custom enough to be able to check Cargo extensions.

justfile
liferooter: It was done on purpose. Using long recipe-specific environment variables makes help messages very ugly and also confuses by being the only thing about running commands that is not printed during recipe execution.

justfile
liferooter: I think having a separate `fmt-check` recipe is better for some reasons:
- Accepting arguments by `fmt` recipe is not very useful because the only useful argument that can be passed to it is `--check`.
- Recipes with arguments cannot be chained (e.g. we can't run `just fmt typos test-doc`), so it makes sense to refrain from using recipes with arguments as much as possible.
- Having a separate `fmt-check` recipe improves autocompletion for this recipe.
- Having a separate `fmt-check` recipe allows to used it as a dependency (like for `pre-commit`) without causing an argument passing hell like what we have in `Makefile` and `gear.sh` now.

justfile
liferooter: Simplification is done.

Argument passing is a bad idea I think.
- It is not very useful, because for running tests for specific crates or with specific options it's better to use `cargo nextest` directly for better control and more explicitness. The point of the recipe is to give us a short command for a common non-trivial testing task, not to give us an universal script runner, for which always having all these options by default is not a good idea.
- The same reasons as with `fmt-check`. Recipes without arguments can be chained, generally cleaner and will never cause argument passing hell.

justfile
liferooter: Fixed

.github/actions/install-cargo-extensions/action.yml
liferooter: Done

PR #4870, opened 2025-09-28
chore(ethexe): Download `BLOCK_HASHES_WINDOW_SIZE` blocks in fast-sync

ark0f:


ethexe/service/src/fast_sync.rs
```diff
@@ -599,6 +600,20 @@ async fn instrument_codes(
     Ok(())
 }
 
+async fn set_tx_pool_data_requirement(
+    observer: &mut ObserverService,
+    db: &Database,
+    latest_committed_block: H256,
+) -> Result<()> {
+    let mut block_hash = latest_committed_block;
+    for _ in 0..OffchainTransaction::BLOCK_HASHES_WINDOW_SIZE {
```
breathx: Can't we batch it by block numbers range?

breathx (approved):

ethexe/observer/src/utils.rs
```diff
@@ -29,218 +29,304 @@ use alloy::{
...
+    DB: OnChainStorageRead + OnChainStorageWrite,
+{
+    async fn load(&self, block: H256, header: Option<BlockHeader>) -> Result<BlockData> {
+        let data = match (self.db.block_header(block), self.db.block_events(block)) {
+            (Some(header), Some(events)) => BlockData {
+                hash: block,
+                header,
+                events,
+            },
+            (Some(_), None) | (None, Some(_)) => unreachable!("inconsistent database state"),
```
grishasobol: 1) It's definetely not unreachable, because state of DB can be changed outside. This is error case.
2) As soon as you added BlockLoader - this situation can be. When two threads have their own block loaders, so one thread write header, then events. In the same time second thread is in this function. Between two moments when thread1 has already set header, but still not set events - this case would appear on thread2.

ethexe/observer/src/utils.rs
```diff
@@ -29,218 +29,304 @@ use alloy::{
...
+                    hash: block,
+                    header,
+                    events,
+                },
+                (Some(_), None) | (None, Some(_)) => unreachable!("inconsistent database state"),
+                (None, None) => {
+                    missing_end_block = Some(bn);
+                    break;
+                }
+            };
```
grishasobol: This match looks weird for me.
To identify whether block has on-chain information in db, we have already block_synced(). If you wanna to skip here loading of already loaded data - much more clear solution can be: use observer Stream:
```
observer.force_block_sync(block);
observer.wait_block_synced(block).await;
```
Using observer stream would be more safe and services style solution.

ethexe/observer/src/utils.rs
```diff
@@ -29,218 +29,304 @@ use alloy::{
...
+        };
+        Ok(data)
+    }
 
-    let mut blocks_data = Vec::new();
+    async fn load_many(&self, range: RangeInclusive<u64>) -> Result<HashMap<H256, BlockData>> {
+        let end_block = self
+            .provider
+            .get_block_by_number((*range.end()).into())
+            .await?
```
grishasobol: You have additional rpc call here, which is not presented in previous solution, making it less efficient.

ethexe/observer/src/utils.rs
```diff
@@ -29,218 +29,304 @@ use alloy::{
...
+        log::trace!("Querying blocks batch in {range:?} range");
 
-    future::try_join_all(batch_futures).await.map(|batches| {
-        batches
+        let batch_futures = range.clone().step_by(MAX_QUERY_BLOCK_RANGE).map(|start| {
+            let end = (start + MAX_QUERY_BLOCK_RANGE as u64 - 1).min(*range.end());
+            self.request_block_batch(start..=end)
+        });
+
+        let batches = future::try_join_all(batch_futures).await?;
```
grishasobol, line 223: Why not FuturesUnordered ?

ethexe/observer/src/utils.rs
```diff
@@ -29,218 +29,304 @@ use alloy::{
...
+
+        Ok(BlockData {
+            hash: block,
+            header,
+            events,
         })
-        .collect();
+    }
+
+    async fn load_many(&self, range: RangeInclusive<u64>) -> Result<HashMap<H256, BlockData>> {
```
grishasobol, line 215: It looks dirty for me: writing blocks data in db and returns in the same time. You can write data in db outside of block loader.

grishasobol (changes requested):
Overall I'm not agree with this approach. Writing data in on-chain storage bypassing observer, is unsafe.

Optimisation in this PR does not make a significant contribution. Observer already tries to upload only not-synced blocks. And in almost all cases "synced" means "header set" and "events set" and vice versa.

ethexe/observer/src/utils.rs
ark0f: Because `try_join_all` takes iterator, so `FuturesUnordered` are nothing more than `Vec`

ethexe/common/src/primitives.rs
```diff
@@ -68,7 +68,7 @@ impl BlockData {
     }
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, Encode, Decode, Default)]
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Encode, Decode, Default)]
```
grishasobol, line 71: Why copy? size = 80 bytes

ark0f: I've searched for recommendation about `Copy` implementation and there is no strict specifications. It's usually something like POD values, semantics and so on

PR #4876, opened 2025-09-29
refactor: rework `Read*` accessors in `core-backend`

### LINKED ISSUE #2164: Remove `DecodeError`
### File Location(s)

`core-backend/common`

### Proposal

Remove `DecodeError` variant and usage of `codec` in general as it is unreachable

liferooter:
Resolves #2164.

- Replace unsound unsafe `ReadAs` and fallible redundant `ReadDecoded` with safe `ReadAs` based on safe transmute from [bytemuck](https://crates.io/crates/bytemuck) crate.
- Replace `ReadDecodedSpecial` with `ReadAsOption` and unify it with `ReadAs`.
- Replace unsound unsafe implementation of `WriteAs` with bytemuck-based implementation.
- Remove unused `MemoryAccessError::DecodeError`.
- Remove `#[repr(packed)]` for some types in `gsys` which doesn't need to be packed for not having padding bytes. Not having padding bytes are now enforced for them by `bytemuck::Pod` derive.

### DIFF
--- a/core-backend/src/accessors.rs
+++ b/core-backend/src/accessors.rs
@@ -29,10 +29,7 @@ use crate::{
 };
 use alloc::vec::Vec;
 use bytemuck::Pod;
-use gear_core::{
-    buffer::{Payload, PayloadSizeError},
-    limited::LimitedVecError,
-};
+use gear_core::{buffer::Payload, limited::LimitedVecError};
 use gear_sandbox::{AsContextExt, Value};
 use gear_sandbox_env::HostError;
 
@@ -461,11 +458,11 @@ impl WriteInGrRead {
     }
 }
 
-impl<T> WriteAs<T> {
+impl<T: Pod> WriteAs<T> {
     pub fn write<Caller, Ext>(
         self,
         ctx: &mut MemoryCallerContext<Caller>,
-        obj: T,
+        obj: &T,
     ) -> Result<(), MemoryAccessError>
     where
         Caller: AsContextExt<State = HostState<Ext, BackendMemory<ExecutorMemory>>>,
--- a/core-backend/src/accessors.rs
+++ b/core-backend/src/accessors.rs
@@ -22,16 +22,19 @@ use crate::{
     BackendExternalities,
     memory::{
         BackendMemory, ExecutorMemory, MemoryAccessError, MemoryAccessRegistry, WasmMemoryRead,
-        WasmMemoryReadAs, WasmMemoryReadDecoded, WasmMemoryWrite, WasmMemoryWriteAs,
+        WasmMemoryReadAs, WasmMemoryWrite, WasmMemoryWriteAs,
     },
     runtime::MemoryCallerContext,
     state::HostState,
 };
 use alloc::vec::Vec;
-use gear_core::{buffer::Payload, limited::LimitedVecError};
+use bytemuck::Pod;
+use gear_core::{
+    buffer::{Payload, PayloadSizeError},
+    limited::LimitedVecError,
+};
 use gear_sandbox::{AsContextExt, Value};
 use gear_sandbox_env::HostError;
-use parity_scale_codec::{Decode, MaxEncodedLen};
 
 const PTR_SPECIAL: u32 = u32::MAX;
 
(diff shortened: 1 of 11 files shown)


breathx (approved):

core-backend/src/accessors.rs
```diff
@@ -117,12 +117,8 @@ pub(crate) struct ReadAs<T> {
     result: Result<T, MemoryAccessError>,
 }
 
-pub(crate) struct ReadDecoded<T: Decode + MaxEncodedLen> {
-    result: Result<T, MemoryAccessError>,
-}
-
-pub(crate) struct ReadDecodedSpecial<T: Decode + MaxEncodedLen + Default> {
-    result: Result<T, MemoryAccessError>,
+pub(crate) struct ReadAsOption<T> {
```
ark0f, line 120:
```suggestion
pub(crate) struct ReadSpecial<T> {
```

I'd leave previous name because of `PTR_SPECIAL`. It's more clear what is going on inside

core-backend/src/memory.rs
```diff
@@ -293,48 +268,19 @@ where
...
-            // It's also safe to construct T from any bytes, because we use the fn
-            // only for reading primitive const-size types that are `[repr(C)]`,
-            // so they always represented from a sequence of bytes.
-            //
-            // Bytes in memory are always stored continuously and without paddings, properly
-            // aligned due to `[repr(C, packed)]` attribute of the types we use as T.
-            let mut_slice = unsafe { slice::from_raw_parts_mut(buf.as_mut_ptr() as *mut u8, size) };
-
-            self.memory.read(ctx.caller, read.ptr, mut_slice)?;
+        if core::mem::size_of::<T>() != 0 {
```
ark0f:
```suggestion
        debug_assert_ne!(mem::size_of::<T>(), 0);
```

We actually never expect ZST

core-backend/src/accessors.rs
breathx: or PTR_SPECIAL could be renamed into something more suitable. Like NULL_PTR / NONE_PTR

core-backend/src/memory.rs
liferooter: We have no way to restrict passing ZSTs in compile-time, so why don't just support them even if they're not very useful?

core-backend/src/accessors.rs
liferooter: `PTR_SPECIAL` was renamed to `NULL_PTR`.

core-backend/src/memory.rs
ark0f: > We have no way to restrict passing ZSTs in compile-time

Instead, we rely on runtime tests despite a compile-time check being way more reliable

> why don't just support them even if they're not very useful?

- As a rule of thumb, the best code is no code
- We don't wonder why they are supported because reading/writing 0 bytes from memory has no semantic meaning and they are never actually used in code

liferooter: > > We have no way to restrict passing ZSTs in compile-time
> 
> Instead, we rely on runtime tests despite a compile-time check being way more reliable

As a rule of thumb, a total function is better than a partial, not relying on runtime tests is better than relying on them.

>     * As a rule of thumb, the best code is no code
We have to check for ZSTs anyway. The only difference is what we do when we find a ZST. We can panic or we can handle it correctly.

>     * We don't wonder why they are supported because reading/writing 0 bytes from memory has no semantic meaning and they are never actually used in code
Nonetheless ZSTs are supported by Rust type system. They have a pretty obvious semantic meaning for reading/writing, which is not reading/writing anything. But you're right that they're pretty useless. Maybe the best solution that makes accidental ZST reading/writing easy to avoid, but still doesn't make them cause unspecified behaviour in release builds, is to keep both support and debug assert, which is a pretty Rusty way to solve such problems, like we have for signed overflow, which is not UB and fully specified, but still cause panics in debug mode. So reading/writing ZST will not automatically cause a bug, but still will be easy to find and remove because it's really useless.

ark0f: > We have to check for ZSTs anyway

I see bytemuck [already handles ZST](https://bonfida.github.io/doc-dex-program/bytemuck/fn.bytes_of_mut.html#)

WASM executor also has no problems with zero-length buffer

> is to keep both support and debug assert

Pretty good solution if I miss something about ZST

liferooter: > WASM executor also has no problems with zero-length buffer

Ok, I didn't know it. Then we can fully remove this check and just leave a debug assertion. The backside is that we will have no good non-hacky way to test the code on ZSTs because tests are run in debug mode by default, so the assertions wil be triggered and code won't be tested further.

liferooter: Ah, just found a solution. We can just remove these assertions in test profile.

liferooter: This solution also have a backside which is the inability to catch these assertions in other tests.

liferooter: Added debug assertions. Let's rely on tests in catching ZSTs.

ark0f (approved):

PR #4878, opened 2025-09-30
refactor(ethexe-network): Resolve TODOs

ark0f:
- Use DB traits instead of `Database` itself
- Use router address in gossip topic name
- Don't manage network key in service itself


ethexe/service/src/config.rs
```diff
@@ -39,9 +39,6 @@ impl Config {
     pub fn log_info(&self) {
         log::info!("💾 Database: {}", self.node.database_path.display());
         log::info!("🔑 Key directory: {}", self.node.key_path.display());
-        if let Some(net_config) = &self.network {
```
breathx, line 42: why

breathx, line 42: why

breathx (approved):

PR #4889, opened 2025-10-04
refactor(ethexe): Make `NetworkService` non-optional

ark0f:


ethexe/service/src/tests/utils/env.rs
```diff
@@ -94,7 +94,7 @@ pub struct TestEnv {
     db: Database,
     /// If network is enabled by test, then we store here:
```
breathx: broken comment?

breathx: bootstrap actually still can be None, isn't it?

ethexe/service/src/lib.rs
```diff
@@ -370,7 +364,7 @@ impl Service {
             let event: Event = tokio::select! {
                 event = compute.select_next_some() => event?.into(),
                 event = consensus.select_next_some() => event?.into(),
-                event = network.maybe_next_some() => event.into(),
```
breathx, line 373: we may remove maybe_next_some() ?

PR #4891, opened 2025-10-05
refactor(ethexe): Move `NetworkMessage` to common

ark0f:


ethexe/common/src/consensus.rs
```diff
@@ -0,0 +1,96 @@
...
+}
+
+impl ToDigest for BatchCommitmentValidationRequest {
+    fn update_hasher(&self, hasher: &mut sha3::Keccak256) {
+        let Self {
+            digest,
+            head,
+            codes,
+            rewards,
+            validators,
```
breathx, line 72: we need to change an order here, isn't it? or rewards and validators bools are incorrect.

breathx: as well as hasher updates below

ethexe/network/src/lib.rs
```diff
@@ -64,13 +66,10 @@ const MAX_ESTABLISHED_INCOMING_PER_PEER_CONNECTIONS: u32 = 1;
...
-#[derive(Eq, PartialEq, derive_more::Debug)]
+#[derive(derive_more::Debug, Eq, PartialEq, Clone)]
 pub enum NetworkEvent {
-    Message {
-        #[debug("{:.8} ({} bytes)", ByteSliceFormatter::Dynamic(data), data.len())]
-        data: Vec<u8>,
-        source: Option<PeerId>,
-    },
+    Message(NetworkMessage),
+    OffchainTransaction(SignedOffchainTransaction),
```
breathx, line 72: to be removed later, right? 

this is also intended to have block producer public key it's addressed. Or will be implemented later?

ethexe/service/src/lib.rs
```diff
@@ -427,42 +414,29 @@ impl Service {
...
-                                            "Failed to process offchain transaction received by p2p: {e}"
-                                        );
-                                    }
-                                }
-                            };
+                Event::Network(event) => match event {
+                    NetworkEvent::Message(message) => {
+                        match message {
+                            NetworkMessage::ProducerBlock(block) => {
+                                consensus.receive_announce(block)?
```
breathx: syntax sugar))) 🔥

ethexe/network/src/lib.rs
ark0f: It will be discussed and reimplemented later

PR #4895, opened 2025-10-07
feat(rpc): subscribe `UserMessageSent` `ProgramStateChanges`

### LINKED ISSUE #4893: node: Support subscribing to program state changes
### Problem to Solve

Each client need to fetch every event to filter what he needs. This is heavy on bandwidth and resources of node and client.

### Possible Solution

add rpc call to subscribe on program state changes.

### Notes

_No response_

ukint-vs:
Resolves #4893  .

RPC `subscribeProgramStateChanges`:
- Add typed `ProgramStateChange` stream support to gsdk.
- Implement `subscribe_program_state_changes` in pallet-gear rpc.
- filter by `program_ids: Option<Vec<H256>>`

notification structure:
```rust
pub struct ProgramStateChange {
    pub block_hash: H256,
    pub program_ids: Vec<H256>,
}
```

example:
```rust
use futures::StreamExt;
use gclient::{GearApi, ProgramStateChange};
use sp_core::H256;

#[tokio::main]
async fn main() -> gclient::Result<()> {
    let api = GearApi::dev().await?;
    let watch = vec![
        H256::from_slice(&hex_literal::hex!("0x00...")),
        H256::from_slice(&hex_literal::hex!("0x01...")),
    ];

    let mut changes = api.subscribe_program_state_changes(Some(watch)).await?;

    while let Some(change) = changes.next().await {
        let ProgramStateChange { block_hash, program_ids } = change?;
        println!("Tracked programs {:?} changed in block {:?}", program_ids, block_hash);
    }

    Ok(())
}

```
---

RPC `subscribeUserMessageSent`:
- Added server-side subscription gear_subscribeUserMessageSent that decodes System::Events once per block, enforces a 5k-block backfill ceiling, limits payload matching patterns to 256 bytes, and drops slow consumers via try_send.
- Exposed typed client support in gsdk (UserMessageSentFilter, PayloadFilter, UserMessageSentSubscription, new API call) with gclient re-exports.

Filter Reference:
```rust
UserMessageSentFilter {
  source: Option<ActorId>,         // match only this sender
  dest: Option<ActorId>,           // match only this recipient
  payload_filters: Vec<PayloadFilter>, // match byte patterns at specific offsets
  from_block: Option<u64>,         // backfill up to 5k blocks
  finalized_only: Option<bool>,    // true = finalized stream
}

PayloadFilter {
  offset: u32,      // byte offset into the payload
  pattern: Vec<u8>, // bytes that must match starting at that offset
}

```
UserMessageSentFilter::with_payload_filter(offset, pattern) adds an entry; with_payload_prefix(prefix) remains as a convenience for offset 0. Omitting fields leaves them unconstrained. from_block triggers historical scanning (subject to node retention).

Notification structure:
```rust
pub struct UserMessageSent {
    /// Block hash that emitted the event.
    pub block: H256,
    /// Index of the event within the block.
    pub index: u32,
    /// Identifier of the emitted message.
    pub id: MessageId,
    /// Message source actor.
    pub source: ActorId,
    /// Message destination actor.
    pub destination: ActorId,
    /// Raw payload bytes.
    pub payload: Vec<u8>,
    /// Attached value.
    pub value: u128,
}
```

Example
```rust
use gclient::{GearApi, PayloadFilter, UserMessageSentFilter};
use futures::StreamExt;
use gear_core::ids::ActorId;

let api = GearApi::dev().await?;
let my_actor = ActorId::try_from(api.account_id().as_ref()).unwrap();

let mut sub = api
    .subscribe_user_message_sent(
        UserMessageSentFilter::new()
            .with_destination(my_actor)
            .with_payload_filter(0, b"PONG")
            .finalized_only(true),
    )
    .await?;

while let Some(event) = sub.next().await.transpose()? {
    println!("reply {:?} with {} value", event.payload, event.value);
}

```

@reviewer-or-team

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -240,7 +240,7 @@ jobs:
             runs-on: macos-latest
           - name: macos-x86_64
             target: x86_64-apple-darwin
-            runs-on: macos-13
+            runs-on: macos-15-intel
           - name: linux-x86_64
             target: x86_64-unknown-linux-gnu
             runs-on: ubuntu-latest


pallets/gear/rpc/src/lib.rs
```diff
@@ -148,16 +172,28 @@ pub struct Gear<C, P> {
...
 impl<C, P> Gear<C, P> {
     /// Creates a new instance of the Gear Rpc helper.
-    pub fn new(client: Arc<C>, allowance_multiplier: u64, max_batch_size: u64) -> Self {
+    pub fn new(
+        client: Arc<C>,
+        allowance_multiplier: u64,
+        max_batch_size: u64,
+        subscription_executor: SubscriptionTaskExecutor,
+    ) -> Self {
+        let mut program_storage_prefix = twox_128(b"GearProgram").to_vec();
```
breathx, line 188: I'd put some warning on pallet or even unit test that name of a pallet ÷ storage slot weren't changed or were updated here

pallets/gear/rpc/src/lib.rs
```diff
@@ -618,4 +655,98 @@ where
             })
         }
     }
+
+    fn subscribe_program_state_changes(
+        &self,
+        pending: PendingSubscriptionSink,
+        program_ids: Option<Vec<H256>>,
```
breathx, line 662: we need some limit for the set of ids

node/service/src/rpc/gear_events.rs
```diff
@@ -0,0 +1,735 @@
...
+}
+
+#[derive(Clone, Debug, serde::Deserialize)]
+pub struct PayloadFilter {
+    pub offset: u32,
+    pub pattern: Vec<u8>,
+}
+
+#[derive(Clone, Debug, serde::Serialize)]
+pub struct UserMessageSentJson {
```
breathx, line 67: reply details are also maybe super useful

gsdk/src/subscription.rs
```diff
@@ -118,6 +125,211 @@ impl BlockEvents {
...
+    pub value: u128,
+}
+
+impl<'de> Deserialize<'de> for UserMessageSent {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        #[derive(Deserialize)]
+        struct RawUserMessageSent {
```
breathx, line 299: btw you may want to use rpc-compatible structs like: gear-core/rpc/RpcValue and sp_core::Bytes

breathx: maybe inside actual impl

gsdk/src/subscription.rs
```diff
@@ -118,6 +125,211 @@ impl BlockEvents {
...
+
+/// Filter options for `gear_subscribeUserMessageSent`.
+#[derive(Clone, Debug, Default, Serialize)]
+pub struct UserMessageSentFilter {
+    /// Only match messages originating from this actor.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub source: Option<H256>,
+    /// Only match messages targeting this actor.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub dest: Option<H256>,
```
breathx:
```suggestion
    pub destination: Option<H256>,
```

node/service/src/rpc/gear_events.rs
```diff
@@ -0,0 +1,788 @@
...
+    pub reply: Option<UserMessageReplyJson>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub ack: Option<bool>,
+}
+
+#[derive(Clone, Debug, serde::Serialize)]
+pub struct UserMessageReplyJson {
+    pub to: [u8; 32],
+    pub code_raw: Bytes,
+    pub code: String,
```
osipov-mit: Wouldn't `codeDescription` (or sth similar) name be more intuitive since it's not a `code` itself but its explanation/transcription (or whatever you can call it)?

ukint-vs: do you mean we should name `code_raw` to `code` and `code` -> `codeDescription`?

osipov-mit: Sth like this. That makes more sense, imho.

ukint-vs: done

PR #4899, opened 2025-10-09
chore(runtime): update weights

StackOverflowExcept1on:
`run_all_benchmarks.sh` script is applied to update weights (branch: `av/wasmer-v6`, commit 4d29f39bb631eb225655280ceb9106c101e14926)

**Note:** If CI fails, try manually updating the heuristic tests: `runtime/vara/src/tests/mod.rs`

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,5 @@
+[env]
+GEAR_WORKSPACE_DIR = { value = "", relative = true }
+
 [alias]
 gbuild = "run -p cargo-gbuild --"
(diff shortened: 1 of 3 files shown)


core/src/gas_metering/schedule.rs
```diff
@@ -1085,11 +1085,11 @@ impl Default for InstrumentationWeights {
     fn default() -> Self {
         Self {
             base: Weight {
-                ref_time: 308799000,
+                ref_time: 1226264561,
```
breathx: x4? :/

StackOverflowExcept1on: this is about 1 ms, and our block is 3 seconds
caused by other commit

breathx (approved):

PR #4900, opened 2025-10-09
feat(ethexe): Validator-only gossipsub topic

ark0f:


ethexe/common/src/crypto/signature.rs
```diff
@@ -174,6 +177,22 @@ impl<T: Sized> SignedData<T> {
     pub fn address(&self) -> Address {
         self.public_key.to_address()
     }
+
+    /// Verifies the data itself using provided public key and signature.
+    pub fn verify(&self) -> SignResult<()>
```
grishasobol: useless, cause signeddata always verified

ethexe/common/src/consensus.rs
```diff
@@ -18,16 +18,18 @@
...
 use gprimitives::CodeId;
 use k256::sha2::Digest as _;
 use parity_scale_codec::{Decode, Encode};
+use sha3::Keccak256;
 
-pub type SignedAnnounce = SignedData<Announce>;
-pub type SignedValidationRequest = SignedData<BatchCommitmentValidationRequest>;
+pub type VerifiedAnnounce = VerifiedData<Announce>;
+pub type VerifiedValidationRequest = VerifiedData<BatchCommitmentValidationRequest>;
+pub type VerifiedValidationReply = VerifiedData<BatchCommitmentValidationReply>;
```
grishasobol, line 32: NOTE:  BatchCommitmentValidationReply already contains signature inside, but this signature is a little bit different so does not suitable for signed data. BatchCommitmentValidationReply itself is suitable as any encodable data

ethexe/consensus/src/lib.rs
```diff
@@ -69,29 +70,26 @@ pub trait ConsensusService:
...
-    /// Outer service have to publish signed announce
-    PublishAnnounce(SignedAnnounce),
-    /// Outer service have to publish signed validation request
-    PublishValidationRequest(SignedValidationRequest),
-    /// Outer service have to publish signed validation reply
-    PublishValidationReply(BatchCommitmentValidationReply),
+    /// Outer service have to publish signed message
+    PublishMessage(SignedValidatorMessage),
     /// Informational event: commitment was successfully submitted, tx hash is provided
-    #[from(skip)]
```
grishasobol, line 92: from(skip) is for code readability - because it's completely unobviously that H256 means CommitmentSubmitted and any String is warning. For readability much better to show in code directly

ethexe/consensus/src/lib.rs
```diff
@@ -69,29 +70,26 @@ pub trait ConsensusService:
...
     fn receive_validation_reply(&mut self, reply: BatchCommitmentValidationReply) -> Result<()>;
 }
 
-#[derive(Debug, Clone, PartialEq, Eq, derive_more::From, derive_more::IsVariant)]
+#[derive(
+    Debug, Clone, PartialEq, Eq, derive_more::From, derive_more::IsVariant, derive_more::Unwrap,
+)]
 pub enum ConsensusEvent {
     /// Outer service have to compute announce
+    #[from]
```
grishasobol, line 87: Why do we need this from?

ark0f: To not use too automatic `#[from(skip)]`

grishasobol: Cannot understand, what do you mean too automatic?

breathx: otherwise macro will provide From(h256) and From(string) which is a little misleading

PR #4906, opened 2025-10-14
refactor(ethexe): move announces chain propagation to consensus

grishasobol:
additionally:
split compute service to subservices

### DIFF
--- a/ethexe/consensus/src/validator/core.rs
+++ b/ethexe/consensus/src/validator/core.rs
@@ -22,14 +22,13 @@ use crate::utils::{self, MultisignedBatchCommitment};
 use anyhow::{Result, anyhow, ensure};
 use async_trait::async_trait;
 use ethexe_common::{
-    Address, Announce, Digest, HashOf, SimpleBlockData, ToDigest, ValidatorsVec,
+    Address, Announce, Digest, HashOf, ProtocolTimelines, SimpleBlockData, ToDigest, ValidatorsVec,
     consensus::BatchCommitmentValidationRequest,
     db::BlockMetaStorageRO,
     ecdsa::PublicKey,
     gear::{
         BatchCommitment, ChainCommitment, CodeCommitment, RewardsCommitment, ValidatorsCommitment,
     },
-    ProtocolTimelines,
 };
 use ethexe_db::Database;
 use ethexe_ethereum::middleware::ElectionProvider;


ukint-vs:
@codex review

ethexe/common/src/mock.rs
```diff
@@ -167,13 +167,13 @@ pub struct PreparedBlockData {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
-pub struct BlockData {
+pub struct BlockFullData {
```
ark0f:
```suggestion
pub struct FullBlockData {
```
To be consistent with `SmallBlockData`

ethexe/compute/src/prepare.rs
```diff
@@ -16,53 +16,193 @@
...
+        if let State::WaitingForBlock = &self.state {
+            let Some(mut block_hash) = self.input.pop_back() else {
+                return Poll::Pending;
+            };
+
+            if !self.db.block_synced(block_hash) {
+                return Poll::Ready(Some(Err(ComputeError::BlockNotSynced(block_hash))));
+            }
+
+            let mut not_processed_blocks_chain = VecDeque::new();
```
kuzminxyz: I think it would be better to move this loop into another function to remove the complexity of `poll_next` implementation.

ethexe/compute/src/codes.rs
ecol-master: in my view of the compute service architecture - `codes` and `prepare` subservices should be one single service, because of prepare service in state of `WaitingForCodes` will do nothing. I think it reduce the complexity of code (for example `enum Event` in `prepare.rs`). 

And i think is great solution to split `compute` and `prepare` because of they able to do things in parallel and by design should be separated.

ethexe/consensus/src/validator/producer.rs
```diff
@@ -93,41 +98,40 @@ impl StateHandler for Producer {
...
-            future: self
-                .ctx
-                .core
-                .clone()
-                .aggregate_batch_commitment(self.block.clone())
-                .boxed(),
-        };
+                Ok(self.into())
+            }
+            State::WaitingAnnounceComputed(expected) => {
```
kuzminxyz, line 129: Please, remove second matching 🥲 to simple if else, much readable for me:

```rust
if *expected == announce_hash {
    self.state = State::AggregateBatchCommitment {
        future: self
            .ctx
            .core
            .clone()
            .aggregate_batch_commitment(self.block.clone(), announce_hash)
            .boxed(),
    };
} else {
    self.warning(format!(
        "Computed announce {announce_hash} is not expected, expected {expected}"
    ));
}
```

ethexe/compute/src/lib.rs
```diff
@@ -25,7 +25,8 @@ use std::collections::HashSet;
 mod compute;
 mod prepare;
 mod service;
-mod utils;
+// mod utils;
```
kuzminxyz: remove commented line

ethexe/compute/src/service.rs
```diff
@@ -239,55 +110,49 @@ impl<P: ProcessorExt> FusedStream for ComputeService<P> {
     }
 }
 
+trait SubService<Output>: Stream<Item = Result<Output>> + Unpin + Send + 'static {
```
kuzminxyz: Maybe should implement the `SubService` trait for subservices? it will help us to remove `SubServiceClosed` error which will never appear, but add unnecessary difficulties.

Just like this:

```rust
pub(crate) trait SubService {
    type Output;
    // in code
    // type Output = Result<AnnounceHash>;
    fn poll_next_some(&mut self, cx: &mut Context<'_>) -> Poll<Self::Output>;
}
```

All implementations will be the same, just only replace `impl Stream` on `impl SubService`.

ecol-master (changes requested):

ethexe/consensus/src/validator/producer.rs
grishasobol: It's impossible like you want 
1) `expected` is not defined
2) there is 3 cases - match is better for that

also match is more powerful code structure, and that could be useful  for future code changes

ethexe/compute/src/service.rs
grishasobol: Make sence

grishasobol: done [remove Stream impl for sub service; use trait SubService only](https://github.com/gear-tech/gear/pull/4906/commits/c406a925d27842bae072f7292bcbdc528749e381)

ethexe/compute/src/lib.rs
grishasobol: done [collect_not_prepared_blocks_chain()](https://github.com/gear-tech/gear/pull/4906/commits/ce680ec52beb6b1227554e404dd77655236e717d)

ethexe/compute/src/prepare.rs
grishasobol: done [collect_not_prepared_blocks_chain()](https://github.com/gear-tech/gear/pull/4906/commits/ce680ec52beb6b1227554e404dd77655236e717d)

ethexe/consensus/src/validator/subordinate.rs
```diff
@@ -207,6 +212,32 @@ impl Subordinate {
...
+            utils::parent_main_line_announce(&self.ctx.core.db, self.block.header.parent_hash)?;
+
+        if parent != announce.parent {
+            self.warning(format!(
+                "Received announce {announce:?} is from invalid branch, expected parent is {parent}",
+            ));
+            return Initial::create(self.ctx);
+        }
+
+        let announce_hash = self.ctx.core.db.set_announce(announce.clone());
```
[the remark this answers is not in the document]
grishasobol: done [remove to_hash() from subordinate](https://github.com/gear-tech/gear/pull/4906/commits/e2902eb622e927b1429310d153fbf7baa17df2c7)

ethexe/compute/src/prepare.rs
```diff
@@ -16,53 +16,204 @@
...
+            // this is the most efficient way of preparing blocks in case of multiple pending blocks.
+            let Some(block_hash) = self.input.pop_back() else {
+                return Poll::Pending;
+            };
+
+            if !self.db.block_synced(block_hash) {
+                return Poll::Ready(Err(ComputeError::BlockNotSynced(block_hash)));
+            }
+
+            let not_prepared_blocks_chain =
```
kuzminxyz, line 111: Maybe also should remove these blocks from `self.input`?

grishasobol: Better to keep guarantees that if block was sent to preparation then event BlockPrepared would be returned (or Error)

PR #4913, opened 2025-10-18
`ethexe`: store validators by era index

ecol-master:
Add storing validators by era index. Now `Observer` is responsible for tracking existing validators in database and if we can setup next validators to database as soon as possible - we just do it.

@grishasobol  @breathx

### DIFF
--- a/ethexe/network/src/lib.rs
+++ b/ethexe/network/src/lib.rs
@@ -36,6 +36,7 @@ use anyhow::{Context, anyhow};
 use ethexe_common::{
     Address,
     ecdsa::PublicKey,
+    injected::SignedInjectedTransaction,
     network::{SignedValidatorMessage, VerifiedValidatorMessage},
     tx_pool::SignedOffchainTransaction,
 };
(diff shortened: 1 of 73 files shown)


ethexe/consensus/src/validator/initial.rs
```diff
@@ -66,11 +66,18 @@ impl StateHandler for Initial {
     fn process_synced_block(self, block_hash: H256) -> Result<ValidatorState> {
         match &self.state {
             State::WaitingForSyncedBlock(block) if block.hash == block_hash => {
+                let timelines = self
```
kuzminxyz: use timelines directly from core.timelines

ethexe/network/src/lib.rs
```diff
@@ -189,8 +189,9 @@ impl NetworkService {
         } = config;
 
         let NetworkRuntimeConfig {
-            genesis_timestamp,
-            era_duration,
+            // TODO: remove this fields from config
```
kuzminxyz: remove these fields

ethexe/network/src/lib.rs
```diff
@@ -130,8 +130,6 @@ impl NetworkConfig {
 
 /// Config from other services
 pub struct NetworkRuntimeConfig {
-    pub genesis_timestamp: u64,
```
breathx, line 135: why don't you need timestamp anymore?

breathx (approved):

ethexe/network/src/lib.rs
kuzminxyz: > why don't you need timestamp anymore?

We don't need timestamp because of it stores in `ProtocolTimelines` (the same for era).

PR #4933, opened 2025-10-30
fix(ethexe/contracts): optimize gas when `_message.call`, fix bug with payload

StackOverflowExcept1on:


ethexe/contracts/src/Mirror.sol
```diff
@@ -374,14 +372,16 @@ contract Mirror is IMirror {
...
+            (bool success,) = _message.destination.call{gas: 500_000, value: _message.value}(payload);
 
             if (!success) {
                 /// @dev In case of failed call, we emit appropriate event to inform external users.
                 emit ReplyCallFailed(_message.value, _message.replyDetails.to, _message.replyDetails.code);
 
                 return;
             }
+        } else {
+            _transferEther(_message.destination, _message.value);
```
breathx, line 384: it was already transferred ??

ethexe/contracts/src/Mirror.sol
```diff
@@ -374,17 +372,19 @@ contract Mirror is IMirror {
                 );
             }
 
-            (bool success,) = _message.destination.call{gas: 500_000}(_message.payload);
+            (bool success,) = _message.destination.call{gas: 500_000, value: _message.value}(payload);
 
             if (!success) {
                 /// @dev In case of failed call, we emit appropriate event to inform external users.
                 emit ReplyCallFailed(_message.value, _message.replyDetails.to, _message.replyDetails.code);
```
breathx, line 381: In case of reply failure it still makes sense to perform transfer of ethers

PR #4934, opened 2025-10-30
feat(utils/gear-wasmer-cache): use native `file.lock/unlock()`

StackOverflowExcept1on:
https://blog.rust-lang.org/2025/08/07/Rust-1.89.0/


breathx (changes requested):
Why should we bump msv for no reason?

StackOverflowExcept1on:
@breathx https://doc.rust-lang.org/stable/std/fs/struct.File.html#method.lock `1.89`

breathx:
> @breathx https://doc.rust-lang.org/stable/std/fs/struct.File.html#method.lock `1.89`

I see, but doesn't current impl work the same way?

ark0f:
Agreed with @breathx about MSRV. I think removed `fs4` does not affect compile time sufficiently

PR #4943, opened 2025-11-04
refactor(gsdk): rework gsdk, part 1

liferooter:
- Update `subxt` to the latest version
- Adapt `gsdk` to `subxt` changes
- Simplify some parts of `gsdk` API
- Rework `gsdk` error types


common/src/gas_provider/node.rs
```diff
@@ -20,14 +20,30 @@ use super::*;
...
+    Debug,
+    Copy,
+    Clone,
+    Hash,
+    Eq,
+    PartialEq,
+    Ord,
+    PartialOrd,
+    Encode,
+    EncodeAsType,
```
ark0f, line 42: Exactly this type doesn't require `EncodeAsType` and `DecodeAsType`. Are they for the future PRs?

gsdk/src/result.rs
```diff
@@ -17,106 +17,86 @@
...
+
+pub use crate::tx_status::{TxError, TxStatusExt, TxSuccess};
+
+use subxt::ext::{scale_encode, subxt_rpcs};
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash, derive_more::Display)]
+#[display("Page {index} of Program {program}")]
+pub struct FailedPage {
+    pub index: u32,
+    pub program: String,
```
ark0f:
```suggestion
    pub program: ActorId,
```

gsdk/src/result.rs
```diff
@@ -17,106 +17,86 @@
...
+    StorageEntryNotFound,
+
+    #[error("subscription has been died")]
+    SubscriptionDied,
+
+    #[error("program has been terminated")]
+    ProgramTerminated,
+
+    #[error("{0} is invalid")]
+    PageInvalid(FailedPage),
```
ark0f:
```suggestion
    InvalidPage(FailedPage),
```

gsdk/src/tx_status.rs
```diff
@@ -0,0 +1,90 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+use crate::{TxInBlock, TxStatus};
+
+/// Sealed trait for [`TxStatus`].
+pub trait IsTxStatus {}
```
ark0f: Not actually sealed. I think `TxStatusExt` doesn't even need to be sealed

gsdk/src/utils.rs
```diff
@@ -97,9 +97,10 @@ pub(crate) fn write_storage_address_root_bytes(addr: &impl Address, out: &mut Ve
 pub(crate) fn storage_address_bytes(
     addr: &impl Address,
     metadata: &Metadata,
-) -> Result<Vec<u8>, Error> {
+) -> Result<Vec<u8>, Box<Error>> {
     let mut bytes = Vec::new();
     write_storage_address_root_bytes(addr, &mut bytes);
-    addr.append_entry_bytes(metadata, &mut bytes)?;
+    addr.append_entry_bytes(metadata, &mut bytes)
+        .map_err(|e| Box::new(e.into()))?;
```
ark0f, line 104: I guess box is here because of compiler lint?

gsdk/src/result.rs
liferooter: Done

gsdk/src/result.rs
liferooter: Done

gsdk/src/tx_status.rs
liferooter: It's actually sealed, since it's an inaccessible public trait used in public API. It should be sealed because the very nature of this trait requires it to be implemented only for `TxStatus` type, to making it sealed disallows to implement it for anything else.

gsdk/src/utils.rs
liferooter: Yes. We'll anyway eventually wrap it with `Box<_>` because we store `Box<subxt::Error>` in `gsdk::Error` to reduce the size of the error variant, but making it `Box<_>` on the level of this function makes the code more locally understandable, so the function doesn't look as a mistake even without knowing of the way it's used, which can be changed eventually.

It's a pity that `subxt::Error` is so huge and still used directly in `subxt`'s `Result`s, so we have to decide whether to wrap it on our side with `Box` every time or to have huge errors too.

common/src/gas_provider/node.rs
liferooter: Yes, looks like I accidentally commited to this branch something from the next part of the `gsdk` rework

liferooter: I've removed these changes from this PR

liferooter: It turned out that the changes are not really leaked from the next PR, but are a valid part of this PR. The new version of `subxt` requires slightly more traits to be derived.

PR #4947, opened 2025-11-05
feat(ethexe): Implement owned value handling

ByteNacked:
Resolves # .

-
-
-

@reviewer-or-team


ethexe/common/src/gear.rs
```diff
@@ -365,6 +365,7 @@ pub struct StateTransition {
     pub exited: bool,
     pub inheritor: ActorId,
     pub value_to_receive: u128,
+    pub value_to_receive_negative_sign: bool,
```
StackOverflowExcept1on, line 383: for reviewers: we represent `value_to_receive` as `u128` and `bool` because each non-zero byte costs 16 gas, and each zero byte costs 4 gas (see https://www.evm.codes/about#gascosts). negative numbers will be stored like this:
``` $ cast
> -1 ether
Type: int256
Hex: 0xfffffffffffffffffffffffffffffffffffffffffffffffffffffff21f494c589c0000
```
so this is optimization on eth side

ethexe/runtime/common/src/journal.rs
```diff
@@ -42,6 +42,22 @@ impl<S: Storage> NativeJournalHandler<'_, S> {
         dispatch: Dispatch,
         delay: u32,
     ) {
+        if delay.is_zero() && !dispatch.value.is_zero() {
+            let source = dispatch.source;
+            // Decrease sender's balance and value_to_receive
+            self.controller.update_state(source, |state, _, transitions| {
+                state.balance -= dispatch.value;
```
StackOverflowExcept1on:
```suggestion
                state.balance = state.balance.checked_sub(dispatch.value()).expect(
                    "Insufficient balance: underflow in state.balance -= dispatch.value()",
                );
```

ethexe/contracts/src/Router.sol
```diff
@@ -485,25 +485,45 @@ contract Router is IRouter, OwnableUpgradeable, ReentrancyGuardTransientUpgradea
...
+        for (uint256 i = 1; i < _transitions.length; i++) {
+            // If negative sign found after positive sign - transitions are not sorted.
+            if (!_transitions[i].valueToReceiveNegativeSign) {
+                isPositiveFound = true;
+                continue;
+            }
+
+            require(
+                !(isPositiveFound && _transitions[i].valueToReceiveNegativeSign),
+                "state transitions must be sorted by valueToReceiveNegativeSign in ascending order"
```
StackOverflowExcept1on:
```suggestion
                "state transitions must be sorted by valueToReceiveNegativeSign in descending order"
```

ByteNacked: I believe “from smaller to bigger” is called ascending order.

StackOverflowExcept1on: `valueToReceiveNegativeSign = true`
it should be `[true, true, false, false]`?

ByteNacked: Ohh, got it, you indeed correct

ethexe/service/src/tests/mod.rs
```diff
@@ -696,6 +696,281 @@ async fn value_in_reply() {
...
+        .unwrap();
+
+    let on_eth_balance = piggy_bank.get_balance().await.unwrap();
+    assert_eq!(on_eth_balance, VALUE_SENT);
+
+    let state_hash = piggy_bank.query().state_hash().await.unwrap();
+    let local_balance = node.db.program_state(state_hash).unwrap().balance;
+    assert_eq!(local_balance, VALUE_SENT);
+
+    env.approve_wvara(piggy_bank_id).await;
```
StackOverflowExcept1on: mb not needed bcz of reverse gas model & ETH value

ethexe/contracts/src/Router.sol
```diff
@@ -485,25 +485,45 @@ contract Router is IRouter, OwnableUpgradeable, ReentrancyGuardTransientUpgradea
...
+            if (!_transitions[i].valueToReceiveNegativeSign) {
+                isPositiveFound = true;
+                continue;
+            }
+
+            require(
+                !(isPositiveFound && _transitions[i].valueToReceiveNegativeSign),
+                "state transitions must be sorted by valueToReceiveNegativeSign in descending order"
+            );
+        }
```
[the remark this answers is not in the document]
ByteNacked: Fixed

ethexe/service/src/tests/mod.rs
ByteNacked: Fixed

ethexe/common/src/gear.rs
breathx: incode comment please

ethexe/contracts/src/Mirror.sol
```diff
@@ -166,11 +160,21 @@ contract Mirror is IMirror {
         implementationSlot.value = _abiInterface;
     }
 
-    // NOTE (breathx): value to receive should be already handled in router.
-    function performStateTransition(Gear.StateTransition calldata _transition) external onlyRouter returns (bytes32) {
+    // NOTE (breathx): value to receive should be already handled in router, expect if negative sign is set.
```
breathx: note may be removed in favor of proper doc

ethexe/contracts/src/MirrorProxy.sol
```diff
@@ -43,10 +43,6 @@ contract MirrorProxy {
         _delegate();
     }
 
-    function transferLockedValueToInheritor() external {
```
breathx, line 46: why was it removed from here?

ethexe/contracts/src/Router.sol
```diff
@@ -485,25 +485,45 @@ contract Router is IRouter, OwnableUpgradeable, ReentrancyGuardTransientUpgradea
...
+    }
+
     function _commitTransitions(Storage storage router, Gear.StateTransition[] calldata _transitions)
         private
         returns (bytes32)
     {
         bytes memory transitionsHashes;
 
+        _checkTransitionsSortedByValueSign(_transitions);
```
breathx: You don't need to iterate it all. I may have here mut bool and on each transition check that sign changed just once

ethexe/processor/src/handling/events.rs
```diff
@@ -128,6 +134,14 @@ impl ProcessingHandler {
                     };
 
                     transitions.modify_transition(actor_id, |transition| {
```
breathx: I'd consider moving it into separate function transitions.claim_value

ethexe/runtime/common/src/transitions.rs
```diff
@@ -185,13 +185,21 @@ impl InBlockTransitions {
...
+
             if !modification.is_noop(new_state.hash) {
                 res.push(StateTransition {
                     actor_id,
                     new_state_hash: new_state.hash,
                     exited: modification.inheritor.is_some(),
                     inheritor: modification.inheritor.unwrap_or_default(),
-                    value_to_receive: modification.value_to_receive,
+                    value_to_receive,
+                    value_to_receive_negative_sign,
```
breathx: much more laconic without code above 
```suggestion
                    value_to_receive: value_to_receive.unsigned_abs()
                    value_to_receive_negative_sign: value_to_receive < 0,
```

examples/value-sender/Cargo.toml
```diff
@@ -17,3 +17,4 @@ gear-wasm-builder.workspace = true
 debug = ["gstd/debug"]
 std = []
 default = ["std"]
+ethexe = ["gstd/ethexe"]
```
breathx: newline

ethexe/runtime/common/src/journal.rs
```diff
@@ -329,16 +365,40 @@ impl<S: Storage> JournalHandler for NativeJournalHandler<'_, S> {
...
-            state.balance += value;
-
-            transitions.modify_transition(to, |transition| transition.value_to_receive += value);
-        });
+        let src_is_prog = self.controller.transitions.is_program(&from);
+        let dst_is_prog = self.controller.transitions.is_program(&to);
+
+        match (src_is_prog, dst_is_prog) {
+            // User to Program or Program to Program value transfer
+            (true, true) | (false, true) => {
```
breathx: (_, true)

ethexe/contracts/src/MirrorProxy.sol
ByteNacked: @StackOverflowExcept1on cc

StackOverflowExcept1on: @breathx `transferLockedValueToInheritor` is called in `_setInheritor`, which is called in `performStateTransition`
https://github.com/gear-tech/gear/blob/39fae879926ece7ac747f2b11501d854fbd178e6/ethexe/contracts/src/Mirror.sol#L170-L185
What's the point of providing `external` method to users if we'll deliver this state to Ethereum and immediately call `transferLockedValueToInheritor`?

breathx: because value may still be locked on router, that belongs to program and user is able to retrieve it to inheritor

ethexe/contracts/src/Router.sol
grishasobol: I think it's maybe for checking before committing any transitions.

Nevertheless this check should be only in tests, because if most of validators are correct then transactions guarantied to be sorted, and if not most of validators are correct - then just checking sorting would not help.

ethexe/runtime/common/src/journal.rs
grishasobol: actually this is just
```rust
if dst_is_prog {
     // do transfer
} else {
     panic!("Transfer to user is not supported");
}
```

ethexe/runtime/common/src/journal.rs
```diff
@@ -42,6 +42,24 @@ impl<S: Storage> NativeJournalHandler<'_, S> {
         dispatch: Dispatch,
         delay: u32,
     ) {
+        if delay.is_zero() && !dispatch.value.is_zero() {
```
grishasobol: Whether we need to support value transfer in scheduled task in case delay is not zero?

ethexe/runtime/common/src/journal.rs
ByteNacked: Yep, but it doesn't provide any additional info in panic case

ethexe/runtime/common/src/journal.rs
ByteNacked: ~~Only decrease local program balance, no need to transfer value from router.~~

Hmm, let me double check

ethexe/processor/src/handling/events.rs
ByteNacked: Fixed

ethexe/runtime/common/src/journal.rs
ByteNacked: Fixed + added test 2df894ba736d182179bfaf1361c14dc7c0cabddc

PR #4950, opened 2025-11-08
build(nix): add some extra tools to `shell.nix`

liferooter:
- `nodejs`: needed for `ethexe` contracts
- `cargo-shear`: to check for unused dependencies, also used in CI


playX18 (approved):
LGTM. What do we need nodejs for though?

PR #4953, opened 2025-11-11
feat: Use `cargo-hakari`

ark0f:
https://docs.rs/cargo-hakari/latest/cargo_hakari/about/index.html#how-much-faster-do-builds-get

Build logs:

[With workspace hack](https://github.com/user-attachments/files/23962506/scratch_38.txt)

[Without workspace hack](https://github.com/user-attachments/files/23962493/scratch_39.txt)

Blocker:

- [x] https://github.com/gear-tech/polkadot-sdk/pull/10


.config/hakari.toml
```diff
@@ -0,0 +1,74 @@
...
+# Format version for hakari's output. Version 4 requires cargo-hakari 0.9.22 or above.
+dep-format-version = "4"
+
+# Setting workspace.resolver = "2" or higher in the root Cargo.toml is HIGHLY recommended.
+# Hakari works much better with the v2 resolver. (The v2 and v3 resolvers are identical from
+# hakari's perspective, so you're welcome to set either.)
+#
+# For more about the new feature resolver, see:
+# https://blog.rust-lang.org/2021/03/25/Rust-1.51.0.html#cargos-new-feature-resolver
+resolver = "2"
```
Lazark0x: Shouldn't we use the same resolver v3 as in the root Cargo.toml?

ark0f: Nice catch

PR #4956, opened 2025-11-12
refactor: get rid of manual codegen in `gsdk`

liferooter:
- Manually generate `.scale` file for `gsdk`, not code.
- Use `subxt::subxt` macro to generate code from `.scale` file during compilation.
- Use original types instead of generated in `gsdk` as much as possible.

### DIFF
--- a/.ethexe.example.local.toml
+++ b/.ethexe.example.local.toml
@@ -29,7 +29,7 @@
 # Ethereum address: 0x1a642f0e3c3af545e7acbd38b07251b3990914f1
 # ```
 # (optional, default: None).
-validator = "02ba5734d8f7091719471e7f7ed6b9df170dc70cc661ca05e688601ad984f068b0"
+validator = "0x02ba5734d8f7091719471e7f7ed6b9df170dc70cc661ca05e688601ad984f068b0"
 
 # Public key of the validator session, if node should act as one.
 # Provided in the format of 33-byte hex string:
@@ -49,7 +49,7 @@ validator = "02ba5734d8f7091719471e7f7ed6b9df170dc70cc661ca05e688601ad984f068b0"
 # Ethereum address: 0x1a642f0e3c3af545e7acbd38b07251b3990914f1
 # ```
 # (optional, default: None).
-validator-session = "02ba5734d8f7091719471e7f7ed6b9df170dc70cc661ca05e688601ad984f068b0"
+validator-session = "0x02ba5734d8f7091719471e7f7ed6b9df170dc70cc661ca05e688601ad984f068b0"
 
 # Max allowed height diff from head for sync directly from Ethereum.
 # (optional, must be non-zero u32, default: 100_000).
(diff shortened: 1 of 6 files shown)


gcli/src/cmd/info.rs
```diff
@@ -38,7 +38,7 @@ pub enum Action {
     Mailbox {
         /// The count of mails for fetching
         #[arg(default_value = "10", short, long)]
-        count: u32,
+        count: usize,
```
breathx, line 41: Why

gcli/src/cmd/program.rs
```diff
@@ -19,7 +19,7 @@
 //! Command `program`.
 use crate::{App, result::Result};
 use clap::Parser;
-use gclient::ext::utils::H256;
+use gclient::ext::subxt::utils::H256;
```
breathx, line 22: Gprimitives to be used

gsdk/scale-gen/src/main.rs
```diff
@@ -0,0 +1,101 @@
+// This file is part of Gear.
```
breathx: I'd consider appending dump of .scale into vara-runtime build.rs

gsdk/vara_runtime.scale
breathx, line 1: may be gitignored

gcli/src/cmd/info.rs
liferooter: Because the only use of that value is to call `.take(n)` on a stream

gcli/src/cmd/program.rs
liferooter: `H256` from `gprimitives` is from a different version of `primitive-types`

gsdk/vara_runtime.scale
liferooter: The reason we use a separate `.scale` file is not to depend on `vara-runtime` crate. Ignoring it will disable users of `gsdk` to use it without manually generating that file.

gsdk/scale-gen/src/main.rs
liferooter: Good idea!

ark0f:
@codex review

gsdk/src/constants.rs
```diff
@@ -18,54 +18,59 @@
...
-        self.query_constant("GearBank", "TreasuryGasFeeShare")
+        self.constant(&gear::constants().gear_bank().treasury_gas_fee_share())
+            .map(|p| p.0)
     }
 
     /// Get treasury tx fee percent.
     pub fn treasury_tx_fee_share(&self) -> Result<Percent> {
-        self.query_constant("GearBank", "TreasuryTxFeeShare")
+        self.constant(&gear::constants().gear_bank().treasury_tx_fee_share())
+            .map(|p| p.0)
```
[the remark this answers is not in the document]
liferooter: Wrong. `.map(|p| p.0)` unwraps `Static<Percent>` into `Percent`, not `Percent` into `u8`.

gclient/src/api/listener/mod.rs
```diff
@@ -19,20 +19,15 @@
 mod iterator;
 mod subscription;
 
-pub use gsdk::metadata::{Event, gear::Event as GearEvent};
+use std::collections::HashSet;
```
ark0f: Shoud be merged with other `use`

gsdk/src/result.rs
```diff
@@ -22,27 +22,9 @@ pub use crate::tx_status::{TxError, TxStatusExt, TxSuccess};
...
-
-    pub fn invalid(self) -> Error {
-        Error::InvalidPage(self)
-    }
-
-    pub fn not_found(self) -> Error {
-        Error::PageNotFound(self)
-    }
-}
+///
```
ark0f: Missing spacing

runtime/vara/build.rs
```diff
@@ -54,11 +55,99 @@ fn main() {
...
 #[cfg(not(feature = "std"))]
 fn main() {
     substrate_build_script_utils::generate_cargo_keys();
 }
+
+#[cfg(all(feature = "std", not(feature = "dev")))]
+fn regenerate_gsdk_scale() {}
+
+#[cfg(all(feature = "std", feature = "dev"))]
+fn regenerate_gsdk_scale() {
```
ark0f, line 72: Consider to use https://github.com/chevdor/subwasm

runtime/vara/build.rs
```diff
@@ -54,11 +55,99 @@ fn main() {
...
+            true,
+            "Metadata_metadata_versions",
+            &[],
+        )
+        .unwrap();
+    let versions = <Vec<u32>>::decode(&mut &versions[..]).unwrap();
+
+    // This value is taken from `subxt`, which
+    // unfortunately doesn't reexport it.
+    let supported_versions = [14, 15, 16];
```
ark0f: `frame-metadata` crate might be suitable

gcli/src/cmd/program.rs
liferooter: Added an explicit `TODO` for that.

runtime/vara/build.rs
liferooter: `subwasm` only has CLI, its library exists only for purpose of using it in CLI. I doubt that depending on external CLI tool is better than extra 50 lines of pretty obvious code.

runtime/vara/build.rs
liferooter: The comment was clarified

PR #4959, opened 2025-11-14
fix(ethexe): submission logic fixes; append mortality for batch commitment

grishasobol:


ethexe/consensus/src/announces.rs
```diff
@@ -413,24 +413,24 @@ fn propagate_one_base_announce(
     // 2. If it includes not committed and not-base announce, which is older than commitment delay limit.
     //
     // We check here till commitment delay limit, because T1 guaranties that enough.
-    let mut predecessor = parent_announce_hash;
+    let mut current_announce_hash = parent_announce_hash;
```
breathx, line 416: Is it current or predecessor than?

ark0f:
@codex review

ethexe/observer/src/sync.rs
```diff
@@ -203,6 +203,11 @@ impl<DB: SyncDB> ChainSync<DB> {
 
             self.db.set_block_synced(hash);
 
+            log::debug!(
```
ark0f:
```suggestion
            log::trace!(
```

ethexe/consensus/src/validator/mod.rs
```diff
@@ -28,16 +28,15 @@
 //!    |
 //!    ├────> Producer
 //!    |         └───> Coordinator
-//!    |                    └───> Submitter
```
ark0f, line 31: File `submitter.rs` is not removed

ethexe/consensus/src/validator/mod.rs
```diff
@@ -529,6 +544,10 @@ struct ValidatorContext {
     pending_events: VecDeque<PendingEvent>,
     /// Output events for outer services. Populates during the poll.
     output: VecDeque<ConsensusEvent>,
+
+    /// Ongoing consensus tasks, if any.
+    #[debug("{}", tasks.len())]
+    tasks: VecDeque<BoxFuture<'static, Result<ConsensusEvent>>>,
```
ark0f:
```suggestion
    tasks: FuturesUnordered<BoxFuture<'static, Result<ConsensusEvent>>>,
```
Or `FuturesOrdered` if order is important

ethexe/consensus/src/lib.rs
```diff
@@ -106,7 +114,8 @@ pub enum ConsensusEvent {
     #[from]
     RequestAnnounces(AnnouncesRequest),
     /// Informational event: commitment was successfully submitted, tx hash is provided
```
ark0f:
```suggestion
    /// Informational event: commitment was successfully submitted
```

ethexe/service/src/tests/utils/env.rs
```diff
@@ -161,6 +158,9 @@ impl TestEnv {
                 if let Some(genesis_timestamp) = genesis_timestamp {
                     anvil = anvil.arg(format!("--timestamp={genesis_timestamp}"));
                 }
+                if continuous_block_generation {
```
ark0f: Master should be merged

ethexe/consensus/src/utils.rs
```diff
@@ -286,17 +287,105 @@ pub fn create_batch_commitment<DB: BlockMetaStorageRO>(
             )
         })?;
 
+    let expiry = chain_commitment
+        .as_ref()
+        .map(|c| calculate_batch_expiry(db, block, c.head_announce, commitment_delay_limit))
+        .transpose()?
+        .and_then(|r| r)
```
ark0f:
```suggestion
        .flatten()
```

ethexe/contracts/src/libraries/Gear.sol
```diff
@@ -238,8 +245,10 @@ library Gear {
         );
     }
 
-    function blockIsPredecessor(bytes32 hash) internal view returns (bool) {
-        for (uint256 i = block.number - 1; i > 0;) {
+    function blockIsPredecessor(bytes32 hash, uint8 expiry) internal view returns (bool) {
+        uint256 start = block.number - 1;
+        uint256 end = expiry >= block.number ? 0 : block.number - expiry;
```
ark0f, line 250: `expiry >= block.number` is always false if network produced more than 255 blocks

ethexe/contracts/src/libraries/Gear.sol
```diff
@@ -238,8 +245,10 @@ library Gear {
...
+        for (uint256 i = start; i >= end;) {
```
ark0f, line 251:
```suggestion
        for (uint256 i = start; i > end;) {
```

If `end=0` and `i=0` then loop is infinite

ethexe/service/src/tests/utils/events.rs
```diff
@@ -120,39 +120,93 @@ impl ServiceEventsListener<'_> {
         id: impl Into<AnnounceId>,
     ) -> HashOf<Announce> {
         let id = id.into();
-        loop {
-            let event = self.next_event().await.unwrap();
+        log::info!("📗 waiting for announce computed: {id:?}");
```
ark0f, line 130:
```suggestion
        log::debug!("waiting for announce computed: {id:?}");
```

ethexe/service/src/tests/mod.rs
```diff
@@ -2789,3 +2820,273 @@ async fn announces_conflicts() {
...
+async fn catch_up_5() {
+    catch_up_test_case(5).await;
+}
+
+async fn catch_up_test_case(commitment_delay_limit: u32) {
+    init_logger();
+
+    assert!(
+        commitment_delay_limit == 3 || commitment_delay_limit == 5,
+        "Only 3 or 5 commitment delay limit is supported"
```
ark0f:
```suggestion
        "Only 3 or 5 commitment delay limit is supported for the test"
```

ethexe/contracts/src/libraries/Gear.sol
grishasobol: and?

ethexe/service/src/tests/utils/events.rs
grishasobol: it's informational test event. should be info

ethexe/ethereum/src/router/mod.rs
```diff
@@ -247,6 +247,18 @@ impl Router {
         commitment: BatchCommitment,
         signatures: Vec<ContractSignature>,
     ) -> Result<H256> {
+        self.commit_batch_pending(commitment, signatures)
```
kuzminxyz, line 250: I think, that here we just need to create pending builder, send transaction, receive its hash and return it into outer service, without waiting to receipt. Because of observer service will tell us whether batch was successfully send or not in next blocks. It will prevent the cases, when transaction was not accepted, but we will wait for its inclusion. 

My possible solution for implementation:
```rust
let pending_tx_builder = self.commit_batch_pending(commitment, signatures).await?;
let tx_hash = pending_tx_builder.inner().tx_hash();
Ok(tx_hash.into())
```

ethexe/consensus/src/validator/mod.rs
```diff
@@ -238,23 +244,34 @@ impl Stream for ValidatorService {
     type Item = Result<ConsensusEvent>;
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        let mut event = None;
         self.update_inner(|mut inner| {
-            // Waits until some event is available or inner futures are not ready.
+            // Waits until inner futures become pending.
             loop {
```
kuzminxyz, line 253: It's not critical for now, but if there are no pending futures, we will wait indefinitely and waste CPU resources.  
Maybe leave a TODO here to consider a redesign or another solution.

ethexe/contracts/src/libraries/Gear.sol
grishasobol: it must be i >= end

1) It would not be infinite becase blockhash(uint256::MAX) would revert immediately
2) This case can only happen if it's batch from reorg branch.

Nevertheless maybe it should be fixed for more correct error representation

ethexe/ethereum/src/router/mod.rs
grishasobol: Maybe yes, maybe no. Anyway this is not the point of this PR. In future we can use commit_batch_pending for submission in coordinator

ethexe/consensus/src/validator/mod.rs
grishasobol: What the waste of CPU resources? As soon as no futures, loop would break immediately

ethexe/consensus/src/validator/mod.rs
grishasobol: strange - maybe the result of incorrect master merging...

ethexe/contracts/src/libraries/Gear.sol
grishasobol: a it's even better - blockhash(0) is 0 and loop would break immediately - and then return false - what we want exactly

ethexe/consensus/src/announces.rs
grishasobol: current block in loop - common naming for iteration

ethexe/consensus/src/utils.rs
```diff
@@ -286,17 +287,105 @@ pub fn create_batch_commitment<DB: BlockMetaStorageRO>(
...
+        return Ok(None);
+    }
+
+    let mut oldest_not_base_announce_depth = (!head_announce.is_base()).then_some(0);
+    let mut current_announce_hash = head_announce.parent;
+
+    if announces_to_check_amount == 1 {
+        // If head announce is not base and older than commitment delay limit - 1, then expiry is only 1.
+        return Ok(oldest_not_base_announce_depth.map(|_| 1));
+    }
```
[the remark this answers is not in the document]
grishasobol: It's impossible because if head announce is commitment_delay_limit-1 blocks behind, then parent announce is committed or not propagated due to propagation rules (see propagate_announces)

ethexe/observer/src/sync.rs
grishasobol: done

ethexe/consensus/src/validator/mod.rs
grishasobol: done

ethexe/consensus/src/validator/mod.rs
grishasobol: done

ethexe/consensus/src/lib.rs
grishasobol: done

ethexe/service/src/tests/utils/env.rs
grishasobol: done

ethexe/consensus/src/utils.rs
grishasobol: done

ethexe/service/src/tests/mod.rs
grishasobol: done

ethexe/contracts/src/libraries/Gear.sol
ark0f: Can you add a comment? Seems to be not obvious to me

PR #4960, opened 2025-11-17
feat(ethexe/cli): add `--net` argument to use network keystore, print peer id

StackOverflowExcept1on:
this simplifies bootnodes setup

### DIFF
--- a/ethexe/cli/src/commands/key.rs
+++ b/ethexe/cli/src/commands/key.rs
@@ -80,7 +80,6 @@ impl KeyCommand {
                     .with_context(|| "failed to generate new keypair")?;
 
                 println!("Public key: {public}");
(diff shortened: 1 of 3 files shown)


ethexe/cli/src/commands/key.rs
```diff
@@ -72,6 +81,14 @@ impl KeyCommand {
 
                 println!("Public key: {public}");
                 println!("Ethereum address: {}", public.to_address());
+
+                if self.network {
```
breathx, line 84: wrap printlns from above under else condition (print them if it's not --network)

StackOverflowExcept1on: ok, I'll wrap only `Ethereum address`

PR #4966, opened 2025-11-17
test(ethexe-service): Implement whole network restore test

ark0f:


ethexe/blob-loader/src/lib.rs
```diff
@@ -59,7 +59,7 @@ pub enum BlobLoaderError {
     #[error("failed to get block by hash: {0}")]
     BlockNotFound(H256),
     #[error("failed to read blob bundle: {0}")]
-    ReadBlob(#[from] reqwest::Error),
```
StackOverflowExcept1on, line 62: why?

ethexe/service/src/tests/utils/env.rs
```diff
@@ -507,16 +511,7 @@ impl TestEnv {
...
-            self.observer_events_publisher()
-                .subscribe()
-                .await
-                .apply_until_block_event(|_| {
-                    blocks_count += 1;
-                    Ok((blocks_count >= blocks_amount).then_some(()))
-                })
-                .await
-                .unwrap();
+            time::sleep(self.block_time * blocks_amount).await;
```
StackOverflowExcept1on: how can you be sure without subscription? what if sub process is lagging?

ethexe/service/src/tests/mod.rs
```diff
@@ -2388,3 +2388,106 @@ async fn injected_tx_fungible_token() {
...
+        .unwrap()
+        .wait_for()
+        .await
+        .unwrap();
+    assert_eq!(res.code_id, ping_code_id);
+    assert_eq!(init_res.payload, b"");
+    assert_eq!(init_res.value, 0);
+    assert_eq!(init_res.code, ReplyCode::Success(SuccessReplyReason::Auto));
+
+    env.approve_wvara(ping_id).await;
```
StackOverflowExcept1on:
```suggestion

```
not needed

ethexe/service/src/tests/mod.rs
```diff
@@ -2388,3 +2388,106 @@ async fn injected_tx_fungible_token() {
...
+        .unwrap()
+        .wait_for()
+        .await
+        .unwrap();
+    assert_eq!(res.code_id, async_code_id);
+    assert_eq!(init_res.payload, b"");
+    assert_eq!(init_res.value, 0);
+    assert_eq!(init_res.code, ReplyCode::Success(SuccessReplyReason::Auto));
+
+    env.approve_wvara(async_id).await;
```
StackOverflowExcept1on:
```suggestion

```

ethexe/blob-loader/src/lib.rs
ark0f: Impossible to find where error occurred

ethexe/service/src/tests/utils/env.rs
ark0f: > what if sub process is lagging?

Non-continuous block generation doesn't have such guarantees, too. We only wait for blocks on RPC node

ethexe/service/src/tests/mod.rs
breathx: Useless approve

ethexe/service/src/tests/mod.rs
```diff
@@ -2388,3 +2388,106 @@ async fn injected_tx_fungible_token() {
...
+
+    let init_res = env
+        .send_message(res.program_id, ping_id.encode().as_slice(), 0)
+        .await
+        .unwrap()
+        .wait_for()
+        .await
+        .unwrap();
+    assert_eq!(res.code_id, async_code_id);
+    assert_eq!(init_res.payload, b"");
```
breathx, line 2927: Please assert message ids on each iteration to be sure this it's not repeated message

breathx (approved):

ethexe/service/src/tests/utils/env.rs
grishasobol: > Non-continuous block generation

Why? Anvil guaraties us that after evm mine call is done, blocks are already generated

What the reason of this code change? Skip blocks - is used in places in tests, where it's important to wait till blocks would be generated.

grishasobol: Nevertheless seems like there is a bug in old implementation - stream can have some stale block events inside which we must not take in account

ethexe/observer/src/lib.rs
```diff
@@ -90,18 +89,32 @@ impl Stream for ObserverService {
     type Item = Result<ObserverEvent>;
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        if self.sync_future.is_none()
+            && let Some(header) = self.block_sync_queue.pop_back()
+        {
+            self.sync_future = Some(self.chain_sync.clone().sync(header).boxed());
+        }
```
grishasobol: What the reason to do this before header stream? It's important to receive new blocks info from observer a soon as possible - in your changes you gives priority to events sync task

ethexe/service/src/tests/utils/events.rs
```diff
@@ -213,56 +214,62 @@ impl ObserverEventsListener {
...
 
-    pub async fn apply_until_block_event<R: Sized>(
+    pub async fn apply_until_block<R: Sized>(
         &mut self,
-        mut f: impl FnMut(BlockEvent) -> Result<Option<R>>,
+        mut f: impl FnMut(H256) -> Result<ControlFlow<R>>,
     ) -> Result<R> {
-        self.apply_until_block_event_with_header(|e, _h| f(e)).await
+        self.apply_until(|event| {
+            let ObserverEvent::BlockSynced(block_hash) = event else {
```
grishasobol: Better to apply antil ObserverEvent::Block().

ethexe/service/src/tests/utils/events.rs
```diff
@@ -213,56 +214,62 @@ impl ObserverEventsListener {
...
-            let header = self
-                .db
-                .block_header(block_hash)
-                .expect("Block header not found");
-            let events = self
-                .db
-                .block_events(block_hash)
-                .expect("Block events not found");
+        let db = self.db.clone();
+        self.apply_until_block(|block_hash| {
```
grishasobol: Here we must use then ObserverEvent::BlockSynced, if you would change apply_until_block

ethexe/observer/src/lib.rs
breathx: Syncing is a highest priority, why not?

grishasobol: Because chain-head info is more important for outer services - so that they could reset their states for a new head as soon as possible. Nevertheless, it's not very important, because sync will be pending on rpc request. But I prefer previous variant of implementaion

Issue #4963, opened 2025-11-17
`ethexe`: rewrite CLI interface


StackOverflowExcept1on:
### Problem to Solve

we have been using ETH as value for some time now, but we haven't updated our Ethereum client and CLI interface. this is also convenient to use for quickly testing all functionality on local node.

- [x] add `--dev` flag to node to simplify dapps development (#5025)
- [x] basic commands (`upload`, `create(-with-abi)`, `query`, `(owned|executable)-balance-top-up`, `send-message`) (#4968)
- [x] add `--watch` and `--json` flag to basic commands (#5040)
- [x] make output more human readable and show txs costs (#5053)
- [x] added `--call-reply` seems unnecessary since in 99% of cases the EOA has no code, remove it (#5067)
- [x] consider adjusting the number of pre-funded accounts in `--dev` mode, create ERC-20 tokens on these accounts (#5068)
- [x] add `--injected` flag for `send-message` command (#5089)
- [x] additional commands (`send-reply`, `claim-value`, `transfer-locked-value-to-inheritor`) (#5094)
- ~~another CLI command that can subscribe to Mirror contract logs and maybe even parse custom events from Solidity notation, maybe also so that it can filter logs?~~

### Possible Solution

add more commands to interact with local node, remove ERC20 approves

PR #4968, opened 2025-11-18
feat(ethexe/cli): revamp most of commands

StackOverflowExcept1on:
Partially solves #4963

### DIFF
--- a/ethexe/cli/Cargo.toml
+++ b/ethexe/cli/Cargo.toml
@@ -32,7 +32,7 @@ hex.workspace = true
 log.workspace = true
 libp2p-identity = { version = "0.2.12", features = ["secp256k1"] }
 serde.workspace = true
-sp-core = { workspace = true, features = ["serde"] }
+sp-core = { workspace = true, features = ["std"] }
 static_init.workspace = true
 tempfile.workspace = true
 tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
--- a/ethexe/service/src/tests/mod.rs
+++ b/ethexe/service/src/tests/mod.rs
@@ -174,7 +174,7 @@ async fn ping() {
     assert_eq!(res.code_id, code_id);
 
     let res = env
-        .send_message(res.program_id, b"PING", 0, false)
+        .send_message(res.program_id, b"PING")
         .await
         .unwrap()
         .wait_for()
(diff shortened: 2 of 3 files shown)


.github/workflows/check.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       - name: "Install: Foundry"
```
breathx, line 56: Why? As example it's better to have real value

ethexe/cli/src/commands/tx.rs
```diff
@@ -235,40 +338,122 @@ impl TxCommand {
...
-        legacy: bool,
+        call_reply: bool,
+        /// Flag to watch for reply from mirror. If false, command will do not wait for reply.
+        #[arg(short, long, default_value = "false")]
+        watch: bool,
     },
 }
+
+#[derive(Debug, Clone)]
+pub struct Payload(Vec<u8>);
```
breathx: You can use spcore Bytes here as well?

ethexe/service/src/tests/mod.rs
```diff
@@ -239,15 +237,15 @@ async fn uninitialized_program() {
     // Case #1: Init failed due to panic in init (decoding).
     {
         let res = env
-            .create_program(code_id, 500_000_000_000_000)
+            .create_program(code_id, H256::zero(), None, 500_000_000_000_000)
```
breathx: I'd split the most used call and call with full arguments and reused second in first one

.github/workflows/check.yml
StackOverflowExcept1on: ?

ethexe/cli/src/commands/tx.rs
StackOverflowExcept1on:
```
^^^^^^^ the trait `AsRef<[u8]>` is not implemented for `sp_core::Bytes`
```

ethexe/cli/Cargo.toml
```diff
@@ -32,9 +32,9 @@ hex.workspace = true
 log.workspace = true
 libp2p-identity = { version = "0.2.12", features = ["secp256k1"] }
 serde.workspace = true
+sp-core = { workspace = true, features = ["serde"] }
```
breathx: std feature

PR #4972, opened 2025-11-19
feat(`vara-eth/consensus`):  prohibit program initialization  through a injected transaction

ecol-master:
@breathx  @grishasobol 

## In this PR:
1. replace `TxValidity` and `TxValidityChecker` into separate module inside consensus - `tx_validation.rs`
2. Implement `TxValidity::UninitializedDestination` variant which will mean that injected message was sent to uninitialized actor. 
3. Update `injected_tx_fungible_token` test. 
4. Move tests from `tx_pool.rs` into `tx_validation.rs`

### DIFF
--- a/ethexe/consensus/src/tx_validation.rs
+++ b/ethexe/consensus/src/tx_validation.rs
@@ -39,7 +39,7 @@ pub enum TxValidity {
     NotOnCurrentBranch,
     /// Transaction's destination [`gprimitives::ActorId`] not found.
     UnknownDestination,
-    /// Transaction's destination [`gprimitives::ActorId`] not initialize.
+    /// Transaction's destination [`gprimitives::ActorId`] not initialized.
     UninitializedDestination,
 }
 
@@ -83,8 +83,11 @@ impl<DB: OnChainStorageRO + AnnounceStorageRO + Storage> TxValidityChecker<DB> {
         };
 
         let Some(state) = self.db.program_state(destination_state_hash.hash) else {
-            // TODO kuzmin: consider in future add a new `TxValidity` veriant or return TxValidity::UnknownDestination.
-            unreachable!("program state must be found by its valid hash.")
+            anyhow::bail!(
+                "program state not found for actor({}) by valid hash({})",
+                tx.data().destination,
+                destination_state_hash.hash
+            )
         };
 
         if state.requires_init_message() {


ethexe/consensus/src/tx_validation.rs
```diff
@@ -0,0 +1,332 @@
...
+    /// Transaction was already include into one of previous [`VALIDITY_WINDOW`] announces.
+    Duplicate,
+    /// Transaction is outdated and should be remove from pool.
+    Outdated,
+    /// Transaction's reference block not on current branch.
+    /// Keep tx in pool in case of reorg.
+    NotOnCurrentBranch,
+    /// Transaction's destination [`gprimitives::ActorId`] not found.
+    UnknownDestination,
+    /// Transaction's destination [`gprimitives::ActorId`] not initialize.
```
grishasobol:
```suggestion
    /// Transaction's destination [`gprimitives::ActorId`] not initialized.
```

ethexe/consensus/src/tx_validation.rs
```diff
@@ -0,0 +1,332 @@
...
+        if self.recent_included_txs.contains(&tx.data().to_hash()) {
+            return Ok(TxValidity::Duplicate);
+        }
+
+        let Some(destination_state_hash) = self.latest_states.get(&tx.data().destination) else {
+            return Ok(TxValidity::UnknownDestination);
+        };
+
+        let Some(state) = self.db.program_state(destination_state_hash.hash) else {
+            // TODO kuzmin: consider in future add a new `TxValidity` veriant or return TxValidity::UnknownDestination.
```
grishasobol:
```suggestion
            // TODO kuzmin: consider in future add a new `TxValidity` variant or return TxValidity::UnknownDestination.
```

ethexe/consensus/src/tx_validation.rs
```diff
@@ -0,0 +1,332 @@
...
+            unreachable!("program state must be found by its valid hash.")
```
grishasobol: it's not unreachable - it's error, because consensus service is not the only one who has access to the database.

ethexe/consensus/src/tx_validation.rs
grishasobol: Also this is not TxValidity::UnknownDestination, because if announce has hash in  program states, then this hash must be presented in CAS DB. Still it's not unreachable, and error must be returned in this case

grishasobol (approved):

PR #4978, opened 2025-11-20
fix: update metadata for prod

ukint-vs:
Resolves # .

-
-
-

@reviewer-or-team


runtime/vara/build.rs
```diff
@@ -66,7 +66,88 @@ fn main() {
 }
 
 #[cfg(all(feature = "std", not(feature = "dev")))]
```
breathx: it may be just 1 func with cfg feature for path inside

liferooter (approved):

PR #4981, opened 2025-11-21
fix!(core): make `PageBuf` use `[u64; N / 8]` instead of `[u8; N]`

liferooter:
Fixes the problem with polkadot-js, which doesn't work with arrays with length more than 2048.

Changes `PageBuf` type info, does not change its encoding.

### DIFF
--- a/core/src/memory.rs
+++ b/core/src/memory.rs
@@ -26,7 +26,6 @@ use alloc::{boxed::Box, format};
 use byteorder::{ByteOrder, LittleEndian};
 use core::{
     fmt::{self, Debug},
-    mem,
     ops::{Deref, DerefMut},
 };
 use numerated::{
@@ -108,12 +107,12 @@ pub struct IntoPageBufError;
 // to mitigate `polkadot-js` limitation of array length,
 // which requires it not to be more than 2048.
 const _: () = {
-    assert!((GearPage::SIZE as usize).is_multiple_of(mem::size_of::<u64>()));
-    assert!(GearPage::SIZE as usize / mem::size_of::<u64>() <= 2048);
+    assert!((GearPage::SIZE as usize).is_multiple_of(size_of::<u64>()));
+    assert!(GearPage::SIZE as usize / size_of::<u64>() <= 2048);
 };
 
 /// Alias for inner type of page buffer.
-pub type PageBufInner = Box<[u64; GearPage::SIZE as usize / mem::size_of::<u64>()]>;
+pub type PageBufInner = Box<[u64; GearPage::SIZE as usize / size_of::<u64>()]>;
 
 /// Buffer for gear page data.
 #[derive(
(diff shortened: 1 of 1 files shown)


core/src/memory.rs
```diff
@@ -167,7 +175,8 @@ impl PageBuf {
 
     /// Returns new page buffer filled with given byte.
     pub fn filled_with(byte: u8) -> PageBuf {
-        Self([byte; GearPage::SIZE as usize].into())
+        let chunk = u64::from_ne_bytes([byte; 8]);
```
breathx, line 177: Use le bytes here as codec is based on them

breathx: actually it doesnt matter, but for consistency

core/src/memory.rs
```diff
@@ -104,8 +104,16 @@ impl Debug for MemoryInterval {
...
+// to mitigate `polkadot-js` limitation of array length,
+// which requires it not to be more than 2048.
+const _: () = {
+    assert!(GearPage::SIZE.is_multiple_of(8));
+    assert!(GearPage::SIZE / 8 <= 2048);
+};
+
 /// Alias for inner type of page buffer.
-pub type PageBufInner = Box<[u8; GearPage::SIZE as usize]>;
+pub type PageBufInner = Box<[u64; GearPage::SIZE as usize / 8]>;
```
StackOverflowExcept1on: mb use `mem::size_of::<u64>` instead of `8`?

liferooter: Fixed by [b6bd2d6](https://github.com/gear-tech/gear/pull/4981/commits/b6bd2d693adc062e2c31f82592bd53af7cc0a1dc)

core/src/memory.rs
liferooter: We have codec that is based on le bytes and bytemuck's safe transmute, which obviously uses ne bytes. I used ne bytes for consistency with the latter, but as you've said, it doesn't matter.

liferooter: I think consistency with transmute is a better consistency, since it's more close abstraction layer to that code.

core/src/memory.rs
```diff
@@ -25,8 +25,8 @@ use crate::{
 use alloc::{boxed::Box, format};
 use byteorder::{ByteOrder, LittleEndian};
 use core::{
-    fmt,
-    fmt::Debug,
+    fmt::{self, Debug},
+    mem,
```
StackOverflowExcept1on: `size_of` is in prelude now
```rust
fn main() {
    dbg!(size_of::<u8>());
}
```
`use core::mem` is not needed

StackOverflowExcept1on (approved):

core/src/memory.rs
liferooter: Fixed by [cf34502](https://github.com/gear-tech/gear/pull/4981/commits/cf3450267dddd3ab6b15b351171335ab6d70d344)

PR #4983, opened 2025-11-21
refactor(ethexe-network): Separate verification logic from the validator list

ark0f:
Preparation for validator discovery

Also:
- Implement `ValidatorListSnapshot`
- Refactor `NetworkRuntimeConfig`
- Refactor `handle_*_event` functions in network service
- Don't use genesis validator set in validator list

### DIFF
--- a/ethexe/service/src/tests/utils/env.rs
+++ b/ethexe/service/src/tests/utils/env.rs
@@ -28,14 +28,14 @@ use crate::{
 };
 use alloy::{
     node_bindings::{Anvil, AnvilInstance},
-    providers::{Provider as _, ProviderBuilder, RootProvider, ext::AnvilApi},
+    providers::{ProviderBuilder, RootProvider, ext::AnvilApi},
     rpc::types::anvil::MineOptions,
 };
 use anyhow::Context;
 use ethexe_blob_loader::{BlobLoader, BlobLoaderService, ConsensusLayerConfig};
 use ethexe_common::{
-    Address, BlockHeader, COMMITMENT_DELAY_LIMIT, CodeAndId, DEFAULT_BLOCK_GAS_LIMIT,
-    SimpleBlockData, ToDigest, ValidatorsVec,
+    Address, COMMITMENT_DELAY_LIMIT, CodeAndId, DEFAULT_BLOCK_GAS_LIMIT, SimpleBlockData, ToDigest,
+    ValidatorsVec,
     ecdsa::{PrivateKey, PublicKey, SignedData},
     events::{BlockEvent, MirrorEvent, RouterEvent},
     network::{SignedValidatorMessage, ValidatorMessage},


ethexe/network/src/validator/list.rs
```diff
@@ -0,0 +1,226 @@
...
+
+        debug_assert_ne!(
+            chain_head_header.height, self.chain_head.header.height,
+            "`set_chain_head` called with the same chain head block"
+        );
+
+        let new_era = self.timelines.era_from_ts(chain_head_header.timestamp);
+        let old_era = self.timelines.era_from_ts(self.chain_head.header.timestamp);
+        if new_era <= old_era {
+            return Ok(None);
```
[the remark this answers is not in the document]
ark0f: @codex it is expected revalidation runs only on new era (e.g. new validator set) or when next validator set arrives before era changed

ethexe/ethereum/src/router/mod.rs
```diff
@@ -397,13 +397,10 @@ impl RouterQuery {
...
     }
 
-    pub async fn validators(&self) -> Result<Vec<LocalAddress>> {
-        self.instance
-            .validators()
-            .call()
-            .await
-            .map(|res| res.into_iter().map(|v| LocalAddress(v.into())).collect())
-            .map_err(Into::into)
+    pub async fn validators(&self) -> Result<ValidatorsVec> {
```
grishasobol: Please use validators_at(latest) instead of this

ethexe/network/src/validator/list.rs
```diff
@@ -0,0 +1,226 @@
...
+#[derive(Debug)]
+pub(crate) struct ValidatorListSnapshot {
+    pub chain_head_ts: u64,
+    pub timelines: ProtocolTimelines,
+    pub current_validators: ValidatorsVec,
+    pub next_validators: Option<ValidatorsVec>,
+}
+
+impl ValidatorListSnapshot {
+    pub(crate) fn is_older_era(&self, snapshot: &Self) -> bool {
```
grishasobol, line 42:
```suggestion
    /// Returns whether `self` era index is less than `snapshot` era index.
    pub(crate) fn is_older_era(&self, snapshot: &Self) -> bool {
```

ethexe/network/src/validator/list.rs
```diff
@@ -0,0 +1,226 @@
...
+
+    pub(crate) fn current_era_index(&self) -> u64 {
+        self.timelines.era_from_ts(self.chain_head_ts)
+    }
+
+    pub(crate) fn block_era_index(&self, block_ts: u64) -> u64 {
+        self.timelines.era_from_ts(block_ts)
+    }
+
+    pub(crate) fn contains_any_validator(&self, address: Address) -> bool {
```
grishasobol, line 57:
```suggestion
    /// Checks if the given address is present in either the current or next era validator set.
    pub(crate) fn contains_any_validator(&self, address: Address) -> bool {
```

ethexe/network/src/validator/list.rs
```diff
@@ -0,0 +1,226 @@
...
+    ) -> anyhow::Result<Option<Arc<ValidatorListSnapshot>>> {
+        let chain_head_header = self
+            .db
+            .block_header(chain_head)
+            .context("failed to get chain head block header")?;
+
+        debug_assert_ne!(
+            chain_head_header.height, self.chain_head.header.height,
+            "`set_chain_head` called with the same chain head block"
+        );
```
grishasobol: Why not equal? What if we receive two blocks with same height because of double announcements on ethereum? Of course this is restricted, but possible case.

ethexe/service/src/lib.rs
```diff
@@ -245,14 +242,21 @@ impl Service {
...
-                runtime_config,
-                &signer,
-                Box::new(RouterDataProvider(router_query)),
-                Box::new(db.clone()),
-            )
-            .with_context(|| "failed to create network service")?;
+            let lastest_block_data = observer
+                .latest_block()
+                .await
+                .context("failed to get lastest block")?;
```
grishasobol:
```suggestion
            let latest_block_data = observer
                .latest_block()
                .await
                .context("failed to get latest block")?;
```

ethexe/service/src/tests/utils/env.rs
```diff
@@ -1169,6 +1177,12 @@ impl Node {
             self.name
         );
 
+        let observer = ObserverService::new(&self.eth_cfg, u32::MAX, self.db.clone())
+            .await
+            .unwrap();
+        let latest_block = observer.latest_block().await.unwrap();
+        let latest_validators = observer.router_query().validators().await.unwrap();
```
grishasobol: please append a provider getter to router query and also we already have router_query in Node fields. Creating observer for such case looks dirty

ethexe/network/src/validator/list.rs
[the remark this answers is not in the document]
grishasobol: Why set_chain_head do not set chain head header in case block is not from next era?

PR #4984, opened 2025-11-22
refactor(gsdk): clean up `gsdk`

liferooter:
Clean up `gsdk` before merging with `gclient`.

### DIFF
--- a/gsdk/src/signer/mod.rs
+++ b/gsdk/src/signer/mod.rs
@@ -41,8 +41,8 @@ mod utils;
 /// Signer representation that provides access to gear API.
 /// Implements low-level methods such as [`run_tx`](`Inner::run_tx`)
 /// and [`force_batch`](`Signer.calls()::force_batch`).
-/// Other higher-level calls are provided by [`Signer::storage`],
-/// [`Signer::calls`], [`Signer::rpc`].
+/// Other higher-level calls are provided by [`Inner::storage`],
+/// [`Inner::calls`], [`Inner::rpc`].
 #[derive(Clone)]
 pub struct Signer(Arc<Inner>);
 


liferooter:
Rebased to master.

Merging with master is not very convenient because the branch I'm currently working on is forked from that branch and maintaining a good history makes my life much easier.

gsdk/src/signer/rpc.rs
```diff
@@ -23,14 +23,21 @@ use gear_core::{
...
     rpc::ReplyInfo,
 };
-use std::sync::Arc;
+use std::{borrow::Cow, sync::Arc};
 use subxt::utils::H256;
 
 /// Implementation of calls to node RPC for [`Signer`].
 #[derive(Clone)]
-pub struct SignerRpc(pub(crate) Arc<Inner>);
+pub struct SignerRpc<'a>(pub(crate) Cow<'a, Arc<Inner>>);
```
ark0f: No need to use refs, `Arc::clone()` is cheap

gclient/src/api/error.rs
```diff
@@ -100,66 +88,22 @@ pub enum Error {
...
-        if let SubxtError::Runtime(DispatchError::Module(m)) = e {
-            return m
-                .as_root_error()
-                .map_or_else(|e| Self::Subxt(Box::new(e)), Self::Runtime);
-        }
-
-        Error::Subxt(Box::new(e))
+impl From<gsdk::ext::subxt_core::Error> for Error {
+    fn from(err: gsdk::ext::subxt_core::Error) -> Self {
+        Self::GearSDK(SubxtError::from(err).into())
```
ark0f, line 95: Error structure is plain in Rust, so there should not be `.into()`

gsdk/src/convert.rs
```diff
@@ -81,3 +93,19 @@ impl<A: IntoSubxt, B> IntoSubxt for sp_runtime::MultiAddress<A, B> {
         }
     }
 }
+
+impl AsGear for EventDetails<GearConfig> {
+    type Target = crate::Event;
+
+    fn as_gear(&self) -> Result<Self::Target> {
+        Ok(self.as_root_event()?)
```
ark0f, line 101: `.as_gear()` and `.as_root_event::<Event>` are almost the same. Consider to remove `AsGear`

gsdk/src/utils.rs
```diff
@@ -28,20 +28,6 @@ use subxt::{
 };
 
 impl Api {
-    /// compare gas limit
-    pub fn cmp_gas_limit(&self, gas: u64) -> Result<u64> {
```
ark0f, line 32: I guess it is unused?

ark0f (changes requested):

gsdk/src/signer/rpc.rs
liferooter: Well, probably we don't need `Arc` here at all.

Removed both `Cow` and `Arc` from `Signer*` types in [fdc3440](https://github.com/gear-tech/gear/pull/4984/commits/fdc34407eb00f31b05c649578cdcd2ca927a5e68)

gsdk/src/convert.rs
liferooter: `.as_root_event::<gsdk::Event>()` requires specifying an event type to decode into, which is potentially more error-prone and less convenient than having an extra method that converts into the only needed and exactly correct type. The same is for `.as_root_error::<gsdk::gear::Error>()`.

gsdk/src/utils.rs
liferooter: Yes, it's unused and has very confusing and pretty useless semantics. Its signature can lead to the idea that it checks gas limit and throws an error if the limit is exceeded, but instead it logs a warning.

gclient/src/api/error.rs
liferooter: `.into()` wraps `subxt::Error` into `gsdk::Error` and is not just a wrapping into an enum variant.

gclient/src/api/error.rs
```diff
@@ -35,21 +31,13 @@ pub enum Error {
     Anyhow(#[from] AError),
     /// A wrapper around [`gsdk::Error`].
     #[error(transparent)]
-    GearSDK(GearSDKError),
+    GearSDK(#[from] GearSDKError),
```
ark0f, line 34:
```suggestion
    GearSdk(#[from] GearSDKError),
```
Abbreviations are not be capitalized in Rust namings

gsdk/src/signer/mod.rs
ark0f, line 51: Such design is actually quite odd because it is implementation detail, so better to proxy `Inner::rpc`, `Inner::storage` methods to `Signer` type itself

ark0f (approved):

gsdk/src/signer/mod.rs
liferooter: `Signer: Deref<Target = Inner>`, so these methods are available on `Signer` itself.

gclient/src/api/error.rs
liferooter: This name is not mine, it was before this PR and will be removed with the whole error type in the next PR.

PR #4988, opened 2025-11-24
chore(ethexe/contracts): use fractions instead of percentages for validator threshold

StackOverflowExcept1on:
most networks use 2/3 threshold and it doesn't make sense to keep percentage on Router


ethexe/contracts/src/Router.sol
```diff
@@ -165,10 +160,6 @@ contract Router is IRouter, OwnableUpgradeable, ReentrancyGuardTransientUpgradea
         return Gear.currentEraValidators(_router()).map[_validator];
     }
 
-    function signingThresholdPercentage() public view returns (uint16) {
```
grishasobol, line 166: what the reason of removing this?

StackOverflowExcept1on: I can add fraction here as `(uint256, uint256)`

grishasobol: Yes, I think it's better to change interface then removing it

StackOverflowExcept1on: done

PR #4992, opened 2025-11-24
fix(ethexe/ethereum): check if transaction was reverted; query revert reason from RPC

grishasobol:


ethexe/ethereum/src/lib.rs
```diff
@@ -280,6 +284,40 @@ impl TryGetReceipt<network::Ethereum> for PendingTransactionBuilder<network::Eth
...
+
+    async fn try_get_receipt_check_reverted(self) -> Result<TransactionReceipt> {
+        let provider = self.provider().clone();
+        let receipt = self.try_get_receipt().await?;
+
+        let try_request_error_reason = async |provider: RootProvider| {
+            let tx = provider
+                .get_transaction_by_hash(receipt.transaction_hash)
+                .await
+                .ok()??;
```
StackOverflowExcept1on, line 296: mb make it return result? (anyhow)

grishasobol: I do not think it's very important to catch this errors as soon as it's only informational call to represent information later in logs or something like that

PR #4993, opened 2025-11-24
chore(ci): Cache all crates by default

ark0f:


ark0f:
Workspace crate are not cached

PR #4994, opened 2025-11-25
fix(ethexe/blob-loader): remove service falling in case code loading error

grishasobol:


ethexe/service/src/lib.rs
```diff
@@ -123,6 +123,7 @@ impl Service {
             ethereum_rpc: config.ethereum.rpc.clone(),
             ethereum_beacon_rpc: config.ethereum.beacon_rpc.clone(),
             beacon_block_time: alloy::eips::merge::SLOT_DURATION,
+            attempts: NonZeroU8::new(3).unwrap(),
```
StackOverflowExcept1on: use `NonZero<u8>`: https://doc.rust-lang.org/stable/std/num/struct.NonZero.html

StackOverflowExcept1on: mb calculate it in const block to avoid unwrap in runtime

ethexe/blob-loader/src/lib.rs
```diff
@@ -110,73 +108,70 @@ struct ConsensusLayerBlobReader {
...
             .get_block_by_hash(block_hash)
             .await?
-            .ok_or(BlobLoaderError::BlockNotFound(H256(block_hash.0)))?;
+            .ok_or(BlockNotFound(H256(block_hash.0)))?;
 
-        // detect anvil by chain id
         let slot = if let Some(chain_id) = tx.chain_id()
             && chain_id == 31337
         {
+            // detect anvil by chain id
```
StackOverflowExcept1on: move comment back (before `let slot`)

ethexe/blob-loader/src/lib.rs
```diff
@@ -319,37 +324,32 @@ impl<DB: Database> BlobLoaderService for BlobLoader<DB> {
...
-    use ethexe_common::gear_core::ids::prelude::CodeIdExt;
 
     #[tokio::test]
-    async fn test_read_code_from_tx_hash() -> Result<()> {
+    async fn test_read_code_from_tx_hash() {
         let consensus_cfg = ConsensusLayerConfig {
             ethereum_rpc: "https://hoodi-reth-rpc.gear-tech.io".into(),
             ethereum_beacon_rpc: "https://hoodi-lighthouse-rpc.gear-tech.io".into(),
             beacon_block_time: Duration::from_secs(12),
+            attempts: NonZeroU8::new(3).unwrap(),
```
StackOverflowExcept1on:
```suggestion
            attempts: const { NonZeroU8::new(3).unwrap() },
```

ethexe/blob-loader/src/lib.rs
```diff
@@ -319,37 +324,32 @@ impl<DB: Database> BlobLoaderService for BlobLoader<DB> {
...
+
+        blobs_reader
+            .read_blob_from_tx_hash(
                 "0xee7f0082b6ad2fb1d409f39e5b169e102c27e4cf86b69a8a4006224cc91b4ae3"
                     .parse()
                     .unwrap(),
-                Some(3),
             )
-            .await?;
-        assert_eq!(code_id, expected_code_id);
```
StackOverflowExcept1on, line 350: point of this code was to take the blob data, calculate the hash (code_id) and compare it with the code_id that was specified in the transaction

grishasobol: code id is not calculated in blob loader any more - it's the task of processor which process the code and check the id and mark code as invalid if code id does not match provided code

ethexe/blob-loader/src/lib.rs
grishasobol: Strange is not a clippy rule

ethexe/service/src/lib.rs
grishasobol: As I can see NonZeroU8  ==  NonZero<u8>
image

ethexe/blob-loader/src/lib.rs
grishasobol: done

ethexe/blob-loader/src/lib.rs
grishasobol: done

ethexe/service/src/lib.rs
StackOverflowExcept1on: `NonZero<T>` is available since 1.79 so why not use it?

ethexe/blob-loader/src/lib.rs
StackOverflowExcept1on: in this case, you can calc code_id from blob content and compare it with the expected_code_id

breathx (approved):

PR #4996, opened 2025-11-26
feat(ethexe): append storage view and IntoBlockId

grishasobol:


ethexe/ethereum/src/lib.rs
```diff
@@ -340,3 +341,26 @@ macro_rules! signatures_consts {
 pub(crate) use signatures_consts;
 
 use crate::wvara::WVara;
+
+/// A helping trait for converting various types into `alloy::eips::BlockId`.
```
StackOverflowExcept1on, line 345: mb open issue in alloy?

StackOverflowExcept1on (approved):

ethexe/ethereum/src/lib.rs
StackOverflowExcept1on: what if use `Into<alloy::eips::BlockId>`? but their representation is `u64` and `FixedBytes<32>`, and we could have `H256` method for conversion to `FixedBytes<32>`

breathx (approved):

PR #5002, opened 2025-11-27
fix(ethexe): not processed codes commitment validation

grishasobol:


ethexe/consensus/src/utils.rs
```diff
@@ -132,14 +132,14 @@ pub fn aggregate_code_commitments<DB: CodesStorageRO>(
     db: &DB,
     codes: impl IntoIterator<Item = CodeId>,
     fail_if_not_found: bool,
-) -> Result<Vec<CodeCommitment>> {
+) -> Result<Vec<CodeCommitment>, CodeId> {
```
kuzminxyz: For me it is strange to return `CodeId` as an error. 

Maybe should return directly  ValidationStatus or create error type `pub struct CodeNotExists(CodeId)` and use it inside of ValidationStatus.

grishasobol: it's a common case in Rust. For example binary_search

https://doc.rust-lang.org/src/core/slice/mod.rs.html#2932-2934

grishasobol: replaced with error

PR #5006, opened 2025-11-29
fix(ethexe-network): `simultaneous_responses_limit` test timeouts

ark0f:


ethexe/network/src/db_sync/mod.rs
```diff
@@ -1156,7 +1163,10 @@ pub(crate) mod tests {
 
         let alice_config = Config::default().with_max_simultaneous_responses(2);
         let (mut alice, _alice_db, _data_provider) = new_swarm_with_config(alice_config).await;
-        let (mut bob, _bob_db, _data_provider) = new_swarm().await;
+        let bob_config = Config::default()
+            // make Bob to wait for Alice much more time to respond, otherwise Bob will drop its request
+            .with_per_request_timeout(Duration::from_hours(1));
```
grishasobol: Could you explain please why 1 hour ? Is it to be on the safe side only?

grishasobol (approved):

ethexe/network/src/db_sync/mod.rs
breathx: as far as I see, so it won't be dropped fast. It could be like 5 mins, actually doesn't matter in tests

ark0f: It's on Bob's side which is `tokio::spawn(bob.loop_on_next());`, so its timeout duration doesn't matter. I could set 1 month or more

PR #5007, opened 2025-12-01
chore: build metadata only if GENERATE_GSDK_METADATA set

ukint-vs:
Resolves # .

-
-
-

@reviewer-or-team


runtime/vara/build.rs
```diff
@@ -73,6 +81,10 @@ fn regenerate_gsdk_scale() {
     use sc_executor_common::runtime_blob::RuntimeBlob;
     use std::{env, fs, path::PathBuf};
 
+    if !should_regenerate_gsdk_metadata() {
+        return;
+    }
```
ark0f: Inline the while function and const and place below `if env::var(...`

PR #5008, opened 2025-12-02
feat(`ethexe-rpc`): Add simple testing for resource freeing after drop subscription

ecol-master:
@grishasobol

### DIFF
--- a/ethexe/rpc/src/tests.rs
+++ b/ethexe/rpc/src/tests.rs
@@ -123,7 +123,7 @@ async fn wait_for_closed_subscriptions(injected_api: InjectedApi) {
 #[tokio::test]
 #[ntest::timeout(20_000)]
 async fn test_cleanup_promise_subscribers() {
-    let listen_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 1728);
+    let listen_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 8002);
     let service = MockService::new(listen_addr).await;
     let injected_api = service.injected_api();
 
@@ -208,10 +208,10 @@ async fn test_cleanup_promise_subscribers() {
 }
 
 // Setup worker-threads=4 to simulate concurrent clients.
-#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+#[tokio::test]
 #[ntest::timeout(120_000)]
 async fn test_concurrent_multiple_clients() {
-    let listen_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 1729);
+    let listen_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::LOCALHOST), 8010);
     let service = MockService::new(listen_addr).await;
     let injected_api = service.injected_api();
 


ethexe/rpc/src/apis/injected.rs
```diff
@@ -37,18 +37,20 @@ pub enum InjectedTransactionAcceptance {
...
 }
 
-#[cfg_attr(not(feature = "client"), rpc(server))]
-#[cfg_attr(feature = "client", rpc(server, client))]
+#[cfg_attr(not(feature = "client"), rpc(server, namespace = "injected"))]
+#[cfg_attr(feature = "client", rpc(server, client, namespace = "injected"))]
 pub trait Injected {
-    #[method(name = "injected_sendTransaction")]
+    /// Just sends an injected transaction.
+    #[method(name = "sendTransaction")]
```
grishasobol, line 45: Naming change must be approved by js part @osipov-mit @breathx

ethexe/rpc/src/apis/injected.rs
```diff
@@ -83,6 +88,11 @@ impl InjectedApi {
             tracing::trace!(promise = ?promise, "rpc promise receiver dropped");
         }
     }
+
+    /// Returns the number of current promise subscribers waiting for promises.
+    pub fn promise_subscribers_count(&self) -> usize {
```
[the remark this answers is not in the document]
grishasobol: if only for testing - use #[cfg(test)]

ethexe/rpc/src/lib.rs
```diff
@@ -89,6 +92,9 @@ impl RpcServer {
 
         let server = Server::builder()
             .set_http_middleware(http_middleware)
+            // Setup WebSocket pings to detect dead connections.
+            // Now it is set to default: ping_interval = 30s, inactive_limit = 40s
+            .enable_ws_ping(PingConfig::default())
```
grishasobol, line 97: interesting wether ping protocol is supported by any client by default, or another words whether any js user for example would successfully interact with rpc for now, or something should be done on js part also

ethexe/rpc/src/apis/injected.rs
kuzminxyz: The methods with namespace will be created like this: "{namespace}_{methodName}", so there are no changes in naming)))

In macro code (https://docs.rs/jsonrpsee-proc-macros/0.26.0/src/jsonrpsee_proc_macros/rpc_macro.rs.html):
```rust
/// Based on the namespace and separator, renders the full name of the RPC method/subscription.
	/// Examples:
	/// For namespace `foo`, method `makeSpam`, and separator `_`, result will be `foo_makeSpam`.
	/// For separator `.`, result will be `foo.makeSpam`.
	/// For no namespace, returns just `makeSpam`.
	pub(crate) fn rpc_identifier<'a>(&self, method: &'a str) -> Cow<'a, str> {
		if let Some(ns) = &self.namespace {
			let sep = self.namespace_separator.as_deref().unwrap_or("_");
			format!("{ns}{sep}{method}").into()
		} else {
			Cow::Borrowed(method)
		}
	}
```


Cargo expand: 
```rust
            fn into_rpc(self) -> jsonrpsee::RpcModule<Self> {
                let mut rpc = jsonrpsee::RpcModule::new(self);
                {
                    let _res = rpc
                        .register_async_method(
                            "injected_sendTransaction",
                            |params, context, _| async move {
                                let (transaction) = if params.is_object() {
                                    #[serde(
                                        crate = "jsonrpsee :: core :: __reexports :: serde"
                                    )]
```

breathx (approved):

PR #5025, opened 2025-12-10
feat(ethexe/cli): add `--dev` flag to run node with predefined accounts

### LINKED ISSUE #4964: `ethexe`: use `FromStr` wherever possible in CLI and configs
### File Location(s)

ethexe/cli/src/commands/tx.rs
ethexe/cli/src/params/mod.rs
ethexe/cli/src/params/node.rs
ethexe/cli/src/params/ethereum.rs
ethexe/cli/src/params/network.rs

### Proposal

since `std::error::Error` is now in `core`, we can use `CodeId`, `ActorId`, and other primitives along with `serde` and `clap`, which removes boilerplate code

StackOverflowExcept1on:
Resolves #4964
Also: deploying contracts to local node is now significantly faster. Maybe we should still use `--slow` when deploying to testnet since it consumes quite a lot of gas and can take up almost the entire block?
`cargo run --package ethexe-cli --release -- run --dev --block-time 1`

### DIFF
--- a/ethexe/cli/src/commands/run.rs
+++ b/ethexe/cli/src/commands/run.rs
@@ -82,18 +82,11 @@ impl RunCommand {
             node.validator = Some(validator_public_key.to_string());
             node.validator_session = Some(validator_public_key.to_string());
 
-            if let Some(ethereum) = self.params.ethereum.as_mut() {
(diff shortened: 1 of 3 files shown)


breathx (approved):
I have a little worry that anvil won't be dropped at some point. Also, don't we need old blobs loading anymore?

PR #5044, opened 2025-12-17
fix!: impl SignedMessage according to EIP-191 for injected tx

breathx:


ethexe/common/src/crypto/signature.rs
```diff
@@ -479,4 +649,45 @@ mod tests {
 
         assert_eq!(contract_signature, decoded);
     }
+
+    #[cfg(feature = "std")]
+    #[test]
+    fn signed_message_and_injected_transactions() {
+        use crate::injected::InjectedTransaction;
+
+        const RPC_INPUT: &str = "{
```
StackOverflowExcept1on, line 658: use raw literals `r#"..."`

StackOverflowExcept1on (approved):

ethexe/common/src/crypto/signature.rs
breathx: Some time soon fix please

PR #5053, opened 2025-12-29
feat(ethexe-cli): make command output more detailed in table format

ukint-vs:
This PR adds more logs and makes output of commands for interacting with Vara.eth much more readable

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -323,12 +323,15 @@ jobs:
         uses: actions/checkout@v4
 
       - name: "Install: Setup linker"
+        if: matrix.name != 'macos-x86_64'
         uses: ./.github/actions/setup-linker
 
       - name: "Install: Rust toolchain"
+        if: matrix.name != 'macos-x86_64'
         uses: ./.github/actions/install-rust
 
       - name: "ACTIONS: Setup caching"
+        if: matrix.name != 'macos-x86_64'
         uses: ./.github/actions/rust-cache
         with:
           us-access-key-id: "${{ secrets.GEAR_CI_S3_US_ACCESS_KEY_ID }}"
@@ -337,7 +340,9 @@ jobs:
           eu-secret-access-key: "${{ secrets.GEAR_CI_S3_EU_SECRET_ACCESS_KEY }}"
           key: "${{ matrix.name }}"
 
-      - run: cargo run -p calc-stack-height --release --locked
+      - name: "Run: Calculate stack height"
+        if: matrix.name != 'macos-x86_64'
+        run: cargo run -p calc-stack-height --release --locked
 
   gsdk-metadata:
     runs-on: ubuntu-latest


ethexe/cli/src/commands/tx.rs
```diff
@@ -596,6 +1019,121 @@ impl TxCommand {
...
+
+fn print_address_with_explorer(label: &str, chain_id: u64, address: Address) {
+    eprintln!("{label}: {address}");
+    if let Some(url) = explorer_address_link(chain_id, address) {
+        eprintln!("{label} explorer: {url}");
+    }
+}
+
+fn explorer_base(chain_id: u64) -> Option<&'static str> {
+    match chain_id {
```
StackOverflowExcept1on: https://docs.rs/alloy-chains/latest/alloy_chains/enum.NamedChain.html

StackOverflowExcept1on: I would remove networks other than mainnet, sepolia, and hoodi

ethexe/cli/src/commands/tx.rs
```diff
@@ -60,6 +60,40 @@ struct TopUpResult {
     formatted_value: String,
 }
 
+#[derive(Debug, Clone)]
+struct UploadResultData {
```
StackOverflowExcept1on: It would make sense to sort the fields in these structs by some criterion (for example, first output data, then optional `explorer_url`, `block_number`)

StackOverflowExcept1on: this should also implement `Serialize` and you can use structs instead of `json!` since macro will break order of fields

ethexe/cli/src/commands/tx.rs
```diff
@@ -596,6 +1045,121 @@ impl TxCommand {
...
+fn print_address_with_explorer(label: &str, chain_id: u64, address: Address) {
+    eprintln!("{label}: {address}");
+    if let Some(url) = explorer_address_link(chain_id, address) {
+        eprintln!("{label} explorer: {url}");
+    }
+}
+
+fn explorer_base(chain_id: u64) -> Option<&'static str> {
+    if chain_id == 31337 {
+        return Some("https://etherscan.io"); // TODO: remove this
```
StackOverflowExcept1on: this is used for testing and should be removed

ethexe/cli/src/commands/tx.rs
```diff
@@ -459,12 +752,14 @@ impl TxCommand {
...
-                        "Topping up executable balance of mirror on Ethereum at {mirror} by {formatted_value} ({raw_value})"
-                    );
+                    eprintln!("Topping up executable balance of mirror on Ethereum:");
+                    eprintln!("  Mirror: {mirror}");
+                    eprintln!("  Value:  {formatted_value} ({raw_value})");
+                    eprintln!();
 
                     let mirror = ethereum.mirror(mirror);
 
+                    // TODO: consider to get receipt from approve tx as well
```
StackOverflowExcept1on, line 756: we currently do not include approval fees in output

ethexe/cli/src/commands/tx.rs
```diff
@@ -147,56 +205,154 @@ impl TxCommand {
             .await
             .with_context(|| "failed to create Ethereum client")?;
 
+        eprintln!("RPC:      {rpc}");
+        if let TxSubcommand::Query { rpc_url, .. } = &self.command {
+            eprintln!("WS RPC:   {rpc_url}");
+        }
```
grishasobol, line 211: I do not know exactly, but I suppose it's possible to print with trailing spaces
```suggestion
        eprintln!("{info:10} {rpc}",  info = "RPC:");
        if let TxSubcommand::Query { rpc_url, .. } = &self.command {
            eprintln!("{info:10} {rpc_url}", info = "WS RPC:");
        }
```

PR #5064, opened 2026-01-13
fix(ethexe-runtime): Do not brick program on OOG init

### LINKED ISSUE #5047: `ethexe`: Refactor `DispatchOutcome::InitFailure` to support structured reasoning and prevent program "bricking" on OOG
## Problem Description
Currently, when a program is initialized with zero or insufficient executable balance, the initialization dispatch fails with `RanOutOfGas`. In the Gear runtime (and specifically in `ethexe`), any `InitFailure` is mapped to `Program::Terminated`.

This results in the program being permanently "bricked." Even if a user later tops up the executable balance, the program remains in the `Terminated` state and cannot be initialized again. While `Terminated` is the correct state for fatal initialization errors (e.g., code panics, invalid logic), it is too strict for funding-related errors which should be retryable.

## Technical Root Cause
The `DispatchOutcome::InitFailure` variant currently uses an opaque `String` to describe the failure reason:
```rust
InitFailure {
    program_id: ActorId,
    origin: ActorId,
    reason: String, // Opaque diagnostic string
}
```
Because the reason is a string, the journal handler cannot reliably distinguish between a transient funding issue (Out of Gas) and a fatal logic error without brittle string matching.

## Affected Components
- `core-processor`: Definition of outcomes and error mapping.
- `ethexe-runtime-common`: Journal handling logic.
- `pallets-gear`: Substrate pallet logic and benchmarking.
- `gtest`: Testing framework simulation.

## Impact
- **UX Improvement:** Users who mis-fund their program initialization can top up and retry without losing the program address.
- **Robustness:** Eliminates the need for brittle string-matching heuristics in the runtime.

### Steps

1. Create new program
2. Send init message -> `RanOutOfGas`
3. New messages will fail and program will be bricked.

### Possible Solution

Refactor the initialization failure logic to use a structured `InitFailureReason` enum. This allows the runtime journal handler to make semantic decisions about whether to terminate the program or keep it active for a retry.

1. **Introduce `InitFailureReason` enum in `core-processor`:**
   ```rust
   pub enum InitFailureReason {
       RanOutOfGas,
       UserspacePanic(String),
       BackendError(String),
       Other(String),
   }
   ```
2. **Update `DispatchOutcome::InitFailure`** to use this enum instead of `String`.
3. **Update `core-processor` mapping logic** to populate the correct variant based on the execution result.
4. **Update Journal Handlers (`pallets-gear`, `ethexe`, `gtest`)** to handle the structured reason. Specifically, `RanOutOfGas` should allow the program to remain `Active` (uninitialized), while other variants should lead to `Terminated`.

### Notes

_No response_

### Relevant Log Output

<p>

```
<Paste your log here>
```

</p>
</details>

ukint-vs:
Resolves #5047 .
- Added structured `InitFailureReason` and propagated it through core-processor
- Updated runtime handling to keep uninitialized programs active on funding/allowance init failures
@reviewer-or-team

### DIFF
--- a/ethexe/runtime/common/src/journal.rs
+++ b/ethexe/runtime/common/src/journal.rs
@@ -263,7 +263,7 @@ impl<S: Storage> JournalHandler for NativeJournalHandler<'_, S> {
 
             self.controller
                 .transitions
-                .maybe_store_injected_reply(&message_id, reply_info);
+                .maybe_store_injected_reply(message_id, reply_info);
         }
 
         if self.controller.transitions.is_program(&destination) {
--- a/.github/actions/install-cargo-xwin/action.yml
+++ b/.github/actions/install-cargo-xwin/action.yml
@@ -12,7 +12,7 @@ runs:
       uses: cargo-bins/cargo-binstall@main
 
     - name: "Install: cargo-xwin"
-      run: cargo +stable binstall --no-confirm cargo-xwin@0.20.2
+      run: cargo +stable binstall --no-confirm cargo-xwin@0.21.2
       shell: bash
 
     - run: |
--- a/.github/actions/rust-cache/action.yml
+++ b/.github/actions/rust-cache/action.yml
@@ -1,12 +1,6 @@
 name: rust-cache
 
 inputs:
-  us-access-key-id:
-    required: true
-    description: US S3 access key ID
-  us-secret-access-key:
-    required: true
-    description: US secret access key
   eu-access-key-id:
     required: true
     description: EU S3 access key ID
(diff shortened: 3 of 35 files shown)


ethexe/runtime/common/src/journal.rs
```diff
@@ -597,9 +597,20 @@ where
                 origin,
                 reason,
             } => {
-                log::trace!("Dispatch {message_id} failed init of program {program_id}: {reason}");
+                log::trace!(
+                    "Dispatch {message_id} failed init of program {program_id}: {:?}",
```
StackOverflowExcept1on:
```suggestion
                    "Dispatch {message_id} failed init of program {program_id}: {reason:?}",
```

StackOverflowExcept1on (approved):

PR #5066, opened 2026-01-14
refactor(core-processor): convert `ProcessErrorCase` into `Payload` instead of cloning

### LINKED ISSUE #4594: Consider to convert `self` into `Payload` to avoid `PanicBuffer` cloning
### File Location(s)

`core-processor/src/processing.rs`

### Proposal

`ProcessCaseError::to_payload()` clones `PanicBuffer` inside because `ProcessCaseError` maybe be used further to print error message for `DispatchOutcome`

liferooter:
Closes #4594.

### DIFF
--- a/.ethexe.example.local.toml
+++ b/.ethexe.example.local.toml
@@ -15,6 +15,10 @@
 # (optional, default: false).
 # dev = false
 
+# Number of pre-funded accounts to generate in dev mode.
+# (optional, must be non-zero u32, default: 10).
+# pre-funded-accounts = 10
+
 # Public key of the validator, if node should act as one.
 # Provided in the format of 33-byte hex string:
 # - `yParityCompressed`, compressed `publicKeyY`,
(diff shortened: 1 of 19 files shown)


core-processor/src/processing.rs
```diff
@@ -247,10 +247,11 @@ impl ProcessErrorCase {
             ProcessErrorCase::ExecutionFailed(reason) => reason.as_simple().into(),
         }
     }
+}
 
-    // TODO: consider to convert `self` into `Payload` to avoid `PanicBuffer` cloning (#4594)
-    fn to_payload(&self) -> Payload {
-        match self {
+impl From<ProcessErrorCase> for Payload {
```
ark0f: I'd make `into_payload` method because such conversion and its semantics are not obvious

ark0f (changes requested):

core-processor/src/processing.rs
liferooter: Fixed by https://github.com/gear-tech/gear/pull/5066/commits/27160ee853bd0c5e8dcc78eb522eda87f88e4b28

PR #5067, opened 2026-01-14
fix(ethexe/cli): remove the `--call-reply` flag for EOA

StackOverflowExcept1on:


ethexe/ethereum/src/mirror/mod.rs
```diff
@@ -180,8 +178,8 @@ impl Mirror {
         &self,
         payload: impl AsRef<[u8]>,
         value: u128,
-        call_reply: bool,
     ) -> Result<PendingTransactionBuilder<network::Ethereum>> {
+        let call_reply = false;
```
kuzminxyz, line 182: Can you pls remind me why we  remove `call_reply` flag?

StackOverflowExcept1on: Because `call_reply` affects how the reply is sent to `destination`: as `Reply` event or as callback function call on destination. However, we work as EOA (externally owned account, account with private key). EOA doesn't have code in 99% of cases except for EIP-7702: https://eips.ethereum.org/EIPS/eip-7702

PR #5077, opened 2026-01-15
refactor: unify `ChargeResult` usage

liferooter:
Add `ChargeResult::is_enough` and `ChargeResult::is_not_enough` methods and use them instead of (in)equality check with a random one of `ChargeResult::Enought` and `ChargeResult::NotEnought`.

### DIFF
--- a/.ethexe.example.local.toml
+++ b/.ethexe.example.local.toml
@@ -15,6 +15,10 @@
 # (optional, default: false).
 # dev = false
 
+# Number of pre-funded accounts to generate in dev mode.
+# (optional, must be non-zero u32, default: 10).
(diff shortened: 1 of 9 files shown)


core/src/gas.rs
ark0f, line 42:
```suggestion
#[derive(Debug, Clone, Copy)]
```
To not face such comparisons again

liferooter: Well, technically such comparison makes sense. Also they can be helpful for some non-trivial combinations of `ChargeResult`s, like tuples or arrays, which can be used to compare multiple results at once and can have more rich and asymmetrical semantics.

PR #5078, opened 2026-01-15
feat(ethexe-processor): charge fee for the first message in announce

liferooter:
This is the initial implementation of the fee for uploading program state transition onto Ethereum.

### DIFF
--- a/core-processor/src/precharge.rs
+++ b/core-processor/src/precharge.rs
@@ -29,7 +29,7 @@ use core::marker::PhantomData;
 use gear_core::{
     code::{CodeMetadata, InstantiatedSectionSizes, SectionName},
     costs::{BytesAmount, ProcessCosts},
-    gas::{ChargeResult, GasAllowanceCounter, GasCounter},
+    gas::{GasAllowanceCounter, GasCounter},
     ids::ActorId,
     message::IncomingDispatch,
 };
@@ -61,6 +61,27 @@ pub enum PreChargeGasOperation {
     /// Obtain program allocations.
     #[display("obtain program allocations")]
     Allocations,
+    /// Other operations.
+    #[display("{description}")]
+    Other {
+        /// Operation description.
+        description: &'static str,
+
+        /// Whether the operation affects block gas allowance.
+        affects_allowance: bool,
+    },
+}
+
+impl PreChargeGasOperation {
+    /// Checks whether the operation affects block gas allowance.
+    pub const fn affects_allowance(&self) -> bool {
+        match self {
+            &Self::Other {
+                affects_allowance, ..
+            } => affects_allowance,
+            _ => true,
+        }
+    }
 }
 
 /// Defines result variants of the precharge functions.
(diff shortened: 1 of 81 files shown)


ethexe/runtime/common/src/lib.rs
```diff
@@ -116,13 +116,27 @@ impl<S: Storage> TransitionController<'_, S> {
...
+    first_message_fee: u64,
+}
+
+/// Processes the program message queue of given type.
+///
+/// Panics if the queue is empty. It's needed to guarantee
+/// that the function always charges for the first message.
+///
+/// Returns journals and the amount of gas burned.
+#[allow(clippy::too_many_arguments)]
```
ark0f, line 133: Create issue that we want to refactor functions with many arguments and reference it in TODO here

ethexe/runtime/common/src/lib.rs
```diff
@@ -250,13 +274,15 @@ where
...
+    ethereum_config: &EthereumConfig,
     program_id: ActorId,
     program_state: &ProgramState,
+    is_first_message: bool,
     instrumented_code: &Option<InstrumentedCode>,
     code_metadata: &Option<CodeMetadata>,
     ri: &RI,
     gas_allowance: u64,
-) -> Vec<JournalNote>
+) -> Result<Vec<JournalNote>, Vec<JournalNote>>
```
ark0f, line 288: Function signature is misleading despite the fact it simplifies code

ethexe/processor/src/handling/run.rs
```diff
@@ -590,13 +600,25 @@ mod chunk_execution_spawn {
...
+        let chunk = chunk
+            .into_iter()
+            .map(|(program_id, state_hash)| {
+                (
+                    program_id,
+                    state_hash,
+                    processed_first_queue.insert(program_id),
+                )
+            })
+            .collect::<Vec<_>>();
```
ark0f: Can be done somewhere inside `prepare_execution_chunks`

core-processor/src/precharge.rs
```diff
@@ -61,6 +61,27 @@ pub enum PreChargeGasOperation {
...
+        description: &'static str,
+
+        /// Whether the operation affects block gas allowance.
+        affects_allowance: bool,
+    },
+}
+
+impl PreChargeGasOperation {
+    /// Checks whether the operation affects block gas allowance.
+    pub const fn affects_allowance(&self) -> bool {
```
ark0f, line 77: What's the purpose of `const`?

ark0f (changes requested):

ethexe/processor/src/handling/run.rs
liferooter: This will decrease logical locality. We mark programs as processed in the same place where we process them. That makes the code more readable, as it reduces extra untrivial connections between different functions.

`prepare_execution_chunks` prepares chunks, `spawn_execution_chunks` spawn them (and also reflects it in `processed_first_queue`).

core-processor/src/precharge.rs
liferooter: This `const` reflects that the function is constant and can be evaluated in compile time. It's totally optional, as we do not use this method in `const` context for now, but having `const` is better than not having it and hurts no one, so I just put it automatically.

ethexe/runtime/common/src/lib.rs
liferooter: It's not misleading, it reflects that the processing must succeed or fail. It's not some kind of "exception", but `Result` is not necessarily connected to some program error, it's used to reflect errors that are not exceptional, not some kind of program failure and can happen and handled normally.

ethexe/runtime/common/src/lib.rs
liferooter: Fixed

ark0f (approved):
Also, add test fee is actually charged once

breathx:
Closing for now — fee charging logic will be reimplemented as part of a broader effort.

PR #5093, opened 2026-01-19
fix(ethexe): re-request blob if blob is empty or id mismatch (at least twice)

grishasobol:


ethexe/blob-loader/src/lib.rs
```diff
@@ -315,6 +326,45 @@ impl<DB: Database> BlobLoaderService for BlobLoader<DB> {
...
+    attempt: u8,
+) -> ReaderResult<Vec<u8>> {
+    // Attempt is only for logging purposes, so use it to avoid unused variable warning
+    let _ = attempt;
+
+    let received_code_id = CodeId::generate(&blob);
+    if blob.is_empty() {
+        log::warn!("received empty blob on attempt #{attempt}");
+        return Err(ReaderError::EmptyBlob);
+    }
```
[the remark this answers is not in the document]
grishasobol: Ok, done

ethexe/blob-loader/src/lib.rs
```diff
@@ -377,7 +424,64 @@ mod tests {
...
+            &mut Some(previously_received_code_id),
+            2,
+        );
+        assert!(matches!(
+            result,
+            Err(ReaderError::CodeIdMismatch {
+                expected,
+                found,
+            }) if expected == code_id && found == blob_code_id
+        ),);
```
StackOverflowExcept1on:
```suggestion
        ));
```

ethexe/blob-loader/src/lib.rs
```diff
@@ -108,19 +113,25 @@ struct ConsensusLayerBlobReader {
 }
 
 impl ConsensusLayerBlobReader {
-    async fn read_blob_from_tx_hash(&self, tx_hash: H256) -> ReaderResult<Vec<u8>> {
+    async fn read_blob(&self, code_id: CodeId, tx_hash: H256) -> ReaderResult<Vec<u8>> {
```
StackOverflowExcept1on:
```suggestion
    async fn read_blob(&self, expected_code_id: CodeId, tx_hash: H256) -> ReaderResult<Vec<u8>> {
```

StackOverflowExcept1on (approved):

ethexe/blob-loader/src/lib.rs
grishasobol: done

ethexe/blob-loader/src/lib.rs
grishasobol: done

ethexe/blob-loader/src/lib.rs
```diff
@@ -315,6 +326,42 @@ impl<DB: Database> BlobLoaderService for BlobLoader<DB> {
...
+        return Err(ReaderError::EmptyBlob);
+    }
+
+    let received_code_id = CodeId::generate(&blob);
+    if *previously_received_code_id == Some(received_code_id) {
+        log::warn!(
+            "received same blob with invalid id on attempt #{attempt}, code id: {received_code_id:?}, consider blob as loaded then",
+        );
+        return Ok(blob);
+    }
```
kuzminxyz, line 354: in my opinion, here should return error if we receive two invalid code ids.

grishasobol: Blob loader is not responsible for code id - correct blob means that rpc returns blob uploaded by user. correct blob does not mean correct code.

grishasobol: We use here code id only to avoid problems on rpc, which can return incorrect blob (return bytes which were not loaded by user) by some reasons.

PR #5133, opened 2026-02-07
feat!(ethexe/contracts): optimization of value claims and hashes

StackOverflowExcept1on:


breathx (approved):
1. This will cause breaking compatibility with current deployment: make sure to simultaneously upgrade contracts and node binaries
2. Why bytes.concat is still persist for value claims?

PR #5145, opened 2026-02-12
feat(ethexe/ethereum): add initial gas benchmarks

### LINKED ISSUE #5176: `ethexe`: implement initial benchamarking of gas consumptions for batch committing

StackOverflowExcept1on:
Resolves #5176

The basic idea is to introduce `SimulationContext`, which contains the context for the mock EVM blockchain. It creates [`revm`](https://github.com/bluealloy/revm) instance within this context and stores the `block_number`, `block_timestamp`, mock smart contract `deployer_address` (`Router`, `Mirror`, `WrappedVara`, etc.), mock smart contract `deployer_nonce`, and validator private keys (we're using 4 validators in testnet).

This `SimulationContext` will then be initialized, and all smart contracts (`Router`, `Mirror`, `WrappedVara`, etc.) and their instrumented versions will be deployed to the mock blockchain. It's also worth noting that about 10 blocks will be created in `SimulationContext` during initialization. Two codes will also be loaded: one will be validated, and the second will be marked as pending validation. Two programs (Mirror) will also be created. One program will have an uninitialized `stateHash` and others, and the second will have an initialized `stateHash`. Basically, all these variations are needed to measure different gas cases (cold slot, hot slot, etc.). The instrumented version for the Router and Mirror smart contacts differs from the regular one in that it uses the `event DebugEvent(uint256 indexed topic0) anonymous`. Since the event is `anonymous`, this allows calling opcode `log1(offset=0, size=0, topic0=some_debug_const)` directly from Solidity, avoiding inline assembly. Next, we use the [`revm::Inspector`](https://docs.rs/revm/latest/revm/trait.Inspector.html) API to measure difference between two `log1` opcodes and thus calculate the execution cost for fixed section of code that is bounded by topics.


ethexe/contracts/src/libraries/Gear.sol
```diff
@@ -387,26 +441,60 @@ library Gear {
         uint256 ts
     ) internal returns (bool) {
         uint256 eraStarted = eraStartedAt(router, block.timestamp);
+        // gas used: ~4883 (enter to function and calculate era started at)
+        // emit DebugEvent(COMMIT_BATCH_AFTER_ERA_STARTED_AT);
+
+        // emit DebugEvent(COMMIT_BATCH_BEFORE_CHECKING_IS_PREVIOUS_ERA_VALIDATION);
         if (ts < eraStarted && block.timestamp < eraStarted + router.timelines.validationDelay) {
+            // TODO: recalc here everything, not sure
```
StackOverflowExcept1on: need to re-bench/write test for that

ethexe/common/src/db.rs
```diff
@@ -238,7 +238,7 @@ mod tests {
     #[test]
     fn ensure_types_unchanged() {
         const EXPECTED_TYPE_INFO_HASH: &str =
-            "8a30677f0819eeffd02548e0e91adf940fa39173e509ad24b50cbb92ce61f907";
+            "1daa6179efabf521b9087edfd0244a64250abc27a9b93255a441e736fb2a6760";
```
StackOverflowExcept1on: Just changing comment in the code changes TypeInfo???

ethexe/ethereum/src/main.rs
```diff
@@ -0,0 +1,138 @@
+// This file is part of Gear.
```
StackOverflowExcept1on, line 1: will be removed, just to demonstrate API
```
cargo run --package ethexe-ethereum --bin ethexe-ethereum
```

ethexe/contracts/src/instrumented/MirrorWithInstrumentation.sol
```diff
@@ -0,0 +1,592 @@
+// SPDX-License-Identifier: UNLICENSED
```
kuzminxyz: Can you please move `*WithInstrumentation.sol` to `instrumened` folder? I think it would be great to have in `contracts/src/` only deployed contracts without patched ones.

ethexe/ethereum/src/main.rs
```diff
@@ -0,0 +1,138 @@
...
+//
+// Copyright (C) 2024-2026 Gear Technologies Inc.
```
kuzminxyz: `2026` for each file in benchmarking?

kuzminxyz (commented):
I viewed pull request. But not approve, because we need to reach the final design in team.

breathx:
Closing for now — not a current priority, can be reopened when the time comes.

PR #5146, opened 2026-02-12
fix(ethexe): fast-sync allow zero state and hashes request splitting

grishasobol:


StackOverflowExcept1on (approved):

ethexe/network/src/db_sync/mod.rs
```diff
@@ -58,7 +58,8 @@ use std::{
 };
 use tokio::sync::{mpsc, oneshot};
 
-const STREAM_PROTOCOL: StreamProtocol = StreamProtocol::new("/ethexe/db-sync/1.0.0");
+const STREAM_PROTOCOL: StreamProtocol =
```
grishasobol: Temporary - in order to be able to connect to current Vara.eth testnet

ethexe/network/src/db_sync/mod.rs
```diff
@@ -58,7 +58,8 @@ use std::{
...
+    StreamProtocol::new(concat!("/ethexe/db-sync/", env!("CARGO_PKG_VERSION")));
```
StackOverflowExcept1on: how it handles major, minor, patch?

StackOverflowExcept1on (approved):

ethexe/network/src/db_sync/mod.rs
StackOverflowExcept1on: What I mean is that we use the workspace version for our main product and also inherit it for ethexe. We now need to remember that this version is used as the network db sync version.

grishasobol: I forget to left a comment - this is temporary, only to be able to connect to current test-net and check fast-sync

breathx:
Closing — fix landed via another PR. Thanks for the work, sorry it sat so long.

PR #5180, opened 2026-02-27
feat(ethexe): Add multicall contract for properly batching SendMessage in node-loader

### LINKED ISSUE #5153: ethexe: add ability to process batch messages in parallel
### Problem to Solve

Currenetly ethexe-node-loader processes each call in a batch sequentially which is quite slow. 

### Possible Solution

During program start, upload multi-call contract to the network and use it to send batch messages all at once.

### Notes

_No response_

playX18:
Resolves #5153

### DIFF
--- a/ethexe/node-loader/src/batch.rs
+++ b/ethexe/node-loader/src/batch.rs
@@ -313,7 +313,7 @@ async fn run_batch_impl(
                 ));
             }
 
-            let created =
+            let (created, tx_count) =
                 create_program_batch_via_multicall(&api, send_message_multicall, &upload_calls)
                     .await?;
 
@@ -326,7 +326,7 @@ async fn run_batch_impl(
                 );
             }
 
-            let wait_for_event_blocks = blocks_window(args.len(), 1, 0);
+            let wait_for_event_blocks = blocks_window(tx_count, 1, 0);
             process_events(
                 api,
                 messages,
@@ -514,7 +514,7 @@ async fn run_batch_impl(
                 ));
             }
 
-            let created =
+            let (created, tx_count) =
                 create_program_batch_via_multicall(&api, send_message_multicall, &upload_calls)
                     .await?;
 
@@ -527,7 +527,7 @@ async fn run_batch_impl(
                 );
             }
 
-            let wait_for_event_blocks = blocks_window(args.len(), 1, 0);
+            let wait_for_event_blocks = blocks_window(tx_count, 1, 0);
             process_events(
                 api,
                 messages,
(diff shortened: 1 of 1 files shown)


ethexe/node-loader/scripts/update-batch-multicall-json.sh
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env sh
+set -eu
```
StackOverflowExcept1on: You can use step ethexe-contracts-pre-commit in Makefile

ethexe/node-loader/src/batch.rs
```diff
@@ -41,11 +48,24 @@ pub mod context;
 pub mod generator;
 pub mod report;
 
+alloy::sol!(
```
StackOverflowExcept1on: move this to crate::abi

ethexe/contracts/src/BatchMulticall.sol
```diff
@@ -0,0 +1,46 @@
...
+                revert InsufficientValue(consumed, msg.value);
+            }
+
+            try IMirror(item.mirror).sendMessage{value: item.value}(item.payload, false) returns (
+                bytes32 messageId
+            ) {
+                success[i] = true;
+                messageIds[i] = messageId;
+            } catch {
+                success[i] = false;
```
StackOverflowExcept1on: the fail is probably very unlikely, maybe in case of out of gas? sendMessage just emits event and hashes something.

ethexe/contracts/src/BatchMulticall.sol
```diff
@@ -0,0 +1,46 @@
...
+            }
+        }
+
+        if (consumed < msg.value) {
+            (bool refunded, ) = payable(msg.sender).call{value: msg.value - consumed}("");
```
StackOverflowExcept1on: casting to payable(..) is not necessary

ethexe/node-loader/scripts/update-batch-multicall-json.sh
playX18: It's also used in CI pipeline. I've also added copying of artifact directly to ethexe-contracts-pre-commit

ethexe/contracts/src/BatchMulticall.sol
```diff
@@ -0,0 +1,93 @@
...
+            address programId = router.createProgram(item.codeId, item.salt, address(this));
+            programIds[i] = programId;
+
+            bytes32 messageId = IMirror(programId).sendMessage{value: item.initValue}(item.initPayload, false);
+            messageIds[i] = messageId;
+
+            if (item.topUpValue > 0) {
+                require(wvara.transferFrom(msg.sender, address(this), item.topUpValue), "wVARA transferFrom failed");
+                require(wvara.approve(programId, item.topUpValue), "wVARA approve failed");
+                IMirror(programId).executableBalanceTopUp(item.topUpValue);
```
StackOverflowExcept1on: I'm surprised this works at all. Possibly because ExecutableBalanceTopUp has higher priority and it happens in same block.

.github/workflows/check.yml
```diff
@@ -95,6 +95,7 @@ jobs:
           cp ./ethexe/contracts/out/Router.sol/Router.json ./ethexe/ethereum/abi
           cp ./ethexe/contracts/out/TransparentUpgradeableProxy.sol/TransparentUpgradeableProxy.json ./ethexe/ethereum/abi
           cp ./ethexe/contracts/out/WrappedVara.sol/WrappedVara.json ./ethexe/ethereum/abi
+          cp ./ethexe/contracts/out/BatchMulticall.sol/BatchMulticall.json ./ethexe/ethereum/abi
```
StackOverflowExcept1on, line 98: I moved it to `./ethexe/ethereum/abi` because we already have Makefile and CI check and don't need another script. For now, it works via `"../ethereum/abi/BatchMulticall.json"`. It might also make sense to move `BatchMulticall` API wrapper to `ethexe-ethereum`.

PR #5095, opened 2026-01-20
feat(ethexe-common): Limit max size of injected transaction payload

ecol-master:

PR #5183, opened 2026-02-27
feat(ethexe): limit max size of injected transaction payload and salt

liferooter:
- Merges back #5095.
- Makes `InjectedTransaction.salt` a `LimitedVec`.

### DIFF
--- a/ethexe/common/src/injected.rs
+++ b/ethexe/common/src/injected.rs
@@ -21,7 +21,7 @@ use alloc::string::{String, ToString};
 use core::hash::Hash;
 use gear_core::{limited::LimitedVec, rpc::ReplyInfo};
 use gprimitives::{ActorId, H256, MessageId};
-use parity_scale_codec::{Decode, Encode};
+use parity_scale_codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
 use sha3::{Digest, Keccak256};
 
@@ -35,7 +35,7 @@ pub const VALIDITY_WINDOW: u8 = 32;
 pub const MAX_INJECTED_TX_PAYLOAD_SIZE: usize = 126 * 1024;
 
 /// Maximum size of injected transaction salt.
-pub const MAX_INJECTED_TX_SALT_SIZE: usize = 256;
+pub const MAX_INJECTED_TX_SALT_SIZE: usize = 32;
 
 #[cfg_attr(feature = "std", derive(serde::Deserialize, serde::Serialize))]
 #[derive(Debug, Clone, Encode, Decode, Eq, PartialEq)]
@@ -69,7 +69,7 @@ pub struct AddressedInjectedTransaction {
 /// IMPORTANT: message id == tx hash == blake2b256 hash of the struct fields concat.
 #[cfg_attr(feature = "std", derive(serde::Deserialize, serde::Serialize))]
 #[cfg_attr(feature = "serde", derive(Hash))]
-#[derive(Debug, Clone, Encode, Decode, TypeInfo, PartialEq, Eq)]
+#[derive(Debug, Clone, Encode, Decode, MaxEncodedLen, TypeInfo, PartialEq, Eq)]
 pub struct InjectedTransaction {
     /// Destination program inside `Vara.eth`.
     pub destination: ActorId,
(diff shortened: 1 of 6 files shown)


ethexe/cli/src/commands/tx.rs
```diff
@@ -978,10 +978,10 @@ impl TxCommand {
 
                         let injected_transaction = InjectedTransaction {
                             destination: raw_actor_id,
-                            payload: payload.0.clone().into(),
+                            payload: payload.0.clone().try_into().unwrap(),
```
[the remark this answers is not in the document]
liferooter: Fixed with https://github.com/gear-tech/gear/pull/5183/commits/2b82a32f7cec5c587d54126a0f5f8a737a89588e

ethexe/cli/src/commands/tx.rs
```diff
@@ -978,10 +978,16 @@ impl TxCommand {
 
                         let injected_transaction = InjectedTransaction {
                             destination: raw_actor_id,
-                            payload: payload.0.clone().into(),
+                            payload: payload.0.clone().try_into().with_context(|| {
+                                format!(
+                                    "Payload size exceeds the maximum allowed size of {} bytes",
+                                    ethexe_common::injected::MAX_INJECTED_TX_PAYLOAD_SIZE
```
StackOverflowExcept1on: import it

ethexe/common/src/injected.rs
```diff
@@ -17,21 +17,25 @@
...
 use sha3::{Digest, Keccak256};
-use sp_core::Bytes;
 
 /// Recent block hashes window size used to check transaction mortality.
 pub const VALIDITY_WINDOW: u8 = 32;
 
+/// Maximum size of single injected transaction payload.
+///
+/// Derived from per-announce injected transactions size limit.
+pub const MAX_INJECTED_TX_PAYLOAD_SIZE: usize = 129_621;
```
StackOverflowExcept1on: I think that's huge number. Mb 64 KiB? Because Ethereum transactions are also limited to 128 KiB and we are heavily dependent on that.

ethexe/rpc/src/tests.rs
```diff
@@ -209,6 +211,8 @@ async fn test_cleanup_promise_subscribers() {
 #[tokio::test]
 #[ntest::timeout(120_000)]
 async fn test_concurrent_multiple_clients() {
+    let _ = tracing_subscriber::fmt::try_init();
```
StackOverflowExcept1on, line 214: use gear-utils
https://docs.rs/gear-utils/latest/gear_utils/fn.init_default_logger.html

ethexe/common/src/injected.rs
liferooter: That's about 126KiB

StackOverflowExcept1on: Ah, ok. Plz use something like `126_... * 1024`.

grishasobol: please use aligned to 4 bytes at least number

liferooter: Done

ethexe/cli/src/commands/tx.rs
liferooter: Done

ethexe/rpc/src/tests.rs
liferooter: Why

liferooter: The initialization of `tracing_subscriber` is pretty simple and obvious itself. Replacing it with `gear_utils::init_default_logger` doesn't make the code readable and do not add any extra behaviour. All it does is adds an extra dev-dependency to `Cargo.toml` and hides the perfectly readable `tracing_subscriber` initialization behind a function with a non-obvious meaning unless you go to its definition to see its redundancy.

StackOverflowExcept1on: It is used in all repository tests.

liferooter: That's not true, we have a lot of `let _ = tracing_subscriber::fmt::try_init()` in our tests

liferooter: `init_debug_logger` is used more frequently, but I don't understand why we should fix usages of `tracing_subscriber::fmt::try_init` instead of usages of `gear_utils::init_debug_logger`.

PR #5189, opened 2026-03-03
fix(ethexe/contracts): transfers to destinations are now optional to fix DoS

StackOverflowExcept1on:


ethexe/contracts/src/Mirror.sol
```diff
@@ -442,7 +459,11 @@ contract Mirror is IMirror {
         inheritor = _inheritor;
 
         /// @dev Transfer all available balance to the inheritor.
-        transferLockedValueToInheritor();
+        (uint128 value, bool success) = _transferLockedValueToInheritor();
+        if (!success) {
+            /// @dev In case of failed transfer, we emit appropriate event to inform external users.
+            emit TransferLockedValueToInheritorFailed(_inheritor, value);
+        }
```
grishasobol, line 466: Why do we need to transfer locked to ethereum if it can be done by anyone later?

ethexe/contracts/src/IMirror.sol
```diff
@@ -95,6 +95,30 @@ interface IMirror {
...
+    /**
+     * @dev Emitted when the program fails to transfer value to destination after failed call
+     *
+     * NOTE:    It's event for USERS:
+     *  it informs about failed transfer of value to destination after failed call.
+     */
+    event ReplyTransferFailed(address destination, uint128 value);
+
+    /**
+     * @dev Emitted when a user fails in claiming value request and doesn't receive balance.
```
grishasobol, line 115: we should support this events in mirror.rs: try_extract_event because it cause unreachable

ethexe/contracts/src/Mirror.sol
StackOverflowExcept1on: Idk, it was initially designed like this

PR #5198, opened 2026-03-04
feat(ethexe/ethereum): add events for failed transfers

StackOverflowExcept1on:
Ref #5189

It would also be nice to have some logging of these events, but I don't know where to put it since it's not `RequestEvent`.

### DIFF
--- a/ethexe/common/src/db.rs
+++ b/ethexe/common/src/db.rs
@@ -260,7 +260,7 @@ mod tests {
     #[test]
     fn ensure_types_unchanged() {
         const EXPECTED_TYPE_INFO_HASH: &str =
-            "c6e5e519f690a94c75602e1bd455828161dba9feabc832a823b72d624b8f6780";
+            "36d0e8436bb8fa8ea920012e1b4b079f9b6a83414e016771afc977568d30f29b";
 
         let types = [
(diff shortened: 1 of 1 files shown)


ethexe/common/src/db.rs
```diff
@@ -238,7 +238,7 @@ mod tests {
     #[test]
     fn ensure_types_unchanged() {
         const EXPECTED_TYPE_INFO_HASH: &str =
-            "8a30677f0819eeffd02548e0e91adf940fa39173e509ad24b50cbb92ce61f907";
```
StackOverflowExcept1on: I hope this doesn't break anything since we just added new events to the database.

breathx: Seems so

grishasobol: please verify this using cli `check` on db snapshot

StackOverflowExcept1on: Should we care about it? Testnet will be re-deployed anyway. If we should, plz write here command how to check db snapshot.

StackOverflowExcept1on: @grishasobol `parity-scale-codec` uses `u8` for first byte when encoding `enum`

Issue #4791, opened 2025-07-24
ethexe: support head != current block hash, have to check head is predecessor of current block


grishasobol:
```
            // TODO +_+_+: support head != current block hash, have to check head is predecessor of current block
```

In file ethexe/consensus/src/validator/participant.rs
https://github.com/gear-tech/gear/blob/8115e1fb636adf470896a070ba97356a31f7f375/ethexe/consensus/src/validator/participant.rs#L156

grishasobol:
Closing as part of ethexe backlog triage: issue has been open for an extended period without an owner or activity, and is not on the current roadmap. Reopen if this becomes relevant.

PR #5214, opened 2026-03-11
feat(`ethexe-consensus`): Limits for BatchCommitment abi encoding

### LINKED ISSUE #5205: `ethexe`: set limits for batch commitment production
Presently any producer is pushing code commitments and announces in chain commitments as many as it possible: all codes waiting for validations and validated by this producer and all not committed announces till best announce for current block. This can cause problems with committing in case of networks downtime: for example after being downtime for 1 day, ethexe validators would try to commit all  transitions and codes which happens for 7200 blocks. That can be impossible to do in one batch commitment.

So, producer should push no more (for example) than 100 codes in one batch, and track the size of batch commitment when pushes announce commitments : for example no more than 64kb of transitions for example.

ecol-master:
Resolves #5205, #4791

@grishasobol @breathx


ethexe/cli/src/params/node.rs
```diff
@@ -137,6 +141,10 @@ impl NodeParams {
                 .block_gas_limit
                 .unwrap_or(DEFAULT_BLOCK_GAS_LIMIT)
                 .min(MAX_BLOCK_GAS_LIMIT),
+            batch_size_limit: self
+                .batch_size_limit
+                .unwrap_or(DEFAULT_BATCH_SIZE_LIMIT)
+                .min(MAX_BATCH_SIZE_LIMIT),
             canonical_quarantine: self.canonical_quarantine.unwrap_or(CANONICAL_QUARANTINE),
```
grishasobol, line 150: I think better just to panic in the case limit is bigger than max batch size. It's more UX friendly, because size is always as you set

grishasobol:
@codex review

ethexe/consensus/src/validator/mod.rs
```diff
@@ -135,6 +138,15 @@ impl ValidatorService {
         config: ValidatorConfig,
     ) -> Result<Self> {
         let timelines = db.config().timelines;
+        let limits = BatchLimits {
+            chain_deepness_threshold: config.chain_deepness_threshold,
+            commitment_delay_limit: config.commitment_delay_limit,
+            batch_size_limit: config.batch_size_limit,
```
grishasobol, line 144: actually it's not completely correct to use the same batch_size_limit for production and validation.

Can be a situation for example when:
1) Producer has limit 120 for example
2) Participant has limit 100
So, producer sends batch with size 110, which is valid, but not for participant.

batch_size_limit - is more about production, then validation. So, I think it's better to use always MAX_BATCH_SIZE_LIMIT for validation

kuzminxyz: I thought I would like a protocol constant for all validators. I think it would be better to use this limit from config to track where validators behavior not the same.

ethexe/cli/src/params/node.rs
```diff
@@ -91,6 +91,10 @@ pub struct NodeParams {
     #[serde(rename = "block-gas-limit")]
     pub block_gas_limit: Option<u64>,
 
+    #[arg(long)]
```
grishasobol, line 95: minor:  doc (will be printed in cli)

ethexe/consensus/src/validator/batch/manager.rs
```diff
@@ -0,0 +1,444 @@
...
+        for commitment in code_commitments {
+            if let Err(err) = batch_filler.include_code_commitment(commitment) {
+                tracing::trace!(
+                    "failed to include all code commitments into batch, because of error={err}"
+                );
+                break;
+            }
+        }
+
+        // TODO: prioritize transitions over code commitments
```
grishasobol: So, why not in this PR ?

ethexe/consensus/src/validator/mod.rs
grishasobol: Done as I can see

ethexe/cli/src/params/node.rs
kuzminxyz: Fixed in [45c22e2](https://github.com/gear-tech/gear/pull/5214/commits/45c22e2fa8cd065dcb667ecd270004a750bdbf64)

ethexe/consensus/src/validator/batch/manager.rs
kuzminxyz: Fixed in [75fcbd6](https://github.com/gear-tech/gear/pull/5214/commits/75fcbd6c1656770d41337123d5f6d755f9746bed)[75fcbd6](https://github.com/gear-tech/gear/pull/5214/commits/75fcbd6c1656770d41337123d5f6d755f9746bed)

PR #5220, opened 2026-03-13
feat(ethexe/contracts): add pause/unpause forge scripts

StackOverflowExcept1on:


ethexe/contracts/script/actions/Pause.s.sol
```diff
@@ -0,0 +1,20 @@
...
+contract PauseScript is Script {
+    function setUp() public {}
+
+    function run() public {
+        uint256 privateKey = vm.envUint("PRIVATE_KEY");
+        address routerAddress = vm.envAddress("ROUTER_ADDRESS");
+
+        vm.startBroadcast(privateKey);
+
+        Router(payable(routerAddress)).pause();
```
kuzminxyz, line 16: why payable here? The `function pause()` on Router do not have a `payable` attr

For `UnpauseScript` the same question

ethexe/contracts/.env.example
```diff
@@ -32,6 +32,16 @@ SYMBIOTIC_OPERATOR_REWARDS_FACTORY="0x0000000000000000000000000000000000000000"
...
 
+# script/MirrorProxy.s.sol
+
+# script/MirrorProxySmall.s.sol
+
+# script/actions/Pause.s.sol
+# ROUTER_ADDRESS="0x0000000000000000000000000000000000000000"
+
+# script/actions/Unpause.s.sol
+# ROUTER_ADDRESS="0x0000000000000000000000000000000000000000"
```
kuzminxyz, line 43: Why do we need multiple `ROUTER_ADDRESS` env variable? This variable can not be reused between scripts?

ethexe/contracts/script/actions/Pause.s.sol
StackOverflowExcept1on: because Router has receive function

ethexe/contracts/.env.example
StackOverflowExcept1on: this is just example file with required fields for some script

PR #5237, opened 2026-03-19
chore(`ethexe-consensus`): Metric for last signed batch commitment

ecol-master:


ethexe/consensus/src/validator/coordinator.rs
```diff
@@ -84,6 +84,7 @@ impl Coordinator {
         mut ctx: ValidatorContext,
         validators: ValidatorsVec,
         batch: BatchCommitment,
+        block: SimpleBlockData,
```
grishasobol, line 87: not important 

append debug_assert!(batch.block_hash == block.hash)

PR #5249, opened 2026-03-25
feat!(ethexe): change announce hash calculation; append min program exec balance for injected txs

grishasobol:


ethexe/db/src/migrations/v2.rs
```diff
@@ -0,0 +1,101 @@
...
+    log::info!("Migration investigation pass started: not modifying any data in database");
+
+    let config = db.kv.config().context("Cannot find db config")?;
+
+    ensure!(
+        config.version == v1::VERSION,
+        "Expected database version {}, but found {}",
+        v1::VERSION,
+        config.version
+    );
```
liferooter: Migration logic already guarantees that either the version in config is exactly `v1::VERSION`, or all previous migrations were executed, so no need to check it again.

ethexe/db/src/migrations/v2.rs
```diff
@@ -0,0 +1,101 @@
...
+        if k.len() != 2 * std::mem::size_of::<H256>() {
+            continue;
+        }
+
+        let block_hash = H256::from_slice(&k[std::mem::size_of::<H256>()..]);
+
+        let Some(BlockSmallData { meta, .. }) = BlockSmallData::decode(&mut v.as_slice()).ok()
+        else {
+            continue;
+        };
```
liferooter: Better to throw error in cases of broken block data encoding

ethexe/db/src/migrations/v2.rs
```diff
@@ -0,0 +1,101 @@
...
+
+        log::trace!("Investigating block {block_hash:?} with meta {meta:?}");
+
+        let Some(announces) = meta.announces else {
+            continue;
+        };
+
+        for announce_hash in announces {
```
liferooter:
```suggestion
        for announce_hash in meta.announces.into_iter().flatten() {
```

ethexe/db/src/migrations/v2.rs
```diff
@@ -0,0 +1,101 @@
...
+            let data = db
+                .cas
+                .read(announce_hash.inner())
+                .ok_or_else(|| anyhow!("found a block with missed announce in set"))?;
```
liferooter:
```suggestion
                .context("found a block with missed announce in set")?;
```

ethexe/db/src/migrations/v2.rs
```diff
@@ -0,0 +1,101 @@
...
+    );
+
+    for announce in announces_to_copy {
+        db.set_announce(announce);
+    }
+
+    db.kv.set_config(DBConfig {
+        version: VERSION,
+        ..config
+    });
```
liferooter, line 127: Consider getting the config by hardcoded key and decoding it manually. This will allow to easily preserve backward compatibility with just replacing `DBConfig` with `v1::DBConfig` in case `DBConfig` type is changed.

ethexe/db/src/migrations/v2.rs
liferooter, line 1: Add a test ensuring the types used in migration hasn't changed, just like `v1` has.

ethexe/db/src/migrations/v2.rs
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+
+    log::info!(
+        "Migration investigation pass finished: found {} announces to copy, starting copy process",
+        announces_to_copy.len()
+    );
+
+    for announce in announces_to_copy {
+        db.set_announce(announce);
```
liferooter, line 121: Consider manually putting the announce in the database by manually calculated key. This will allow to easily preserve backward compatibility and not to break the migration with changes in the database API and keys.

ethexe/db/src/migrations/v2.rs
```diff
@@ -0,0 +1,84 @@
...
+
+pub const VERSION: u32 = 2;
+
+pub async fn migration_from_v1(_: &InitConfig, db: &RawDatabase) -> Result<()> {
+    // Changes from version 1 to version 2:
+    // Copy announces data to KV
+
+    log::info!("Migration investigation pass started: not modifying any data in database");
+
+    const BLOCK_SMALL_DATA_PREFIX: u64 = 0x00;
```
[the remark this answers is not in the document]
liferooter: Exacly the oposite, hardcoding parts of the current behaviour in migrations allows them not to be fragile, as the migration should not change its behaviour with future changes in the database logic.

ethexe/common/src/primitives.rs
```diff
@@ -95,7 +95,20 @@ pub struct Announce {
...
+        let Announce {
+            block_hash,
+            parent,
+            gas_allowance,
+            injected_transactions,
+        } = self;
+
+        let transitions_hashes = injected_transactions
+            .iter()
+            .map(|tx| utils::hash(&tx.encode()))
```
breathx: Are you sure which hash we will use here? We operate both Blake and keccak for txs

ethexe/common/src/primitives.rs
```diff
@@ -95,7 +95,20 @@ pub struct Announce {
...
         // # Safety because of implementation
-        unsafe { HashOf::new(H256(utils::hash(&self.encode()))) }
+        let Announce {
+            block_hash,
+            parent,
+            gas_allowance,
+            injected_transactions,
+        } = self;
+
+        let transitions_hashes = injected_transactions
```
kuzminxyz: Invalid, must be:
```rust
let transactions_hashes = injected_transactions.iter().map(|tx| tx.data().to_hash()).collect::<Vec<_>>();
```

kuzminxyz:
```rust
impl InjectedTransaction {
    /// Returns the hash of [`InjectedTransaction`].
    pub fn to_hash(&self) -> HashOf<InjectedTransaction> {
        // Safe because we hash corresponding type itself
        let bytes = [
            self.destination.as_ref(),
            self.payload.as_ref(),
            &self.value.to_be_bytes(),
            &self.reference_block.0,
            self.salt.as_ref(),
        ]
        .concat();
        unsafe { HashOf::new(gear_core::utils::hash(&bytes).into()) }
    }
}
```
Transaction hash is calculated not over parity_scale_codec bytes

ethexe/db/src/migrations/v2.rs
grishasobol: it's not error case - it means that we found some data with the same prefix, but it's not block small data. It's possible case

ethexe/db/src/migrations/v2.rs
grishasobol: It does not check that after migration version from DB config is increased, but I move the check there

grishasobol: moved

ethexe/db/src/migrations/v2.rs
grishasobol: ok, I decided that it's really better to throw error

ethexe/db/src/migrations/v2.rs
grishasobol: done

ethexe/db/src/migrations/v2.rs
grishasobol: I added a static check for the version (2) for which this migration is implemented, if something changed and the version as well, then programmer would receive a compilation  error

ethexe/db/src/migrations/v2.rs
grishasobol: done

ethexe/db/src/migrations/v2.rs
grishasobol: see above

ethexe/db/src/migrations/v2.rs
```diff
@@ -0,0 +1,151 @@
...
+        let block_hash = H256::from_slice(&k[std::mem::size_of::<H256>()..]);
+
+        let BlockSmallData { meta, .. } = BlockSmallData::decode(&mut v.as_slice())
+            .context("failed to decode BlockSmallData during migration")?;
+
+        log::trace!("Investigating block {block_hash:?} with meta {meta:?}");
+
+        for announce_hash in meta.announces.into_iter().flatten() {
+            let announce = get_announce_from_cas(announce_hash)
+                .context(format!("cannot get announce by {announce_hash:?}"))?;
```
liferooter:
```suggestion
                .with_context(|| format!("cannot get announce by {announce_hash:?}"))?;
```
to only run formatting code in case of error

ethexe/common/src/primitives.rs
grishasobol: yep, checked

ethexe/common/src/primitives.rs
grishasobol: announce hash changed

ethexe/db/src/migrations/v2.rs
grishasobol: done

PR #5252, opened 2026-03-26
tests(ethexe/processor): add determenistic event-ordering test

### LINKED ISSUE #5251: tests(ethexe/processor): test that txs, events, and tasks are processed in deterministic order
### Problem to Solve

Right now there is no test that they are processed in some concrete order. It would be nice to have one to ensure stability of ethexe processing.

### Possible Solution

Write a test that implements a WAT program with wait_for(1) in handle(). The test first forces program to wait, then sends txs + wake and we verify that txs are processed first, and wake is processed second and then queues are processed third (by sending reply after wake) 

### Notes

_No response_

playX18:
Resolves #5251


ethexe/processor/src/tests.rs
```diff
@@ -1496,6 +1496,254 @@ async fn call_wake_with_delay_is_unsupported() {
...
+    );
+
+    // === Block 3: All three phases via process_programs ===
+    // - Phase 1 (injected + events): injected tx + canonical message populate queues
+    // - Phase 2 (tasks): WakeMessage moves waiting dispatch back to canonical queue
+    // - Phase 3 (queues): injected queue first, then canonical queue
+    let block3 = chain.blocks[3].to_simple();
+    assert_eq!(block3.header.height, wake_block_height);
+
+    let injected_tx = InjectedTransaction {
```
kuzminxyz, line 1626: Maybe should add `InjectedTransaction::mock(actor_id)` and use it here?
I think it would be useful mock impl

ecol-master (approved):
looks good to me

ethexe/processor/src/tests.rs
playX18: might be a part of our planned refactor of test suite.

PR #5263, opened 2026-03-28
feat(ethexe): bump foundry toolchain, add configurable gas, use only EIP-7594 txs

StackOverflowExcept1on:
Now we need to use the new version of Foundry (for local dev):

```bash
foundryup -i nightly-c07d504b4ae67754584f4e05ff0c547a43c50f7b
```

### DIFF
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -4,6 +4,138 @@
 
 Review pull requests for the whole `gear` workspace with a balanced posture: prioritize correctness, safety-critical behavior, and verification gaps before maintainability comments.
 
+## Repository Context
+
+> **Note:** The canonical source for architecture documentation is [CLAUDE.md](../CLAUDE.md).
+> This section is a subset kept in sync for Gemini Code Assist, which only reads `styleguide.md`.
+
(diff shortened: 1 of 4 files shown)


ethexe/ethereum/src/router/mod.rs
```diff
@@ -143,16 +143,11 @@ impl Router {
...
             .instance
             .requestCodeValidation(code_id.into_bytes().into());
-        let builder = if chain_id == 31337 {
-            // TODO: remove when https://github.com/foundry-rs/foundry/pull/12404 is merged
-            builder.sidecar(SidecarBuilder::<SimpleCoder>::from_slice(code).build()?)
-        } else {
-            builder.sidecar_7594(SidecarBuilder::<SimpleCoder>::from_slice(code).build_7594()?)
-        };
+        let builder =
+            builder.sidecar_7594(SidecarBuilder::<SimpleCoder>::from_slice(code).build_7594()?);
```
StackOverflowExcept1on, line 150: I also want to experiment with EIP-7594 txs here: [`@vara-eth/api`](https://github.com/gear-tech/gear-js/tree/main/apis/vara-eth) (API is currently broken on testnet and we could backport https://github.com/wevm/viem/pull/4022 to fix it). The thing is that Foundry updates quite slowly and I switched it to `nightly` so that we could use only one transaction type for blobs - EIP-7594.

PR #5280, opened 2026-04-03
feat(ethexe/contracts): add fee for upload code, add method for withdrawing exec balance

StackOverflowExcept1on:


ethexe/contracts/src/Router.sol
```diff
@@ -287,6 +294,16 @@ contract Router is
...
 
+    function withdrawExecutableBalance() external onlyOwner {
+        IERC20 _wrappedVara = IERC20(wrappedVara());
+
+        uint256 value = _wrappedVara.balanceOf(address(this));
+        require(value > 0, ZeroValueTransfer());
+
+        bool success = _wrappedVara.transfer(msg.sender, value);
+        require(success, WithdrawExecutableBalanceFailed());
+    }
```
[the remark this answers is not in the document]
StackOverflowExcept1on: we currently do not use `_commitRewards` at production

ethexe/contracts/src/Router.sol
```diff
@@ -33,6 +33,8 @@ contract Router is
     // keccak256(abi.encode(uint256(keccak256("router.storage.Transient")) - 1)) & ~bytes32(uint256(0xff))
     bytes32 private constant TRANSIENT_STORAGE = 0xf02b465737fa6045c2ff53fb2df43c66916ac2166fa303264668fb2f6a1d8c00;
 
+    uint256 public constant CODE_COMMITMENT_GAS = 13_551;
```
StackOverflowExcept1on, line 36: from benchmarks PR

ethexe/ethereum/src/deploy.rs
```diff
@@ -160,6 +160,9 @@ impl EthereumDeployer {
 
 // Private implementation details
 impl EthereumDeployer {
+    /// Fixes "out of gas" error for unknown reason, maybe need to report to alloy?
+    const LOOKUP_GENESIS_HASH_GAS: u64 = 100_000;
```
kuzminxyz, line 172: Yes, make sense, I don't find similar issues alloy repo right now.

But, I guess, it happens, because we use the nightly build of foundry (maybe I am wrong)

kuzminxyz (approved):

breathx (changes requested):
I don’t see any need to store or update the gas price, since we won’t be converting it directly, either now or in the future.

Moreover, the withdrawal of exec balances is high risk operation since it could brake further committing logic so I'd keep it in upcoming commitments reward.

Fee for code uploading should be charged in WVARA with some constant value that could be updated on migrations.

PR #5321, opened 2026-04-08
feat(ethexe-consensus): mini-announces for instant injected TX promises

ukint-vs:
## Summary

- Injected TX end-to-end latency was 0.4-12.4s (avg ~6.4s). Processing is fast (~400ms), the bottleneck was the 0-12s wait for the next ETH block
- After the first announce computes, the producer polls the TX pool every `producer_delay` interval. If TXs found, creates a batched mini-announce. If empty, creates batch commitment immediately
- Promise delivery latency: ~400ms average (0-200ms poll wait + ~400ms computation)
- Canonical events skipped for same-block announces (compute layer fix)
- **Block-aware CDL**: All CDL-bounded loops now count block transitions instead of announce hops, fixing 7 functions that assumed 1 block = 1 announce

## Architecture

```
Block N arrives
     │
produce_announce()
parent = best_parent_announce()
     │
┌────▼───────────────────────┐
│ WaitingAnnounceComputed    │◄──── mini-announce computed
└────┬───────────────────────┘
     │
┌────▼───────────────────────┐
│ ReadyForMiniAnnounce       │
│ (poll timer running)       │
└──┬──────────┬──────────┬──┘
   │          │          │
 Timer fires  │    New head arrives
   │          │          │
 Pool check   │    AggregateBatchCommitment
   │          │          │
 ┌─▼──┐      │     ┌────▼────────┐
 │TXs?│      │     │ Coordinator │
 └┬───┘      │     └─────────────┘
  │Yes: mini-announce → compute → back to Ready
  │No: AggregateBatchCommitment → Coordinator → Initial
```

### Producer (`ethexe/consensus/src/validator/producer.rs`)

- **Timer-based polling**: `ReadyForMiniAnnounce` state with `poll_timer`. Every tick, checks TX pool via `select_for_announce`. TXs found → `produce_mini_announce_with_txs` (batched). Empty → `AggregateBatchCommitment`.
- **`process_new_head` override**: When new block arrives in `ReadyForMiniAnnounce` or `AggregateBatchCommitment`, buffers it in `next_block` and creates batch commitment. Passes `next_block` through Coordinator to Initial.
- **No `process_injected_transaction` override**: TXs go to pool via `DefaultProcessing`, timer picks them up.

### Coordinator (`ethexe/consensus/src/validator/coordinator.rs`)

- **`next_block` buffering**: Overrides `process_new_head` to buffer the block instead of dying. Passes `next_block` to Initial after submission.
- **Timeout on second head**: If a second new head arrives while still waiting for signatures, gives up and transitions to Initial. Prevents hang when threshold can't be met.

### Subordinate (`ethexe/consensus/src/validator/subordinate.rs`)

- **`ReadyForMoreAnnounces` state**: Accepts mini-announces from the producer after first announce computes.
- **`replay_pending_events`**: After entering `ReadyForMoreAnnounces`, replays pending events oldest-first so parent announces are processed before children.
- **VR handling**: Validation request from producer → enqueue as pending → `Participant::create`. Non-validators drop the VR.

### Compute (`ethexe/compute/src/compute.rs`)

- **Skip canonical events**: If parent announce has the same `block_hash`, canonical Ethereum events were already processed. Pass empty events, only process injected transactions.

### Announces correctness (`ethexe/consensus/src/announces.rs`, `batch/utils.rs`)

Mini-announces chain within the same block (parent-child), but all CDL-bounded loops assumed 1 announce = 1 block. Fixed 7 functions:

| Function | Fix |
|----------|-----|
| `propagate_announces` | `leaf_announces()` filter prevents exponential growth |
| `propagate_one_base_announce` | Block-aware CDL expiry check |
| `find_announces_common_predecessor` | Leaf filter + block-aware iteration |
| `best_announce` | Score across CDL blocks, not CDL announce hops |
| `best_parent_announce` | Base-filter to get cross-block parents only |
| `recover_announces_chain_if_needed` | Block-aware recovery loop |
| `calculate_batch_expiry` | Block-aware depth counting |

**Behavioral change**: Non-base announces now live 1 block longer than before. The old code expired at CDL-1 blocks distance (off-by-one vs S1 which says `<= CDL`). This is intentional and matches the protocol spec. Two existing test assertions updated.

## Trade-offs

- **Batch commitment deferred**: Created when poll timer finds empty pool or new head arrives. Adds ~200ms vs immediate creation. Promise delivery unaffected (happens during computation).
- **Coordinator timeout**: Second new head kills Coordinator. If threshold wasn't met, that batch is lost. Next block's producer creates a new batch via `collect_not_committed_predecessors`.

## Known issues (tracked)

- #5342: Batch commitment lost when new head arrives during `WaitingAnnounceComputed` (pre-existing)
- #5343: Synced/prepared events lost when consensus state is not Initial (pre-existing)

## Test plan

- [x] `cargo nextest run -p ethexe-consensus` — 85 tests pass (7 new for chained mini-announces)
- [x] `cargo nextest run -p ethexe-compute` — 21 tests pass
- [x] `cargo clippy -p ethexe-consensus` — clean
- [x] Codex adversarial challenge — no new issues in announces.rs changes
- [x] 1000-case proptests for announce propagation
- [ ] Manual test: submit injected TX via RPC, verify promise arrives in ~400ms


ethexe/consensus/src/validator/producer.rs
```diff
@@ -140,12 +185,23 @@ impl StateHandler for Producer {
...
                 Poll::Ready(Ok(Some(batch))) => {
                     tracing::debug!(batch.block_hash = %batch.block_hash, "Batch commitment aggregated, switch to Coordinator");
-                    return Coordinator::create(self.ctx, self.validators, batch, self.block)
-                        .map(|s| (Poll::Ready(()), s));
+                    let next_block = self.next_block.take();
+                    let state = Coordinator::create(self.ctx, self.validators, batch, self.block)?;
+                    let state = match next_block {
+                        Some(block) => state.process_new_head(block)?,
+                        None => state,
+                    };
```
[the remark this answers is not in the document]
ukint-vs: Fixed. The code now checks `state.is_initial()` before passing `next_block`. For threshold>1, the Coordinator keeps running to collect validation replies, and the next block arrives naturally via the service event loop.

ethexe/consensus/src/validator/producer.rs
```diff
@@ -129,6 +140,40 @@ impl StateHandler for Producer {
...
+                    .batch_manager
+                    .clone()
+                    .create_batch_commitment(self.block, last_announce_hash)
+                    .boxed(),
+            };
+            Ok(self.into())
+        } else {
+            DefaultProcessing::new_head(self, block)
+        }
+    }
```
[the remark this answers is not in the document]
ukint-vs: This is pre-existing behavior, not introduced by this PR. The original code also had `WaitingAnnounceComputed` with default `process_new_head` → Initial. The mini-announce changes didn't modify this path.

ethexe/consensus/src/validator/subordinate.rs
```diff
@@ -71,18 +72,17 @@ impl StateHandler for Subordinate {
...
-                if self.is_validator {
-                    Participant::create(self.ctx, self.block, self.producer)
-                } else {
-                    Initial::create(self.ctx)
-                }
+                // Enter waiting state for more announces (mini-announces).
+                // The Participant transition happens when the validation request arrives.
+                self.state = State::ReadyForMoreAnnounces;
+                Ok(self.into())
             }
```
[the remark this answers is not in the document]
ukint-vs: Fixed. Added `replay_pending_events()` which drains pending events after entering `ReadyForMoreAnnounces`. Any mini-announces or VRs that arrived during computation are now processed immediately.

ethexe/consensus/src/validator/producer.rs
```diff
@@ -249,18 +307,84 @@ impl Producer {
...
+            parent: last_announce_hash,
+            gas_allowance: Some(self.ctx.core.block_gas_limit),
+            injected_transactions,
+        };
+
+        let (announce_hash, newly_included) =
+            self.ctx.core.db.include_announce(announce.clone())?;
+        if !newly_included {
+            return Ok(self.into());
+        }
```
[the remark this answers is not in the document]
ukint-vs: This is intentional. If the announce was already included (duplicate from RPC abuse), there's no computation to wait for. Transitioning to `WaitingAnnounceComputed` would block forever waiting for a computation event that will never come. Staying in `ReadyForMiniAnnounce` is the correct fallback.

ecol-master:
@ukint-vs The idea looks good and make sense, but now in `ethexe/compute/src/compute.rs` for each announce we take from database the ethereum events which correspond to a `announce.block_hash`. 

So in current implementation for each mini `Announce` you will execute the same ethereum events which were executed already in the first announce for the block. 

So, for me, this is now the major issue that should be fixed before the normal PR review.

But, still, the idea to allow producer collect transaction within a slot looks good.

ecol-master:
I also checked the state switching in subordinate, producer and participant. 
From the first view looks normally. Follow the current consensus rules

ukint-vs:
Fixed in 6c02fd70b. `prepare_executable_for_announce()` now checks if the parent announce has the same `block_hash`. If so, canonical events are skipped (empty `events` vec) — only injected transactions are processed. The parent already executed the block's canonical events, and the mini-announce inherits its `program_states` and `schedule` from the parent.

ethexe/consensus/src/validator/producer.rs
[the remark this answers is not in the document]
ukint-vs: Documented as TODO in code (producer.rs:188-193). Pre-existing behavior, not introduced by mini-announces. Also added a batch timer in `ReadyForMiniAnnounce` (af65feeb9) so batch commitment fires on a deadline without waiting for the next block.

ethexe/consensus/src/validator/subordinate.rs
```diff
@@ -105,15 +115,21 @@ impl StateHandler for Subordinate {
...
+            _ if request.address() == self.producer => {
+                tracing::trace!(
+                    "Receive validation request from producer: {request:?}, saved for later."
+                );
+                self.ctx.pending(request);
+                Ok(self.into())
+            }
+            _ => DefaultProcessing::validation_request(self, request),
         }
     }
```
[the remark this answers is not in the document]
ukint-vs: Fixed in f3c2e76c1. Non-validators in `ReadyForMoreAnnounces` now drop the VR instead of re-enqueuing it.

ethexe/consensus/src/validator/coordinator.rs
```diff
@@ -55,6 +57,12 @@ impl StateHandler for Coordinator {
         self.ctx
     }
 
+    fn process_new_head(mut self, block: SimpleBlockData) -> Result<ValidatorState> {
+        // Buffer the new head instead of dying. Process it after submission.
+        self.next_block = Some(block);
```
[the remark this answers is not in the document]
ukint-vs: Intentional. The latest block is always the correct one to process — older buffered blocks are stale. Same pattern used in `AggregateBatchCommitment`. No warning needed since this is normal operation under fast block production.

ethexe/consensus/src/validator/producer.rs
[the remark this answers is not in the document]
ukint-vs: Updated in 102613a10. Coordinator now overrides `process_new_head` to buffer the block instead of dying. When submission completes, `next_block` is passed to Initial. Works for all threshold values.

ecol-master:
@ukint-vs I re-reviewed PR. The work is closely to working solution, but still exists some issues.

When I found:
1. `Producer::process_announce`(producer.rs) now ignores the announce when `State::WaitingAnnounceComputed`
2. Issue connected with `Coordinator::process_new_head`
3. Constant `MAX_MINI_ANNOUNCES_PER_BLOCK`  in `producer.rs` can be removed, because if you will have `MAX_MINI_ANNOUNCES_PER_BLOCK + 1` transaction in block - producer will stop producing announces. Instead of this I propose my view of this solution: add to Producer timer, for example for 200ms, and every 200ms poll local transaction poll in purpose to find new unique transactions. If occurs some new transactions - create mini announce.

Also, I create a commit ([d0a58bc](https://github.com/gear-tech/gear/pull/5321/commits/d0a58bcba15cf1c51b4ff56a90a03847b99af88c)) where I make tests pass. 
It still do not fix the problem, just help AI to detect where consensus logic has degraded

ethexe/consensus/src/validator/subordinate.rs
```diff
@@ -105,15 +115,26 @@ impl StateHandler for Subordinate {
...
+            State::ReadyForMoreAnnounces
+                if request.address() == self.producer && self.is_validator =>
+            {
+                self.ctx.pending(request);
+                Participant::create(self.ctx, self.block, self.producer)
+            }
+            State::ReadyForMoreAnnounces if request.address() == self.producer => {
+                // Non-validator: VR is meaningless, drop it to avoid recycle loop
+                // in replay_pending_events.
+                Ok(self.into())
```
ecol-master, line 170: Maybe should be `Initial::create()`. If I correct, for subordinate nothing to wait when it receive validation request?

ethexe/service/src/tests/mod.rs
```diff
@@ -1446,26 +1462,26 @@ async fn multiple_validators() {
     }
 
     log::info!("📗 Stop validator 1 and check, that ethexe is not working after");
-    validators[1].stop_service().await;
+    // validators[1].stop_service().await;
```
grishasobol: Why commented?

ethexe/consensus/src/validator/producer.rs
```diff
@@ -304,130 +471,263 @@ mod tests {
...
+        // After computed announce, producer enters ReadyForMiniAnnounce.
+        // Batch commitment is deferred to the next block's cycle.
+        assert!(state.is_producer());
+        let producer = state.unwrap_producer();
+        assert!(producer.state.is_ready_for_mini_announce());
     }
 
     #[tokio::test]
     #[ntest::timeout(3000)]
-    async fn threshold_one() {
```
grishasobol, line 325: Why test removed

ethexe/consensus/src/validator/producer.rs
```diff
@@ -304,130 +471,263 @@ mod tests {
...
+        assert!(
+            producer.state.is_waiting_announce_computed(),
+            "Expected WaitingAnnounceComputed after mini-announce, got {:?}",
+            producer.state
+        );
     }
 
     #[tokio::test]
     #[ntest::timeout(3000)]
-    async fn threshold_two() {
```
grishasobol, line 380: Why test is removed?

grishasobol (changes requested):
All deleted tests and test cases, must be returned back. At least one integration test for mini announces should be added to ethexe-service tests

ethexe/consensus/src/validator/subordinate.rs
ukint-vs: Refactored to timer-based polling per your suggestion:
  1. Removed process_injected_transaction override and MAX_MINI_ANNOUNCES_PER_BLOCK — poll timer now checks the pool every producer_delay interval, batching all pending TXs into one mini-announce
  2. When timer fires and pool is empty → batch commitment created immediately (no more waiting for next ETH block)
  3. Coordinator now buffers next_block via process_new_head override so blocks aren't lost during signature collection (4 validators live)

  Latest commit: 5532e51

ukint-vs: The subordinate still needs to enter Participant to validate the batch commitment from the producer. Going to Initial would skip validation entirely. The VR is enqueued as pending before `Participant::create` so it's consumed correctly.

ethexe/consensus/src/validator/producer.rs
ukint-vs: restored

ethexe/consensus/src/validator/producer.rs
ukint-vs: restored

ethexe/service/src/tests/mod.rs
ukint-vs: That was from ecol-master's debug commit (d0a58bc). Un-commented and working now in 57d025749. The issue was that Coordinator hung forever when threshold couldn't be met (not enough validators responding). Fixed by timing out Coordinator on the second new head.

ukint-vs:
All 3 issues addressed:

1. `process_injected_transaction` override removed, replaced with timer-based polling per your proposal (39c75fb). Poll timer checks the pool every `producer_delay` interval, batching all pending TXs into one mini-announce. When pool is empty → batch commitment created immediately.

2. Coordinator now buffers `next_block` via `process_new_head` override and times out on second new head if signatures can't be collected (57d025749). Validator restart test un-commented and passing.

3. `MAX_MINI_ANNOUNCES_PER_BLOCK` removed. Timer naturally rate-limits.

Pre-existing issues (batch lost during `WaitingAnnounceComputed`, synced/prepared events lost outside Initial) tracked in #5342 and #5343.

ethexe/consensus/src/validator/subordinate.rs
```diff
@@ -97,6 +100,13 @@ impl StateHandler for Subordinate {
                 let (announce, _pub_key) = verified_announce.into_parts();
                 self.send_announce_for_computation(announce)
             }
+            State::ReadyForMoreAnnounces
```
grishasobol: Where is check of parent? New mini announce must be above latest

ethexe/consensus/src/validator/subordinate.rs
```diff
@@ -105,15 +115,26 @@ impl StateHandler for Subordinate {
...
-            Ok(self.into())
-        } else {
-            DefaultProcessing::validation_request(self, request)
+        match &self.state {
+            State::ReadyForMoreAnnounces
+                if request.address() == self.producer && self.is_validator =>
+            {
+                self.ctx.pending(request);
+                Participant::create(self.ctx, self.block, self.producer)
+            }
```
grishasobol: what if producer send mini announce MA, then validation request VR, but due to network issues VR would come before MA?

ethexe/consensus/src/validator/subordinate.rs
ukint-vs: Fixed in fc0ea47. `ReadyForMoreAnnounces` now tracks `latest_announce_hash` (the just-computed announce). Mini-announces are only accepted if `announce.parent == latest_announce_hash`. If parent doesn't match, falls through to `DefaultProcessing::announce_from_producer` which queues as pending with a warning.

ethexe/consensus/src/validator/subordinate.rs
ukint-vs: Fixed in fc0ea47. Before entering Participant, we now check if the VR's `head` announce is computed locally (`db.announce_meta(h).computed`). If not computed (MA hasn't arrived yet), the VR is saved to pending and subordinate stays in `ReadyForMoreAnnounces`. When the MA arrives and computes, `replay_pending_events` replays the VR, and by then the head is computed.

Edge case: if VR is deferred and no more MAs arrive, VR sits in pending until next ETH block resets to Initial (~12s). Batch fails threshold and is recovered by `collect_not_committed_predecessors`. Acceptable for a rare network reordering scenario.

ukint-vs:
Superseded by #5352 (two-phase compute). Same goal, simpler approach: depth-1 with canonical-only compute instead of depth-2 mini-announces. Eliminates CDL patches, subordinate state explosion, and gossip reorder complexity.

ukint-vs:
Reopening: this is the approach that actually delivers 400ms promise latency. Looking for ways to simplify the implementation (reduce CDL patches, subordinate complexity).

breathx:
Will be superseded by the Malachite consensus migration — closing to keep things tidy.

PR #5341, opened 2026-04-13
VIEW ONLY: vara-eth testnet re-genesis

grishasobol:


ethexe/contracts/src/Router.sol
```diff
@@ -124,6 +124,14 @@ contract Router is
         // All protocol data must be removed - so leave it zeroed in new router.
     }
 
+    /// @dev Re-genesis: set a new genesis block and reset the latest committed batch.
+    /// All programs, codes, validators, and settings are preserved.
+    function reGenesis(Gear.GenesisBlockInfo memory newGenesis) public onlyOwner {
+        Storage storage router = _router();
+        router.genesisBlock = newGenesis;
+        router.latestCommittedBatch = Gear.CommittedBatchInfo({hash: bytes32(0), timestamp: 0});
```
StackOverflowExcept1on, line 91: I'd prefer different approach: reset `router.genesisBlock` and `router.latestCommittedBatch` and set the current block height as we do in the constructor. Then, anyone should call `lookupGenesisHash` in next block. This prevents the admin from setting any block hash they want.

grishasobol:
No more needed - test-net was updated

PR #5347, opened 2026-04-15
feat: initial error PR | refactor  error

### LINKED ISSUE #4364: ethexe(rpc): return standard json rpc error codes
### Problem to Solve

Error code used right now is `8000` which is valid, but not standard.

### Possible Solution

Replace with standard codes for application/server errors (from -32099 to -32000)

kuzminxyz:
#4364

@gear-tech/dev


breathx:
Closing — the base branch is gone. Reopen against master if still needed.

PR #5363, opened 2026-04-20
feat(crypto): native crypto + hash syscalls on Vara and ethexe

ukint-vs:
Adds 7 `gr_*` syscalls with byte-identical ABI on Vara and ethexe.

## Syscalls

| Primitive            | Shape                                                          |
|----------------------|----------------------------------------------------------------|
| `gr_blake2b_256`     | infallible, base + per-byte                                    |
| `gr_sha256`          | infallible, base + per-byte                                    |
| `gr_keccak256`       | infallible, base + per-byte                                    |
| `gr_sr25519_verify`  | infallible, explicit Schnorrkel ctx, base + per-transcript-byte |
| `gr_ed25519_verify`  | infallible, base + per-byte                                    |
| `gr_secp256k1_verify`| infallible, `malleability_flag: u32` (0 permissive / 1 strict low-s) |
| `gr_secp256k1_recover` | fallible (err u32), same malleability flag                   |

## Architecture

One shared wrapper in `core/backend/funcs`. Both networks enter through it,
and it charges gas once upstream via `CostToken` / `SyscallWeights`.

From there:

- **Vara** calls a native `Externalities` impl in `core/processor/ext` that
  runs `sp_core::{sr25519,ed25519,ecdsa}`, `libsecp256k1`, and `schnorrkel`
  directly in the runtime.
- **Ethexe** overrides the same `Externalities` methods in
  `ethexe/runtime/common/ext` to call new `ext_*_v1` host imports, which
  `ethexe/processor` services via `wasmtime::Linker::func_wrap` using the
  same crypto crates — natively, not inside the WASM runtime blob.

Ethexe crypto methods are intentionally **not** routed through `delegate!`.
Delegating would run `sp_core` compiled into the ethexe-runtime WASM,
which is the slow path the syscall exists to escape.

Cross-network policy (e.g. `SECP256K1_N_HALF`, `is_low_s`) lives in a single
`gear_core::crypto` module so `verify` and `recover` can never disagree on
the same `(sig, flag)`.

## Verified

- `demo-crypto` KATs (14): blake2b/sha256/keccak256 standard vectors,
  ed25519/sr25519 valid+tampered, sr25519 ctx tests (matching/mismatched/empty/substrate),
  secp256k1 high-s consistency, boundary tests, zero-length inputs, invalid-v
  rejection, `SECP256K1_N_HALF` constant
- `demo-crypto` gas_delta test exists and passes
- `make fmt`, `make clippy-gear`, `make typos` clean
- `vara-runtime syscall_weights_test` green (82-field count)

## Deferred

1. **Weights are `Weight::zero()` placeholders.** Real benchmarks blocked on
   pre-existing polkadot-sdk `runtime-benchmarks` feature bit-rot (reproduces
   on master).
2. `make test` + `make clippy-examples` fail with pre-existing
   polkadot-sdk / macOS issues (reproduced on master).


breathx:
Closing for now — work will continue in a fresh PR.

PR #5367, opened 2026-04-21
feat(gear/core): strip WASM custom sections from `InstrumentedCode`

ukint-vs:
## Summary

Strip non-name WASM custom sections (for example `sails:idl` and `producers`) from persisted `InstrumentedCode`, while preserving the original uploaded bytes for RPC readers and preserving the WASM `name` section for readable trap backtraces.

Custom sections are not consumed by sandbox execution, but keeping them in `InstrumentedCode` inflates the bytes loaded into Wasmer/Wasmtime. `OriginalCode` remains unchanged, so `gear_readWasmCustomSection` and IDL consumers continue to work.

### Shared instrumentation path

- Added `Module::strip_custom_sections()` in `protocol/wasm-instrument`.
- The method clears non-name custom sections and keeps `name_section`.
- `Code::try_new_internal` calls it after instrumentation and before serializing/storing `InstrumentedCode`.
- Avoided extra section cloning in `Module::serialize` while touching this path.

### Vara / pallet-gear

- Bumped `Schedule::InstructionWeights::version` from `1900` to `1910`.
- This uses the existing lazy re-instrumentation path for deployed programs; no storage migration is added.
- Added a pallet regression that simulates stale `InstrumentedCode` containing `sails:idl` and verifies re-instrumentation strips it.
- Regenerated `vara/sdk/gsdk/vara_runtime.scale`.

### ethexe

- Added an `ethexe-processor` regression test proving the shared `Processor::process_code` path preserves `OriginalCode` and strips `InstrumentedCode`.
- No ethexe DB key, runtime version, RPC signature, or migration changes are included.

## Migration story

**Vara:** lazy re-instrumentation through the existing `needs_reinstrumentation` / `reinstrument_code` path after the instruction weights version bump.

**ethexe:** no migration in this PR. The previous ethexe DB/version/migration approach was removed from scope after review feedback.

## Tests

- `protocol/wasm-instrument`: `strip_custom_sections_clears_custom_but_keeps_name`, `strip_custom_sections_on_empty_module_is_noop`.
- `protocol/core`: `instrumented_code_strips_custom_sections_but_original_keeps_them`.
- `ethexe-processor`: `instrumented_code_strips_custom_sections`.
- `pallet-gear`: `stripping_reduces_instrumented_code_len`.
- `pallet-gear`: `reinstrumentation_strips_custom_sections`.

## Verification

```bash
./scripts/update-gsdk-metadata.sh
cargo nextest run -p gear-wasm-instrument -E 'test(strip_custom_sections)'
cargo nextest run -p gear-core -E 'test(instrumented_code_strips_custom_sections_but_original_keeps_them)'
cargo nextest run -p ethexe-processor -E 'test(instrumented_code_strips_custom_sections)'
cargo nextest run -p pallet-gear -E 'test(stripping_reduces_instrumented_code_len) | test(reinstrumentation_strips_custom_sections)'
env TERM=xterm make fmt-check
git diff --check
```

## Not in scope

- ethexe `InstrumentedCode` DB key layout changes.
- ethexe runtime `VERSION` / `RUNTIME_ID` changes.
- ethexe DB migrations or eager re-instrumentation.
- Re-running pallet runtime weight benchmarks on canonical hardware.


ethexe/db/src/migrations/v3.rs
```diff
@@ -0,0 +1,83 @@
...
+    let deleted = old_keys.len();
+    for key in old_keys {
+        // SAFETY: these entries are unreadable under the new 3-tuple key
+        // layout and the return value is intentionally discarded.
+        unsafe {
+            db.kv.take(&key);
+        }
+    }
+
+    log::info!(
```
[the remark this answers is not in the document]
ukint-vs: This is a false positive:

1. **No compilation error.** `log` is declared as a direct workspace dep of `ethexe-db` at `ethexe/db/Cargo.toml:23` (`log.workspace = true`). `log::info!` resolves via the crate path and does not require a `use log` import. Local `cargo check -p ethexe-db` passes.
2. **Logging convention.** The `migrations` module consistently uses `log::*!`:
   - `ethexe/db/src/migrations/v2.rs:46, 71, 115`
   - `ethexe/db/src/migrations/init.rs:37, 40, 59, 63, 78, 98`

   Switching v3 to `tracing::info!` would diverge from its siblings in the same module. The `tracing` convention lives in `ethexe/db/src/database.rs` — not a crate-wide rule.

Keeping `log::info!` to stay consistent with the other migration files.

ukint-vs:
@claude opus review

ukint-vs:
Addressed the hardcoded `(1, 2)` mock literal in `ddfc0bc27`:

- `ethexe-common::mock` now exports `MOCK_RUNTIME_ID` and `MOCK_VERSION` constants and uses them at the call site instead of the literal.
- `ethexe-runtime-common` picks up `ethexe-common` with the `mock` feature as a dev-dep and adds a `#[cfg(test)]` assertion that `RUNTIME_ID == MOCK_RUNTIME_ID && VERSION == MOCK_VERSION`.

The dep-cycle constraint (runtime-common → common, not the reverse) is unchanged, so the mock can't import the real constants directly. But now a future bump of `VERSION` or `RUNTIME_ID` that forgets to update the mock fails that test loudly instead of silently drifting.

Follow-up per @claude's review.

grishasobol (commented):
update the branch, you have side changes

ethexe/common/src/db.rs
```diff
@@ -76,8 +76,13 @@ pub trait CodesStorageRO {
     fn original_code_exists(&self, code_id: CodeId) -> bool;
     fn original_code(&self, code_id: CodeId) -> Option<Vec<u8>>;
     fn program_code_id(&self, program_id: ActorId) -> Option<CodeId>;
-    fn instrumented_code_exists(&self, runtime_id: u32, code_id: CodeId) -> bool;
-    fn instrumented_code(&self, runtime_id: u32, code_id: CodeId) -> Option<InstrumentedCode>;
+    fn instrumented_code_exists(&self, runtime_id: u32, version: u32, code_id: CodeId) -> bool;
+    fn instrumented_code(
+        &self,
+        runtime_id: u32,
+        version: u32,
```
grishasobol: What the reason of adding second version? As soon as version can be changed only when runtime is changed, then we can use only one here, runtime version or instrumentation version

ethexe/runtime/common/src/lib.rs
```diff
@@ -64,7 +64,7 @@ mod transitions;
 
 // TODO: consider format.
 /// Version of the runtime.
-pub const VERSION: u32 = 1;
+pub const VERSION: u32 = 2;
 pub const RUNTIME_ID: u32 = 1;
```
grishasobol: what the difference between version and runtime id ?

ethexe/db/src/migrations/v5.rs
```diff
@@ -0,0 +1,83 @@
...
+/// (runtime_id, code_id) → (runtime_id, version, code_id). Drop any entries
+/// stored under the old 2-tuple layout so the cluster re-instruments them on
+/// next code observation. `OriginalCode` is preserved in CAS, so reprocessing
+/// does not need to re-fetch.
+///
+/// Note: this migration only cleans stale DB state. It does not itself
+/// re-instrument existing codes — that relies on the normal code processing
+/// pipeline observing the code again. Operators upgrading a live node should
+/// expect a brief window where existing programs return
+/// `MissingInstrumentedCodeForProgram` until their codes are re-observed.
```
grishasobol: we do not have any re-instrumentation mechanism right now

grishasobol (changes requested):

ukint-vs:
Addressed @grishasobol's review in `cbee4d651`.

**Both "what's the difference between version and runtime_id" / "use only one here" comments**: collapsed. The `InstrumentedCode` key is now `(version, code_id)` — single `u32` discriminant. Dropped the `RUNTIME_ID` constant entirely (it had no callers outside DB-key sites) and kept `VERSION`, which is also what `Code::try_new` consumes as `instruction_weights_version` — so it carries real semantic weight. Trait signatures and the mock follow.

The public RPC `code_getInstrumented(runtime_id, code_id)` keeps its signature for backward compat; the `runtime_id` parameter is ignored internally and explicitly documented.

**"we do not have any re-instrumentation mechanism right now"**: fixed the v5 doc to stop claiming the compute pipeline re-instruments on next observation. It doesn't — `RouterEvent::CodeUploaded` only fires for fresh uploads. Doc now states honestly that programs whose entries get wiped will surface `MissingInstrumentedCodeForProgram` until the same `OriginalCode` is uploaded to chain again, which is acceptable pre-mainnet. Migration logic changed accordingly: old and new key shapes now have the same byte length (4 + 32), so the length-filter no longer helps — every entry under the discriminant is stale post-bump and gets wiped unconditionally.

Tests green: `ethexe-*` 134/134, `gear-core`/`gear-wasm-instrument` 151/151, `pallet-gear::stripping_reduces_instrumented_code_len` passes.

breathx:
Closing for now. Implementation is mostly done — main thing left is addressing @grishasobol's feedback. Good reference if someone wants to finish this.

ukint-vs:
removed the ethexe runtime/version/DB-key/migration changes, so the comments about `VERSION` vs `RUNTIME_ID`, the extra instrumentation version, and the migration/re-instrumentation path should now be outdated. The current diff only touches the shared WASM instrumentation path, Vara re-instrumentation trigger/tests, the ethexe processor regression test, and regenerated `vara_runtime.scale`.

ukint-vs:
@claude review

ethexe/processor/src/tests.rs
```diff
@@ -300,6 +300,56 @@ async fn handle_new_code_valid() {
     );
 }
 
+#[tokio::test]
+async fn instrumented_code_strips_custom_sections() {
```
breathx: could be just protocol/core code mod unit test

protocol/core/src/code/mod.rs
```diff
@@ -1253,4 +1255,82 @@ mod tests {
             ))
         ));
     }
+
+    /// Walks a WASM binary and returns `true` if it contains a custom section
+    /// with the given name.
+    fn has_custom_section(wasm: &[u8], name: &str) -> bool {
```
breathx: it's very heavy check, can't be done in orig code on init parse and/or creation?

vara/pallets/gear/src/schedule.rs
```diff
@@ -899,7 +899,7 @@ impl<T: Config> Default for InstructionWeights<T> {
         // See below for the assembly listings of the mentioned instructions.
         type W<T> = <T as Config>::WeightInfo;
         Self {
-            version: 1900,
+            version: 1910,
```
breathx:
```suggestion
            version: 1901,
```

protocol/core/src/code/mod.rs
breathx: or mv to tests if only for them

breathx (changes requested):
otherwise lgtm

breathx (approved):

grishasobol (approved):

PR #5369, opened 2026-04-21
fix(ci): bump Claude review model to Opus 4.7

grishasobol:
Opus 4.6 is listed as legacy; Anthropic recommends migrating to 4.7. Since April 16 the review workflow fails with an SDK auth error shortly after process start — bumping the model is the safest first step before pinning the action version.


grishasobol:
@claude review

ukint-vs (approved):

PR #5381, opened 2026-04-28
feat(ethexe/blob-loader): extend BlobLoader tests and add mock BlobReader

### LINKED ISSUE #4674: `ethexe`: write tests for BlobLoaderService
### Problem to Solve

.

### Possible Solution

.

### Notes

_No response_

playX18:
Resolves #4674


StackOverflowExcept1on:
Maybe we'd rather not have MockBlobLoader, but rather test various cases of ConsensusLayerBlobReader? For example: beacon rpc disconnect, the case when chain id != anvil, blob re-request, invalid json data.

ethexe/blob-loader/src/lib.rs
```diff
@@ -383,14 +393,140 @@ fn handle_blob(
 #[cfg(test)]
 mod tests {
```
kuzminxyz: tests are (895 - 394) * 100 / 895 =55,9 % of file, so maybe move them into `tests.rs`?

ethexe/blob-loader/src/lib.rs
```diff
@@ -383,14 +393,140 @@ fn handle_blob(
...
+            provider: ProviderBuilder::default()
+                .connect(&ethereum_rpc)
+                .await
+                .expect("test reader should connect to ethereum rpc"),
+            http_client: Client::new(),
+            config: ConsensusLayerConfig {
+                ethereum_rpc,
+                ethereum_beacon_rpc,
+                beacon_block_time,
+                attempts: const { NonZero::new(3).unwrap() },
```
kuzminxyz: why `const`?)

ethexe/blob-loader/src/lib.rs
```diff
@@ -383,14 +393,140 @@ fn handle_blob(
...
+            },
+        }
+    }
+
+    async fn run_beacon_server(responses: Vec<String>) -> (String, Arc<Mutex<Vec<String>>>) {
```
kuzminxyz: what?? It is like a mock beacon server?

The purpose of this function is to simulated the invalid RPC cases?

kuzminxyz (commented):
From my view there is a number of useless tests, for example, tests connected with fake beacon server. I think we do not need to test de-/serialise errors.

ethexe/blob-loader/src/lib.rs
playX18: Addressed in DM; Added doc comment for this func. TLDR; it is a necessity to test ConsensusLayerBlobReader re-try policies, failures and so on

breathx (approved):

PR #5389, opened 2026-04-29
feat(ethexe/observer): test observer resubscription logic

### LINKED ISSUE #4568: `ethexe`: test creating a new subscription in case when Receiver becomes invalid
Subscription object in `alloy` is a wrapper on top of tokio Receiver.
When stream is terminated Receiver and Subscription become invalid.

So we need to write the test with stream terminating and test the correct creation a new subscription. `

playX18:
Resolves #4568


StackOverflowExcept1on (approved):

kuzminxyz (approved):

kuzminxyz (commented):
Maybe also should add small check that subscription starts with previous block

PR #5390, opened 2026-04-29
feat(ethexe/observer): use alloy ChunkedEvent for batched log queries

grishasobol:
## Summary

- Replace the hand-rolled 256-block chunking around `eth_getLogs` in `EthereumBlockLoader::load_many` with [`alloy::contract::ChunkedEvent`](https://docs.rs/alloy/latest/alloy/contract/struct.ChunkedEvent.html). The helper attempts the full range first, then splits into configurable-size windows queried concurrently, then falls back to per-block queries on failures — so it gets us a faster best case and a more robust worst case for free.
- Block-header batching (eth_getBlockByNumber) is kept as before and now runs in parallel with the log query via `try_join`, so we save the back-pressure between the two.
- Adds an ignored `bench::bench_load_many` test that times `load_many` against an in-process Anvil so the speedup is reproducible locally.

## Numbers (local Anvil, 5 runs each)

| Range | Old (custom chunked) | New (alloy ChunkedEvent) |
| --- | --- | --- |
| 1500 blocks | avg 131.0 ms | avg 120.5 ms (~8% faster) |
| 5000 blocks | avg 445.1 ms | avg 439.5 ms (~1% faster) |

The bigger win on smaller ranges is the optimistic full-range query that ChunkedEvent attempts first — Anvil never refuses it, so chunking overhead disappears entirely. On real RPCs that reject large `getLogs` ranges, ChunkedEvent's per-block fallback still gives us robustness we did not have before.

## Test plan

- [x] `cargo nextest run -p ethexe-observer`
- [x] `cargo nextest run -p ethexe-ethereum`
- [x] `cargo clippy -p ethexe-observer --tests`
- [x] Manual benchmark: `cargo nextest run -p ethexe-observer --run-ignored only --no-capture bench::bench_load_many`


grishasobol:
@claude opus review

grishasobol:
@claude fix and delete appended benches  - they are not needed now (only for oneshot difference calculation)

PR #5393, opened 2026-04-29
feat(ethexe/node-loader): add per-batch watchdog, SIGTERM, heartbeat

grishasobol:
- Wrap each batch in a 180s `tokio::time::timeout` so a hung RPC call drops the worker's batch instead of stalling the pool forever.
- Listen for `SIGTERM` in addition to `SIGINT` so `docker stop` triggers the same graceful drain + summary as `Ctrl+C`.
- Emit an INFO heartbeat every 10s with `completed/failed/in_flight` counters so a stalled pool is visible without grepping DEBUG.

### DIFF
--- a/ethexe/node-loader/src/batch.rs
+++ b/ethexe/node-loader/src/batch.rs
@@ -36,6 +36,7 @@ use std::{
     collections::{BTreeMap, BTreeSet},
     marker::PhantomData,
     sync::Arc,
+    time::Duration,
 };
 use tokio::sync::{
     RwLock,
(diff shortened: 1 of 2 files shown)


grishasobol:
@claude fix review and also fix windows problems https://github.com/gear-tech/gear/actions/runs/25118845247/job/73613584380?pr=5393

playX18 (approved):

ethexe/node-loader/src/batch.rs
```diff
@@ -117,6 +118,13 @@ type WorkerBatchFuture =
 
 const MAX_MULTICALL_CALLDATA_BYTES: usize = 120 * 1024;
 
+/// Per-batch watchdog: drop and reschedule a batch if a hung RPC call parks
+/// the worker. Generous: code validation alone takes ~14 s for 5 codes.
+const BATCH_TIMEOUT: Duration = Duration::from_secs(180);
```
playX18, line 123: might be worth making batch timeout somewhat dynamic, but 180s can be fine if batch size is not that large.

grishasobol: ok, I will change it if it would be enough

PR #5486, opened 2026-05-20
fix(ethexe-rpc): verify compact promise validators

### LINKED ISSUE #5385: SignedCompactPromise is not verified against the validator set
**What is broken?**
In `ethexe/rpc/src/apis/injected/promise_manager.rs`, received `SignedCompactPromise` items are accepted without verifying the signature against the known validator set.

**Steps to reproduce**
Submit a `SignedCompactPromise` signed by a non-validator key.

**Expected behavior**
Promise is rejected; only validator-signed promises are accepted.

**Actual behavior**
Promise is accepted regardless of who signed it.

playX18:
Closes #5385

  ## Summary

  Reject `SignedCompactPromise` values in `ethexe-rpc` unless the signer belongs to the current or next known validator set. The same validator check is applied when returning
  cached transaction promises, so stale invalid compact promises are not exposed through RPC.

  ## How to test

  cargo nextest run -p ethexe-rpc --features client

  ## Notes

  None.

  ## Checklist

  - [x] PR title follows Conventional Commits (`type(scope): description`)
  - [x] Single logical change
  - [x] Tests added or updated (if logic changed)
  - [x] Docs updated (if needed)


ethexe/rpc/src/apis/injected/promise_manager.rs
```diff
@@ -149,6 +155,60 @@ impl PromiseSubscriptionManager {
...
+                "current validator set not found for tx receipt validator check"
+            );
+            return false;
+        };
+
+        let signer_is_known = current_validators.contains(&address)
+            || current_era
+                .checked_add(1)
+                .and_then(|next_era| self.db.validators(next_era))
+                .is_some_and(|next_validators| next_validators.contains(&address));
```
grishasobol: I cannot understand this logic with next era validators:
For example election is already done - and we know next era validators, but current era is not yet finished. Why should we take in account next era validators here? what the reason for that?

I think you should take take in account not next, but previous era validators in the begining of a new era.

ethexe/rpc/src/apis/injected/server.rs
```diff
@@ -147,7 +147,13 @@ impl InjectedApi {
         tx_hash: HashOf<InjectedTransaction>,
     ) -> RpcResult<Option<SignedTxReceipt>> {
         match self.db.receipt(tx_hash) {
-            Some(receipt) => Ok(Some(receipt)),
+            Some(receipt) => {
+                if !self.manager.tx_receipt_signed_by_known_validator(&receipt) {
+                    return Ok(None);
+                }
+
+                Ok(Some(receipt))
```
grishasobol: what the reason of checking already checked receipts (stored in db only if checked)?

grishasobol (changes requested):

playX18:
@claude review

grishasobol (approved):

PR #5493, opened 2026-05-21
fix(gear/lazy-pages): forward faults outside managed regions

### LINKED ISSUE #5492: lazy-pages fault handler aborts the node on faults from threads without a runtime context
### Problem

The lazy-pages fault handler is installed process-wide (`sigaction` on Linux/macOS, a vectored exception handler on Windows). When a fault is raised on a thread that never entered a lazy-pages WASM execution, that thread has no runtime context, and `user_signal_handler` returns `Error::GlobalContext(ContextError::RuntimeContextIsNotSet)`.

`handle_sigsegv` (`protocol/lazy-pages/src/sys/unix.rs`) and `exception_handler` (`protocol/lazy-pages/src/sys/windows.rs`) only forward `OutOfWasmMemoryAccess` and `WasmMemAddrIsNotSet` to the previous handler / OS exception chain. A `GlobalContext` error hits the fallback branch and the handler `panic!`s.

Panicking inside a fault handler is not async-signal-safe: it turns any unrelated fault into a node abort (`fatal runtime error: failed to initiate panic, error 3, aborting`).

Additionally, on unix `old_sig_handler` returns `false` when the previously installed disposition is `SIG_DFL`/`SIG_IGN` (or nothing was captured), which also leads the caller to `panic!` — so even forwarding `GlobalContext` is not enough on its own.

Observed on ethexe testnet validators: a SIGSEGV raised by RocksDB on an RPC thread (which never ran WASM) was turned into a panic-driven abort instead of a clean crash.

### Expected

A fault on a thread with no lazy-pages runtime context is not a lazy-pages page fault. The handler must forward it — to the previously installed handler / the OS exception chain, falling through to the default action — instead of panicking inside the handler.

### Notes

Related to #5051 (broader async-signal-safety violations in the handler). This issue is the specific foreign-fault path that aborts the node; it does not fix the underlying foreign SIGSEGV, only stops lazy-pages from masking it as a handler panic.

### LINKED ISSUE #5496: lazy-pages fault handler runs non-async-signal-safe code before classifying the fault
### Problem

The lazy-pages fault handler (`handle_sigsegv` on unix, `exception_handler` on Windows) dispatches straight into `user_signal_handler`, which accesses thread-local storage, borrows a `RefCell` and logs (allocating) — before it knows whether the fault is a lazy-pages page fault at all.

None of those operations are async-signal-safe. They are only safe when the interrupted thread was executing WASM and faulted on lazy-pages-managed memory: such a thread holds no allocator or logger lock. For a fault raised anywhere else — e.g. a SIGSEGV on an RPC thread (see #5492) — the interrupted code may hold the allocator or logger lock, so doing this work in the handler risks a deadlock or a panic.

### Expected

The handler must decide whether the fault is inside lazy-pages-managed WASM memory using only async-signal-safe operations, and run the non-async-signal-safe processing only for a confirmed in-region fault. A fault outside managed memory must be forwarded immediately, without touching thread-locals, the `RefCell` or the logger.

grishasobol:
## Summary

The lazy-pages fault handler is installed process-wide. It used to dispatch every fault — including ones raised on threads that never entered a lazy-pages WASM execution — straight into `user_signal_handler`, which accesses thread-local storage, borrows a `RefCell` and logs (allocating). None of that is async-signal-safe, and on a foreign fault (e.g. a SIGSEGV from RocksDB on an RPC thread) the handler `panic!`ed — aborting otherwise-healthy validator nodes (`fatal runtime error: failed to initiate panic, error 3, aborting`).

The handler now **classifies the fault before doing any non-async-signal-safe work**:

- **`ACTIVE_WASM_REGION`** (`lib.rs`) — a `const`-initialized thread-local recording the WASM memory range lazy-pages currently protects on the thread. The `const` form is a plain static TLS slot (no lazy initialization, no allocation, no destructor), so it is safe to read from the handler. It is kept in sync at the whole-buffer protect/unprotect sites (`initialize_for_program`, `set_lazy_pages_protection`, `unset_lazy_pages_protection`, `change_wasm_mem_addr_and_size`).
- **`handle_sigsegv` (unix) / `exception_handler` (Windows)** — read only the fault address and that TLS slot. An address **outside** the managed region is not a lazy-pages page fault: it is forwarded immediately (to the previously installed handler / the OS exception chain) without touching `LAZY_PAGES_CONTEXT`, the `RefCell` or the logger. Only an **in-region** fault — provably on a thread that was executing WASM and holds no allocator/logger lock — proceeds to the full processing.
- **`old_sig_handler` (unix)** — when there is no chainable previous handler (`SigDfl`/`SigIgn`, or nothing captured at install time), restore the default disposition and report success instead of returning `false` (which made the caller `panic!`). The re-executed faulting instruction is then terminated by the kernel's default action.
- An in-region fault that `H::handle` still declines is a lazy-pages **invariant violation**, not a foreign fault — it `panic!`s (safe here, on a clean thread; the backtrace points at the bug) rather than being forwarded.
- Adds a Linux regression test (`tests/sigsegv_handler.rs`); existing unit tests that protect memory directly now register their region.

This does **not** fix the underlying foreign SIGSEGV — it stops lazy-pages from masking it. A foreign fault now dies via the kernel's default action (→ core dump), and the handler no longer runs non-async-signal-safe code for it.

Closes #5492
Closes #5496


grishasobol:
@claude opus review

protocol/lazy-pages/tests/sigsegv_handler.rs
```diff
@@ -0,0 +1,125 @@
...
+//! validator nodes.
+//!
+//! This test raises a SIGSEGV on a thread running no lazy-pages execution
+//! and asserts the process dies cleanly via the default action, not via a
+//! handler panic.
+//!
+//! The handler intercepts SIGSEGV only on Linux (on other unixes it is
+//! installed for SIGBUS), so the regression is reproducible on Linux only.
+
+#![cfg(unix)]
```
ark0f:
```suggestion
#![cfg(target_os = "linux")]
```

ark0f:
@codex review

protocol/lazy-pages/src/sys/unix.rs
```diff
@@ -80,22 +77,35 @@ where
...
+        // is safe in this context.
         let is_write = ucontext_get_write(ucontext as *mut _);
         let exc_info = ExceptionInfo {
             fault_addr: addr as *mut _,
             is_write,
         };
 
         if let Err(err) = H::handle(exc_info) {
-            let old_sig_handler_works = match err {
-                Error::OutOfWasmMemoryAccess | Error::WasmMemAddrIsNotSet => {
```
ark0f, line 91: Consider making these 2 errors panic

protocol/lazy-pages/src/sys/windows.rs
```diff
@@ -25,16 +22,23 @@ where
     let is_access_violation =
         unsafe { (*exception_record).ExceptionCode == EXCEPTION_ACCESS_VIOLATION };
     let num_params = unsafe { (*exception_record).NumberParameters };
+    // Not an access violation — not a lazy-pages page fault. Hand it back to
+    // the OS exception chain without running any non-async-signal-safe work
```
ark0f: Windows does not have `async-signal-safety` definition for their exception handlers. The only meaningful safety remarks I found in Microsoft docs: https://learn.microsoft.com/en-us/windows/win32/api/winnt/nc-winnt-pvectored_exception_handler#remarks

ark0f (approved):

protocol/lazy-pages/src/sys/windows.rs
grishasobol: fixed

PR #5508, opened 2026-05-22
refactor(vara.eth): introduce typed MB/EB hashes, move MB to ethexe-common

### LINKED ISSUE #5507: ethexe: typed MB/EB hashes via HashOf<MB> and HashOf<EB>
## Problem

`H256` is the lingua franca for many distinct hash kinds across ethexe — Ethereum block hashes, malachite envelope hashes, payload digests, etc. Mixing them at signature level is dangerous; the most visible footgun today is the MB hash vs the EB (Ethereum block) hash, which both appear in `CompactMb`, in `MalachiteEvent`, and in the `MbStorageRO/RW` trait surface as bare `H256`. Compiler can't help when a caller swaps one for the other.

## Goal

- Add `ethexe-common` as a dep of `ethexe-malachite-core` (currently leaf-isolated).
- Move `ethexe_malachite_core::Block` to `ethexe_common::malachite::MB` (carrying `parent_hash: HashOf<MB>`, `height: u64`, `payload: BlockPayload`, `reserved: [u8; 64]`). `BlockPayload` moves with it.
- Restructure `CompactMb`: same fields as `MB` but with `payload_hash: H256` in place of `BlockPayload`. Move it to `ethexe_common::malachite` alongside `MB`.
- Replace `H256` with `HashOf<MB>` everywhere it represents a malachite block hash: trait sigs in `MbStorageRO/RW`, `DBGlobals::latest_finalized_mb_hash`, `Externalities` callbacks, `CommitCertificate::block_hash`, `MalachiteEvent::*::mb_hash`, etc.
- Introduce `pub type EB = SimpleBlockData;` and a `SimpleBlockData::typed_hash() -> HashOf<EB>` that lifts the existing Ethereum block hash into the typed wrapper without recomputing.

`HashOf<T>` encodes/decodes as 32 bytes (phantom is `#[codec(skip)]`), so the wire format stays byte-identical to bare `H256`.

## Non-goals

- No data migrations — DB can change shape freely on this branch.

## Acceptance

- `ethexe-common::malachite` exports `MB`, `BlockPayload`, `CompactMb`, `EB`.
- `MbStorageRO/RW` keys take `HashOf<MB>` not `H256`.
- `DBGlobals::latest_finalized_mb_hash`, `CompactMb::parent`, `Externalities` callback hash args all typed.
- All ethexe-* tests pass.

grishasobol:
## Summary

- Move the consensus block envelope (`ethexe_malachite_core::Block` → `ethexe_common::malachite::MB`), the versioned `BlockPayload`, and `CompactMb` into `ethexe-common` so the executor crates can use them without depending on the consensus layer.
- `CompactMb` now matches `MB`'s shape with the opaque payload replaced by `payload_hash` (renamed from `transactions_hash`).
- Type-safe hashes for boundary fields: `HashOf<MB>` on `CompactMb.parent`, `BlockMeta.last_committed_mb`, `DBGlobals.{latest_finalized_mb_hash, latest_computed_mb_hash}`; `HashOf<EB>` on `MbMeta.last_advanced_eb`, `BlockMeta.last_committed_eb`, `DBGlobals.{start_block_hash, latest_prepared_eb_hash}`, `DBConfig.genesis_block_hash`.
- New `pub type EB = SimpleBlockData` + `SimpleBlockData::typed_hash()` / `eb_hash_of(H256)` to lift the bare Ethereum block hash into the typed wrapper.
- `HashOf<T>` is `Encode + Decode` as a bare 32-byte hash (phantom is `#[codec(skip)]`), so the on-wire/on-disk layout is byte-identical to before — only the static type at the field changes.
- `MbStorageRO/RW` trait method parameters intentionally keep `mb_hash: H256`. Threading `HashOf<MB>` through every call site in compute/cli/consensus/processor/etc. would cascade through dozens of files for no extra safety gain — the boundary structs already catch the EB-vs-MB confusion that motivated the refactor. Call sites pass `field.inner()` where needed.

Fixes #5507.

## Test plan

- [x] `cargo nextest run -p ethexe-common -p ethexe-db -p ethexe-malachite-core -p ethexe-malachite -p ethexe-compute -p ethexe-consensus` — 248/248 pass (incl. multi-validator integration tests).
- [x] `cargo fmt`, `cargo clippy --workspace --tests` — clean.
- [x] `ensure_types_unchanged` updated with the new `EXPECTED_TYPE_INFO_HASH` (CompactMb shape changed).
- [ ] CI green.

### DIFF
--- a/ethexe/cli/src/commands/check.rs
+++ b/ethexe/cli/src/commands/check.rs
@@ -337,7 +337,7 @@ impl Checker {
                 pb.inc(1);
             };
 
-            if current_compact_mb.parent == H256::zero() {
+            if current_compact_mb.parent.is_zero() {
                 break;
             }
             current_mb = current_compact_mb.parent;
--- a/ethexe/cli/src/commands/check.rs
+++ b/ethexe/cli/src/commands/check.rs
@@ -19,7 +19,6 @@ use ethexe_db::{
 };
 use ethexe_processor::{Processor, ProcessorConfig};
 use ethexe_runtime_common::FinalizedBlockTransitions;
-use gprimitives::H256;
 use indicatif::{ProgressBar, ProgressStyle};
 use std::{collections::HashSet, path::PathBuf};
 
--- a/ethexe/cli/src/commands/dump.rs
+++ b/ethexe/cli/src/commands/dump.rs
@@ -4,7 +4,7 @@
 use crate::params::{MergeParams, Params};
 use anyhow::{Context, Result};
 use clap::{Parser, Subcommand};
-use ethexe_common::db::GlobalsStorageRO;
+use ethexe_common::{EB, HashOf, db::GlobalsStorageRO};
 use ethexe_db::{Database, RawDatabase, RocksDatabase, dump::StateDump};
 use gprimitives::H256;
 use std::path::{Path, PathBuf};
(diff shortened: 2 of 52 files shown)


grishasobol:
Pushed refactor to type DB trait signatures with `HashOf<EB>`/`HashOf<MB>` end-to-end (32 files, +628/-411). `MbStorageRO/RW`, `OnChainStorageRO/RW` and `BlockMetaStorageRO/RW` now take typed values; `.inner()` only at the explicit boundary with external systems (chain events, malachite-core API, test fixtures).

CI: https://github.com/gear-tech/gear/actions/runs/26400404637

PR #5510, opened 2026-05-25
test(vara.eth): pick claim-MB stateHash in process_outgoing_action tests

grishasobol:
## Problem

Three tests in `ethexe-service` (`mailbox`, `value_send_program_to_user_and_replied`, `value_send_program_to_user_and_claimed`) were racy and failed intermittently with `OutgoingActionMerkleRootNotFound`.

`Mirror._merkleRoots[stateHash]` is set inside `performStateTransition` to the merkle root of `value_claims` for that specific transition. Empty `value_claims` → stored value is `bytes32(0)` → `processOutgoingAction` reverts.

Tests called `mirror.claim_value(mid)` then sent a no-op "force-fold" `send_message(pid, b"")` to push the MB tick. When the claim and the no-op fell into **different** MBs (which depends on timing), the latest stateHash on chain pointed at the no-op's MB whose merkleRoot was zero — and `mirror.query().state_hash()` returned exactly that stateHash.

## Fix

- New helper `find_value_claim_proof` subscribes to the local observer event stream, watches `Mirror.StateChanged` for the actor, and for each new stateHash probes `db.outgoing_actions(stateHash)` to find the one whose actions contain the claimed `message_id`.
- All three tests now create the receiver **before** the trigger and pull `(state_hash, total_leaves, leaf_index, proof)` from the helper instead of latest-state + hardcoded `(1, 0, vec![])`.
- `mailbox` separately re-derives the surrounding `mb_hash` via `db.mb_outcome(mb_hash)` — `Router.MBCommitted` is emitted **after** `Mirror.StateChanged` in the same eth block, so a single-pass `latest_mb_hash` tracker misses it.

## Verification

Five back-to-back release runs of all three tests, all green:

```
=== run 1 === Summary 3 passed
=== run 2 === Summary 3 passed
=== run 3 === Summary 3 passed
=== run 4 === Summary 3 passed
=== run 5 === Summary 3 passed
```

Each run: `cargo nextest run --release -p ethexe-service -E 'test(mailbox) | test(value_send_program_to_user_and_replied) | test(value_send_program_to_user_and_claimed)'`.

### DIFF
--- a/ethexe/service/src/tests/mod.rs
+++ b/ethexe/service/src/tests/mod.rs
@@ -6,8 +6,9 @@
 pub(crate) mod utils;
 
 use crate::tests::utils::{
-    EnvNetworkConfig, InfiniteStreamExt, NodeConfig, TestEnv, TestEnvConfig, TestingEvent,
-    TestingNetworkEvent, TestingRpcEvent, ValidatorsConfig, init_logger, stop_nodes, test_info,
+    EnvNetworkConfig, InfiniteStreamExt, NodeConfig, ObserverEventReceiver, TestEnv, TestEnvConfig,
+    TestingEvent, TestingNetworkEvent, TestingRpcEvent, ValidatorsConfig, init_logger, stop_nodes,
+    test_info,
 };
 use alloy::{
     primitives::U256,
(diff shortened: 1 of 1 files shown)


grishasobol:
CI dispatched: https://github.com/gear-tech/gear/actions/runs/26392533779

PR #5511, opened 2026-05-25
fix(ethexe):  Injected Transaction hash implementation

### LINKED ISSUE #5417: InjectedTransaction ToDigest implementation produces incorrect hash
[injected.rs](ethexe/common/src/injected.rs)

The problem connected with `ToDigest` implementation for `InjectedTransaction`: 

```rust
impl ToDigest for InjectedTransaction {
    fn update_hasher(&self, hasher: &mut Keccak256) {
        let Self {
            destination,
            payload,
            value,
            reference_block,
            salt,
        } = self;

        destination.into_bytes().update_hasher(hasher);
        payload.update_hasher(hasher);
        value.to_be_bytes().update_hasher(hasher);
        reference_block.0.update_hasher(hasher);
        salt.update_hasher(hasher);
    }
}
```

Current hash implementation uses concatenation of fields bytes. This implementation allows create two different transactions with the same hash (shuffling bytes between value and payload for example). 

The pull request must change the mechanism of hash calculation.

kuzminxyz:
Closes #5417 

## Notes
The transaction hash is changed. This pull request must be applied before regenesis (with malachite PR).


ethexe/common/src/injected.rs
```diff
@@ -74,8 +77,16 @@ pub struct InjectedTransaction {
     pub salt: LimitedVec<u8, MAX_INJECTED_TX_SALT_SIZE>,
 }
 
-impl ToDigest for InjectedTransaction {
-    fn update_hasher(&self, hasher: &mut Keccak256) {
+// Destination + payload_hash + value + ref_block + salt_hash
+const INJECTED_TX_HASH_SIZE: usize =
```
grishasobol: naming tells me that it's "hash size", which is actually missleading. Maybe better something like:         

```suggestion
const INJECTED_TX_HASHABLE_SIZE: usize =
```

ethexe/common/src/injected.rs
```diff
@@ -74,8 +77,16 @@ pub struct InjectedTransaction {
...
+    size_of::<ActorId>() + 32 + size_of::<u128>() + size_of::<H256>() + 32;
```
grishasobol:
```suggestion
    size_of::<ActorId>() + size_of::<H256>() + size_of::<u128>() + size_of::<H256>() + size_of::<H256>();
```

ethexe/common/src/injected.rs
```diff
@@ -74,8 +77,16 @@ pub struct InjectedTransaction {
...
+
+impl InjectedTransaction {
+    /// Helper function that returns bytes of [InjectedTransaction]
+    /// that will be hashed by blake2b256 or keccak256.
+    fn to_hashable_bytes(&self) -> Vec<u8> {
+        let mut hashable_buffer = Vec::with_capacity(INJECTED_TX_HASH_SIZE);
```
grishasobol: What the reason to use Vector of const size instead of const size array here? redundant allocations

ethexe/common/src/injected.rs
```diff
@@ -113,6 +116,12 @@ impl InjectedTransaction {
     }
 }
 
+impl ToDigest for InjectedTransaction {
+    fn update_hasher(&self, hasher: &mut Keccak256) {
+        let hashable_bytes = self.to_hashable_bytes();
+        hasher.update(hashable_bytes);
+    }
+}
```
grishasobol, line 154:
```suggestion
}

```

grishasobol (approved):
Also better to append test (failed before and passed after your patch)

ethexe/common/src/injected.rs
kuzminxyz: Fixed in [06a5731](https://github.com/gear-tech/gear/pull/5511/commits/06a5731ae87532b659ad56af4fbb7e2f6105a028)

ethexe/common/src/injected.rs
kuzminxyz: Fixed in [06a5731](https://github.com/gear-tech/gear/pull/5511/commits/06a5731ae87532b659ad56af4fbb7e2f6105a028)

ethexe/common/src/injected.rs
kuzminxyz: Fixed in [06a5731](https://github.com/gear-tech/gear/pull/5511/commits/06a5731ae87532b659ad56af4fbb7e2f6105a028)

ethexe/common/src/injected.rs
kuzminxyz: Fixed in [06a5731](https://github.com/gear-tech/gear/pull/5511/commits/06a5731ae87532b659ad56af4fbb7e2f6105a028)

PR #5543, opened 2026-05-31
[diagnostic, do-not-merge] repro WAL advisory-lock race on master

grishasobol:
This is a **diagnostic PR**, not for merging.

It modifies \`seven_validators_full_network_restart\` so the WAL advisory-lock race
(\`MalachiteService::shutdown\` returns before the upstream Malachite WAL writer
\`std::thread\` releases \`flock\` on \`consensus.wal\`) is reliably surfaced on CI.

Expected outcome: \`build / workspace (release)\` fails with

\`\`\`
service starts: building Malachite engine: Actor panicked during startup
'Failed to acquire exclusive advisory lock: the file is already locked'
\`\`\`

at \`ethexe/malachite/core/tests/multi_validators.rs:start_service\`. The companion
branch \`gsobol/ethexe/malachite-wal-fix\` applies the production fix in
\`MalachiteService::shutdown\` and is expected to pass under the same conditions.

See https://github.com/gear-tech/gear/actions/runs/26678194627/job/78634373130 for the
original master-side flake this repros.

### DIFF
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -63,18 +63,19 @@ jobs:
       - name: "Build: malachite-core tests"
         run: cargo build -p ethexe-malachite-core --tests --release
 
-      - name: "Test: malachite seven_validators_full_network_restart"
-        # Run the target test 5 times to make a flake-free verdict cheap.
-        # Without the fix this is expected to fail on the very first run;
-        # with the fix all 5 iterations must pass.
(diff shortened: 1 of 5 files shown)


grishasobol:
Superseded by #5546 (the actual fix). This diagnostic branch only exists as a trail of how the WAL advisory-lock race was isolated.

PR #5544, opened 2026-05-31
[diagnostic, do-not-merge] fix for WAL advisory-lock race + same repro

grishasobol:
**Diagnostic PR** — paired with #5543.

Carries the same race-repro test modification AND the production fix in
\`MalachiteService::shutdown\` (poll \`consensus.wal\` advisory lock until
the upstream WAL writer std::thread releases it, before returning).

Expected: \`build / workspace (release)\` is green — 5/5 iterations of
\`seven_validators_full_network_restart\` pass.

The final fix-only PR will drop the repro test stress and the CI strip;
this branch is only for proving the diagnosis.

### DIFF
--- a/.github/workflows/PR.yml
+++ b/.github/workflows/PR.yml
@@ -3,7 +3,7 @@ name: PR & Merge queue
 on:
   pull_request:
     branches: [master]
-    types: [labeled, synchronize]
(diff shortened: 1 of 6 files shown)


grishasobol:
Superseded by #5546 (the actual fix). This was the prove-the-fix branch (same probe test + fix applied).

PR #5560, opened 2026-06-03
fix(vara.eth): handle program code reinstrumentation failure gracefully

### LINKED ISSUE #5559: Reinstrumentation failure halts the whole block instead of failing only the affected program
## Problem

When a program's instrumented code is missing for the current runtime (e.g. after a runtime / instrumentation version bump), the processor lazily re-instruments the original code during execution. If re-instrumentation fails (the code is rejected under the current rules), the processor returns a hard `MissingInstrumentedCodeForProgram` error that propagates up and **aborts the entire block** — taking down every other program in the block with it.

## Goal

Isolate the failure to the affected program:

- a failed (re)instrumentation MUST NOT abort the block;
- record the failure in the code metadata (`InstrumentationFailed { version }`) so it is not retried on every subsequent run;
- reply to the dispatch with an explicit `UnavailableActor(ReinstrumentationFailure)` error.

grishasobol:
Closes #5559.

## What

When a program's instrumented code is missing for the current runtime, the processor lazily (re)instruments the original code during execution. Previously a failed (re)instrumentation returned a hard `MissingInstrumentedCodeForProgram` error that **aborted the whole block**.

This PR makes a failed (re)instrumentation non-fatal and isolated to the program:

- `instrumented_code_and_metadata` returns `Ok(None)` instead of erroring, so the block keeps running;
- the failure is recorded as `InstrumentationStatus::InstrumentationFailed { version }` so it is not retried on later runs;
- the dispatch is routed to `process_reinstrumentation_error`, replying with `UnavailableActor(ReinstrumentationFailure)`.

## Tests

Processor integration tests through `process_queues`:

- (previous-runtime instrumented code: yes / no) × (current (re)instrumentation: succeeds / fails) — 4 cases, asserting code metadata + the error / normal reply;
- `reinstrumentation_not_retried_when_already_failed` — a recorded failure short-circuits without re-instrumenting (proved with a *valid* code that would otherwise instrument successfully).

### DIFF
--- a/ethexe/compute/src/codes.rs
+++ b/ethexe/compute/src/codes.rs
@@ -54,7 +54,7 @@ impl<P: ProcessorExt> CodesSubService<P> {
                 );
                 debug_assert!(
                     self.db
-                        .instrumented_code_exists(ethexe_runtime_common::VERSION, code_id),
+                        .instrumented_code_exists(ethexe_runtime_common::RUNTIME_ID, code_id),
                     "Instrumented code {code_id:?} must exist in database"
                 );
             }
@@ -75,7 +75,7 @@ impl<P: ProcessorExt> CodesSubService<P> {
                     {
                         db.set_original_code(&code);
                         db.set_instrumented_code(
-                            ethexe_runtime_common::VERSION,
+                            ethexe_runtime_common::RUNTIME_ID,
                             code_id,
                             instrumented_code,
                         );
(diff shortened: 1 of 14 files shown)


playX18 (approved):

ethexe/compute/src/codes.rs
```diff
@@ -54,7 +54,7 @@ impl<P: ProcessorExt> CodesSubService<P> {
                 );
                 debug_assert!(
                     self.db
-                        .instrumented_code_exists(ethexe_runtime_common::VERSION, code_id),
+                        .instrumented_code_exists(ethexe_runtime_common::RUNTIME_ID, code_id),
```
breathx, line 57: let's already use schedule version here and in similar places

breathx (changes requested):

ethexe/compute/src/codes.rs
breathx: upd: for easiness you can just reexport it from runtime

breathx (approved):

PR #5598, opened 2026-06-22
fix(vara.eth):  batch-commitment refactor; correct validators rotation in malachite

grishasobol:
Finalizes the ethexe consensus **batch-commitment refactor** and **fixes validator rotation across eras** in the malachite layer. Compiles, clippy-clean, ethexe unit/integration tests pass.

## 1. Batch commitment

A batch is what validators agree on and submit to the Router. It bundles a few optional parts plus the chain progress:

```mermaid
flowchart TB
  B["BatchCommitment<br/>(block_hash, timestamp, expiry, previous_batch)"]
  B --> C["chain_commitment?"]
  B --> CO["code_commitments[]"]
  B --> V["validators_commitment?"]
  B --> R["rewards_commitment?"]
  C --> H["head MB"]
  C --> T["state transitions"]
  C --> E["last_advanced_eth_block"]
```

Normal (non-empty) chain commitments already worked. This PR fixes the **edge cases** where building the chain commitment could stall or loop:

- After a **validator-set handover**, a freshly joined validator may not have the compact block of the already-committed anchor MB. The walk now stops at the anchor's *hash* instead of dereferencing it.
- When the finalized chain doesn't descend from the committed anchor, the genesis zero-MB self-parent caused an **infinite walk** — it now stops at genesis.
- Empty **checkpoint** chain commitments are only emitted once the uncommitted gap passes a threshold, instead of riding along with other commitments.
- `is_strict_descendant_eth_block` — genesis-anchor equality fix.

## 2. Validator rotation across eras

Time is split into **eras**, and each era has its own on-chain validator set. A batch is verified against the set of the era its block falls into. A node must therefore hold the validator set for **every era it touches**, not just the current one.

```mermaid
flowchart LR
  subgraph "era N"
    VN["validators(N)"]
  end
  subgraph "era N+1"
    VN1["validators(N+1)"]
  end
  subgraph "era N+2"
    VN2["validators(N+2)"]
  end
  VN --> VN1 --> VN2
```

Fixed here:
- Malachite only allows the **first block of a new era** when advancing into it, so rotation happens at a clean boundary.
- On sync, the observer walks the freshly loaded chain and **backfills the validator set for each era it spans** (stops early once an era is already known), so a node that joined mid-chain can verify historical certificates.

## 3. `MbMeta` + DB migration (v1 → v2)

`MbMeta` gained a `finalized` flag and `last_advanced_eb` became optional:

```
v1:  { computed, last_advanced_eb: H256 }
v2:  { computed, finalized, last_advanced_eb: Option<H256> }
```

A v1 → v2 migration re-encodes every record: `last_advanced_eb` is wrapped in `Some(..)`, and `finalized` is backfilled `true` for every MB reachable from the latest finalized MB.

## Deliberately out of scope

Full **cross-era membership** — where a node that joined after several rotations can resolve and observe *any* historical era from a complete pubkey table (vote/propose only in elected eras, observe otherwise) — is a consensus feature, not a bug fix, and is left as a follow-up. The current resolver uses the node's own per-era set with a live-set fallback, which is enough because a node only runs consensus for eras it is elected to.

### DIFF
--- a/ethexe/compute/src/compute.rs
+++ b/ethexe/compute/src/compute.rs
@@ -252,8 +252,8 @@ pub fn prepare_executable_for_mb(
 /// Walk the MB's `Operations` list and prepare processor input.
 ///
 /// Synthetic block height/timestamp come from `last_advanced_eb` (the latest
-/// EB pinned by this MB or any ancestor); if none, fall back to the router's
-/// genesis block from [`ConfigStorageRO::config`].
+/// EB pinned by this MB or any ancestor); the zero sentinel falls back to the
+/// router's genesis block from [`ConfigStorageRO::config`].
 fn build_executable_data(
     db: &Database,
     operations: Operations,
@@ -596,6 +596,11 @@ mod tests {
                 operations_hash,
             },
         );
+        // Mimic `process_mb_proposal`, which always populates `last_advanced_eb`
+        // before an MB is computed.
+        db.mutate_mb_meta(mb_hash, |meta| {
+            meta.last_advanced_eb = Some(H256::zero());
+        });
     }
 
     /// `seed_mb` plus the malachite-side bookkeeping: record the advanced EB
--- a/ethexe/compute/src/compute.rs
+++ b/ethexe/compute/src/compute.rs
@@ -259,19 +259,20 @@ fn build_executable_data(
     operations: Operations,
     program_states: ethexe_common::ProgramStates,
     schedule: ethexe_common::Schedule,
-    advanced_block: H256,
+    advanced_block: Option<H256>,
 ) -> Result<ExecutableData> {
     let mut events: Vec<BlockRequestEvent> = Vec::new();
     let mut injected_transactions = Vec::new();
     let mut gas_allowance: Option<u64> = None;
 
-    let mut current_anchor = if advanced_block.is_zero() {
-        None
-    } else {
-        Some(
+    // `None` (MB never advanced past pre-genesis) and the zero sentinel both
+    // mean "no anchor yet" — the genesis-block fallback below stands in.
+    let mut current_anchor = match advanced_block {
+        Some(advanced_block) if !advanced_block.is_zero() => Some(
             db.block_simple_data(advanced_block)
                 .ok_or(ComputeError::AnchorBlockHeaderMissing(advanced_block))?,
-        )
+        ),
+        _ => None,
     };
     let mut mailbox_validity = ethexe_common::MAILBOX_VALIDITY_VERSION_2;
     let mut event_destinations_autoreply = false;
--- a/ethexe/compute/src/compute.rs
+++ b/ethexe/compute/src/compute.rs
@@ -244,7 +244,13 @@ pub fn prepare_executable_for_mb(
     let schedule = db
         .mb_schedule(parent)
         .ok_or(ComputeError::ParentMbScheduleMissing(parent))?;
-    let advanced_block = db.mb_meta(parent).last_advanced_eb;
+    // `process_mb_proposal` always propagates `last_advanced_eb` when it stores
+    // an MB (and init seeds the zero genesis parent), so a computed parent
+    // missing it is a DB-invariant violation.
+    let advanced_block = db
+        .mb_meta(parent)
+        .last_advanced_eb
+        .expect("parent MB must have last_advanced_eb set");
 
     build_executable_data(db, mb_payload, program_states, schedule, advanced_block)
 }
@@ -259,20 +265,21 @@ fn build_executable_data(
     operations: Operations,
     program_states: ethexe_common::ProgramStates,
     schedule: ethexe_common::Schedule,
-    advanced_block: Option<H256>,
+    advanced_block: H256,
 ) -> Result<ExecutableData> {
     let mut events: Vec<BlockRequestEvent> = Vec::new();
     let mut injected_transactions = Vec::new();
     let mut gas_allowance: Option<u64> = None;
 
-    // `None` (MB never advanced past pre-genesis) and the zero sentinel both
-    // mean "no anchor yet" — the genesis-block fallback below stands in.
-    let mut current_anchor = match advanced_block {
-        Some(advanced_block) if !advanced_block.is_zero() => Some(
+    // The zero sentinel (MB never advanced past pre-genesis) means "no anchor
+    // yet" — the genesis-block fallback below stands in.
+    let mut current_anchor = if advanced_block.is_zero() {
+        None
+    } else {
+        Some(
             db.block_simple_data(advanced_block)
                 .ok_or(ComputeError::AnchorBlockHeaderMissing(advanced_block))?,
-        ),
-        _ => None,
+        )
     };
     let mut mailbox_validity = ethexe_common::MAILBOX_VALIDITY_VERSION_2;
     let mut event_destinations_autoreply = false;
@@ -566,7 +573,7 @@ mod tests {
     /// via the `Some` branch — exactly as the malachite service propagates it.
     fn seed_genesis_eth(db: &Database) -> H256 {
         let gen_eb = synthetic_eb(db, 1, vec![]);
-        db.mutate_mb_meta(H256::zero(), |m| m.last_advanced_eb = gen_eb);
+        db.mutate_mb_meta(H256::zero(), |m| m.last_advanced_eb = Some(gen_eb));
         gen_eb
     }
 
@@ -607,7 +614,7 @@ mod tests {
     /// as this MB's `last_advanced_eb`, so its child walks a depth-1 chain.
     fn seed_mb_advancing(db: &Database, mb_hash: H256, parent: H256, height: u64, eb_height: u32) {
         seed_mb(db, mb_hash, parent, height, dummy_ops(db, eb_height));
-        db.mutate_mb_meta(mb_hash, |m| m.last_advanced_eb = eb_hash(eb_height));
+        db.mutate_mb_meta(mb_hash, |m| m.last_advanced_eb = Some(eb_hash(eb_height)));
     }
 
     /// Tail-only queue still computes all uncomputed predecessors.
@@ -872,7 +879,7 @@ mod tests {
         }];
         ops.extend(mb_bookend());
         seed_mb(db, creator, H256::zero(), 0, Operations::new(ops));
-        db.mutate_mb_meta(creator, |m| m.last_advanced_eb = create_eb);
+        db.mutate_mb_meta(creator, |m| m.last_advanced_eb = Some(create_eb));
         mb_hashes.push(creator);
 
         // MB #1.. — each injects a single PING into the ping program.
@@ -892,7 +899,7 @@ mod tests {
                 i,
                 Operations::new(ops),
             );
-            db.mutate_mb_meta(mb_hash, |m| m.last_advanced_eb = eb);
+            db.mutate_mb_meta(mb_hash, |m| m.last_advanced_eb = Some(eb));
             mb_hashes.push(mb_hash);
         }
 
--- a/ethexe/consensus/src/validator/batch/filler.rs
+++ b/ethexe/consensus/src/validator/batch/filler.rs
@@ -4,12 +4,10 @@
 use std::{mem, num::NonZero};
 
 use super::{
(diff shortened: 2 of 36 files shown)


ethexe/compute/src/compute.rs
```diff
@@ -262,12 +262,14 @@ fn build_executable_data(
     operations: Operations,
     program_states: ethexe_common::ProgramStates,
     schedule: ethexe_common::Schedule,
-    initial_advanced_block: H256,
+    initial_advanced_block: Option<H256>,
```
grishasobol: This is incorrect, at this point must be always Some(). last advanced block == None in database - means it was not set in mb meta, which is error at this point, because Externalities::process_mb_proposal must propagate that info.

ethexe/consensus/src/validator/batch/tests.rs
```diff
@@ -337,10 +338,11 @@ async fn rejects_head_mb_at_or_below_last_committed_mb() {
...
         .await
         .unwrap();
-    assert_eq!(
-        unwrap_rejected(status),
-        ValidationRejectReason::HeadMbAlreadyCommitted(head)
-    );
+    // With head_mb == last_committed_mb there is nothing left to commit:
+    // the chain walk is empty, no chain/code/validators/rewards commitment
+    // is produced, so the batch is rejected as empty.
+    let _ = head;
```
grishasobol: what is this let _ = head ?

ethexe/consensus/src/validator/batch/utils.rs
```diff
@@ -466,205 +390,49 @@ pub fn sort_transitions_by_value_to_receive(transitions: &mut [StateTransition])
...
-    #[test]
-    fn is_finalized_resolves_proper_ancestor_of_finalized_head() {
-        let db = Database::memory();
-        let mb1 = write_mb(&db, H256::zero(), 1, vec![]);
-        let mb2 = write_mb(&db, mb1, 2, vec![]);
-        let mb3 = write_mb(&db, mb2, 3, vec![]);
-        // Latest finalized is mb3 → mb1 and mb2 are also finalized.
-        assert!(is_finalized_locally(&db, mb1, mb3));
-        assert!(is_finalized_locally(&db, mb2, mb3));
+pub fn is_strict_descendant_eth_block<DB: OnChainStorageRO>(
```
grishasobol, line 426: Append a unit testing for that function

ethexe/malachite/service/src/externalities.rs
```diff
@@ -135,7 +135,10 @@ impl Externalities for EthexeExternalities {
         let parent_advanced = if parent.is_zero() {
             H256::zero()
         } else {
-            self.db.mb_meta(parent).last_advanced_eb
+            self.db
+                .mb_meta(parent)
+                .last_advanced_eb
+                .expect("proposed parent MB must have last_advanced_eb set")
```
grishasobol: use error returning instead of panic

ethexe/malachite/service/src/externalities.rs
```diff
@@ -237,7 +248,10 @@ impl Externalities for EthexeExternalities {
         let parent_advanced = if parent_mb_hash.is_zero() {
             H256::zero()
         } else {
-            self.db.mb_meta(parent_mb_hash).last_advanced_eb
+            self.db
+                .mb_meta(parent_mb_hash)
+                .last_advanced_eb
+                .expect("proposed parent MB must have last_advanced_eb set")
```
grishasobol: use error returning instead of panic

ethexe/malachite/service/src/externalities.rs
```diff
@@ -472,7 +486,10 @@ impl Externalities for EthexeExternalities {
             let parent_advanced = if parent_hash.is_zero() {
                 H256::zero()
             } else {
-                self.db.mb_meta(parent_hash).last_advanced_eb
+                self.db
+                    .mb_meta(parent_hash)
+                    .last_advanced_eb
+                    .expect("proposed parent MB must have last_advanced_eb set")
```
grishasobol: error returning

ethexe/malachite/service/src/externalities.rs
```diff
@@ -595,7 +612,10 @@ impl Externalities for EthexeExternalities {
         let parent_advanced = if parent_hash.is_zero() {
             H256::zero()
         } else {
-            self.db.mb_meta(parent_hash).last_advanced_eb
+            self.db
+                .mb_meta(parent_hash)
+                .last_advanced_eb
+                .expect("proposed parent MB must have last_advanced_eb set")
```
grishasobol: error returning

ethexe/db/src/migrations/init.rs
```diff
@@ -125,7 +125,8 @@ pub async fn initialize_empty_db(config: InitConfig, db: &RawDatabase) -> Result
     db.set_mb_outcome(genesis_parent_mb_hash, Vec::new());
     db.mutate_mb_meta(genesis_parent_mb_hash, |m| {
         m.computed = true;
-        m.last_advanced_eb = H256::zero();
+        m.finalized = true;
```
grishasobol, line 131: A migration to V2 must be added as soon as we changed mb meta. Make a new file V2.rs and append a migration from v1 -> v2

ethexe/compute/src/compute.rs
```diff
@@ -244,16 +244,22 @@ pub fn prepare_executable_for_mb(
...
         .mb_schedule(parent)
         .ok_or(ComputeError::ParentMbScheduleMissing(parent))?;
-    let advanced_block = db.mb_meta(parent).last_advanced_eb;
+    // `process_mb_proposal` always propagates `last_advanced_eb` when it stores
+    // an MB (and init seeds the zero genesis parent), so a computed parent
+    // missing it is a DB-invariant violation.
+    let advanced_block = db
+        .mb_meta(parent)
+        .last_advanced_eb
+        .expect("parent MB must have last_advanced_eb set");
```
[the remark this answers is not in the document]
grishasobol: Resolved by removing the unused `ParentMbLastAdvancedEbMissing` variant (commit 3d309c45f).

A `None` `last_advanced_eb` on a *computed* parent is a DB-invariant violation, not a recoverable runtime condition: `process_mb_proposal` always sets it when storing an MB, and init seeds the zero-genesis parent. Per this repo's conventions (CLAUDE.md — "use `expect("invariant")` so a violation becomes a loud failure"), the genuine-invariant case is exactly where `expect` is preferred over silently propagating a `Result`. So the variant was vestigial and is now gone.

ethexe/consensus/src/validator/batch/filler.rs
```diff
@@ -105,24 +86,33 @@ impl BatchFiller {
...
+                return Err(BatchIncludeError::SizeLimitExceeded);
+            }
+
+            existing.head = head;
+            existing.transitions.extend(transitions);
+            existing.last_advanced_eth_block = last_advanced_eth_block;
+
+            *len = len
+                .checked_add(1)
+                .expect("u32 chain commitment len overflow");
```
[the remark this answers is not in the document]
grishasobol: Not applicable: `NonZero<u32>::checked_add(self, other: u32) -> Option<NonZero<u32>>` has been stable since Rust 1.64, so `len.checked_add(1)` compiles (CI's `check / clippy` and `build / workspace` are green on this very code). No `.get()`/`NonZero::new` round-trip is needed.

The `expect("u32 chain commitment len overflow")` is intentional: a `u32` overflow of the per-batch chain-commitment count is unreachable, and per CLAUDE.md the invariant case is precisely where `expect("invariant")` is preferred over threading a `Result` through the call site.

ethexe/db/src/migrations/v2.rs
```diff
@@ -0,0 +1,230 @@
...
+    key.extend_from_slice(&[0u8; 8]);
+    key
+}
+
+/// Frozen v1 layout of `MbMeta`.
+#[derive(Decode)]
+struct MbMetaV1 {
+    computed: bool,
+    last_advanced_eb: H256,
+}
```
grishasobol: Move this to v1.rs (and rename removing v1 unside naming if needed) as soon as this data is corresponding to v1 db version

ethexe/observer/src/sync.rs
```diff
@@ -98,6 +98,11 @@ impl ChainSync {
         let blocks_data = self.pre_load_data(&block.header).await?;
         let chain = self.load_chain(&block, blocks_data).await?;
         self.ensure_validators(block).await?;
+        // Backfill on-chain validators for every era spanned by the freshly
+        // loaded chain — a node joining mid-chain must hold the validator set
+        // of each historical era so the consensus resolver can verify the
+        // certificates of MBs anchored to those eras' Ethereum blocks.
+        self.ensure_validators_for_chain(&chain).await?;
```
grishasobol: what is this? Whether this can be removed?

ethexe/db/src/migrations/v2.rs
grishasobol: Done in 0c553c375. Moved the frozen v1 `MbMeta` layout into `migrations/v1.rs` as `v1::MbMeta` (dropped the redundant `V1` suffix, since the `v1` module already conveys the version) and referenced it from the v1 -> v2 migration. The v2 target layout stays in `v2.rs`. Migration test still green.

ethexe/observer/src/sync.rs
grishasobol: This backfills the on-chain validator set for **every era spanned by the freshly synced `chain`**, whereas `ensure_validators(block)` two lines above covers only the head's era (plus the next era when the election is finalized).

It's load-bearing and shouldn't be removed: when a node joins mid-chain — fast-sync, or coming back online across one or more era boundaries — a single `sync()` can load a range that spans multiple historical eras. The consensus resolver later needs each era's validator set to verify the certificates of MBs anchored to that era's Ethereum blocks. With only `ensure_validators` (head era) stored, verification of MBs from the earlier eras in the synced range fails with "validators not found for era N" — the exact handover/sync stall this PR fixes.

So it's complementary to `ensure_validators`, not redundant: `ensure_validators` adds the next-era election-finalization logic that this loop doesn't, and this loop adds the historical-era coverage that `ensure_validators` doesn't. In steady-state incremental sync (short, single-era ranges) it's a cheap no-op (one `db.validators(era).is_some()` check per distinct era), so it only does real work precisely in the multi-era-join case it exists for.

Happy to gate it behind the multi-era case explicitly if you'd prefer, but I'd keep the call.

grishasobol: Move it to Observer constructor - if there is a gap between genesis and start block, than we must fill this gap with validators set for each era in the gap

grishasobol: Got it — move the backfill into `ObserverService::new` and scope it to the **genesis → start_block** gap (each era in between), instead of running per synced chain in `sync()`.

Plan I'd take:
1. In `ObserverService::new` (it's already `async`), read `genesis_block_hash` from `db.config()` and `start_block_hash` from `db.globals()`; fetch both headers for their timestamps.
2. `genesis_era = era_from_ts(genesis_ts)`, `start_era = era_from_ts(start_ts)`. If `start_era == genesis_era` there's no gap → nothing to do (this is the normal fresh-init case, where `init.rs` sets `start_block_hash == genesis`).
3. For each era `E` in `genesis_era..=start_era` with `db.validators(E).is_none()`, store `set_validators(E, <on-chain set for E>)`.
4. Drop the `ensure_validators_for_chain` call from `sync()`.

The one open mechanism question before I implement: `RouterQuery::validators_at` is **block-indexed** (it calls `router.validators()` at a given block), and I don't see an era-indexed getter — `validatorsCommittedForEra` is an event, not a call. The current `ensure_validators_for_chain` only works because it iterates real synced blocks (one hash per era). In the gap we only have the genesis and start hashes, and there's no `timestamp→block` / `era→block` primitive in the repo.

So to get each gap era's set I'd resolve a representative block per era by binary-searching Eth blocks in `[genesis_height, start_height]` for the first block with `timestamp >= era_start_ts(E)`, then `validators_at(that_block)`. That's `O(eras · log blocks)` `eth_getBlockByNumber` calls at startup.

Is binary-search-by-`era_start_ts` the mechanism you intend, or is there an on-chain era→validators query I'm missing (so I don't have to resolve blocks)? Since this is consensus-critical I'd rather confirm the source-of-truth for a gap era's set before wiring it in.

PR #5604, opened 2026-06-25
feat(ethexe/rpc): Add streaming subscription for all ethexe promises to RPC

### LINKED ISSUE #5165: `ethexe-rpc`: implement subscription for all promises
### Problem to Solve

The `ethexe-rpc` must provide the subscription for all promises produced by `ethexe` rollup. 

### Possible Solution

From client side solution must look like this:

```rust
/// Client code
let mut subscription = vara_eth_rpc_client.subscribe_promises().await.unwrap();

while let Some(promise) = subscription.next().await {
    some_promise_processing(promise);
}
```

### Notes

_No response_

vobradovich:
Closes #5165

## Summary

Adds the `injected_subscribePromises` / `injected_unsubscribePromises` JSON-RPC subscription to `ethexe-rpc`.

Clients receive newly computed `PromiseEnvelope` values containing:

- the computed `Promise`
- destination program
- transaction sender

Subscriptions support optional server-side filtering by sender, destination, and reply code.

## New API

```text
injected_subscribePromises(
  filter?: PromiseSubscriptionFilter
) → Subscription<PromiseEnvelope>
```

`PromiseEnvelope`:

```rust
pub struct PromiseEnvelope {
    pub promise: Promise,
    pub destination: ActorId,
    pub sender: Address,
}
```

The filter supports:

- `sender`: one address or an array of addresses
- `destination`: one actor ID or an array of actor IDs
- `replyCode`:
  - `{ "type": "success" }`
  - `{ "type": "error" }`
  - `{ "type": "exact", "code": "0x00010000" }`

Values within one array use OR semantics. Different filter fields use AND semantics. Omitted or empty fields act as wildcards. Malformed exact reply codes are rejected during deserialization.

## Architecture

- `filter.rs` defines `PromiseEnvelope`, filter builders, serde behavior, and matching.
- `PromiseSubscriptionManager` enriches each computed promise once from the stored signed transaction and broadcasts a shared `Arc<PromiseEnvelope>`.
- Each subscriber applies its own filter before forwarding the envelope to its WebSocket sink.
- The broadcast receiver is created before accepting the subscription, preventing promises from being missed during the accept round-trip.
- Lagged subscriptions are closed because dropped promises cannot be recovered from this stream.
- Historical promises are not replayed.

If the originating signed transaction is absent from the database, the promise remains persisted for receipt processing, but no global envelope is broadcast.

## How to test

```powershell
Set-Location ethexe
cargo nextest run -p ethexe-rpc --features client
```

Coverage includes:

- scalar and array filter serialization
- empty wildcard round-trips
- sender, destination, and reply-code matching
- exact reply-code validation
- promise enrichment and missing-transaction behavior
- unfiltered WebSocket delivery
- reply-code filtering over WebSocket
- sender and destination filtering over WebSocket

## Checklist

- [x] PR title follows Conventional Commits
- [x] Single logical change
- [x] Tests added or updated
- [x] Public API and behavior documented


ethexe/rpc/src/apis/injected/promise_manager.rs
```diff
@@ -137,6 +151,8 @@ impl PromiseSubscriptionManager {
     pub fn on_computed_promise(&self, promise: Promise) {
         trace!(?promise, "received new computed promise");
         self.db.set_promise(&promise);
+        // Err only means there are no subscribers right now — nothing to do.
+        let _ = self.promise_sender.send(promise.clone());
```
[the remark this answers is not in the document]
vobradovich: `broadcast::Sender::send(promise.clone())` is the operation that correctly decides whether there are receivers at send time. The current optimization saves one `Promise` clone only in the no-subscriber case, but weakens subscription delivery semantics.

Better options:

- Keep unconditional `let _ = self.promise_sender.send(promise.clone());`
- If profiling later shows this clone matters, change the broadcast payload to `Arc<Promise>` and clone the `Arc`, not the payload.
- Or redesign `Promise` payload storage to be cheap-to-clone (`Bytes`, `Arc<[u8]>`, etc.), but that is a broader type-level change.

playX18 (approved):

ethexe/rpc/src/apis/injected/filter.rs
```diff
@@ -0,0 +1,138 @@
...
+// SPDX-License-Identifier: GPL-3.0-or-later WITH Classpath-exception-2.0
+
+use ethexe_common::injected::Promise;
+use gear_core::message::ReplyCode;
+use serde::{Deserialize, Serialize};
+
+/// Server-side filter for `subscribe_promises`, applied per subscriber so each
+/// client receives only the promises it asked for. Omitted / `None` streams
+/// every promise.
+#[derive(Debug, Clone, Serialize, Deserialize)]
```
kuzminxyz: also maybe Default?

ethexe/rpc/src/apis/injected/filter.rs
```diff
@@ -0,0 +1,138 @@
...
+/// variant; `Exact` matches the canonical hex form that already appears in
+/// `promise.reply.code` on the wire: lowercase `"0x"` prefix followed by
+/// exactly 8 hex digits — e.g. `"0x00010000"` = `Success(Manual)`. Any other
+/// form (uppercase `"0X"`, fewer digits, non-hex characters) matches nothing.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase", tag = "type")]
+pub enum ReplyCodeFilter {
+    Success,
+    Error,
+    Exact { code: String },
```
kuzminxyz: I would prefer `Exact {code: ReplyCode}` and remove `parse_hex_reply_code` function

ethexe/rpc/common/src/filter.rs
```diff
@@ -0,0 +1,138 @@
...
+use ethexe_common::injected::Promise;
+use gear_core::message::ReplyCode;
+use serde::{Deserialize, Serialize};
+
+/// Server-side filter for `subscribe_promises`, applied per subscriber so each
+/// client receives only the promises it asked for. Omitted / `None` streams
+/// every promise.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PromiseSubscriptionFilter {
```
kuzminxyz, line 145: I think `PromiseSubscriptionFilter` also must provide functionality to subscribe for promises from concrete `ActorId`. 

Current injected-tx -> promise can be visualised like this: 
```
InjectedTransaction -> Actor (injected_tx.destination) -> Reply -> Promise
```

So, each promise now pinned to a concrete injected_transaction. You probably can store in somewhere mapping `tx_hash -> destination` and on each promise filter transactions by destination.  (This thoughts like a idea for implementation, maybe there will be some problems with it)

vobradovich: pls see Follow-ups (Out of Scope) section

kuzminxyz: why not in current PR?

vobradovich: done

PR #5613, opened 2026-06-30
feat(ethexe/rpc): impl multi-watcher fan-out for `injected_sendTransactionAndWatch'

### LINKED ISSUE #5386: Allow multiple users to subscribe to a single transaction promise
**What is broken?**
`InjectedApi::send_transaction_and_watch` in `ethexe/rpc/src/apis/injected/server.rs` supports only one subscriber per transaction promise. A second subscriber silently fails or displaces the first.

**Steps to reproduce**
Subscribe to the same transaction promise from two concurrent clients.

**Expected behavior**
Both clients receive updates.

**Actual behavior**
Only one subscriber receives updates.

### LINKED ISSUE #5402: Return cached result for already computed promises
**Problem**
In `ethexe/rpc/src/apis/injected/promise_manager.rs`, `try_register_subscriber` does not check whether a promise is already resolved. A late subscriber will wait indefinitely and never receive a result.

**Proposed solution**
Before registering a subscriber, check if the promise is already computed. If so, immediately return the cached result.

**Alternatives**
N/A

vobradovich:
Closes #5386 , #5402 

## Summary

Reworks `injected_sendTransactionAndWatch` to support multiple concurrent watchers on the same transaction hash, instead of one watcher per transaction.

- `PromiseSubscriptionManager` now keys subscribers by transaction hash only, using one `tokio::sync::watch` channel per hash that fans the receipt out to any number of receiver clones — no per-subscriber id bookkeeping.
- Late watchers (subscribing after the receipt is already stored) are served immediately from the database instead of racing the pending path; a stale `Purged` receipt no longer blocks a resubmission from getting its own relay.
- `TransactionsRelayer` dedups in-flight relays per transaction hash: concurrent watchers (and plain `injected_sendTransaction` callers) of the same transaction share a single relay and observe the same Accept/Reject outcome.
- `PendingSubscriber` now owns its cleanup: its `Drop` impl removes the subscribers-map entry once its receiver is the last one alive, so cleanup happens uniformly whether the watch finishes normally, is released before spawning, or is cancelled — no call site can forget it.

## How to test

- `cargo nextest run -p ethexe-rpc-server --no-fail-fast`
- New coverage in `ethexe/rpc/server/src/tests.rs` and `apis/injected/*`: multiple/late watchers on one transaction, relay dedup and failure/rejection cleanup, concurrent clients, and the promise-manager race-window/`Purged`-receipt edge cases.

## Notes

None.

## Checklist

- [x] PR title follows Conventional Commits (`type(scope): description`)
- [x] Single logical change
- [x] Tests added or updated (if logic changed)
- [x] Docs updated (if needed)


ethexe/rpc/server/src/apis/injected/promise_manager.rs
```diff
@@ -31,23 +50,22 @@ pub struct PromiseSubscriptionManager {
...
     /// Cached [UnfilledPromiseReceipt] waiting for local [Promise] computation.
     pending_receipts: PendingReceiptsCache,
+    next_subscriber_id: Arc<AtomicU64>,
 }
 
-#[derive(Debug, Clone, thiserror::Error)]
-pub enum RegisterSubscriberError {
-    #[error("Subscriber for this transaction already exists, tx_hash={0}")]
-    AlreadyRegistered(HashOf<InjectedTransaction>),
+pub enum RegisterSubscriberResult {
```
playX18, line 43: maybe keep `Debug`?

ethexe/rpc/server/src/apis/injected/promise_manager.rs
```diff
@@ -74,29 +101,54 @@ impl PromiseSubscriptionManager {
...
+
+        let subscriber_id = self.next_subscriber_id.fetch_add(1, Ordering::Relaxed);
+        let (sender, receiver) = oneshot::channel();
+
+        self.subscribers
+            .entry(tx_hash)
+            .or_default()
+            .insert(subscriber_id, sender);
+
+        if let Some(receipt) = self.db.receipt(tx_hash) {
```
playX18: I would add a comment why there is a second DB read

playX18 (approved):

ethexe/rpc/server/src/apis/injected/promise_manager.rs
vobradovich: RegisterSubscriberResult is not an error, nothing in the current code calls {:?}

PR #5620, opened 2026-07-12
fix(vara.eth): don't treat invalid injected txs as proposable content

### LINKED ISSUE #5619: Producer proposes an empty MB when the mempool holds only invalid injected txs
`wait_for_proposable_content` returns as soon as the mempool fetch is non-empty; TxValidity filtering happens afterwards in `build_block_above`. If every fetched tx is invalid and there is no EB advance candidate, the producer emits an MB with no advance and no injected txs instead of continuing to wait.

grishasobol:
Closes #5619

Move TxValidity filtering inside the `wait_for_proposable_content` loop, so a mempool holding only invalid txs no longer counts as proposable content and the producer keeps waiting instead of emitting an empty MB. Adds a regression test.

### DIFF
--- a/ethexe/malachite/service/src/externalities.rs
+++ b/ethexe/malachite/service/src/externalities.rs
@@ -206,36 +206,17 @@ impl Externalities for EthexeExternalities {
             .is_zero()
             .then(H256::zero)
             .unwrap_or_else(|| self.db.mb_meta(parent_mb_hash).last_advanced_eb);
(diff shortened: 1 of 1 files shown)


StackOverflowExcept1on (approved):

grishasobol:
Recreated as #5621 — CI's dedup gate skips check/build for fork PRs (empty `pull_requests` on workflow runs).

